PR #17, opened 2015-05-18
Rename scenario tests to names that make sense

iamjasonp:
* Port over scenario tests
* Rationalize names somewhat

This is a first cut of porting some more scenario tests over and rationalizing the names to make more sense.


roncain:
Closing this PR after making repo public

PR #22, opened 2015-05-20
Fix build script to use default web proxy to download remote files.

PatrickMcDonald:


roncain:
Thanks for the suggestion.  We're looking at it now.

iamjasonp:
Thanks @PatrickMcDonald! 

If you're interested, similar fixes probably need to be made to the [dotnet/corefx](https://github.com/dotnet/corefx/) and [dotnet/coreclr](https://github.com/dotnet/coreclr) repos.

PatrickMcDonald:
dotnet/corefx#38 is already done, I'll have a look at dotnet/coreclr.

PatrickMcDonald:
It seems the corefx change was reverted with dotnet/corefx#788, I wasn't aware of that. See dotnet/corefx#963

iamjasonp:
Ahh, I was looking for why I didn't see corefx#38 and was grovelling through the commits for it. 

We synced up the build scripts and proj files between dotnet/wcf and dotnet/corefx a few days before our initial commit, so our build scripts should be reasonably similar. Once dotnet/corefx#963 gets fixed, that change should come over here too.

PatrickMcDonald:
Good stuff, I think the change should be relatively simple, I'll have a look there.

iamjasonp:
Awesome - thanks!

PR #25, opened 2015-05-20
Support more than one parameter to build.cmd for OuterLoop Tests

hongdai:
*Previously we only support passing OuterLoop parameter:
       /p:WithCategories=OuterLoop
* The change adds parsing command line parameters.
   We will be able to pass more than one parameter, such as
   build /p:Coverage=true /p:WithCategories=OuterLoop

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -6,8 +6,26 @@ setlocal
 ::       means that that rebuilding cannot successfully delete the task
 ::       assembly. 
 
-echo %2 | find /I "outerloop"
-if "%errorlevel%" equ "0" (
+set outloop=false
+
+:START_CMDLINE_PARSE
+set SWITCH=%1
+if {%SWITCH%} == {} goto :END_CMDLINE_PARSE
+set VALUE=%2
+
+if /i "%SWITCH%" == "/p:WithCategories" (
+  if /i "%VALUE%" == "OuterLoop" (
+      set outloop=true
+      goto :END_CMDLINE_PARSE
+  )
+)
+
+SHIFT
+SHIFT
+goto :START_CMDLINE_PARSE
+:END_CMDLINE_PARSE
+
+if "%outloop%" equ "true" (
 	start /wait BuildWCFTestService.cmd
 )
 
@@ -36,8 +54,7 @@ set _buildlog=%~dp0msbuild.log
 set _buildprefix=echo
 set _buildpostfix=^> "%_buildlog%"
 
-echo %2 | find /I "outerloop"
-if "%errorlevel%" equ "0" (
+if "%outloop%" equ "true"  (
         pushd setupfiles
         start /wait RunElevated.vbs SetupWCFTestService.cmd
         popd
@@ -63,8 +80,7 @@ echo.
 findstr /ir /c:".*Warning(s)" /c:".*Error(s)" /c:"Time Elapsed.*" "%_buildlog%"
 echo Build Exit Code = %BUILDERRORLEVEL%
 
-echo %2 | find /I "outerloop"
-if "%errorlevel%" equ "0" (
+if "%outloop%" equ "true"  (
 	pushd setupfiles
 	start /wait RunElevated.vbs CleanupWCFTestService.cmd
 	popd
--- a/build.cmd
+++ b/build.cmd
@@ -8,25 +8,14 @@ setlocal
 
 set outloop=false
 
-REM this command line processing is a temporary step until the build tools provide a proper hook. 
+REM this is a temporary step until the build tools provide a proper hook. 
 REM it will need to deal with multiple include and exclude categories.
 REM See dotnet/corefx#1477
-:START_CMDLINE_PARSE
-set SWITCH=%1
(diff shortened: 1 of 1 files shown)


build.cmd
```diff
@@ -6,8 +6,26 @@ setlocal
...
-echo %2 | find /I "outerloop"
-if "%errorlevel%" equ "0" (
+set outloop=false
+
+:START_CMDLINE_PARSE
+set SWITCH=%1
+if {%SWITCH%} == {} goto :END_CMDLINE_PARSE
+set VALUE=%2
+
+if /i "%SWITCH%" == "/p:WithCategories" (
```
mconnew: This is only allowing for parameters to be passed with the format '/p:WithCategories OuterLoop' and won't work if it's passed as '/p:WithCategories=OuterLoop'

build.cmd
```diff
@@ -6,8 +6,26 @@ setlocal
...
+  if /i "%VALUE%" == "OuterLoop" (
+      set outloop=true
+      goto :END_CMDLINE_PARSE
+  )
+)
+
+SHIFT
+SHIFT
```
mconnew: If an earlier argument is passed using an =, but then the /p:WithCategories argument is passed with a space, this will cause the /p:WithCategories to be missed. E.g. if the parameters are "/p:Coverage=true /p:WithCategories OuterLoop" then the double shift will shift the "/p:Coverage=true" and the "/p"WithCategories" values off the argument parameters leaving just "OuterLoop".

build.cmd
hongdai: The script engineer apparently treat '=' as a command line parameter separator as well. If you pass in '/p:WithCategories=OuterLoop', %1 will be /p:WithCategories, %2 will be OuterLoop.

I have verified it work with '/p:WithCategories=OuterLoop'.

build.cmd
hongdai: Again the script engine treats '=' as a command line separator.

Also, does msbuild support /p:WithCategories OuterLoop? Everything passed into build.cmd will be passed into msbuild.

iamjasonp: I believe msbuild only does parameters where you do /p:ParameterName=Value - no spaces unless it's quoted. 

For example, 

* `/p:WithCategories=OuterLoop` will make parameter `WithCategories` equal "OuterLoop"
* `/p:WithCategories=Outer Loop` will make parameter `WithCategories` equal to "Outer"
* `/p:WithCategores="outer loop"` will make parameter `WithCategories` equal to "outer loop" 

Reminder that "OuterLoop" is case *insensitive*

roncain: It should be noted that this command line processing is a temporary step until the build tools provide a proper hook.  it will need to deal with multiple include and exclude categories, work if you just msbuild a separate .csproj or F5 from VS.
See https://github.com/dotnet/corefx/issues/1477

hongdai: Good to know! I have added a comment to the build.cmd.

build.cmd
mconnew: Awesome, I learnt something new today!

build.cmd
mconnew: My initial concern still stands though. You could pass /m on the command line to get it to use multiple cores and that would be a single parameter and get the command line parsing out of sync.

hongdai: Agreed. Since the build tools will provide a proper hook to deal with categories. I'll make it simple by just looking for string /p:WithCategories=OuterLoop in the command line.

roncain: For simplicity (and because build tools will replace this need) it might be best to just detect that OuterLoop appears on the command line. We should support usage like /p:WithCategories=InnerLoop,Outerloop.

hongdai: To reduce confusion, I'd think we should be consistent with current build tools usage, i.e. support the same set of parameters current build tools support.

build.cmd
```diff
@@ -6,8 +6,18 @@ setlocal
...
 
-echo %2 | find /I "outerloop"
-if "%errorlevel%" equ "0" (
+set outloop=false
+
+REM this is a temporary step until the build tools provide a proper hook. 
+REM it will need to deal with multiple include and exclude categories.
+REM See dotnet/corefx#1477
+
+echo %* | findstr /i /C:"/p:WithCategories=OuterLoop"  1>nul
```
roncain: This will break for syntaxes like /p:Withcategories=Custom,OuterLoop. I really think simpler is better. The appearance of OuterLoop anywhere on the cmd line is good enough for now. It would even survive a rename of WithCategories. We should actively chat on the 1477 issue so they know our requirements.

hongdai: /p:Withcategories=Custom,OuterLoop is invalid for msbuild and current build tool. i.e current build tool does not support such syntaxes. You need to do /p:Withcategories=Custom /p:WithcategoriesOuterLoop

roncain:
LGTM.  Thanks Hong.  Feel free to comment on https://github.com/dotnet/corefx/issues/1477 to make sure whatever hooks we need in the next version of build tools are discussed. Specifically, we will want the selfhost service to launch from build.cmd, msbuild of a random .csproj, and F5.

PR #27, opened 2015-05-21
Rename and clean up scenario tests for more consistency

iamjasonp:
* Port over scenario tests
* Rationalize names somewhat
* Remove packages.config references from ServiceModel.Scenarios.Common.csproj
* Split BasicBindingTests.cs into BasicHttp, Custom, and NetHttp binding test files
* Change exception messages in TypedProxyTests to state "timed out" rather than "failed".


iamjasonp:
cc: @roncain 
See comments in #17 - that PR was closed when making dotnet/wcf public and we were advised that downstream forks may get dissociated.

src/System.Private.ServiceModel/tests/Scenarios/Http/TransportChannel/BasicBindings/NetHttpBindingTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        StringBuilder errorBuilder = new StringBuilder();
+        NetHttpBinding binding = new NetHttpBinding();
+
+        bool success = false;
+        ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_NetHttp));
+        IWcfService serviceProxy = factory.CreateChannel();
+        string result = serviceProxy.Echo(testString);
+        success = string.Equals(result, testString);
+
+        Assert.True(success, string.Format("Error: expected response from service: '{0}' Actual was: '{1}'", testString, result));
```
Maxwe11: Why not `Assert.Equal`?

src/System.Private.ServiceModel/tests/Scenarios/Http/TransportChannel/BasicBindings/BasicHttpBindingTests.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            BasicHttpBinding binding = new BasicHttpBinding(BasicHttpSecurityMode.None);
+            ScenarioTestHelpers.RunBasicEchoTest(binding, Endpoints.HttpBaseAddress_Basic, variationDetails, errorBuilder);
+        }
+        catch (Exception ex)
+        {
+            errorBuilder.AppendLine(String.Format("Unexpected exception was caught: {0}", ex.ToString()));
+        }
+
+        Assert.True(errorBuilder.Length == 0, String.Format("Test Case: {0} FAILED with the following errors: {1}", testCaseName, errorBuilder));
```
Maxwe11: What benefits from this practice?

iamjasonp:
@Maxwe11 - Thanks for taking a look! 

Prior to going fully open, we had code that ran under a non-xunit framework, and you'll notice places where we use `Assert.True` instead of `Assert.Equal` or `Assert.Null`, etc. Hence, as we bring over some more test code, you'll see places where we don't use these helpers. 

One place where we currently don't use `Assert.Equal` or `Assert.Null` ... etc. is where we are trying to output a custom error message, such as 

    Assert.True(errorBuilder.Length == 0, String.Format("Test Case: {0} FAILED with the following errors: {1}", testCaseName, errorBuilder)); 

What do you think about usages like this? 

We're trying to avoid this as much as possible, but sometimes, the errors output by `Assert.XXXX` simply aren't good enough, and only `Assert.True` has an overload where we can specify an error string. 

If you do see something that needs to change, please do let us know or feel free to submit a PR

There will be more pushes into this PR - in response to your comments as well as while I scrub through. 

Cheers!  :smiley:

src/System.Private.ServiceModel/tests/Scenarios/Http/TransportChannel/BasicBindings/NetHttpBindingTests.cs
iamjasonp: Hopefully https://github.com/dotnet/wcf/pull/27#issuecomment-104207515 addresses this concern?

src/System.Private.ServiceModel/tests/Scenarios/Http/TransportChannel/BasicBindings/BasicHttpBindingTests.cs
iamjasonp: In this case, probably not much benefit. https://github.com/dotnet/wcf/pull/27#issuecomment-104207515 for a more detailed explanation of why this exists today.

roncain:
LGTM
cc: @StephenBonikowsky  for his feedback
I agree this is a multistep cleanup to get it merged back into master and available for others to work with

StephenBonikowsky:
The logic Ron and I worked through when we were first organizing the tests was that under the TransportChannel whether for Http or Tcp we should have a minimum of "buckets" unless as we proceed we find an obvious gap. Since the tests you want to move into these new files are simple round-trips of either a standard binding or a very simple custom binding they should go under the ServiceModelLayer bucket. Can you move them there?

src/System.Private.ServiceModel/tests/Scenarios/Http/TransportChannel/BasicBindings/BasicHttpBindingTests.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Threading.Tasks;
+using TestTypes;
+using Xunit;
+
+public static class Http_TransportChannel_BasicBindings_BasicHttpBindingTests
+{
+
+    [Fact]
+    [OuterLoop]
+    public static void DefaultSettings_EchoString_RoundTrips()
```
StephenBonikowsky: DefaultSettings_EchoString_RoundTrip()

Only making one round-trip.

src/System.Private.ServiceModel/tests/Scenarios/Http/TransportChannel/BasicBindings/CustomBindingWithHttpTransportTests.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Threading.Tasks;
+using TestTypes;
+using Xunit;
+
+public static class Http_TransportChannel_BasicBindings_CustomBindingWithHttpTransportTests
+{
+    // Client and Server bindings setup exactly the same using default settings.
+    [Fact]
+    [OuterLoop]
+    public static void SameBinding_Https_Soap12_EchoString_RoundTrips()
```
StephenBonikowsky: SameBinding_Https_Soap12_EchoString_RoundTrip()

Only making one round-trip.

src/System.Private.ServiceModel/tests/Scenarios/Http/TransportChannel/BasicBindings/NetHttpBindingTests.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using TestTypes;
+using Xunit;
+
+public static class Http_TransportChannel_BasicBindings_NetHttpBindingTests
+{
+    [Fact]
+    [OuterLoop]
+    public static void DefaultSettings_EchoString_RoundTrips()
```
StephenBonikowsky: DefaultSettings_EchoString_RoundTrip()

Only making one round-trip.

src/System.Private.ServiceModel/tests/Scenarios/Http/TransportChannel/BasicBindings/BasicHttpBindingTests.cs
roncain: I think the intent was to say "it round trips successfully" so perhaps we need a different naming pattern. RoundTrip_String_Using_Defaults?

iamjasonp:
@StephenBonikowsky, referring to your comment https://github.com/dotnet/wcf/pull/27#issuecomment-104425433:
I don't think these tests should move - they are basic binding tests that aren't explicitly do not try to pry open into components of the ServiceModel layer to test them. My thinking is that "ServiceModel layer" refers to things like the typed proxy, client base, message inspectors, behaviors, message formatters, etc. 

As for separate files - the move to separate files from one file (BasicBindingTests) was suggested in https://github.com/dotnet/wcf/pull/17#commitcomment-11240890. Fundamentally, there are a lot more things one can do with each binding (tweaking settings, say), which suggests that they can go into separate files. We just haven't quite gotten there yet.

StephenBonikowsky:
:shipit:

PR #41, opened 2015-05-27
Merge working scenario feature branch to master 

iamjasonp:
The scenario feature branch was getting a little messy with some errant commits and merges that weren't squashed (mainly responses to PR feedback). 

We didn't want to squash all the way in order to preserve some git blame history, so this is the middle ground that we were able to strike between squashing and preserving all commits. 

There were some funny merge errors as we were bringing the changes over - hence an additional fixup checkin (in case I messed something up - easy to revert/blame)


roncain:
:shipit:

mconnew:
:shipit:

Issue #50, opened 2015-05-28
Reorganize the existing scenario tests


roncain:
PR https://github.com/dotnet/wcf/pull/41 effectively moved our experimental scenario branch into master in preparation for allowing further reorganizing and improving in the open. However the test organization in that branch pre-dated this WCF repo becoming public and therefore was not available for community feedback.

The goal of this current issue is to publically reorganize and rename these existing tests into a more rational structure that is similar to how the long-standing WCF samples are organized. Once this reorganization has been reviewed and completed, we will be able to publish guidance for adding new scenario tests.

StephenBonikowsky:
For reference the WCF Samples on MSDN can be found here...
3.5: https://msdn.microsoft.com/en-us/library/vstudio/ms751532(v=vs.90).aspx
4.0/4.5: https://msdn.microsoft.com/en-us/library/dd483346(v=vs.110).aspx

So based on the previously released WCF Samples below is the new proposed structure with modifications where it felt right.

Notes...
- This only shows the file structure for where existing tests go. As more tests are added the file structure will widen.
- These are Scenario tests not Unit tests.
- The CS files are actual checked-in files, some of them need to be renamed, we should do that as a separate PR after this one is in.
- The test cases in the CS files also need to be revisited for renaming purposes, I also think that should be done as a separate PR.

Feedback would be appreciated, such as whether some directories should go deeper. For instance for Http and Tcp Bindings, Transport Security involves more than just ClientCredentialType so I took the directory one layer deeper to add other transport related scenarios for each Binding.

└───Scenarios
    ├───Binding
    │   ├───Custom
    │   │       CustomBindingWithHttpTransportTests.cs
    │   │
    │   ├───Http
    │   │   │   BasicHttpBindingTests.cs
    │   │   │   NetHttpBindingTests.cs
    │   │   │
    │   │   └───TransportSecurity
    │   │           ClientCredentialTypeTests.cs
    │   │
    │   └───Tcp
    │       └───TransportSecurity
    │               ClientCredentialTypeTests.cs
    │
    ├───Client
    │   ├───ChannelLayer
    │   │       DuplexChannelShapeTests.cs
    │   │       RequestReplyChannelShapeTests.cs
    │   │
    │   ├───ClientBase
    │   │       ClientBaseTests.cs
    │   │
    │   ├───ExpectedExceptions
    │   │       ExpectedExceptionTests.cs
    │   │
    │   └───TypedClient
    │           TypedProxyTests.cs
    │
[...]

PR #58, opened 2015-06-01
Reorganizing the scenario tests as described in Issue #50.

StephenBonikowsky:
Please see Issue #50 for additional details.


hongdai:
I pulled the changes and was able to run the tests successfully on my machine!

However, it is much slower than before because we have many more projects to build and run. How about we create only one project at the scenario root level and include all tests from the subdirectories?

On the console, it will only output total tests passed/failed info, individual test result is available in the testsResults.xml file.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/project.json
```diff
@@ -0,0 +1,62 @@
...
+    "xunit": "2.0.0-beta5-build2785",
+    "xunit.abstractions.netcore": "1.0.0-prerelease",
+    "xunit.assert": "2.0.0-beta5-build2785",
+    "xunit.core.netcore": "1.0.1-prerelease",
+    "xunit.netcore.extensions":  "1.0.0-prerelease-*"
+  },
+  "frameworks": {
+    "dnxcore50": {}
+  }
+}
```
roncain: Nit: GitHub will show this icon whenever the file is missing a final newline, so ultimately we need a newline here (and anywhere else this icon shows).

roncain:
LGTM

I recommend we merge as soon as we get another confirmation.

I also verified I can build and run this PR, and the CI machine's OuterLoop run succeeded.  I also verified the SLN solution folder layout looks good and that you can build everything from VS.

Nit: GitHub wants a newline at the end of every file, and some of these are missing them.  But I see this was already the case in some of the renamed and copied files.  I recommend we deal with that as a separate Issue+PR so we get all of them at once, including any others in existing code.

The normal way to close the original issue would be to include "Fixes #nnn" as the last line in the PR commit message.  No need to change this PR, because we can manually close the issue and link back to this PR.

I recommend we do *not* make a common csproj to shorten build time.  That is the job of the build tool work being discussed at https://github.com/dotnet/corefx/issues/1477.  I prefer we organize for clarity, like this PR does.

Next step is to ensure we have open issues for file and method renaming and any other scenario cleanup work

mconnew:
LGTM :shipit:

PR #63, opened 2015-06-02
Enable Kerberos and Ntlm authentication for HTTP

mconnew:
The HTTP transport was relying on TokenProvider's to provide the ICredentials for HttpClient. The code for Kerberos is complex and platform specific requiring P/Invokes which may not be available, such as in .Net Native. The TokenProvider logic was throwing an exception preventing Kerberos from being used when the absence of a Kerberos TokenProvide only blocks usage with message security. TokenProviders are only needed if using message security. For transport authentication, they are not needed so we can remove them and allow HttpClient to manage the transport authentication.


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -581,9 +503,8 @@ protected override void OnClose(TimeSpan timeout)
 
             protected internal override async Task OnCloseAsync(TimeSpan timeout)
             {
+                var timeoutHelper = new TimeoutHelper(timeout);
                 PrepareClose(false);
-                TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
```
roncain: Just checking -- this change is bug fix to make the PrepareClose happen inside the timeout interval, right?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/ClientCredentialsSecurityTokenManager.cs
```diff
@@ -70,89 +70,9 @@ internal protected bool IsIssuedSecurityTokenRequirement(SecurityTokenRequiremen
 
         public override SecurityTokenProvider CreateSecurityTokenProvider(SecurityTokenRequirement tokenRequirement)
         {
-            return this.CreateSecurityTokenProvider(tokenRequirement, false);
+            throw ExceptionHelper.PlatformNotSupported("CreateSecurityTokenProvider(SecurityTokenRequirement tokenRequirement) not supported");
```
roncain: Could you briefly describe which scenario will hit this PNSE?

roncain:
LGTM
The only comments are question for my own understanding

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: Yes

src/System.Private.ServiceModel/src/System/ServiceModel/Security/ClientCredentialsSecurityTokenManager.cs
mconnew: No new scenarios will hit a PNSE. In the original code, every code path that could be taken to get a TokenProvider always resulted in a PNSE. This changes removes all the checks to decide which line of code would be the lucky winner to throw a PNSE. The other methods in this class all have the exact same pattern of throwing PNSE with this format of message. We only need a TokenProvider with message security. If/when we implement that, this code will end up looking quite different anyway.

PR #64, opened 2015-06-02
Add project.lock.json for refactored scenario tests

roncain:
Following CoreFx's example, all project.lock.json files are
added to the repo to ensure we build against the right versions.


roncain:
Withdrawing this PR in favor of more comprehensive one to set locked=true

PR #65, opened 2015-06-02
Copy common target files from CoreFx to restore build tools

roncain:
The CoreFx repro has updated its common targets files to
restore the build tools.  We track whenever these change
and keep our repo up to date. This is necessary to permit
building the OSS projects in the TFS mirror.


SajayAntony:
:shipit: 
I wonder if the buildtools should or is already a submodule to dotnet/wcf?

roncain:
The buildtools themselves live in https://github.com/dotnet/buildtools so are separate from CoreFx.  However, the targets files that invoke them are in CoreFx and occasionally change -- usually to take a newer version or take advantage of newer functionality.  And the TFS mirrored versions require some of this new functionality to build properly.  Because we want our TFS mirrored sources to build directly from this OSS, we need to update to the latest targets.  It's also easier to take the CoreFx versions without modification than to keep a modified copy and try to keep it in sync.

hongdai:
LGTM.

PR #68, opened 2015-06-03
Set locked = true for all project.lock.json

roncain:
sets the 'locked' value to true for all project.lock.json files.
Also adds project.lock.json files for refactored scenario tests.

Setting the 'locked' value prevents package restore from
attempting to resolve package versions again and update
the project.lock.json file.

This both locks down the actual versions as well as prevents
the TFS mirror from attempting to write to a readonly file.


StephenBonikowsky:
:shipit:

SajayAntony:
:shipit:

hongdai:
LGTM

PR #71, opened 2015-06-05
Enable conditional building for Net Native

### LINKED ISSUE #60: Modify project file as needed to allow conditional building for NET Native
We want the OSS codebase to be able to build both CoreCLR and NET Native versions.  The default will be CoreCLR but using MsBuild properties set externally we want the ability to build for NET Native too. An example of this is https://github.com/dotnet/corefx/tree/master/src/System.IO.FileSystem

roncain:
These changes allow the main product to be built for
NET Native conditionally.  This is not the default but
is used when building the NuGet packages in the TFS mirror.

This commit also adds an embedded rd.xml when the project
is compiled for NET Native.

Fixes #60


hongdai:
Does your change enable running tests for Net Native? i.e. can I use build.cmd kick off a Net Native run?

roncain:
No, this does not enable building or testing against NET Native yet.  The changes here are only intended to make it possible to build for NET Native conditionally.  The TFS mirror will use this when building for NET Native. I will open new issues to enable testing against NET Native from OSS.

hongdai:
Thanks for explaining it!
LGTM.

Issue #62, opened 2015-06-02
Rename existing scenario tests where needed.


StephenBonikowsky:
The existing set of scenario tests have been moved a couple of times now, they were named based on their original locations which may no longer make sense for some of them. Revisit the tests and re-name the ones that need it.

SajayAntony:
Your commit message should include something like closes #issue. See [closing issues](https://help.github.com/articles/closing-issues-via-commit-messages/)

PR #79, opened 2015-06-10
Numerous changes to scenario tests per Issue #62

StephenBonikowsky:
This includes test class renames.
Test method renames.
Renaming one file.
Moved a couple types out of common code to the same location as the tests that need them.
Updated several tests that were using abstracted helper methods located in common code to just do what they need to do right in the test itself. These types of helper methods make more sense for unit tests than scenario tests, the abstraction is not worth a few extra lines of code.
This pull request fixes Issue #62


SajayAntony:
This breaks the history for the updated file and I would push back but since the contents are pretty small it might be ok this time. 
:shipit:

StephenBonikowsky:
I used "git mv" to rename it but after editing the contents of the newly moved file it turned into a delete and add. Maybe it needs to be modified first and then moved?

SajayAntony:
Lets just break up the commit next time.

iamjasonp:
There's no great way to move files from what I've found - git basically tracks each move as a delete/add, and "knows" when a file looks similar enough to call it a move.

PR #84, opened 2015-06-12
Update buildtools to 1.0.25-prerelease-00053

iamjasonp:
Fixes packages restore issues. 
This meant that workspaces where the packages/ directory wasn't deleted would continue to build
fine, but as soon as a clean build was done, nothing would succeed anymore.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -10,7 +10,7 @@
 
   <!-- Build Tools Versions -->
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


iamjasonp:
@roncain @zhenlan - please review after the fact. 
Merging this without an ACK from you - we need this to unblock CI as everything is failing today.

PR #86, opened 2015-06-16
Change some pragmas to format supported by internal build tools

iamjasonp:
There are some internal builds that use a slightly different compiler version,
which means that while this code works with build.cmd from Git, it doesn't
work with the toolchain we use for these builds.

These changes will mean that we build on the build systems we're aware of.


roncain:
LGTM.  I tested the patch via build.cmd as well as with the internal build system that required this change.

zhenlan:
LGTM. Thanks!

Issue #29, opened 2015-05-21
Provide a mechanism to pass command line parameters into tests


roncain:
Our scenario tests will require the ability to redirect the URL of the test service to alternate machines.  One can also imagine other test inputs that might feed some [Theory] based test.
One technique to investigate is a pre-build step that accepts MSBuild properties of a known format, generates a temporary .cs file with the information (a Dictionary?), and adds that file to the @(Compile) collection.  The tests could be written to consult the dictionary for expected names.  Also we could create a [Theory] style test that pulls multiple values.  The implementation should not be tied only to scenario tests but available to all tests.

This would be more generally useful than just for WCF, so we should consider working with https://github.com/dotnet/buildtools

mconnew:
Pull request #89 will need to be reverted as part of this change

roncain:
I have a working prototype that injects MSBuild properties into a dictionary available at runtime.  I will get it ready for review shortly.  We will make the fiddler proxy an option that can be polled by our test helpers.

SajayAntony:
I responded on the PR but should have added the info here. Basically consider falling back to GetEnvironmentVariable during runtime is critical since having to run msbuild each time seems like a productivity drain.

mconnew:
@SajayAntony, I would prefer it be the other way around. That is, if the variable has been defined, check for an environment variable first and if it exists use that, otherwise use the compiled in version. If the environment variable is a fallback mechanism, you would still need to edit and recompile to override it anyway. Then when you are done, that change would need to be reverted. At which point you might as well have just changed the value in the source and not worried about using an environment variable.

SajayAntony:
@mconnew I didn't know that that build would hardcode the values. My impression was that build args would not be populated unless it is a build. Are you looking for a way to override build props?

roncain:
I was thinking to implement as Matt suggested, namely populate the name/value pairs at build time from MSBuild properties, but give precedence to Env vars at runtime.  That means the build is guaranteed to populate them with either the defaults or from MSBuild properties.  But if you want to just change something and re-run without building, just set the Env var.  It gives the nice behavior of allowing a build machine to avoid Env vars, but also allows a developer to manually override without rebuilding.

SajayAntony:
@roncain @mconnew  sounds good. So lets just settle on Env vars takes precedence over .props

PR #89, opened 2015-06-16
Add support for tests to use fiddler proxy

mconnew:
With this change, you can specify /p:UseFiddler=true to build.cmd and the special hostname
localhost.fiddler will be used. This causes the localhost proxy bypass to be skipped and the
requests will be put through the defined proxy server. Fiddler then strips off the .fiddler
from the hostname and passes on the request with the Host being just localhost.
Some tests fail when using fiddler as the failure mode changes and there's not a lot we can
do about that, but this does give greater flexibility to debugging product/test issues.

### DIFF
--- a/src/System.ServiceModel.Tests.Common/src/BaseAddress.cs
+++ b/src/System.ServiceModel.Tests.Common/src/BaseAddress.cs
@@ -5,6 +5,29 @@ using System;
 
 public static class BaseAddress
 {
+#if USE_FIDDLER
+    // Base address for testing nonexistent endpoints
+
+    public const string HttpServerBaseAddress = "http://localhost.fiddler:8081/";
+
+    // Base address for HTTP endpoints
+    public const string HttpBaseAddress = "http://localhost.fiddler:8081/WindowsCommunicationFoundation";
+
+    // Base address for HTTPS endpoints
+    public const string HttpsBaseAddress = "https://localhost.fiddler:44285/WindowsCommunicationFoundation";
+
+    // Base address for HTTPS endpoints with Basic Authentication 
+    public const string HttpsBasicBaseAddress = "https://localhost.fiddler:44285/WindowsCommunicationFoundation";
+
+    // Base address for HTTPS endpoints with Digest Authentication
+    public const string HttpsDigestBaseAddress = "https://localhost.fiddler:44285/WindowsCommunicationFoundation";
+
+    // Base address for HTTPS endpoints with NT Authentication
+    public const string HttpsNtlmBaseAddress = "https://localhost.fiddler:44285/WindowsCommunicationFoundation";
+
+    // Base address for HTTPS endpoints with Windows Authentication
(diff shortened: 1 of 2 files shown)


src/System.ServiceModel.Tests.Common/src/BaseAddress.cs
```diff
@@ -5,6 +5,29 @@
 
 public static class BaseAddress
 {
+#if USE_FIDDLER
+    // Base address for testing nonexistent endpoints
```
roncain: This approach is really a special case of what I intended in https://github.com/dotnet/wcf/issues/29.  So if we do this approach, we will need to back it out when we do it right -- which I think consists of propagating an MSBuild property ("WcfBaseURL"?) into the build.  It also is affected by the Bridge conversations, because the URI's will be provided by Bridge.  So it probably means we need to be able to specify the URL of the Bridge itself, as well as the base address we want Bridge to prepend if it chooses localhost.  This means we should really drive this requirement into the Bridge discussion, so that we can best choose how Bridge gets configured.

roncain: BTW, my thinking went something like this -- using the new lightweight name-value pair config API's, write an MSBuild task that generates code at build time to populate the name-value pairs directly from available MSBuild properties.  Tests could them poll these name-value pair for things like Bridge URL, Fiddler prefix, etc.

mconnew: I definitely agree we need a general solution to this problem. The test Bridge host will be driving the hostname used so long term it makes sense to have this problem solved via the config mechanism we use there. I had an itch I needed to scratch today which was get fiddler working on localhost for our existing outer loop tests so figured I would generalize the solution so others could easily benefit (and I wouldn't need to keep a local copy of these changes in sync :smiley: .) I have no problem with this being a stop-gap solution. I would like to see us not need to use fiddler in the long run as I would eventually like to have some inspection mechanism built into either the bridge or the client.

hongdai: Fiddler is very helpful when debugging issues related to requests. I'd vote we check this in now and modify it when we have the long term solution.

iamjasonp:
This looks like it'd be a good addition to our current (very minimal) arsenal of tools we can use to diagnose issues - it seems self-contained enough too that we could back it out pretty easily when the bridge comes into play. 

LGTM.

src/System.ServiceModel.Tests.Common/src/BaseAddress.cs
iamjasonp: I agree with @hongdai, @mconnew - it's a stopgap solution to deal with an issue for now. I think the bridge discussion could take a while, so in the meantime, let's make the tests more debuggable for all.

roncain: Please update the Bridge bug to request configurability of base address to allow things like this.  Also please update #29 to remove this code when we do it right.

roncain:
LGTM, subject to my replacing this shortly with #29

PR #102, opened 2015-06-17
Fix ability to replay http content as well as code cleanup

### LINKED ISSUE #48: Digest authentication with Http transport hangs
`HttpClient` doesn't support pre-authentication with Digest authentication and requires the request entity body to be resent. Currently the Http transport doesn't support re-sending of the request message content after it has been sent once. The result is the client asynchronously await's on a `Task` which will never complete.

mconnew:
We couldn't replay the HttpContent because of the way the ForwardingStream worked. This caused problems with digest authentication as pre-authentication isn't possible with WinHttp (the http client behind the corefx implementation of HttpClient).  This is achieved with a new MessageContent class which encapsulates a WCF Message and can write directly to a stream provided by HttpClient multiple times.
Code cleanup allows easier error handling and extending the error cases in future changes. The send and receive code paths are separated more cleanly. This wasn't possible before because of the HttpOutput/HttpInput abstraction we were dealing with which was inherited from Desktop code. The net change is a reduction of around 700 lines of code.
Fixes #48, #6 and #5


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -647,42 +648,182 @@ public void Fault(RequestChannel channel)
                             throw;
                         }
                     }
+                }
 
-                    try
+                private bool PrepareMessageHeaders(Message message)
```
mconnew: This code came from various methods in HttpInput placed into the correct order here.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpRequestMessageHelper.cs
```diff
@@ -0,0 +1,367 @@
...
+using System.Net;
+using System.Net.Http;
+using System.ServiceModel.Diagnostics.Application;
+using System.ServiceModel.Security;
+using System.Threading.Tasks;
+using System.Xml;
+
+namespace System.ServiceModel.Channels
+{
+    internal class HttpResponseMessageHelper
```
mconnew: A lot of the code here came from HttpInput, with some conversion to async where necessary.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -0,0 +1,367 @@
...
+using System.Net;
+using System.Net.Http;
+using System.ServiceModel.Diagnostics.Application;
+using System.ServiceModel.Security;
+using System.Threading.Tasks;
+using System.Xml;
+
+namespace System.ServiceModel.Channels
+{
+    internal class HttpResponseMessageHelper
```
mconnew: Most of this code came from HttpInput, with some changes made to make things async where appropriate.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
```diff
@@ -0,0 +1,266 @@
...
+
+        public Message Message
+        {
+            get
+            {
+                return _message;
+            }
+        }
+
+        private void PrepareContentHeaders()
```
mconnew: This is the code from HttpInput that set headers on the HttpContent. As we have the message available to us, the HttpContent class should be the one to set the properties.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,351 @@
...
+            }
+
+            internal int Write(byte[] srcBuffer, int srcOffset, int srcCount)
+            {
+                if (_readOnly)
+                {
+                    throw new InvalidOperationException("BufferContainer.Write");
+                }
+
+                // TODO: Parameter validation
```
SajayAntony: Why not just to it instead of a TODO?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing && !_disposed)
+            {
+                _disposed = true;
+                if(_buffer.Array != null)
+                {
+                    _bufferManager.ReturnBuffer(_buffer.Array);
```
SajayAntony: reset it first and then return to the BufferManager.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
+﻿using System.Diagnostics.Contracts;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.ServiceModel.Channels
+{
+    internal class ProducerConsumerStream : Stream
```
SajayAntony: I would recommend reordering the methods to have Write/WriteAsync and Read/ReadAsync one after the other. There are bunch of properties sprinkled in between them currently.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
```diff
@@ -0,0 +1,266 @@
...
+            _stream = new ProducerConsumerStream();
+            _encodingTask = _messageEncoder.WriteMessageAsync(_message, _stream);
+            return Task.FromResult(_stream);
+        }
+
+        protected override async Task SerializeToStreamAsync(Stream stream, TransportContext context)
+        {
+            _stream = new ProducerConsumerStream();
+            await _messageEncoder.WriteMessageAsync(_message, _stream);
+            await _stream.CopyToAsync(stream, 8192);
```
SajayAntony: smoking :gun: with 8KB constant. Move it up so that we know its a ```StreamedMessageContent.DefaultBufferSize``` or in someway call it out.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -513,18 +500,12 @@ protected override IAsyncRequest CreateAsyncRequest(Message message)
...
                 // SecurityTokenContainer doesn't exist so was excluded from method signature
-                return Task.FromResult(this.Factory.GetHttpRequestMessage(to, via, cancelToken, false));
-            }
-
-
-            public virtual bool WillGetWebRequestCompleteSynchronously()
-            {
-                return true;
+                return Task.FromResult(this.Factory.GetHttpRequestMessage(to, via));
             }
```
roncain: Nit: sometime the code uses 'this.Factory' and sometimes just 'Factory'.  Recommend be consistent throughout with 'this'.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -564,32 +548,56 @@ public HttpClientChannelAsyncRequest(HttpClientRequestChannel channel)
...
-                        HttpOutput httpOutput = HttpOutput.CreateHttpOutput(_httpRequestMessage, _factory, request, _factory.IsChannelBindingSupportEnabled, _factory.GetHttpClient(), _factory._authenticationScheme, _timeoutHelper);
+                        bool suppressEntityBody = PrepareMessageHeaders(request);
 
+                        if (!suppressEntityBody)
+                        {
+                            _httpRequestMessage.Content = MessageContent.Create(_factory, request, _timeoutHelper);
+                        }
+
+                        await SendPreauthenticationHeadRequestIfNeeded();
                         bool success = false;
```
roncain: Just curious -- why is SendPreauthenticationHeadRequest not inside the try/catch?  Can it throw, and if it does shouldn't we abort the channel?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -601,9 +609,9 @@ private void Cleanup()
                     if (_httpRequestMessage != null)
                     {
                         _timeoutHelper.CancelCancellationToken(false);
-                        this.TryCompleteHttpRequest(_httpRequestMessage);
+                        TryCompleteHttpRequest(_httpRequestMessage);
+                        _httpRequestMessage.Dispose();
                     }
```
roncain: Should we null out _httpRequestMessage after the Dispose?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
roncain: It's still difficult to know why we're seeing both red and green for code that came from somewhere else.  I'm not sure what I can ignore and what I need to review carefully.  Anything you can do to identify areas we can ignore would be a huge help.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -0,0 +1,367 @@
...
+                try
+                {
+                    toAbsent = (message.Headers.To == null);
+                }
+                catch (XmlException)
+                {
+                }
+                catch (CommunicationException)
+                {
+                }
```
roncain: There are several catch/ignores here around seemingly innocuous property sets.  Could you add comments what's going on and why these exceptions can be ignored?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
```diff
@@ -0,0 +1,266 @@
...
+            }
+            else
+            {
+                return new BufferedMessageContent(request, factory.MessageEncoderFactory.Encoder, factory.BufferManager);
+            }
+        }
+    }
+
+    internal class StreamedMessageContent : MessageContent
+    {
```
roncain: I'd consider making these classes public solely for unit testing.  If would be extremely useful to unit test these

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
...
+            if (offset < 0 || offset > buffer.Length)
+            {
+                throw new ArgumentOutOfRangeException("offset");
+            }
+            if (count < 0 || count > buffer.Length - offset)
+            {
+                throw new ArgumentOutOfRangeException("count");
+            }
+            Contract.EndContractBlock();
+
```
roncain: I think we either need to opt all-in for things like Contract.EndContractBlock() or opt all-out.  Once you opt in, then we probably need Contract.Requires and Contract.Assert on internal methods as well. I personally like use of Contract, but only if consistently embraced.  Enabling static analysis without an all-in adoption yields only noise.  Your call to leave this in or not.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
...
+        }
+
+        public override bool CanWrite
+        {
+            get { return !_disposed; }
+        }
+
+        public override long Length
+        {
+            get { throw new NotImplementedException(); }
```
roncain: Aren't these going to generate warnings when we build under TFS.  NotImplementedByDesign?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
...
+            if (count < 0 || count > buffer.Length - offset)
+            {
+                throw new ArgumentOutOfRangeException("count");
+            }
+            Contract.EndContractBlock();
+
+            if (_disposed)
+            {
+                return 0;
+            }
```
roncain: Why not throw if attempting to read from a disposed content like this?
And if returning a value is acceptable, is zero the right one (vs. -1)?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
...
+
+            _buffer.TrySetResult(new WriteBufferWrapper(buffer, offset, count));
+            int totalBytesRead = 0;
+            do
+            {
+                int bytesRead = await _dataAvail.Task;
```
roncain: Should we break the read loop if the cancellation token requests cancellation?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
...
+            {
+                return !(a == b);
+            }
+
+            internal int Write(byte[] srcBuffer, int srcOffset, int srcCount)
+            {
+                if (_readOnly)
+                {
+                    throw new InvalidOperationException("BufferContainer.Write");
+                }
```
roncain: Should this be a resource string?  Or can we reuse any other kind of readonly exception message?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelHelpers.cs
```diff
@@ -1672,20 +80,12 @@ private static Exception CreateUnexpectedResponseException(HttpRequestException
...
             return TraceResponseException(
                 new ProtocolException(SR.Format(SR.UnexpectedHttpResponseCode,
-                (int)response.StatusCode, statusDescription), responseException));
-        }
-
-        public static Exception CreateNullReferenceResponseException(NullReferenceException nullReferenceException)
-        {
-            return TraceResponseException(
-                new ProtocolException(SR.NullReferenceOnHttpResponse, nullReferenceException));
+                (int)response.StatusCode, statusDescription)));
```
iamjasonp: Did you mean to take out the `,responseException` in the `SR.Format`?

iamjasonp: Oh, my bad, the method signature changed on this one, no longer passed in.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -0,0 +1,367 @@
...
+            Exception result = null;
+            AddProperties(message);
+
+            // check if user is receiving WS-1 messages
+            if (message.Version.Addressing == AddressingVersion.None)
+            {
+                bool actionAbsent = false;
+                try
+                {
+                    actionAbsent = (message.Headers.Action == null);
```
iamjasonp: I don't know that `Message.Headers` guaranteed to be non-null (it's abstract, so it really depends on the implementer). 
This would be a really weird place to null ref if it happened.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -0,0 +1,367 @@
...
+
+        private async Task<Stream> GetStreamAsync()
+        {
+            var content = _httpResponseMessage.Content;
+            Stream contentStream = null;
+            _contentLength = -1;
+            if (content != null)
+            {
+                contentStream = await content.ReadAsStreamAsync();
+                _contentLength = content.Headers.ContentLength ?? -1;
```
iamjasonp: We may have to be careful about the use of the `??` operator as it's relatively new, and I recall that we may have to watch our uses of these newer keywords/operators. Please be sure this compiles in all the environments we need to work on prior to merging the PR.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
...
+                set
+                {
+                    if (_offset == value)
+                    {
+                        return;
+                    }
+                    if (_readOnly)
+                    {
+                        throw new ArgumentException("value");
+                    }
```
iamjasonp: I would put the `_readonly` check ahead of the `_offset == value` check.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
...
+                set
+                {
+                    if (_count == value)
+                    {
+                        return;
+                    }
+                    if (_readOnly)
+                    {
+                        throw new ArgumentException("value");
+                    }
```
iamjasonp: ditto on the ordering of `_readonly` check.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
...
+            {
+                // Doing _readOnly comparison first as main use case is comparing against
+                // BufferContainer.EmptyContainer which is the only readonly instance of BufferContainer.
+                // If either instance is the read only, then this field is sufficient for comparison.
+                if (_readOnly || other._readOnly)
+                {
+                    return _readOnly == other._readOnly;
+                }
+
+                if (other._buffer == _buffer && other._offset == _offset)
```
iamjasonp: NIT: this check for buffer/offset/count was just slightly strange while I was hunting for the things you were comparing to make sure you covered everything :smile:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -0,0 +1,350 @@
...
+        }
+
+        public override void Flush()
+        {
+            _dataAvail.TrySetResult(0);
+        }
+
+        public override long Seek(long offset, SeekOrigin origin)
+        {
+            throw new NotImplementedException();
```
iamjasonp: We will run into problems here when building via TFS. I think you want `NotImplemented.ByDesign`

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -564,32 +548,56 @@ public HttpClientChannelAsyncRequest(HttpClientRequestChannel channel)
...
+                        await SendPreauthenticationHeadRequestIfNeeded();
                         bool success = false;
+
                         try
                         {
-                            _httpResponseMessageTask = await httpOutput.SendAsync(_httpRequestMessage);
-                            //this.channelBinding = httpOutput.TakeChannelBinding();
-                            await httpOutput.CloseAsync();
+                            _httpResponseMessage = await _httpClient.SendAsync(_httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, _timeoutHelper.CancellationToken);
+                            _httpRequestMessage.Dispose();
```
roncain: It would be useful to comment why success is assumed here if no exception is thrown.  Is there nothing in the status code or headers that we care about?

Also, shouldn't we set _httpResponseMessage to null after the Dispose -- just to avoid holding onto disposed objects?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -697,6 +838,31 @@ private void TryCompleteHttpRequest(HttpRequestMessage request)
                         _channel.OnHttpRequestCompleted(request);
                     }
                 }
+
+                private async Task SendPreauthenticationHeadRequestIfNeeded()
+                {
+                    if (_factory.AuthenticationScheme == AuthenticationSchemes.Anonymous)
+                    {
+                        return;
+                    }
```
iamjasonp: NIT: Above could be replaced with 

    if (!AuthenticationSchemeMayRequireResend)
    {
        return; 
    }

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -647,42 +648,182 @@ public void Fault(RequestChannel channel)
...
+                            {
+                                // this will be taken care of by System.Net when we make the request
+                            }
+                            else if (string.Compare(name, "proxy-connection", StringComparison.OrdinalIgnoreCase) == 0)
+                            {
+                                throw ExceptionHelper.PlatformNotSupported("proxy-connection");
+                            }
+                            else if (string.Compare(name, "range", StringComparison.OrdinalIgnoreCase) == 0)
+                            {
+                                // we don't support ranges in v1.
```
roncain: Question: if we don't support it, why don't we throw PNSE like we do for proxy-connection?
Is there a reason we don't support it?  If so, a comment would be good.
Also, nit -- I'd avoid fictitious versions like 'v1' and just say we don't currently support it.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -647,42 +648,182 @@ public void Fault(RequestChannel channel)
...
+                            else
+                            {
+                                if (!_httpRequestMessage.Headers.TryAddWithoutValidation(name, value))
+                                {
+                                    throw FxTrace.Exception.AsError(new InvalidOperationException(SR.Format(
+                                                    SR.CopyHttpHeaderFailed,
+                                                    name,
+                                                    value,
+                                                    _httpRequestMessage.Headers.GetType().Name)));
+                                }
```
roncain: Is the GetType().Name safe in NET Native?  I've seen that usage throw for missing metadata in other places because the toolchain could not infer its use.  Excellent reason we need negative tests that trigger these exceptions.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -0,0 +1,367 @@
...
+        private long _contentLength;
+
+        public HttpResponseMessageHelper(HttpResponseMessage httpResponseMessage, HttpChannelFactory<IRequestChannel> factory)
+        {
+            Contract.Assert(httpResponseMessage != null);
+            Contract.Assert(httpResponseMessage.RequestMessage != null);
+            _httpResponseMessage = httpResponseMessage;
+            _httpRequestMessage = httpResponseMessage.RequestMessage;
+            _factory = factory;
+            _encoder = factory.MessageEncoderFactory.Encoder;
```
roncain: Nit: Contract.Assert(factory != null)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -0,0 +1,367 @@
...
+
+            if (!hasContent)
+            {
+                if (_encoder.MessageVersion == MessageVersion.None)
+                {
+                    message = new NullMessage();
+                }
+                else
+                {
+                    return null;
```
roncain: Question: what is the distinction between a NullMessage and a null message value?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -0,0 +1,367 @@
...
+            {
+                Stream contentStream = await GetStreamAsync();
+                if (contentStream != null)
+                {
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ProtocolException(SR.HttpContentTypeHeaderRequired));
+                }
+                return false;
+            }
+            else if(_contentLength!=0)
+            {
```
roncain: What happens if _contentLength==-1 here?

I also wonder if _contentLength and _contentType should be pre-iniitialized when content==null.  It feels odd for a validate method to initialize some instance fields conditionally and leave the others as they were.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: I just discovered yesterday via an internal email thread (not related to WCF but still using HttpClient), some web services will fail a request for a HEAD request but accept the POST request. This means we shouldn't abort on a failure to issue a HEAD request as the real request might still succeed. I'm going to wrap it in a try/empty catch block with a comment explaining why.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: It's the request message I'm calling `Dispose()` on here, not the response message. We don't want to null the field out just yet as we might still want to retrieve some properties from the object. Calling `Dispose()` causes the contained `HttpContent` object to be disposed, which in our case will return buffers to the `BufferManager` and dispose the contained `Stream`. If no exception has been thrown, that means the entire request stream has been sent to the server, and the HTTP response headers have been received. This means we don't need the request object any more. I'll add a short comment.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: TryCompleteHttpRequest currently doesn't do anything. It uses an interlocked compare exchange to ensure _channel.OnHttpRequestCompleted is only called once. OnHttpRequestCompleted was used for cleaning up some native client certificate security allocations that we don't do at the moment. That code shouldn't throw. If a change in the future is made such that it can and does throw, the worst that could happen is the stream and allocated buffer in the request object won't get cleaned up until the GC runs and the buffer won't get returned to the buffer pool.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: It's not feasible to match up the red to the green as the original code was like spaghetti and took a lot of effort to unravel. If a step has been missed out or is in the wrong order, we should pick it up in our tests as the requests need to be very precisely structured and if something is out of place, the service will reject it.  
I would focus on the green code and check for error handling, confusing code where a clarifying comment would be helpful, and bad code practices such as not guarding for null's etc.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: I'll fix the comment. This was copied verbatim from the existing code which came from the desktop. It would seem reasonable to throw an exception here as it's a header where a customer has explicitly provided it and we're silently dropping it instead of basically saying that makes no sense. But throwing an exception here would be different behavior from the desktop. This header is a strange one, as the idea of the range header is to be able to pick up where you left off if a connection dropped, or to make multiple concurrent requests to improve bandwidth, but that doesn't make sense for SOAP. For example, making the exact same request twice to an endpoint using message security (I know we don't support that yet, but we do on desktop) can result in a different response, and piecing those pieces of response together isn't possible. Also it would require the same service request to be executed twice. The only time range makes sense is with a static resource. The ByteStreamMessageEncoder is the only way this could even make sense, and we haven't included that.

mconnew: Basically my question is, do you think we should throw an exception or continue to silently drop it like we do on the desktop clr?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: What are your thoughts on putting the constant string "System.Net.Http.Headers.HttpRequestHeaders" in here? It's only ever going to be that type.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: Agreed

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelHelpers.cs
mconnew: For background on why this was needed, `HttpWebRequest` would throw an exception for some HTTP status codes which means we would have an exception to set as the inner exception for the `ProtocolException`. With `HttpClient`, you get an `HttpResponseMessage` with the relevant status code and no exception. This means a lot of the failure modes for the Http transport no longer occur with an exception and instead are detected by inspecting the response code.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
mconnew: When the MessageVersion is None, it is acceptable to receive a response with no body as nothing dictates that you need a body. A NullMessage represents a message with no content but can still contain other metadata such as message properties. For example, HttpResponseMessageProperty containing the response status code and headers. Other message versions require a SOAP message be returned so having no body isn't acceptable and is an error condition. In those cases, representing as a NullMessage doesn't reflect the situation so null is returned which means "we didn't receive a response soap message". I don't know why null is returned instead of throwing an exception (that's what I would do if I was designing this from scratch), but this is existing behavior, the semantics of which is understood by calling code.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
mconnew: The code for Message.MessageHeaders reads as follows:
```csharp
        public abstract MessageHeaders Headers { get; } // must never return null
```
So although external code could potentially implement this returning null, I think it's safe to presume all ServiceModel implementations will never have this value be null. This is the same behavior as on desktop.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
mconnew: I'll add comments. This is existing code so I had to investigate. Basically message.Headers.Action causes an xml document to be read with an XmlDictionaryReader, which could trigger an XmlException if the xml is malformed. CommunicationException is the base exception that all WCF exceptions derive from. This would mean some WCF code somewhere else would be throwing an exception. I can't tell you what type, this is a catch-all for anything WCF related.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
mconnew: I'll switch it to use the older ? style.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: Agreed. Although unit testing might be difficult as HttpContent uses `protected abstract` methods to allow custom behavior while not allowing encapsulating an existing HttpContent inside another. This prevents the HttpContent class from being easily used from external code. I might need to get creative in how the tests are written.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: I realized it was not needed. We can pass the original stream into WriteMessageAsync, there's no need for an intermediate stream. The passed in stream might need to be wrapped in a BufferedStream depending on it's implementation, but that will show up in perf tests later if it's an issue.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: What do you mean? Do you mean zero out the buffer before returning it? Or do you mean do the `_buffer = default(ArraySegment<byte>)` first?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: Looking at other streams, it looks like I actually need to use `NotSupportedException`. According to [msdn](https://msdn.microsoft.com/en-us/library/system.notsupportedexception%28v=vs.110%29.aspx), this is the correct exception to use.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: Some Contract methods work all the time, and other only when doing a special compilation with the symbol CONTRACTS_FULL enabled. I don't believe we have that symbol enabled and EndContractBlock() is conditionally compiled with that symbol so it ends up a noop. I'm removing this as it's effectively dead code.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: As per response to Jason, `NotSupportedException` is the correct exception so I'm changing it to that.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: The only way to signal that there is nothing more to write is by calling `Dispose()`. From the writers perspective, once it has written everything and the final `WriteAsync()` has completed, the stream has been completed and it's free to call `Dispose()`. From the reader's perspective, the way it is signaled that a steam has finished and there's nothing else left is when a call to `Read()`/`ReadAsync()` returns zero bytes. [Here](https://msdn.microsoft.com/en-us/library/system.io.stream.read(v=vs.110).aspx) in the msdn documentation for the return value for Stream.Read it says:
> The total number of bytes read into the buffer. This can be less than the number of bytes requested if that many bytes are not currently available, or zero (0) if the end of the stream has been reached.  

Think of the scenario where `ReadAsync()` is called with a buffer which can hold 1024 bytes. `WriteAsync()` is called with 512 bytes, which is then copied to the read buffer then returns. The writer then calls `Dispose()` to signal it's done, but there's still a pending `ReadAsync()`. Okay, we can trigger the read to complete with a partially filled buffer (512 bytes), but the reader has no way of knowing that there's no more data. It could have been that a timer expired and wrote what it had so far. So the reader will call `ReadAsync()` again. If this throws an `ObjectDisposedException`, that could abort whatever it is that the reader is trying to do. I could add an extra state variable that after `Dispose()` is called, it allows one more final `Read` call which returns 0 bytes, after which it will throw `ObjectDisposedException`. What do you think? As it stands, further calls to `Write()` will throw `ObjectDisposedException` and further calls to `Read` will return immediately with 0, stating we're at the end of the stream. As no real resources are actually held by the class, I feel it's okay to have this half-disposed nature.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: That is actually a very difficult thing to get right. If the read cancellation token is cancelled, there needs to be a mechanism to prevent any more writing into the current pending buffer. So I would need to track when a write begins and ends so that the read doesn't throw from the cancellation while the buffer continues to be written to. Then there needs to be a mechanism to replace the current buffer with WriteBufferWrapper.EmptyContainer in a thread safe cache coherent manner so that another incoming write doesn't trample over the cancelling read.  We also have the problem of a read or a write could be asynchronously waiting for the corresponding opposite operation to happen which never comes so we need a way to cancel the await on the passed in cancellation token. Checking the code which will use this, it doesn't look like this will ever be cancelled from the read side as this code path will only be used on .Net Native.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: I've modified all the properties to be getters only. No code was using the set code path and the only getter that's being used is Count. It feels wrong to only expose count though, so I'm leaving all three getters even though only one of them is actually used.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: This equals code is on the hot code path so it's slightly different than you would normally see, but it is correct. There's some fast paths through the code to minimize the number of compares.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: This isn't a public type so the only way this would get hit would be a code bug. Someone using WCF shouldn't be able to this code to run. Do you think I should change this to a Contract.Assert instead?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
iamjasonp: I think we can use "System.Net.Http.Headers.HttpRequestHeaders" - might be safer as Native is known for being weird if it can't resolve metadata

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
iamjasonp: If this is the case, would we be in a situation where the HEAD would be rejected, the POST would work, and we'd have to re-POST? Would that just be handled?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelHelpers.cs
iamjasonp: Thanks for the explanation!

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
iamjasonp: I'm fine with that then if the general presumption is that `message.Headers` is non-null. We might want to add an Assert, but I'll leave that up to you.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
iamjasonp: I think leaving all the getters is fine.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
iamjasonp: Yep, makes sense, it was just a little hard to read, that's all.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
iamjasonp: ACK.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
roncain: Makes sense (given the odd conflation of close+dispose we have to deal with). Do these objects ever get into a state where they really do need to throw ObjectDisposedException due to having discarded other data structures?  Just wondering if we need to distinguish between Disposed-but-you-can-interact-with-it versus Disposed-and-you-can't-use-it-anymore-for-any-reason.  Don't block this PR answering this question.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
roncain: It isn't a public type, but is it reachable from public API's? (Like someone calls some HttpContent public API that ultimately ends up calling this method).  If the answer is 'no', I'd favor Contract.Assert.  BTW, we have an ObjectIsReadOnly resource string if you decide the exception should remain.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
roncain: I'm also fine with a more generic error message that only says an error occurred copying to the request headers. I'm not sure how valuable the type name of the headers is to the user. The alternative is not using the TryAddWithoutValidation and capturing whatever exception gets thrown by the Add.  No matter what happens, we will throw for an invalid header here.

roncain:
I am ready to approve subject only to the responses to the open questions.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
SajayAntony: `_buffer = default(ArraySegment<byte>)` 
It would be good if the buffer manager zero'd out the buffer before handing it out.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: I like having a full exception here as if a bug happens around this code, it will be very hard to work out what happened after the fact. I'll use the ObjectIsReadOnly resource string.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: @roncain, I like your idea of not using the Try variant. Right now there's no way to know what the problem was. It's possible the Add will give a more meaningful exception. We still need to wrap it in an InvalideOperationException though for consistent exception handling, and the resource string still requires a type name be provided.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
mconnew: _contentLength and _contentType already have the correct default values (0 and null) in the case that content==null. _contentLength==-1 means the content length wasn't specified, so presume chunked encoding.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: Done. The sync code paths will now wrap the async calls in Try/Catch(OperationCanceledException) and throw an ObjectDisposedException as the class is effectively disposed at that point as it's unusable any more.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: Fixed (not the zero-ing out, that would need careful designing to prevent a perf problem).

roncain:
LGTM

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -601,9 +620,10 @@ private void Cleanup()
                     if (_httpRequestMessage != null)
                     {
                         _timeoutHelper.CancelCancellationToken(false);
-                        this.TryCompleteHttpRequest(_httpRequestMessage);
+                        TryCompleteHttpRequest(_httpRequestMessage);
+                        _httpRequestMessage.Dispose();
+                        _httpRequestMessage = null;
                     }
```
iamjasonp: Consider snapshotting _httpRequestMessage prior to setting to null, e.g., 

```
if (_httpRequestMessage != null) 
{
    var httpRequestMessageSnapshot = _httpRequestMessage;
    _httpRequestMessage  = null; 
    // stuff
   _httpRequestMessage.Dispose(); 
}
```

iamjasonp:
Just checking: are you going to turn on the auth tests at the same time as committing this?

Otherwise, LGTM. 

 :shipit: says: Please squash the commits as you see appropriate though - at least get rid of the "changes based on PR" commits

mconnew:
@iamjasonp, I can't turn on the auth tests as we need service side support outside of a domain. In theory you can do digest authentication without using a domain, but without that being set up on the service side, we can't test this. Digest is the only authentication method on coreclr (which uses WinHttp) which can't do pre-authentication.

iamjasonp:
@mconnew okay, that makes sense. I was wondering if Basic could be turned on, but it can wait a little bit or be split out.

iamjasonp:
@dotnet-bot test this please

iamjasonp:
Tested this locally on my box with outerloop tests too - no breaks. Merging.

PR #106, opened 2015-06-19
Eliminate use of xunit Assert.IsAssignableFrom

roncain:
The NET Native version of xunit does not support
Assert.IsAssignableFrom, so this change removes that
usage and does explicit IsAssignable checks.


iamjasonp:
To clarify, the .NET Native build is now building from this set of sources, whereas prior to 18 June, the .NET Native sources were built off a set of forked sources. 

LGTM.

PR #107, opened 2015-06-19
Remove use of ServiceContractAttribute.SessionMode

roncain:
This property is not part of the public contract but was
being used inadvertantly by a unit test.  Removes that usage.


hongdai:
LGTM.

iamjasonp:
LGTM

PR #109, opened 2015-06-22
Remove Constants.cs and LocalConstants.cs it's no longer used

iamjasonp:
Sigh. 

Got pwned by Constants.cs pointing to a server that is internal. It appears that this set of servers is no longer used out in Git-land, so proposing getting rid of these files to save others the grief I faced while trying to why a test was not hitting the server as expected :cry:

Before we do get rid of them, we need to confirm that these server addresses are no longer used even in the builds for .NET Native, so please do not merge until this has been double checked.


roncain:
LGTM.
I built and ran all tests in this PR, and I exercised the new experimental code to build and test this in NET Native and CoreCLR using this PR patch.  All (expected) tests pass and there were not compile errors.

StephenBonikowsky:
LGTM

Issue #104, opened 2015-06-18
Create end-to-end Duplex scenario test


roncain:
Task is to update the standalone test service to expose a duplex service and to add basic end-to-end tests to verify it works properly.

iamjasonp:
I'm actually working on something like this to verify #90 / #91...

StephenBonikowsky:
@iamjasonp  I used what you started as the basis for the pull request I just made. Opened Issue111 for product bug.

StephenBonikowsky:
Also, there are a few little things in this PR that I really should clean-up before it gets merged but after you and I spending the last couple of days on this I wanted to get something out there before I left today that can be used to reproduce the product issue.

PR #112, opened 2015-06-23
Add Duplex callback service to WcfTestService (Issue #104)

StephenBonikowsky:


src/System.Private.ServiceModel/src/Extensions/ReflectionExtensions.cs
```diff
@@ -92,7 +92,7 @@ public static bool IsValueType(this Type type)
         }
         public static InterfaceMapping GetInterfaceMap(this Type type, Type interfaceType)
         {
-            throw ExceptionHelper.PlatformNotSupported();
+            return type.GetTypeInfo().GetRuntimeInterfaceMap(interfaceType);
```
hongdai: How is the product change related to the purpose of this checkin?

mconnew: This is a one line product fix needed to get the test to pass. For pragmatic reasons, it's worth making that one change rather than check in the test, have the test fail, open a new issue, then submit a separate pull request just for this one line.

iamjasonp: Yep - this is a change I made while working on the test+product earlier. 
This change allows us to look for the interface(s) implemented by a given type.

iamjasonp:
Wait, what the heck? I must have managed to send you a patch that wasn't up to my latest commit, as there are changes in there that I no longer have in my branch.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/Client.ClientBase.Tests.csproj
```diff
@@ -33,6 +33,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="ClientBaseTests.cs" />
+    <Compile Include="DuplexClientBaseTests.cs" />
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
 </Project>
```
roncain: Nit: that little red icon says you need to add a newline at the end of the file (even if it was already missing when someone else checked it in.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.ServiceModel.Channels;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+public static class DuplexClientBaseTests
+{
+    static TimeSpan s_WaitTimeout = TimeSpan.FromSeconds(10);
+
```
roncain: Nit: in future we may want this to have a more descriptive name ("MaxTestWaitTime?") and move it into the common scenarios folder.  I suspect we will use it again.  We can wait for the 2nd use before moving it.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+        finally
+        {
+            if (duplexService != null && duplexService.State != CommunicationState.Closed)
+            {
+                duplexService.Abort();
+            }
+        }
+
+        Assert.True(success, "Test timed out.");
```
roncain: I think it would be clearer to move this Assert up right after the WaitOne because "success" only means it didn't time out.  For example, if it does timeout, I'd like to see that assert fire, not the exception that might occur from the Close(). I might consider an Assert to fail if the finally requires an Abort.   That would be useful to know if that path ever executes. Not required -- your call.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
```diff
@@ -0,0 +1,69 @@
...
+                return _resetEvent;
+            }
+        }
+
+        public void OnPingCallback()
+        {
+            _resetEvent.Set();
+        }
+    }
+}
```
roncain: newline needed

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/Program.cs
```diff
@@ -12,14 +12,32 @@ private static void Main(string[] args)
...
+
+                DumpEndpoints(wcfservicehost, userNameService, wcfDuplexService);
+
                 Console.WriteLine("Press ENTER to close services...");
                 Console.ReadLine();
             }
         }
+
+         static void DumpEndpoints(params ServiceHost[] hosts)
+         {
```
roncain: Great idea

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/IWcfDuplexService.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.ServiceModel;
+
+namespace WcfService
+{
+    [ServiceContract (CallbackContract = typeof(IWcfDuplexServiceCallback))]
+    public interface IWcfDuplexService
+    {
+        [OperationContract]
+        void Ping();
+    }
```
roncain: Can we make Ping and OnPingCallback take a parameter, and let Ping echo it back to the callback? The test can ping a GUID and check it gets it back.  Matching up the parameters of the callback operations is an important path to test of the client layer.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/IWcfDuplexService.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.ServiceModel;
+
+namespace WcfService
+{
+    [ServiceContract (CallbackContract = typeof(IWcfDuplexServiceCallback))]
+    public interface IWcfDuplexService
```
roncain: FYI, I opened https://github.com/dotnet/wcf/issues/113 to have a shared file for these interfaces that both the tests and selfhost service can share.  We shouldn't have to maintain 2 copies of every interface.  No need to solve that issue with this PR.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/App.config
```diff
@@ -151,6 +154,15 @@
                     </baseAddresses>
                 </host>
             </service>
+            <service name="WcfService.WcfDuplexService">
+                <endpoint address="tcp-nosecurity-callback" binding="netTcpBinding" bindingConfiguration="nettcpBindingCallback"
+                          name="nettcp-defaultBinding" contract="WcfService.IWcfDuplexService" />
+                <host>
+                    <baseAddresses>
+                        <add baseAddress="net.tcp://*:810/WindowsCommunicationFoundation" />
```
iamjasonp: In the latest patch, there should be no need to set up a new baseAddress. I had merged it back into the same endpoint for simplicity (i.e., `*:809`).

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/IWcfDuplexService.cs
iamjasonp: :+1:

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/Program.cs
iamjasonp: :+1: on `DumpEndpoints`. 

Not important: Is there a way that `DumpEndpoints` can just enumerate through all ServiceHosts present on the system (or app domain) rather than have to pass this information in?

SajayAntony: @iamjasonp I don't think its possible without walking the heap or attaching something like an on open handler. I just did this to keep it simple and get an output of all the endpoints when the exe starts up.

mconnew:
This isn't the correct thing to review. A git rebase master dropped one of Stephen's commits. I've recovered it for him but it will be tomorrow before he can update this PR.

src/System.Private.ServiceModel/src/Extensions/ReflectionExtensions.cs
iamjasonp: To clarify: the test isn't passing for me right now, I need that one line change to get past the `PlatformNotSupportedException`.

SajayAntony:
:shipit:

roncain:
LGTM

PR #115, opened 2015-06-24
Bridge test changes

dmetzgar:
Test changes to use the bridge


src/System.ServiceModel.Tests.Common/src/OuterLoopTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            // Placeholder for releasing the resource
+        }
+
+        private void MakeConfigRequest()
+        {
+            using (HttpClient httpClient = new HttpClient())
+            {
+                httpClient.BaseAddress = new Uri(BaseAddress);
+                var content = new StringContent(
+                    @"{ resourcesDirectory : ""C:\\dotnet\\wcf\\src\\System.Private.ServiceModel\\tests\\Scenarios\\SelfHostWcfService\\bin\\Debug"" }",
```
iamjasonp: Make sure the resources directory isn't hard coded like this

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.csproj
```diff
@@ -5,12 +5,27 @@
...
+    <UpdateInterval>7</UpdateInterval>
+    <UpdateIntervalUnits>Days</UpdateIntervalUnits>
+    <UpdatePeriodically>false</UpdatePeriodically>
+    <UpdateRequired>false</UpdateRequired>
+    <MapFileExtensions>true</MapFileExtensions>
+    <ApplicationRevision>0</ApplicationRevision>
+    <ApplicationVersion>1.0.0.%2a</ApplicationVersion>
+    <IsWebBootstrapper>false</IsWebBootstrapper>
+    <UseApplicationTrust>false</UseApplicationTrust>
+    <BootstrapperEnabled>true</BootstrapperEnabled>
```
iamjasonp: What are these changes from?

dmetzgar: Interesting. VS must have added them. 

-----Original Message-----

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/ConfigBackup.xml
```diff
@@ -0,0 +1,176 @@
...
+        propagateActivity="true">
+        <listeners>
+          <add name="traceListener"
+            type="System.Diagnostics.XmlWriterTraceListener"
+            initializeData="ServerTraces.svclog" />
+        </listeners>
+      </source>
+    </sources>
+  </system.diagnostics>
+</configuration>
```
SajayAntony: Is this some kind of autogenerated file?

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/ConfigBackup.xml
```diff
@@ -0,0 +1,176 @@
+﻿<configuration>
+  <!-- 
+    This is a copy of the old app.config for reference.
+    My intention is to replace all the XML config with code.
```
SajayAntony: From an earlier conversations it was called out that having config was easier. If the others feel OK then I think it is fine. But being able to pull from config is definitely helpful in some cases, However given k doesn't have config support it might be ok to have the server also avoid config.

src/System.ServiceModel.Tests.Common/src/OuterLoopTests.cs
```diff
@@ -0,0 +1,89 @@
...
+        protected string EndpointAddress
+        {
+            get;
+            private set;
+        }
+
+        private string BaseAddress
+        {
+            // Would like to pull this address either from msbuild properties, env vars, or 
+            // configuration passed into xunit.
```
SajayAntony: This should come from the bridge.

dmetzgar: Are you talking about the BaseAddress? That's the address to the bridge. The EndpointAddress is coming from the bridge.

-----Original Message-----

SajayAntony: The property name is  :confounded: I would consider renaming it to indicate its the resource base address or bridge address or something like that.

src/System.ServiceModel.Tests.Common/src/OuterLoopTests.cs
SajayAntony: @roncain this along with the address for bridge should be passed in here somehow ..I guess you already have a work item for this.

roncain: I am currently working on issue #29 to make build-time parameters accessible to tests or the bridge.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
```diff
@@ -4,17 +4,20 @@
...
 using System.ServiceModel.Channels;
+using System.ServiceModel.Tests.Common;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using TestTypes;
 using Xunit;
 
-public static class Binding_Http_BasicHttpBindingTests
+public class Binding_Http_BasicHttpBindingTests : OuterLoopTests
```
roncain: I'm not sure I like the idea of using a base class for this.  I think we will multiple categories of tests that will want access to the bridge, and requiring a specific base class may interfere with other inheritance needs in the future.  Can we not just make a public Bridge type and let the tests interact with a singleton of it?

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
```diff
@@ -4,17 +4,20 @@
...
 {
+    public Binding_Http_BasicHttpBindingTests() : base("WcfService.TestResources.BasicHttpResource") { }
+
     [Fact]
     [OuterLoop]
-    public static void DefaultSettings_Echo_RoundTrips_String()
+    public void DefaultSettings_Echo_RoundTrips_String()
```
roncain: I presume the elimination of static here also reflects all the OuterLoopTests are instance-based now.  Is that strictly necessary? What are the pro's and con's of doing this versus interacting with a singleton Bridge instance?

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.csproj
```diff
@@ -44,27 +61,47 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
+    <Reference Include="WcfTestBridgeCommon">
+      <HintPath>..\..\..\..\..\..\bifrost\src\WcfTestBridgeCommon\bin\Debug\WcfTestBridgeCommon.dll</HintPath>
```
roncain: Will this ultimately become a P2P reference to a project inside the WCF repo?

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/ConfigBackup.xml
roncain: I don't entirely follow what is proposed here.  @dmetzgar can you describe why the config info is being shrunk and why the "backup" config file is being populated?

src/System.ServiceModel.Tests.Common/src/OuterLoopTests.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        protected string EndpointAddress
+        {
+            get;
+            private set;
+        }
+
+        private string BaseAddress
+        {
+            // Would like to pull this address either from msbuild properties, env vars, or
```
roncain: The MSBuild properties will be accessible via a public static singleton, probably in the common test project.

src/System.ServiceModel.Tests.Common/src/OuterLoopTests.cs
```diff
@@ -0,0 +1,89 @@
...
+                try
+                {
+                    var response = httpClient.PostAsync("/config/", content).Result;
+                    if (!response.IsSuccessStatusCode)
+                        throw new Exception("Unexpected status code: " + response.StatusCode);
+                }
+                catch (Exception exc)
+                {
+                    throw new Exception("Bridge is not running", exc);
+                }
```
roncain: If the Bridge becomes a singleton, it can record the first failure and fail each test that asks for a resource -- maybe something like Assert.True(Bridge.IsRunning, "The Bridge is not running");

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
SajayAntony: The other option would with a TestFixture which is the xunits way of initializing stuff and its cleaner since its an interface and so we can have multiple if needed -  E.g. https://github.com/SajayAntony/bifrost/blob/master/src/tests/Bridge.Tests/ResourceControllerTests.cs#L15

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/ConfigBackup.xml
dmetzgar: No. I'm just keeping it there as a reference since I haven't converted everything from config into code.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/ConfigBackup.xml
dmetzgar: Both Matt and Stephen expressed that they preferred creating the bindings in code rather than config. My intention is that there would be no recognition of a config file in the bridge. The backup xml file here is just my reference as I'm converting this over.

src/System.ServiceModel.Tests.Common/src/OuterLoopTests.cs
dmetzgar: Noted

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
dmetzgar: The fixture would be my preferred approach as well.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
dmetzgar: I think this goes back to the earlier comment about using a fixture.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.csproj
dmetzgar: Yes, I haven't determined where I want to put the bridge code yet.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/ConfigBackup.xml
iamjasonp: It sounds reasonable. The reason why we did this in config to begin with was so that it'd be easier to maintain with multiple people trying to modify a single test server. Now that the code is committed into the repo and the server is running on developer machines, I see few benefits of keeping server configuration in config.

src/System.ServiceModel.Tests.Common/src/OuterLoopTests.cs
iamjasonp: Agreed... property of `BaseAddress` and `EndpointAddress` :point_right: :person_frowning: :frowning: :sob: :cry: :crying_cat_face: :-1:

I was bitten really badly previously by having `Constants.SomeAddress` and `Endpoints.SomeAddress` look exactly the same in IntelliSense but point to different places.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
roncain: I'd like to push back a bit for 2 reasons:
1. One goal was for Bridge to not impact the tests themselves.  I thought our only point of contact with Bridge would be the existing (static) test helpers.  Taking over the base class for a set of tests is fairly impactful and something that needs some careful consideration. We may have other reasons to want a shared base class other than for the Bridge, so I'd hate to close off that option unless strictly necessary.
2.  I don't believe TestFixture is available to us in NET Native xunit. That version of xunit is severely pared back, so I predict that approach will not work when we run these tests in our NET Native testing environment.

I would like to hear what the benefits are of a base class or test fixture over a simple static singleton (BridgeClient?).

SajayAntony: Is there any other per test initialization strategy. One time initialization needs to happen somewhere. If these aren't available in native then that's an issue. The brigde would be initialized only one time but each test needs to Ping it. So if the Ping is the check then a static helper like GetOrCreate("servicename") might be all it takes.

roncain: We don't have a test initialization story yet.

For the Bridge initialization I was thinking of something along the lines of a statically available instance of some singleton bridge client, like:
```c#
public class BridgeClient {
    public static BridgeClient Instance {get{...}}
    private void ConfigureIfNeeded() {...}
    public string GetOrCreateService(string serviceName) {
       ConfigureIfNeeded();
       ... use HttpClient to interact with Bridge ...
}

public static class Endpoints {
    public static string HttpBaseAddress_Basic{
       return BridgeClient.Instance.GetOrCreateService("wcfservice") + "/Basic";
    }
}
```
Then the existing tests would just work without modification. The BridgeClient could just Assert that the bridge is running so when it isn't tests fail with a simple message like "Unable to contact the test bridge at URL xyz".

SajayAntony: I think I like your idea of having a class called `Endpoints`  and it could abstract abstract the parsing and handling of return Json etc and the callsite only gets back a valid URL.  

```csharp
 [OuterLoop]
{
   string variationDetails = "Client:: NetHttpBinding/DefaultValues\nServer:: NetHttpBinding/DefaultValues";

    ChannelFactory<IWcfService> factory = 
                          new ChannelFactory<IWcfService>(binding, new 
                                 EndpointAddress(Endpoints.HttpBaseAddress_NetHttp));
}
```
But constructing the URL using `"/Basic"` seems odd since the server returns a part of the URL and the client builds the rest. I would have expected the server to return an array of endpoints and you pick out one based on a name. This would avoid issues of hitting a wrong service a non-existent endpoint for the service. Find could be an extension to search the returned array type

```csharp
public static class Endpoints {

#region wcfservice endpoints 
    // Return the endpoint with matching names.
    // Find might throw an Invalid Endpoint name 
    public static EndpointAddress HttpBaseAddress_Basic{
          return GetEndpoint("wcfservice", "basic");
    }

private static Url GetEndpoint(string servicename, string endpointname)
{
       var endpoints = BridgeClient.GetOrCreateService(servicename);
       var endpoint = endpoints.FirstOrDefault(e => String.Compare(e.name,endpointname,true) == 0);
       if(endpoint != null){
           return new Url(endpoint.ListenUri);
       }

      throw new Exception("Endpoint not initialized.");
}
#endregion 
}
```

dmetzgar: If the class fixture is not available in .Net native I think there is still an initialization story with the constructor and dispose method, which is the preferred method for xUnit. At least running the tests with corerun that part works. I used a base class so that I don't have to add a constructor/dispose method to each test. But I have no problem doing that. I'll find out if the class fixture works at all.

SajayAntony: I think the point is the minimum about of change to the tests so the changing of base classes and making them non-static . 
I think for the initial change we should avoid having to change the callsites and maybe investigate these approaches after these light up and the rest of the team is unblocked.

dmetzgar: I don't think changing the test classes from static to instance and adding a constructor is a big change. It's also necessary for initialization. Let me put together another PR incorporating the suggestions I've received so far.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.cs
```diff
@@ -4,13 +4,14 @@
...
+using System.ServiceModel.Tests.Common;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using TestTypes;
 using Xunit;
 
-public static class CustomBindingTests
+public class CustomBindingTests : IClassFixture<BridgeTestFixture>
 {
```
roncain: I am fairly sure the xunit wrapper we use in ToF does not support IClassFixture. I would like to hear the reasons why all the tests need to modified to initialize like this, as opposed to the static BridgeClient approach.

dmetzgar:
@roncain @SajayAntony 

PR is updated with latest changes. There are a couple issues I'm looking into. One is that Newtonsoft.Json is copied to every test project because it doesn't get copied to the output folder otherwise. I don't see much harm in this except the modifications to the project.json.lock files inflate the size of the PR. Another is in how base addresses are used in some tests. There is an assumption that a ServiceHost is listening at a base address and will return an exception since the request URL does not include an endpoint. I think the fix is easy but need to verify.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/.nuget/NuGet.Config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <solution>
+    <add key="disableSourceControlIntegration" value="true" />
+  </solution>
+</configuration>
```
iamjasonp: NIT: Need newline. 
Question: do we really need to check in the Nuget.exe binary? Can we not piggyback off what's already in the repo to do this?

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.csproj
```diff
@@ -49,22 +50,58 @@
...
     <Compile Include="WcfService.cs" />
-    <Compile Include="WcfUserNameService.cs" />
   </ItemGroup>
   <ItemGroup>
     <None Include="App.config" />
   </ItemGroup>
+  <ItemGroup>
+    <None Include="ConfigBackup.xml">
+      <SubType>Designer</SubType>
+    </None>
```
iamjasonp: Given this is a backup of config file that's not used, is this necessary here?

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.csproj
```diff
@@ -49,22 +50,58 @@
...
+    <BootstrapperPackage Include=".NETFramework,Version=v4.5">
+      <Visible>False</Visible>
+      <ProductName>Microsoft .NET Framework 4.5 %28x86 and x64%29</ProductName>
+      <Install>true</Install>
+    </BootstrapperPackage>
+    <BootstrapperPackage Include="Microsoft.Net.Framework.3.5.SP1">
+      <Visible>False</Visible>
+      <ProductName>.NET Framework 3.5 SP1</ProductName>
+      <Install>false</Install>
+    </BootstrapperPackage>
```
iamjasonp: 3.5 needed?

src/System.Private.ServiceModel/tests/System.Private.ServiceModel.Tests.csproj
```diff
@@ -14,7 +14,7 @@
     <ProjectGuid>{BB3B114D-68D7-4E36-9EBE-127E694E6EC9}</ProjectGuid>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="**\*.cs" Exclude="Scenarios\SelfHostWcfService\*" />
+    <Compile Include="**\*.cs" Exclude="Scenarios\SelfHostWcfService\**;Scenarios\Bridge\**" />
```
iamjasonp: Small merge conflict ahead: you may want to coordinate on this; https://github.com/dotnet/wcf/pull/120#discussion-diff-33398115R17 removes this in favour of excluding the entire Scenarios\ directory.

iamjasonp:
> Another is in how base addresses are used in some tests. There is an assumption that a ServiceHost is listening at a base address and will return an exception since the request URL does not include an endpoint. I think the fix is easy but need to verify.

Let's identify the tests and just fix them. I think the "listening at base address" issue is an artifact from back when we started with one endpoint... 

We should make the endpoint fixes so it's easier to understand what the test needs.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/.nuget/NuGet.Config
roncain: We should not check-in any binaries. Let's find a way to restore NuGet.exe and packages.  Note: we definitely need to try building this in the TFS mirror.  There is an explicit pass to restore build tools, and we need to ensure any other restoration logic triggers prior to doing the actual build.

src/System.ServiceModel.Tests.Common/src/BaseAddress.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.ServiceModel.Tests.Common;
 
 public static class BaseAddress
 {
```
roncain: Nit: BaseAddress is currently used by unit tests too, and we don't want the Bridge to be required to run unit tests. No actual client to server communication occurs over the raw BaseAddress URLs.  Endpoints.cs was meant to be used by scenario tests that needed a running server.  It effectively layered over BaseAddress.  I think we could just revert BaseAddress to its old form now that Endpoints does not depend on it anymore (and let the unit tests run without the Bridge.

src/System.ServiceModel.Tests.Common/src/BridgeTestFixture.cs
```diff
@@ -0,0 +1,148 @@
...
+                httpClient.BaseAddress = new Uri(BridgeBaseAddress);
+                var content = new StringContent(
+                        string.Format(@"{{ name : ""{0}"" }}", resourceName),
+                        Encoding.UTF8,
+                        "application/json");
+                try
+                {
+                    var response = httpClient.PutAsync("/resource/", content).Result;
+                    if (!response.IsSuccessStatusCode)
+                        throw new Exception("Unexpected status code: " + response.StatusCode);
```
roncain: I think I see that the Bridge will return a 500 with a more informative body that described the problem. It might be useful to deserialize any response body and put it in the exception message.
Not required for this PR.

src/System.ServiceModel.Tests.Common/src/System.ServiceModel.Tests.Common.csproj
```diff
@@ -14,9 +14,6 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <ProjectGuid>{E896294A-AB4A-4AF5-A01C-A19E3972EFF9}</ProjectGuid>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(UseFiddler)' == 'true'">
```
roncain: Note: (no action required) I opened #129 to reenable the Fiddler option using TestProperties

src/System.ServiceModel.Tests.Common/src/BridgeTestFixture.cs
```diff
@@ -0,0 +1,148 @@
...
+    public static class BridgeTestFixture 
+    {
+        private static Dictionary<string, string> _Resources = new Dictionary<string, string>();
+        private static BridgeState _BridgeStatus = BridgeState.NotStarted;
+        private static Dictionary<string, string> _BaseAddresses = null;
+
+        private static string BridgeBaseAddress
+        {
+            // TODO: Pull this address from msbuild props, env vars, or config passed into xunit.
+            get { return "http://localhost:44283"; }
```
roncain: Go ahead and leave this hard-coded -- I'll make the change to use TestProperties after you check-in

PR #116, opened 2015-06-24
Fix build warnings

roncain:
Build warnings were produced due to fields not being CLS-compliant
and Equals being overridden without overriding GetHashcode.
These warnings caused the build to fail when building the
TFS mirrored sources with warnings treated as errors.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
@@ -15,9 +15,11 @@ namespace System.ServiceModel.Channels
 {
     public abstract class MessageContent : HttpContent
     {
+#pragma warning disable 3008	// not CLS-compliant
         protected Message _message;
         protected MessageEncoder _messageEncoder;
         protected Stream _stream = null;
+#pragma warning restore 3008	// not CLS-compliant
         private bool _disposed;
 
         public MessageContent(Message message, MessageEncoder messageEncoder)
@@ -75,7 +77,6 @@ namespace System.ServiceModel.Channels
                     }
                     else if (string.Compare(name, "content-type", StringComparison.OrdinalIgnoreCase) == 0)
                     {
-                        MediaTypeHeaderValue contentType;
                         if (SetContentType(value))
                         {
                             wasContentTypeSet = true;
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
@@ -190,9 +190,9 @@ namespace System.ServiceModel.Channels
             }
         }
 
-#pragma warning disable 660,661  // Hashcode not needed, Equals overriden for fast path comparison of EmptyContainer
+#pragma warning disable 659,660,661  // Hashcode not needed, Equals overriden for fast path comparison of EmptyContainer
         private struct WriteBufferWrapper : IEquatable<WriteBufferWrapper>
-#pragma warning restore 660,661
+#pragma warning restore 659,660,661
         {
             public static readonly WriteBufferWrapper EmptyContainer = new WriteBufferWrapper(null, -1, -1, true);
 


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -190,9 +190,9 @@ private static void CancelAndDispose(object state)
             }
         }
 
-#pragma warning disable 660,661  // Hashcode not needed, Equals overriden for fast path comparison of EmptyContainer
+#pragma warning disable 659,660,661  // Hashcode not needed, Equals overriden for fast path comparison of EmptyContainer
         private struct WriteBufferWrapper : IEquatable<WriteBufferWrapper>
-#pragma warning restore 660,661
+#pragma warning restore 659,660,661
```
iamjasonp: Question only: 

According to MSDN: 
> CS0659:  'class' overrides Object.Equals(object o) but does not override Object.GetHashCode()
> CS0660: 'class' defines operator == or operator != but does not override Object.Equals(object o)
> CS0661: 'class' defines operator == or operator != but does not override Object.GetHashCode()

Seems to me that CS0660 is no longer needed here? 
Also, what's the effect of the `pragma warning disable 659` on Line 228? Is it needed in both the class definition and the method?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
```diff
@@ -15,9 +15,11 @@ namespace System.ServiceModel.Channels
 {
     public abstract class MessageContent : HttpContent
     {
+#pragma warning disable 3008	// not CLS-compliant
         protected Message _message;
         protected MessageEncoder _messageEncoder;
         protected Stream _stream = null;
+#pragma warning restore 3008	// not CLS-compliant
```
iamjasonp: This solves our short-term problem - will this present a longer-term problem if we don't fix these compliance issues? e.g., will there be an issue if WCF is compiled on other platforms?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
roncain: @mconnew , could you comment on these observations.  These are changes to files you added.

iamjasonp:
I'm currently validating this change. 
To be clear, the comments above are *not blocking*, but something we may want to discuss at a later time.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: The compliance issue is that protected fields can't start with an underscore. What's the naming standard for protected fields? I would rather change the field names rather then disable the warning.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: It certainly looks like 660 isn't needed as 659 implies Object.Equals(object o) was overridden. I agree with Jason that it should just be 659 and 661.

iamjasonp:
LGTM for fixing the break. 
We may want to consider looking at the issues commented on after the immediate fire is taken out.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: If they are left as not CLS compliant, this simply means that those fields can't be accessed by code written in some other languages which only support CLS compliant assemblies. As this is only public to enable writing unit tests and isn't exposed on the contract, in practice this non-CLS compliance doesn't negatively affect any scenario.

roncain:
Accepting the PR as is to unbreak the build.  Opening a new issue to deal with these comments in a followup PR.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: I just re-read the comment by @iamjasonp and I'm confused why the disable on 228 didn't work. That was what Visual Studio complained about and VS created the quick fix for me. I wonder if we are running into problems of different compilers requiring the warning to be disabled in different places. I'm okay with having it all at the class level as it's not a large class.

Issue #75, opened 2015-06-08
Reorganize unit tests


roncain:
The unit tests need to be organized slightly.  At a high level I recommend we adopt these conventions:
 * (Like today) keep unique folders corresponding to each contract (e.g. src\S.SM.Http\tests)
 * (Like today) the tests in these contract-specific folders will test only types that belong to that contract.
 * (New) these contract-specific tests will test *all* public members of the types in the contract, even if those members themselves are not part of the contract.
 * (Like today) keep a single folder under S.Private.SM/tests for all tests for types not in any contract
 * (Like today) folders under the unit tests will follow the normal namespace/folder naming conventions, and each test class will test exactly one type (by appending the word "Test" to the name of the type).

Changes needed to get there:
 * System.ServiceModel.Tests.Common should move from root down into tests/common folder
 * Any existing tests under S.Private.SM for types belonging to a contract should move to their respective contract-specific folder
 * We should create a new folder called "unit" parallel to "Scenarios" and move all (not in contract) unit tests currently under S.Private.SM\tests to be under "unit" instead

When this refactor is complete, "src\System.Private.ServiceModel\tests" will contain these 3 folders:
 * unit
 * scenarios
 * common

And "src" will contain these folders:
 * System.Private.ServiceModel\tests
 * System.ServiceModel.Primitives\tests
 * System.ServiceModel.Http\tests
 * System.ServiceModel.NetTcp\tests
 * System.ServiceModel.Security\tests
 * System.ServiceModel.Duplex\tests

StephenBonikowsky:
Issue completed with PR #141, closing.

PR #120, opened 2015-06-25
Adds TestProperties class to capture MSBuild properties

### LINKED ISSUE #29: Provide a mechanism to pass command line parameters into tests

roncain:
A new TestProperties class is added together with a custom
task to generate a C# file to populate a dictionary with
MSBuild properties defined at build time.

This makes all the build-time MSBuild properties available
to tests at runtime.  This is necessary because the frameworks
on which these tests run do not support configuration files
or environment variables.

Fixes #29

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/TestPropertiesTest.cs
+++ b/src/System.Private.ServiceModel/tests/Common/TestPropertiesTest.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.ServiceModel.Tests.Common;
+using Xunit;
+
+public static class TestPropertiesTest
+{
+    [Fact]
+    public static void TestProperties_Property_Names_Are_Initialized()
+    {
+        // Test property names are auto-generated.
+        // This test will fail to compile if these names are not generated.
+        Assert.NotNull(TestProperties.BridgeResourceFolder_PropertyName);
+        Assert.NotNull(TestProperties.BridgeUrl_PropertyName);
+        Assert.NotNull(TestProperties.UseFiddlerUrl_PropertyName);
+    }
+
+    [Fact]
+    public static void TestProperties_PropertyNames_Property_Returns_List()
+    {
+        IEnumerable<string> propertyNames = TestProperties.PropertyNames;
+        Assert.NotNull(propertyNames);
+        Assert.True(propertyNames.Any());
+    }
+
+    [Fact]
+    public static void TestProperties_All_Properties_Have_Values()
+    {
+        IEnumerable<string> propertyNames = TestProperties.PropertyNames;
+        foreach(var name in propertyNames)
+        {
+            string value = TestProperties.GetProperty(name);
+            Assert.True(value != null, String.Format("Property '{0}' should not be null.", name));
+        }
+
+    }
+
+    [Fact]
+    public static void TestProperties_Throw_ArgumentNullException()
+    {
+        Assert.Throws<ArgumentNullException>(() => TestProperties.GetProperty(null));
+    }
+
+    [Fact]
+    public static void TestProperties_Throw_KeyNotFoundException()
+    {
+        Assert.Throws<KeyNotFoundException>(() => TestProperties.GetProperty("NotAProperty"));
+    }
+}
--- a/src/System.Private.ServiceModel/tests/System.Private.ServiceModel.Tests.csproj
+++ b/src/System.Private.ServiceModel/tests/System.Private.ServiceModel.Tests.csproj
@@ -14,7 +14,7 @@
     <ProjectGuid>{BB3B114D-68D7-4E36-9EBE-127E694E6EC9}</ProjectGuid>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="**\*.cs" Exclude="Scenarios\SelfHostWcfService\*" />
+    <Compile Include="**\*.cs" Exclude="Scenarios\**\*" />
   </ItemGroup>
   <ItemGroup>
     <None Include="project.json" />
--- a/src/System.ServiceModel.Tests.Common/src/testproperties.targets
+++ b/src/System.ServiceModel.Tests.Common/src/testproperties.targets
@@ -1,0 +1,135 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- 
+    CreateTestPropertiesFile:
+      This task generates a C# file to initialize the TestProperties.Properties
+      dictionary with the provided name/value pairs.
+      FileName :          the full name of the file to generate
+      TestPropertyItems : item collection of name/value pairs to generate
+                          name = item.ItemSpec
+                          value = item's metadata for "Value" or "DefaultValue" if "Value" is blank
+  -->
+  <UsingTask TaskName="CreateTestPropertiesFile" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v12.0.dll">
+    <ParameterGroup>
+      <FileName ParameterType="System.String" Required="true" />
+      <TestPropertyItems ParameterType="Microsoft.Build.Framework.ITaskItem[]" Required="true" />
+    </ParameterGroup>
+    <Task>
+      <Reference Include="System" />
+      <Reference Include="System.IO" />
+      <Code Type="Fragment" Language="cs">
+        <![CDATA[
+        
+        // In some build environments, paths such as $(OutputPath) are
+        // not created prior to starting the build.  So we may find
+        // we need to create the folder for the generated file.
+        string outputFolder = Path.GetDirectoryName(FileName);
+        if (!Directory.Exists(outputFolder))
+        {
+            Log.LogMessage("Creating output folder {0}", outputFolder);
+            Directory.CreateDirectory(outputFolder);
+        }
+
+        StringBuilder sb = new StringBuilder();
+	      sb.AppendLine(
+ @"using System.Collections.Generic;
+namespace System.ServiceModel.Tests.Common
+{
+    public static partial class TestProperties
+    {");
+        
+        // Generate a string for each property name.
+        // Code accessing the properties should use these generated names.
+        foreach (var item in TestPropertyItems)
+        {
+            sb.AppendLine(String.Format("        public static readonly string {0}_PropertyName = \"{0}\";", item.ItemSpec));
+        }
+        
+        // Generate the Initialize partial method to populate the dictionary
+        sb.AppendLine(
+@"        static partial void Initialize(Dictionary<string, string> properties)
+        {");
+                
+        foreach (var item in TestPropertyItems)
+        {
+            string propertyName = item.ItemSpec;
+            string propertyValue = item.GetMetadata("Value");
+            if (String.IsNullOrEmpty(propertyValue))
+            {
+                propertyValue = item.GetMetadata("DefaultValue");
+            }
+         
+            // Convert the property value into a legal string literal.     
+            StringBuilder literal = new StringBuilder();       
+            foreach (var c in propertyValue)
+            {
+                switch (c)
+                {
+                    case '\'': literal.Append(@"\'"); break;
+                    case '\"': literal.Append("\\\""); break;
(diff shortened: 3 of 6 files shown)


src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,215 @@
...
+      <PrefixNames ParameterType="System.String" Required="true" />
+    </ParameterGroup>
+    <Task>
+      <Reference Include="System" />
+      <Reference Include="System.IO" />
+      <Using Namespace="System.Reflection" />
+      <Code Type="Fragment" Language="cs">
+        <![CDATA[
+        
+        // Project properties are available via Microsoft.Build.Execution.ProjectInstance.Properties.
```
roncain: Reviewers: the use of Reflection to obtain the properties appears to be the only known way to do it.  The algorithm here shows up on StackOverflow as a known way to accomplish this.

src/System.Private.ServiceModel/tests/System.Private.ServiceModel.Tests.csproj
```diff
@@ -14,7 +14,7 @@
     <ProjectGuid>{BB3B114D-68D7-4E36-9EBE-127E694E6EC9}</ProjectGuid>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="**\*.cs" Exclude="Scenarios\SelfHostWcfService\*" />
+    <Compile Include="**\*.cs" Exclude="Scenarios\**\*" />
```
mconnew: What does it mean to exclude "Scenarios\**\*"? It looks to me like it means to exclude any files which don't have an extension. As the include only pulls in things with a .cs extension, it would seem to me that this exclude would be a no-op?

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+      <FileName ParameterType="System.String" Required="true" />
+      <TestPropertyItems ParameterType="Microsoft.Build.Framework.ITaskItem[]" Required="true" />
+    </ParameterGroup>
+    <Task>
+      <Reference Include="System" />
+      <Reference Include="System.IO" />
+      <Code Type="Fragment" Language="cs">
+        <![CDATA[
+        
+        // In some build environments, paths such as $(OutputPath) are
```
mconnew: Is there a standard msbuild task which will do this?

src/System.Private.ServiceModel/tests/System.Private.ServiceModel.Tests.csproj
roncain: That include was pulling all the scenario tests into the unit tests for S.P.SM -- it's been a nuisance bug I finally decided to fix.  Before this change, an F5 on this unit test project attempts to run all the scenario tests too, and they fail.  After the change, F5 runs only the unit tests under this folder.  Issue #75 is responsible for refactoring the folder structure to separate the unit tests from the scenario tests better.

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+        // we need to create the folder for the generated file.
+        string outputFolder = Path.GetDirectoryName(FileName);
+        if (!Directory.Exists(outputFolder))
+        {
+            Log.LogMessage("Creating output folder {0}", outputFolder);
+            Directory.CreateDirectory(outputFolder);
+        }
+
+        StringBuilder sb = new StringBuilder();
+	      sb.AppendLine("using System.Collections.Generic;");
```
mconnew: A cleaner way to do multiline append is this:
```csharp
    sb.AppendLine(
@"using System.Collections.Generic;
namespace System.ServiceModel.Tests.Common
{
    public partial class TestProperties
    {");
```

src/System.ServiceModel.Tests.Common/src/testproperties.targets
roncain: Could you elaborate on what "this" is?  I'm not aware of msbuild tasks that do everything in this custom task.

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+	      sb.AppendLine("namespace System.ServiceModel.Tests.Common");
+	      sb.AppendLine("{");
+	      sb.AppendLine("    public partial class TestProperties");
+	      sb.AppendLine("    {");
+        
+        // Generate a string for each property name.
+        // Code accessing the properties should use these generated names.
+        foreach (var item in TestPropertyItems)
+        {
+            sb.AppendLine(String.Format("        public static readonly string {0}_PropertyName = \"{0}\";", item.ItemSpec));
```
mconnew: Is there any validation needed to make sure item.ItemSpec only contains valid keyword values? Or are we going to rely on the failure when compiling the generated code to catch this?

roncain: Currently, we control exactly the contents of the ItemSpec value (done later in this same file).  The value needs to be scrubbed because it can come in from arbitrary MSBuild properties.  So I don't think validation is required.

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+    <!-- 
+      This item collection gives the name and value for each property to generate.
+      New items can be added here or in any other project or targets file.
+      The Value metadata contains the value to use, and DefaultValue contains
+      the value to use if Value is empty.  This permits properties to be
+      specified externally to override the defaults here.
+      Example: 'msbuild /p:UseFiddlerUrl=true' will override the default false.
+    -->
+    <ItemGroup>
+      <TestProperty Include="BridgeResourceFolder">
```
roncain: Every property here is just a suggestion.  We could add, change or remove any of these.  I expect the list to grow.

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+    <!-- Generate the file (will be a NOP if file exists and doesn't need to change) -->
+    <CreateTestPropertiesFile FileName ="$(GeneratedTestPropertiesName)" TestPropertyItems="@(TestProperty)"/>
+    
+    <!-- Add the generated file to the @(Compile) collection so it gets built. -->
+    <ItemGroup>
+      <Compile Include="$(GeneratedTestPropertiesName)" />
+    </ItemGroup>
+
+  </Target>
+</Project>
```
roncain: Incidentally, this PR has been tested under ToF and against NET Native.

src/System.Private.ServiceModel/tests/Common/TestPropertiesTest.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.ServiceModel.Tests.Common;
+using Xunit;
+
+public static class TestPropertiesTest
+{
+    [Fact]
+    public static void TestProperties_Are_Initialized()
+    {
+        Assert.True(TestProperties.Instance.Properties.ContainsKey(TestProperties.BridgeResourceFolder_PropertyName),
```
roncain: This unit test is necessary to demonstrate both that the code-gen happens and also to show what usage will look like by test code.  The actual values being tested does not have to be exhaustive.

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+                propertyValue = item.GetMetadata("DefaultValue");
+            }
+         
+            // Convert the property value into a legal string literal.     
+            StringBuilder literal = new StringBuilder();       
+            foreach (var c in propertyValue)
+            {
+                switch (c)
+                {
+                    case '\'': literal.Append(@"\'"); break;
```
mconnew: Most of this initial items are unprintable characters with a value < 0x20. Simply removing them from this list will cause it to be emitted with the Unicode value. E.g. \f would be encoded to \u000f and simplify the code. Less code is generally better :)

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+                        if (c >= 0x20 && c <= 0x7e)
+                        {
+                            // ASCII printable character
+                            literal.Append(c);
+                        }
+                        else
+                        {
+                            // As UTF16 escaped character
+                            literal.Append(@"\u");
+                            literal.Append(((int)c).ToString("x4"));
```
mconnew: Integer is being formatted anyway, so could reduce these two lines to:
```csharp
literal.AppendFormat(@"\u{0:x4}", (int)c);
```

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+                        }
+                        break;
+                }
+            }
+            
+            string setStatement = String.Format("            properties[\"{0}\"] = \"{1}\";", propertyName, literal.ToString());
+            sb.AppendLine(setStatement);
+        }
+            
+	      sb.AppendLine("        }");
```
mconnew: Multi-line single append using @

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+	      sb.AppendLine("    }");
+	      sb.AppendLine("}");
+        
+        // Do not rewrite file if there are no changes.  This permits incremental builds.
```
mconnew: Put the test properties in a separate .props file then have the CreateTestProperties target have the .props file as an input and the generated filename as the output. The change check will happen automatically based on timestamp comparison between input and output and the target won't even run if the output is up to date.

src/System.ServiceModel.Tests.Common/src/testproperties.targets
```diff
@@ -0,0 +1,167 @@
...
+  <Target Name="CreateTestProperties">
+    <!-- 
+      This item collection gives the name and value for each property to generate.
+      New items can be added here or in any other project or targets file.
+      The Value metadata contains the value to use, and DefaultValue contains
+      the value to use if Value is empty.  This permits properties to be
+      specified externally to override the defaults here.
+      Example: 'msbuild /p:UseFiddlerUrl=true' will override the default false.
+    -->
+    <ItemGroup>
```
mconnew: These should probably be in a separate .props file.

src/System.ServiceModel.Tests.Common/src/testproperties.targets
roncain: Makes sense.  I left the ones most likely to appear in real use, just for readability in the generated file and MSBuild log

src/System.ServiceModel.Tests.Common/src/testproperties.targets
roncain: I did move the property item collection and property name out to a separate .props file so that the .targets file contains only the Task.  But I don't believe it is possible to make MsBuild understand when the output of the CreateTestProperties target would be up to date.  Its inputs come from both the .props file but also from MSBuild properties.  If I set the environment variable UseFiddlerUrl to true and rebuilt, it would be necessary to regenerate the file.

SajayAntony:
I'm curious why we don't fall back to EnvironmentVariables.  - 

https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Environment.cs#L601
I also see CoreFx having a tests here - 
https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/tests/System/Environment.GetEnvironmentVariable.cs#L99

src/System.Private.ServiceModel/tests/System.Private.ServiceModel.Tests.csproj
iamjasonp: +1. This issue was causing a lot of grief, and I had removed it from any local repo I had, but didn't submit it for PR. Also helps with load performance.

roncain:
After sign-offs, I would like to do a final squash before merging.

SajayAntony:
:shipit:

PR #121, opened 2015-06-26
Add test coverage for Async APIs with NetTcpBinding

### LINKED ISSUE #97: Need scenario tests for Async service calls for NetTcpBinding
We currently only have Async coverage for http bindings.

hongdai:
* None of the tests are passing because of product issue 78

Fixes #97

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/TypedClient/TypedProxyTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/TypedClient/TypedProxyTests.cs
@@ -21,103 +21,44 @@ public static class TypedProxyTests
 
     private const string action = "http://tempuri.org/IWcfService/MessageRequestReply";
     private const string clientMessage = "[client] This is my request.";
-    static TimeSpan maxTestWaitTime = TimeSpan.FromSeconds(10);
 
     [Fact]
     [OuterLoop]
     public static void ServiceContract_TypedProxy_AsyncBeginEnd_Call()
     {
-        // This test verifies a typed proxy can call a service operation asynchronously using Begin/End
-        StringBuilder errorBuilder = new StringBuilder();
+        CustomBinding customBinding = new CustomBinding();
+        customBinding.Elements.Add(new TextMessageEncodingBindingElement());
+        customBinding.Elements.Add(new HttpTransportBindingElement());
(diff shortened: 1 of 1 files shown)


hongdai:
My opinion is using helpers is fine if the helpers are in the same file and it significantly reduced the number of duplicated code. I see they are still self-contained if within the same file.

roncain:
@iamjasonp , @mconnew , @StephenBonikowsky, what are your thoughts on the user of helpers vs duplicating code?  I favor the style where we don't use helpers, and you can tell what a test does without navigating to some other method.  I thought we already did some work to break up existing helpers. It does lead to code duplication but I've found it to be easier to understand and maintain.

StephenBonikowsky:
That is how I have been writing all the previous tests. I don't believe duplicated code is a big deal, and I also prefer not to navigate away from the test case to some other method. In our past WCF test infrastructure this was normal but in our new OSS context I think each test should be as self-contained as possible.

As a separate point, this is how we have been writing the tests and even though going forward tests will be authored by all of us shouldn't we be trying to keep a consistent style and pattern?

SajayAntony:
:shipit:  That said this is how we might be able to continue organizing while aligning with the samples. 

I see there are 2 pivots here - Binding+InvocationPattern
I would expect us to group by binding ie transport, since that selection is critical in a networking stack.

Binding | Invocation
---------|-------------
Http | Sync
Http | Async+Begin
Http | Async+Task
NetTcp | Task
NetTcp | Async+Begin
NetTcp | Async+Task

Basically split the TypedClientTests.cs into `TypeClientHttpTests.cs` & `TypedClientTcpTests.cs` I would not be too against about duplicating them in their respective files since it would allow us to disable or even split Tcp/Http tests since I think we would also have duplex in NetTcp. 
These are thoughts for later and does not block this PR.

hongdai:
Since most people do not mind duplicating test code, I'll duplicate them in there respective files as Sajay suggested.

StephenBonikowsky:
:rocket:

iamjasonp:
A little late to the party, but I too would prefer (in OSS context) duping code rather than using helpers if at all possible. In general, we want our tests should be short/sweet enough anyway such that there shouldn't be a *huge* amount of code duplicated around. 

We might reconsider if there's a particularly gnarly set of tests, but that's for another day.

PR #122, opened 2015-06-26
Revert behaviour when adding a null/empty string to a http header 

iamjasonp:
This is a regression from commit ddcd6e4. 

App testing found that if we attempt to add a header with a null or empty string value to the _httpRequestMessage.Headers collection, an exception is thrown by
`System.Net.Http.Headers.HttpHeaderParser.ParseValue(...)`

which causes a failure in the app.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
@@ -794,18 +794,13 @@ namespace System.ServiceModel.Channels
                             }
                             else
                             {
-                                try
+                                if (!_httpRequestMessage.Headers.TryAddWithoutValidation(name, value))
                                 {
-                                    _httpRequestMessage.Headers.Add(name, value);
-                                }
-                                catch (Exception addHeaderException)
-                                {
-                                    throw FxTrace.Exception.AsError(new InvalidOperationException(SR.Format(
-                                                    SR.CopyHttpHeaderFailed,
-                                                    name,
-                                                    value,
-                                                    HttpChannelUtilities.HttpRequestHeadersTypeName),
-                                                    addHeaderException));
+                                        throw FxTrace.Exception.AsError(new InvalidOperationException(SR.Format(
+                                                SR.CopyHttpHeaderFailed,
+                                                name,
+                                                value,
+                                                _httpRequestMessage.Headers.GetType().Name)));
                                 }
                             }
                         }


iamjasonp:
@mconnew can you please take a look?

iamjasonp:
Basically - the behaviour reverts back to the way it was here: https://github.com/dotnet/wcf/pull/102/files#diff-7f0b8f789a5eff4c91c2a8a2753685ceL1328 

@mconnew, was there a reason why you chose to implement this in the way that's currently committed?

mconnew:
I'll add more details in a little bit, but the current code is correct, this is a test bug.

mconnew:
In my previous change, @roncain  noted the prior behavior and suggested not to use the Try variant and swallow any errors. I believe this is the correct behavior as if someone has explicitly added a header to be sent and there's a problem with that header, we shouldn't silently continue without sending it. This newer behavior makes it explicit that there was a problem with what they were requesting and it needs to be fixed as we can't send the message that they requested (which includes headers). I believe the benefit of this has already been shown as the app in question was trying to send an empty header which isn't possible.

iamjasonp:
I admit I didn't catch the previous note from @roncain since the PR was so big :smile: 

Is this behaviour accepted in desktop? If WCF on the desktop throws today, then it's clearly a test bug, but otherwise, we will need to consider the implications when someone builds the same app on desktop vs. CoreCLR/.NET Native. This would lead to a compat issue if desktop silently swallows today.

roncain:
I made the original suggestion to throw but agree we should do this only if desktop does something similar.  Whichever way the desktop behaves, we should make our code behave, and have a test that validates the behavior

mconnew:
We had a similar issue with an internal app when doing compatibility testing. The application was setting an invalid user agent. On desktop we just accepted the invalid user agent string. As the user agent string is a known header and now parsed by `System.Net.Http.Headers.ProductInfoHeaderValue`, an exception is thrown because of the invalid header. The question came down to whether we should provide stronger validation and fail now that we can, or accept any value for any header and send an invalid header across the network violating the relevant RFC for the HTTP protocol. The app compat group for .Net Native decided to accept a breaking change as this provides developers with better validation.  
This issue is the same. The header in question is the Authorization header which isn't allowed to be null. It's better to throw an exception than to allow an HTTP protocol violation to occur. Just because we did a worse job validating the provided request headers when using HttpWebRequest doesn't mean we shouldn't do a better job validating now that we can.

SajayAntony:
@mconnew do you want to merge or close this or are we waiting on @iamjasonp ?

iamjasonp:
Let's wait for the compatibility regression tests to complete before making a decision.

iamjasonp:
They're complete, and it appears there were no breaks with this change. With that, I'm reasonably convinced that we can the new throwing behavior is acceptable. Closing this PR.

PR #124, opened 2015-06-26
Enable building and debugging of test projects from visual studio

SajayAntony:
Once dotnet/buildtools#181 is fixed we should be able to remove this.


roncain:
LGTM

PR #131, opened 2015-07-01
Temporarily make package restore not run concurrently

roncain:
An issue with Dnu package restore running concurrently is preventing
builds from working correctly.  This change is temporary and allows
the build to succeed.  An issue will be opened to undo this change
when Dnu is fixed.


hongdai:
LGTM

PR #133, opened 2015-07-01
Bridge

dmetzgar:
Contains the bridge for facilitating outer loop testing. The bridge hosts the WcfService. There are some minor test changes needed in order to use the bridge and those are contained in this PR. The build.cmd has also changed to run the bridge instead of running the WcfService.exe directly.


src/test.props
```diff
@@ -8,5 +7,8 @@
 -->
   <PropertyGroup Condition=" $(MSBuildProjectName.EndsWith('.Tests')) AND '$(BuildingInsideVisualStudio)' == 'true' ">
     <OutputPath>$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), build.proj))\\bin\\tests\\Windows_NT.$(Platform).$(Configuration)\$(MSBuildProjectName)\dnxcore50</OutputPath>
-  </PropertyGroup>  
+  </PropertyGroup>
+  <PropertyGroup Condition="$(MSBuildProjectName.EndsWith('.Tests'))">
+    <XunitOptions>-parallel none</XunitOptions>
```
SajayAntony: Does this stop parallel execution for all tests?

src/System.ServiceModel.Tests.Common/src/BridgeTestFixture.cs
```diff
@@ -0,0 +1,148 @@
+﻿using Newtonsoft.Json;
+using System.Collections.Generic;
+using System.IO;
+using System.Net.Http;
+using System.Text;
+
+namespace System.ServiceModel.Tests.Common
+{
+    public static class BridgeTestFixture
```
SajayAntony: Would rename this class to BridgeClient to avoid confusion with TestFixtures.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/.nuget/NuGet.targets
```diff
@@ -0,0 +1,144 @@
...
+                }
+                catch (Exception ex) {
+                    Log.LogErrorFromException(ex);
+                    return false;
+                }
+            ]]>
+            </Code>
+        </Task>
+    </UsingTask>
+</Project>
```
SajayAntony: I think the exe would be an issue. Just update the targets to get File above directory.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/Bridge.sln
```diff
@@ -0,0 +1,54 @@
+﻿
```
SajayAntony: Would recommend moving bridge under tests/tools or something Scenarios/tools.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/.nuget/NuGet.targets
iamjasonp: :+1: - let's just use an upstream targets and/or nuget.config, otherwise it's just an additional place we have to update the build tools when we do that (and it's reasonably often)

src/System.Private.ServiceModel/tests/Scenarios/Bridge/Bridge/Bridge.csproj
```diff
@@ -0,0 +1,106 @@
...
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Bridge</RootNamespace>
+    <AssemblyName>Bridge</AssemblyName>
+    <TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
+    <OutputPath>..\bin\</OutputPath>
+    <RestorePackages>true</RestorePackages>
+    <PackageOutputDir>..\..\..\..\..\packages</PackageOutputDir>
```
iamjasonp: We've previously defined a PackagesDir in dir.props, could you use that?

src/System.Private.ServiceModel/tests/Scenarios/Bridge/Bridge/App.config
```diff
@@ -0,0 +1,26 @@
...
+        <assemblyIdentity name="System.Web.Http" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-5.2.3.0" newVersion="5.2.3.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.Owin" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-3.0.1.0" newVersion="3.0.1.0" />
+      </dependentAssembly>
+    </assemblyBinding>
+  </runtime>
+</configuration>
```
iamjasonp: NIT: A whole bunch of files don't have newlines at the end, please insert when you see this little symbol at the end of a file.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/Bridge/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,35 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Bridge")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("Bridge")]
+[assembly: AssemblyCopyright("Copyright ©  2015")]
```
iamjasonp: I think we've avoided including AssemblyInfo.cs files since the copyrights are... weird. At least could you remove thing that could resemble a change in licencing terms for the Bridge.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/WcfTestBridgeCommon/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("WcfTestBridgeCommon")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("Microsoft Corporation")]
+[assembly: AssemblyProduct("WcfTestBridgeCommon")]
+[assembly: AssemblyCopyright("Copyright © Microsoft Corporation 2015")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
```
iamjasonp: Ditto, probably best to get rid of anything copyright-related in this file. If you look at equivalent files in corefx, they are minimal, e.g.: https://github.com/dotnet/corefx/blob/fe33ff4bcbe4a3b4b47854388e68358a29bac93d/src/System.Dynamic.Runtime/src/Properties/AssemblyInfo.cs

src/System.Private.ServiceModel/tests/Scenarios/Bridge/Web/Models/ConfigurationExtensions.cs
```diff
@@ -0,0 +1,76 @@
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
```
iamjasonp: NIT: please sort+remove.

src/System.ServiceModel.Tests.Common/src/packages.config
```diff
@@ -1,7 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<packages>
-  <package id="xunit" version="2.0.0-beta5-build2785" />
-  <package id="xunit.abstractions.netcore" version="1.0.0-prerelease" />
-  <package id="xunit.assert" version="2.0.0-beta5-build2785" />
-  <package id="xunit.core.netcore" version="1.0.0-prerelease" />
-</packages>
```
iamjasonp: Huh. Interesting that this packages.config isn't referenced by the csproj.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.csproj
```diff
@@ -54,19 +55,51 @@
     <Compile Include="IWcfRestartService.cs" />
     <Compile Include="IWcfService.cs" />
     <Compile Include="MahojongTypes.cs" />
-    <Compile Include="Program.cs" />
```
iamjasonp: Intentional? I don't seem to see a removal of a Program.cs file

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.csproj
```diff
@@ -5,19 +5,19 @@
...
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
     <ProjectGuid>{CBE9909F-4CF7-4AD0-9307-BD0DFBB1105E}</ProjectGuid>
-    <OutputType>Exe</OutputType>
+    <OutputType>Library</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>WcfService</RootNamespace>
     <AssemblyName>WcfService</AssemblyName>
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
+    <OutputPath>..\Bridge\bin\WcfService\</OutputPath>
```
iamjasonp: It would be best if any build artifacts and build output were placed into the bin/ directory rather than anywhere under src/.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/Bridge/Bridge.csproj
roncain: Regarding OutputPath, we need a convention for where our tools binaries land so they are not inside the enlistment.  One possibility would be create a new targets file at the root (imported by override.targets) that sets some properties to define this This would also be an ideal place to set $(BridgeResourceFolder) so that we can have all resources build their outputs to the same known location where the Bridge looks.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/Bridge/Properties/AssemblyInfo.cs
roncain: The build targets generate AssemblyInfo's for the other projects.  I don't know what is required for Bridge to tap into those targets (look for versioning.targets after restoring build tools).  We can file an issue and revisit this after checkin

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.csproj
roncain: See earlier comment.  I think we need our own wcf.targets at the root included by override.targets that sets where all tools and resources output artifacts will go via properties.  Example, set $(WcfToolsOutputPath) to some specific place in bin.  Set $(BridgeResourceFolder) at the same time so that TestProperties automatically shares where the resource artifacts are put when they are built.  This would also allow a build to place the resource folder entirely different, like a public share.

src/test.props
dmetzgar: There is a bug either in xunit or dotnet core where running more than one test at a time causes a race condition. The threads will get stuck in the call to send the HTTP request to the bridge. Even when I put a lock around the methods so only one thread is making an HTTP request at a time, it will still get stuck indefinitely on the one thread trying to make the request. Perhaps an alternative is to only apply this option when running OuterLoop tests.

SajayAntony: Yes would special case this for outerloop only for performance reasons. 
I would like to check this in an open a bug to enable parallel execution.

src/System.Private.ServiceModel/tests/Scenarios/Bridge/.nuget/NuGet.targets
dmetzgar: I have to plead the fifth here. The only reason this file is here is because I want the solution to allow nuget package restore. I don't understand what you're suggesting.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/TestResources/EndpointResource.cs
```diff
@@ -0,0 +1,78 @@
...
+            ServiceHost host;
+            if (!currentHosts.TryGetValue(Address, out host))
+            {
+                lock (currentHostLock)
+                {
+                    if (!currentHosts.TryGetValue(Address, out host))
+                    {
+                        host = new ServiceHost(typeof(ServiceType));
+                        // URI assumes localhost but that instead should be passed in
+                        host.AddServiceEndpoint(
```
roncain: Agree should not be localhost.  Can this be the machine name where the resource itself is running?  That would allow the x-plat scenario where Bridge and service are running on a different machine.  Don't block this PR on this issue, we can come back and fix this later.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.csproj
dmetzgar: I guess it's not necessary to remove this in case you want to run the WcfService without the bridge. But since I've already changed it to a library instead of a console app I'll just remove it.

roncain:
I have successfully built and run the PR locally.  I think this is ready to prepare for checkin, and additional changes can happen afterwards. I recommend the next steps are:
 1. @dmetzgar - merge this branch with all the changes in master and create a new PR.
 2. me -- verify PR can build successfully when building against clean TFS state (verifies package restore of OSS projects)
 3. me -- verify these tests pass in the NET Native experimental test environment if I start Bridge manually.

iamjasonp:
+1. I have run this PR locally as well (from OSS only, not TFS), and I think this is ready. We can iron out issues later as we see them.

PR #135, opened 2015-07-02
Add override.targets and wcf.targets

roncain:
Adds override.targets to resemble how CoreFx includes multiple
product specific targets at build time.

Also adds wcf.targets that defines common properties and targets
necessary for wcf projects.


roncain:
@SajayAntony @dmetzgar, @iamjasonp @mconnew , this PR is meant to allow the Bridge PR #133 have environment variables for where to generate its output artifacts for both tools and resources. It follows the pattern used by CoreFx to include optional product specific targets

SajayAntony:
We should be able to merge override and wcf.targets into one. since we don't have multiple imports would reduce the number of targets to go after.

SajayAntony:
:shipit:

PR #136, opened 2015-07-02
Fix DuplexClientBaseTests use of Close()

roncain:
DuplexClientBaseTests used ClientBase<T>.Close() which is public
but not available in the public contract.  So the test fails when
run against NET Native.  This change allows it to function in both
CoreCLR and NET Native.


SajayAntony:
:ship:

Issue #139, opened 2015-07-06
Create test directory for test common code.


StephenBonikowsky:
Test common code is now located: ...\System.Private.ServiceModel\tests\Common\src

Need to create the directory: ...\System.Private.ServiceModel\tests\Common\test and put the file, "...\System.Private.ServiceModel\tests\Common\TestPropertiesTest.cs" in it, these are tests that validate certain test infrastructure features.

zhenlan:
Done by @StephenBonikowsky already.

Issue #140, opened 2015-07-06
Compartmentalize Scenario and Unit common code.


StephenBonikowsky:
Scenario tests should only require a reference to Scenario test common code.
Unit tests should only require a reference to Unit test common code.

StephenBonikowsky:
Fixed by #235

PR #141, opened 2015-07-06
Fixing Issue75

StephenBonikowsky:
The first commit is for adding the two project reference tokens to 'test.props' and updating .csproj files to use them. Unfortunately I missed a few .csproj files and had to update them in one of the subsequent commits.

The second commit is to move the big common code project into ...\System.Private.ServiceModel\tests\Common\src

The final commit is to move the unit tests that were under System.Private.ServiceModel\tests into a new "Unit" subdirectory below there.

Fixing this Issue led to the creation of the following two new issues...
- Create test directory for test common code. #139
- Compartmentalize Scenario and Unit common code. #140


SajayAntony:
@StephenBonikowsky Do your commits have "Fixes #139 and #140" somewhere so that they auto resolve and link the commit with the PR?

StephenBonikowsky:
I didn't mean that this PR fixed those other two Issues, just that in the process of getting this PR ready I created two new issues that need to be fixed after this PR.

SajayAntony:
:shipit:

roncain:
Before I merge, let me prepare the TFS mirrored location so that it agrees with these new locations.  This will avoid a build breakage when nightly test runs occur.

iamjasonp:
LGTM, subject to you having thought about @roncain's comment: 

> An alternate place to put these is wcf.targets at the root. I already started putting the names of other commonly used folders there. Not required, just an alternative.

I'll let @roncain decide when it's safe to merge.

roncain:
LGTM.  Doing merge now and simultaneously updating TFS

Issue #87, opened 2015-06-16
EndpointNotFoundException scenario test fails periodically on CI machine


roncain:
Every 2nd day or so, this test fails.  It is the only failure.

MESSAGE:
Expected exception: EndpointNotFoundException, actual: System.IO.IOException
+++++++++++++++++++
STACK TRACE:
at ExpectedExceptionTests.UnknownUrl_Throws_EndpointNotFoundException() in d:\j\workspace\dotnet_wcf_outerloop\src\System.Private.ServiceModel\tests\Scenarios\Client\ExpectedExceptions\ExpectedExceptionTests.cs:line 315

iamjasonp:
Looks like this is a high priority test to fix when fixing #74.

shmao:
This is an NCL issue. I've come up with a repro which involves with System.Net APIs only. Opened a issue for NCL team.

roncain:
Tracking https://github.com/dotnet/corefx/issues/4934

shmao:
The NCL team confirmed that this was caused by a bug in HttpClient and they would investigate a fix. Marked this issue as 'tracking'.

shmao:
According to @davidsh 's comment in https://github.com/dotnet/corefx/issues/4934 ,

> It has been determined by the WinHTTP team that the IOException is being caused by a bug/race condition in native WinHTTP. There is no current workaround. Even ignoring the error from WinHttpWriteData() will not allow a subsequent WinHttpReceiveResponse() to work, despite the response headers actually being transmitted on the wire from the server to the client.

>So, the fix for this will require a Windows OS fix in a future Windows 10 update

zhenlan:
dotnet/corefx#4934 has been fixed. This can be closed. @StephenBonikowsky can you check if we have any test cases disabled by this issue? Thanks.

StephenBonikowsky:
Test case is active and passing.

Issue #74, opened 2015-06-08
Improve assert messages for unexpected exceptions


roncain:
Many of the scenario tests detect unexpected exceptions but only report their type name, which does not give enough information, especially for CI failures.

We need to report unexpected exceptions in a consistent way.  This includes:
- The expected exception type
- The actual exception type
- (Newline and) exception.ToString() of the actual exception caught

If xunit asserts provide all this, they should be used.  Otherwise we need a test helper.
All unit tests and scenario tests should be updated to use this style.

iamjasonp:
xunit should show exception messages properly. The offending code is this: 

    src\System.Private.ServiceModel\tests\Scenarios\Client\ExpectedExceptions\ExpectedExceptionTests.cs

        catch (Exception e)
        {
            if (e.GetType() != typeof(EndpointNotFoundException))
            {
                Assert.True(false, string.Format("Expected exception: {0}, actual: {1}", "EndpointNotFoundException", e.GetType()));
            }

            if (!e.Message.Contains(expectExceptionMsg))
            {
                Assert.True(false, string.Format("Expected exception message: {0}, actual: {1}", expectExceptionMsg, e.Message));
            }
            return;
        }

I wonder in this case, we are actually reading into the expected exception message; perhaps we can make our own helper here to deal with the case where we expect an exception message.

roncain:
As long as we get the stacktrace of the unexpected exception, and any inner exception, and can test the message, we'll have what we need. I'd use xunit instead of helpers where possible. The goal should be that we can diagnose intermittent test failure in the lab or submitted as an issue. I actually think testing only exception type is a mistake (think ArgumentNullException thrown from somewhere other than you expect.)

iamjasonp:
I've sometimes found that if you're having to dig into the exception message in test, there's a way to test in a more precise way (e.g., split up a block of code into multiple `Assert` calls), or the product has been designed with exceptions that cover too broad a set of errors :smile: 

That being said, I agree that there are many times we may want to verify the message - the example you gave makes sense, as does the testcase cited earlier. I think though that we need to move away from catching the exception and let xunit do as much of its magic as possible. 

Digging slightly into xunit, one pattern we might be able to use is: 

    Assert.Throws<ArgumentNullException>("paramName", () => { /* action */ }); 
    
... if the Exception.paramName is set. 
To verify an exception message: 

    var exception = Assert.Throws<EndpointNotFoundException>( () => { /* action */ });
[...]

PR #146, opened 2015-07-08
Change some ExpectedExceptionTests to show better exception messages

iamjasonp:
We had another failure in CI (http://dotnet-ci.cloudapp.net/job/dotnet_wcf_outerloop/396/) where we've been bitten by the lack of a good exception message. Well, no more! (for this test, at least). 

This PR changes some of the tests. Should help us uncover what's going on with #87 and is a start on #74

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -138,7 +138,7 @@ public static class ExpectedExceptionTests
     // The client should throw a TimeoutException
     [Fact]
     [OuterLoop]
-    public static void TimeoutTest_SendTimeout5Seconds()
+    public static void SendTimeout_For_Long_Running_Operation_Throws_TimeoutException()
     {
         BasicHttpBinding binding = new BasicHttpBinding();
         binding.SendTimeout = TimeSpan.FromMilliseconds(5000);
@@ -161,13 +161,13 @@ public static class ExpectedExceptionTests
 
         // want to assert that this completed in > 5 s as an upper bound since the SendTimeout is 5 sec
         // (usual case is around 5001-5005 ms) 
-        Assert.InRange<long>(watch.ElapsedMilliseconds, 5000, 6000);
+        Assert.InRange<long>(watch.ElapsedMilliseconds, 5000, 10000);
     }
 
     // SendTimeout is set to 0, this should trigger a TimeoutException before even attempting to call the service.
     [Fact]
     [OuterLoop]
-    public static void TimeoutTest_SendTimeout0Seconds()
+    public static void SendTimeout_Zero_Throws_TimeoutException_Immediately()
     {
         BasicHttpBinding binding = new BasicHttpBinding();
         binding.SendTimeout = TimeSpan.FromMilliseconds(0);
@@ -188,9 +188,9 @@ public static class ExpectedExceptionTests
             watch.Stop(); 
         }
 
-        // want to assert that this completed in < 0.5 s as an upper bound since the SendTimeout is 0 sec
+        // want to assert that this completed in < 2 s as an upper bound since the SendTimeout is 0 sec
         // (usual case is around 1 - 3 ms) 
-        Assert.InRange<long>(watch.ElapsedMilliseconds, 0, 500);
+        Assert.InRange<long>(watch.ElapsedMilliseconds, 0, 2000);
     }
 
     [Fact]
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.ServiceModel;
 using System.Text;
 using Xunit;
@@ -12,41 +14,22 @@ public static class ExpectedExceptionTests
     [OuterLoop]
     public static void NotExistentHost_Throws_EndpointNotFoundException()
     {
-        string nonExistHost = "http://nonexisthost/WcfService/WindowsCommunicationFoundation";
-        //On .NET Native retail, exception message is stripped to include only parameter
-        string expectExceptionMsg = nonExistHost;
+        string nonExistentHost = "http://nonexisthost/WcfService/WindowsCommunicationFoundation";
 
-        try
+        BasicHttpBinding binding = new BasicHttpBinding();
+        binding.SendTimeout = TimeSpan.FromMilliseconds(10000);
+
+        EndpointNotFoundException exception = Assert.Throws<EndpointNotFoundException>(() =>
         {
-            //This test can also hang for other test infrastructure related reasons, adding clock based timeout in addition to the innner SendTimeout.
-            DateTime start = DateTime.Now;
-            while (DateTime.Now.Subtract(start).Seconds < 15)
+            using (ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(nonExistentHost)))
             {
-                BasicHttpBinding binding = new BasicHttpBinding();
-                //Setting a timeout as this test takes 500 seconds to finish when it fails.
-                binding.SendTimeout = TimeSpan.FromMilliseconds(10000);
-                using (ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(nonExistHost)))
-                {
-                    IWcfService serviceProxy = factory.CreateChannel();
-                    string response = serviceProxy.Echo("Hello");
-                }
+                IWcfService serviceProxy = factory.CreateChannel();
+                string response = serviceProxy.Echo("Hello");
             }
-        }
-        catch (Exception e)
-        {
-            if (e.GetType() != typeof(EndpointNotFoundException))
-            {
-                Assert.True(false, string.Format("Expected exception: {0}, actual: {1}", "EndpointNotFoundException", e.GetType()));
-            }
+        });
 
-            if (!e.Message.Contains(expectExceptionMsg))
-            {
-                Assert.True(false, string.Format("Expected exception message: {0}, actual: {1}", expectExceptionMsg, e.Message));
-            }
-            return;
-        }
-
-        Assert.True(false, "Expected EndpointNotFoundException, but no exception thrown.");
+        // On .Net Native retail, exception message is stripped to include only parameter
+        Assert.Contains(nonExistentHost, exception.Message, StringComparison.OrdinalIgnoreCase);
     }
 
     [Fact]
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -10,47 +12,6 @@ public static class ExpectedExceptionTests
 {
     [Fact]
     [OuterLoop]
-    public static void NotExistentHost_Throws_EndpointNotFoundException()
```
iamjasonp: I don't really understand this test - it seems to do exactly what `UnknownUrl_Throws_EndpointNotFoundException` does, except with the funky while loop that doesn't make sense to me. Is there something I'm missing? I ended up amalgamating the two tests into one [Theory] based test. 

One endpoint is for a *host* that doesn't exist, one is for an *endpoint* that doesn't exist on a host that does.

roncain: Be aware Theory is not supported yet in the experimental NET Native test environment, so it will make the test a NOP in NET Native.  I recommend avoiding Theory for now except for strictly data driven tests.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -292,32 +253,25 @@ public static void UnexpectedException_Throws_FaultException()
         Assert.True(false, "Expected FaultException<FaultDetail> exception, but no exception thrown.");
     }
 
-    [Fact]
+    [Theory]
```
roncain: As mentioned elsewhere, Theory won't run in our NET Native test environment.  Unfortunate, but I recommend an explicit loop over the theory data for data driven tests or separate tests if they really are different scenarios.

iamjasonp: Got rid of the [Theory]. Sadness that the .NET Native environment doesn't handle it well.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: Thanks. Got rid of the [Theory] in this set of tests.

iamjasonp:
**Note:** The goal here was not to change every test in the class; I just took the low hanging fruit where it is likely to fail in the same way `ExpectedExceptionTests.UnknownUrl_Throws_EndpointNotFoundException` did. Hopefully this is sufficient to give us a hint as to why this (and related) tests fail seemingly randomly in CI. 

Also, @StephenBonikowsky if you could take a look too... thanks!

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -157,28 +140,28 @@ public static void NonExistentAction_Throws_ActionNotSupportedException()
...
+            watch.Stop();
         }
 
-        if (!exceptionThrown)
-        {
-            Assert.True(false, "Expected TimeoutException was not thrown nor was any other exception thrown.");
-        }
+        // want to assert that this completed in > 5 s as an upper bound since the SendTimeout is 5 sec
+        // (usual case is around 5001-5005 ms) 
+        Assert.InRange<long>(watch.ElapsedMilliseconds, 5000, 6000);
```
iamjasonp: Added this check to ensure that we really do fail pretty much in ~5 sec.  See below, there is another test that has limits between 0-500 ms. 

The wide range is so that timing issues don't cause false failures. In this case, we know that we will fail in no *less than* 5 s.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -12,41 +14,22 @@ public static class ExpectedExceptionTests
...
-            }
-        }
-        catch (Exception e)
-        {
-            if (e.GetType() != typeof(EndpointNotFoundException))
-            {
-                Assert.True(false, string.Format("Expected exception: {0}, actual: {1}", "EndpointNotFoundException", e.GetType()));
-            }
+        BasicHttpBinding binding = new BasicHttpBinding();
+        binding.SendTimeout = TimeSpan.FromMilliseconds(10000);
```
roncain: Nit: we should put timeout literals in a common place and use them from the tests.  Not needed for this PR but we can clean these up as we discover them.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -157,28 +140,28 @@ public static void NonExistentAction_Throws_ActionNotSupportedException()
     [OuterLoop]
     public static void TimeoutTest_SendTimeout5Seconds()
     {
```
roncain: Nit: I've always wanted this test method to be retitled, because it doesn't describe what it does.  It should be something like Long_Running_Operation_Throws_TimeoutException.  Not required for this PR, but let's gradually make the method names describe what they do.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -157,28 +140,28 @@ public static void NonExistentAction_Throws_ActionNotSupportedException()
...
-            proxy.EchoWithTimeout("Hello");
-        }
-        catch (TimeoutException)
-        {
-            exceptionThrown = true;
+            var exception = Assert.Throws<TimeoutException>(() =>
+            {
+                IWcfService proxy = factory.CreateChannel();
+                watch.Start();
+                proxy.EchoWithTimeout("Hello");
```
roncain: Not now, but this test should always have passed in a timeout value to the service (example: its own SendTimeOut * 2).  The server's operation should not have a hard-coded wait time that magically is slightly longer than the value in this test.  Not needed for this PR.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
roncain: I think the InRange so tightly restricted will cause this test to fail periodically in the labs when load is high.  Do we strictly care about the upper bound, as long as we know it is at least >= SendTimeout?

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -186,28 +169,28 @@ public static void TimeoutTest_SendTimeout5Seconds()
     [OuterLoop]
     public static void TimeoutTest_SendTimeout0Seconds()
     {
```
roncain: Nit: another name that could be improved.  Something like Zero_SendTimeout_Throws_TimeoutException_Immediately.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -186,28 +169,28 @@ public static void TimeoutTest_SendTimeout5Seconds()
...
+            watch.Stop(); 
         }
 
-        if (!exceptionThrown)
-        {
-            Assert.True(false, "Expected TimeoutException was not thrown nor was any other exception thrown.");
-        }
+        // want to assert that this completed in < 0.5 s as an upper bound since the SendTimeout is 0 sec
+        // (usual case is around 1 - 3 ms) 
+        Assert.InRange<long>(watch.ElapsedMilliseconds, 0, 500);
```
roncain: I'm concerned about the InRange here too.  Predicting execution times of lab machines usually makes tests break periodically.  I think it would be okay to simply verify the time is not greater than the default SendTimeout (i.e. verify we did timeout "sooner" than the normal timeout).

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: Hmm. We walk a fine line here, since we want to make sure that the 5s time is what we hit, not some other setting. This test should make sure that we set a timeout and it's adhered to, which is why the range should be as tight as possible without causing false failures.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: This is probably something we need the bridge for? Once a service is up, we can't dynamically change the timeout value per request.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: But then now we're in the business of verifying timeouts in the order of 10s? 

I dunno. In theory, the CI machines should be fairly well-equipped to handle load, and local machines should also be able to handle this. 

Maybe we give 2 seconds leeway... that's probably good enough for verifying a client scenario. @StephenBonikowsky, what are your thoughts on this upper bound value? 

I maintain that we should check this value, but I'm open to loosening the bounds. I do think that a 10s bound is probably a little too much. (Yes, we're having a discussion on choosing a magic number now...)

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: ... as long as they're contained within the same project :smile:. I don't want to be hunting through multiple projects to find a send timeout, they should be at least F12 resolvable. 

Putting a send timeout in `System.ServiceModel.Tests.Common` for all tests under the sun seems a little overkill to me in test code.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: Changed... not quite your wording exactly though.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: Oh nevermind. I'm being dumb... yes. The service method should just wait a number of seconds. 

However, we run into an interesting case if the number passed then ends up exceeding the service's Send/ReceiveTimeout...

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
roncain: I mostly meant there should be one "this is the longest time any test can legally wait" value that we can increase of lab machine load increases.  Hey, it belongs in TestProperties and we could drive it in from the CMD line.  No need to do that for this PR.

iamjasonp: Ah, I see what you mean now. Wouldn't each test would now have to be aware of this number? 

I feel like this might be something we push for in the xunit harness instead :)

StephenBonikowsky:
:shipit:

iamjasonp:
Squashed and added some better descriptions to the commits.

PR #148, opened 2015-07-09
Update README.md with myget.org package links

iamjasonp:


roncain:
LGTM. I presume the reason for this change is to resemble recent changes to CoreFx's README.

iamjasonp:
Yes - it'd be good to show that our packages are indeed being pushed up to the myget feeds

mconnew:
LGTM

PR #150, opened 2015-07-09
Escape URLs in dir.props

roncain:
The double-slashes for URL's in dir.props needed
to be escaped to match CoreFx's changes.


iamjasonp:
LGTM. 
Was there any hint as to why only the `http://` had to be escaped, and not the remaining slashes? I couldn't figure it out...

mconnew:
LGTM

Issue #138, opened 2015-07-02
Implement Task based Duplex callback contracts


SajayAntony:
The duplex callback issue tracks only synchronous invocation and not the other patterns. We need to implement the task based pattern for CallBackContracts on the client. #1

mconnew:
Fixed in PR 165

PR #151, opened 2015-07-09
Scenario test for Issue #138

StephenBonikowsky:
* Test won't pass until Issue 138 is fixed.
* Validated that it worked on Desktop.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/TypedClient/TypedProxyDuplexTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/TypedClient/TypedProxyDuplexTests.cs
@@ -1,0 +1,69 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using TestTypes;
+using Xunit;
+
+public static class TypedProxyDuplexTests
+{
+    // ServiceContract typed proxy tests create a ChannelFactory using a provided [ServiceContract] Interface which...
+    //       returns a generated proxy based on that Interface.
+    // ChannelShape typed proxy tests create a ChannelFactory using a WCF understood channel shape which...
+    //       returns a generated proxy based on the channel shape used, such as...
+    //              IRequestChannel (for a request-reply message exchange pattern)
+    //              IDuplexChannel (for a two-way duplex message exchange pattern)
+
+    [Fact]
+    [ActiveIssue(138)]
+    [OuterLoop]
+    public static void ServiceContract_TypedProxy_AsyncTask_CallbackReturn()
+    {
+        DuplexChannelFactory<IWcfDuplexTaskReturnService> factory = null;
+        Guid guid = Guid.NewGuid();
+
+        NetTcpBinding binding = new NetTcpBinding();
+        binding.Security.Mode = SecurityMode.None;
+
+        DuplexChannelServiceCallback callbackService = new DuplexChannelServiceCallback();
+        InstanceContext context = new InstanceContext(callbackService);
+
+        using (factory = new DuplexChannelFactory<IWcfDuplexTaskReturnService>(context, binding, new EndpointAddress(Endpoints.Tcp_NoSecurity_Callback_Address)))
+        {
+            IWcfDuplexTaskReturnService serviceProxy = factory.CreateChannel();
+
+            Task<Guid> task = serviceProxy.Ping(guid);
+
+            Guid returnedGuid = task.Result;
+
+            if (guid != returnedGuid)
+            {
+                Assert.True(false, String.Format("The sent GUID does not match the returned GUID. Sent: {0} Received: {1}", guid, returnedGuid));
+            }
+
+            factory.Close();
+
+            if (factory != null && factory.State != CommunicationState.Closed)
+            {
+                factory.Abort();
+            }
+        }
+    }
+
+    public class DuplexTaskReturnServiceCallback : IWcfDuplexTaskReturnCallback
+    {
+        public Task<Guid> ServicePingCallback(Guid guid)
+        {
+            // This returns the guid to the service which called this callback.
+            // We could return Task.FromResult(guid) but that means we could execute the 
+            // completion on the same thread. But if someone is using a task it means they 
+            // would potentially have the completion on another thread.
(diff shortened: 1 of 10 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Client/TypedClient/TypedProxyDuplexTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            IWcfDuplexTaskReturnService serviceProxy = factory.CreateChannel();
+
+            Task<Guid> task = serviceProxy.Ping(guid);
+
+            Guid returnedGuid = task.Result;
+
+            if (guid != returnedGuid)
+            {
+                Assert.True(false, String.Format("The sent GUID does not match the returned GUID. Sent: {0} Received: {1}", guid, returnedGuid));
```
iamjasonp: NIT: `Assert.True(guid == returnedGuid, "comment")`. Consider `Assert.Equal` - it should provide enough information from xunit without the custom error message

iamjasonp: Hmm, I'm also not sure that you're testing the value that's returned from the callback?

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/App.config
```diff
@@ -167,6 +167,10 @@
                 <endpoint address="tcp-nosecurity-callback" binding="netTcpBinding" bindingConfiguration="nettcpBindingCallback"
                           name="nettcp-defaultBinding" contract="WcfService.IDuplexChannelService" />
             </service>
+            <service name="WcfService.DuplexChannelCallbackReturnService">
+                <endpoint address="tcp-nosecurity-callback" binding="netTcpBinding" bindingConfiguration="nettcpBindingCallback"
+                          name="nettcp-defaultBinding" contract="WcfService.IWcfDuplexTaskReturnService" />
+            </service>
```
iamjasonp: You will need to coordinate with @dmetzgar with this, as this has changed fundamentally with the introduction on the bridge.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/IWcfDuplexService.cs
```diff
@@ -31,4 +32,17 @@ public interface IDuplexChannelCallback
...
+    public interface IWcfDuplexTaskReturnService
+    {
+        [OperationContract]
+        Task<Guid> Ping(Guid guid);
+    }
+
+    public interface IWcfDuplexTaskReturnCallback
+    {
+        [OperationContract]
+        Task<Guid> ServicePingCallback(Guid guid);
```
iamjasonp: Is it sufficient here to return a `Task` rather than a `Task<Guid>`?

src/System.Private.ServiceModel/tests/Scenarios/Client/TypedClient/TypedProxyDuplexTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            factory.Close();
+
+            if (factory != null && factory.State != CommunicationState.Closed)
+            {
+                factory.Abort();
+            }
```
iamjasonp: This `if (factory != null && factory.State != CommunicationState.Closed` block should be put inside a finally. 

We want to abort if there's been an exception in any of the code above; if there's been an error, this code to Abort will never execute as this code would be skipped in an exception flow case.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/IWcfDuplexService.cs
StephenBonikowsky: I could but then I would have to put that Guid back into a Task in the service operation in order to return it back to the client. Seems easier/simpler this way?

src/System.Private.ServiceModel/tests/Scenarios/Client/TypedClient/TypedProxyDuplexTests.cs
StephenBonikowsky: Assert.Equal looks good, thanks.

"guid" is first sent to the service operation, then passed to the callback server side, the server callback sends it to the callback on the client which just returns it back to the server callback, the service operation then returns it to the client as the return value of the initial call to the service operation as "returnedGuid", at least that's the intention, unless I wired something together wrong.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/App.config
roncain: Depending who merges first, @dmetzgar may need to update his bridge branch to accommodate thisnew endpoint.

roncain:
LGTM, subject to @iamjasonp comments being addressed.  I think the Task<Guid> return is fine and demonstrates a real scenario of taking the whole callback sequence async.  To clarify, any service operation returning Task<T> doesn't actually serialize back a Task.  It schedules the resulting T to be returned when the Task completes on the server side.

StephenBonikowsky:
I need to update with the feedback and I also found a couple things I missed. Please don't merge this PR.

src/System.Private.ServiceModel/tests/Scenarios/Client/TypedClient/TypedProxyDuplexTests.cs
StephenBonikowsky: Ok, so I will not use the using statement and will use a try/finally instead.

StephenBonikowsky:
Need a new base address because I needed a new service, needed a new service because I needed to set the ConcurrencyMode on the ServiceBehavior which I couldn't have done on either of the other two Duplex services because that would have messed up the scenario being tested with those services. I couldn't see a way around this.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/IWcfDuplexService.cs
iamjasonp: Yeah, I guess this is testing the request-reply path at the same time as the callback path, not a bad thing; let's just leave it as-is.

SajayAntony:
Closing the PR since I squashed and committed as per @StephenBonikowsky 
29171d84ae2a9c0e3f4aff7ef2c0c528b637dd5e
5a082865de3d972a31877dc236f386d04e0588a9

PR #156, opened 2015-07-10
Changing ExpectedExceptionTests to not use Assert.Contains

iamjasonp:
Our internal xunit-like framework for .NET Native doesn't support the use of Assert.Contains,
so for now, moving any Assert.Contains to using Assert.True instead. 

At a later date, we can consider moving back to Assert.Contains since the messages look nicer.


roncain:
LGTM

Issue #91, opened 2015-06-16
Implement DuplexChannelFactory


roncain:
As per https://github.com/dotnet/wcf/issues/1 implement DuplexChannelFactory

iamjasonp:
Tests written, #161

PR #152, opened 2015-07-09
Enable duplex service contracts

mconnew:
Fixes #1

Issue #90, opened 2015-06-16
Implement DuplexClientBase


roncain:
As per issue https://github.com/dotnet/wcf/issues/1, implement DuplexClientBase

SajayAntony:
Won't #111 be blocked by this?

iamjasonp:
@SajayAntony Yes, it's kind of a circular reference. Basically we are using work done in #111 to validate anything we do here.

iamjasonp:
Tests written.

PR #161, opened 2015-07-13
Unit and scenario tests for DuplexClientBase and DuplexChannelFactory

iamjasonp:


iamjasonp:
Tests fix #91, #90. Implementation was #152

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -125,31 +125,74 @@ public static void ClientMessageInspector_Verify_Invoke()
 
     [Fact]
     [OuterLoop]
-    public static void ClientBaseOfT_Call()
+    public static void ClientBaseOfT_HttpTransport_Call()
     {
```
roncain: Nit: my bad for choosing a poor name initially.  Would prefer something like ClientBaseOfT_Synchronous_RoundTrip_Call_Using_UsingHttpTransport or something like that

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -125,31 +125,74 @@ public static void ClientMessageInspector_Verify_Invoke()
...
+            string result = serviceProxy.Echo("Hello");
+            Assert.Equal("Hello", result);
 
-            MyClientBase client = new MyClientBase(customBinding, new EndpointAddress(BaseAddress.HttpBaseAddress));
-            IWcfServiceGenerated serviceProxy = client.ChannelFactory.CreateChannel();
+            Assert.Equal(CommunicationState.Opened, client.State);
 
-            string result = serviceProxy.Echo("Hello");
-            if (!string.Equals(result, "Hello"))
+            client.Close();
```
roncain: Close() is not in the public contract.  This test fails to compile in our CoreCLR/NET Native test environment.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -125,31 +125,74 @@ public static void ClientMessageInspector_Verify_Invoke()
...
+        try
         {
-            errorBuilder.AppendLine(String.Format("Unexpected exception was caught: {0}", ex.ToString()));
-        }
+            string result = serviceProxy.Echo("Hello");
+            Assert.Equal("Hello", result);
 
-        Assert.True(errorBuilder.Length == 0, string.Format("Test Scenario: ClientBaseOfTCall FAILED with the following errors: {0}", errorBuilder));
+            Assert.Equal(CommunicationState.Opened, client.State);
+
```
roncain: Nit: testing the communication state seems out of scope for this test.  It would be better to create a separate test to verify communication state as affected by ClientBase<T>.  It wouldn't even need to make a call to the service. The purpose of this current test is really just verifying a synchronous round-trip call via ClientBase<T> succeeds.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
```diff
@@ -15,10 +13,45 @@ public static class DuplexClientBaseTests
...
+            // "Contract requires Duplex, but Binding 'BasicHttpBinding' doesn't support it or isn't configured properly to support it"
+            Assert.True(exception.Message.Contains("BasicHttpBinding")); 
+        }
+        finally
+        {
+            if (duplexService != null && duplexService.State != CommunicationState.Closed)
+            {
+                duplexService.Abort();
+                Assert.Equal(CommunicationState.Closed, duplexService.State);
+            }
```
roncain: Similar comment -- it is not really the purpose of this test to verify Abort sets the communication state.  It is worthwhile having a test like that, but I'd prefer not to entangle it with a test verifying DuplexClientBase<T> throwing due to Http.

src/System.ServiceModel.Primitives/tests/ServiceModel/ChannelFactoryTest.cs
```diff
@@ -118,10 +130,12 @@ public static void CreateChannel_Of_IRequestChannel_Using_BasicHttpBinding()
             if (factory != null)
             {
                 factory.Close();
+                Assert.Equal(CommunicationState.Closed, factory.State);
```
roncain: Same comment -- this test is not testing factory.Close()'s affect on communication state.  I'd rather put that in a separate test if we want to verify it.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
```diff
@@ -0,0 +1,384 @@
...
+        Binding binding = new NetTcpBinding();
+        EndpointAddress remoteAddress = null;
+
+        DuplexChannelFactory<IWcfDuplexService> factory = new DuplexChannelFactory<IWcfDuplexService>(callback, binding, remoteAddress);
+        try {
+            Assert.Throws<InvalidOperationException>(() =>
+               {
+                   Assert.Equal(CommunicationState.Created, factory.State);
+                   factory.Open();
+                   factory.CreateChannel();
```
roncain: Not a problem but this test reminds me we need to test async opens and async closes.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    [Fact]
+    // valid address, but the scheme is incorrect
+    public static void CreateChannel_ExpectedNetTcpScheme_HttpScheme_ThrowsUriFormat()
+    {
+        WcfDuplexServiceCallback callback = new WcfDuplexServiceCallback();
+        Binding binding = new NetTcpBinding();
+        Assert.Throws<ArgumentException>(() =>
+        {
```
roncain: Nit: I think all ArgumentException tests should validate exception.ParamName.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
```diff
@@ -0,0 +1,384 @@
...
+        DuplexChannelFactory<IWcfDuplexService> factory = new DuplexChannelFactory<IWcfDuplexService>(callback, binding, (EndpointAddress)null);
+
+        Assert.Throws<InvalidOperationException>(() =>
+        {
+            factory.CreateChannel(); 
+        });
+    }
+
+    [Fact(Skip = "SecureConnectionDuplexSession is not supported yet.")]
+    public static void CreateChannel_Using_NetTcpBindingDefaults()
```
roncain: This should be [ActiveIssue(nnn)] with the open issue for this.  Otherwise this test will fail in our CoreCLR/NETNative test environment.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
```diff
@@ -0,0 +1,384 @@
...
+
+            Assert.Equal(CommunicationState.Created, factory.State);
+            Assert.Equal(CommunicationState.Created, factory2.State);
+
+            var exception = Assert.Throws<InvalidOperationException>(() =>
+            {
+                // Create the channel.
+                channel = factory.CreateChannel();
+                Assert.True(typeof(IWcfDuplexService).GetTypeInfo().IsAssignableFrom(channel.GetType().GetTypeInfo()),
+                    String.Format("Channel type '{0}' was not assignable to '{1}'", channel.GetType(), typeof(IDuplexChannel)));
```
roncain: I  don't think the IsAssignable test belongs here.  By definition the CreateChannel must return a type assignable to IWcfDuplexService.  It would be a runtime error if the DispatchProxy layer attempted to return something that wasn't.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
```diff
@@ -0,0 +1,384 @@
...
+            {
+                factory2.Abort();
+                Assert.Equal(CommunicationState.Closed, factory2.State);
+            }
+        }
+    }
+
+    [Fact]
+    public static void CreateChannel_Of_IDuplexChannel_Using_NetTcpBinding()
+    {
```
roncain: This test is really validating that 2 calls to CreateChannel creates unique channel instances.  I'd rename it to reflect that.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexClientBaseTest.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using System.Threading.Tasks;
+using Xunit;
+
+public class DuplexClientBaseTest
+{
+    [Fact]
+    public static void CreateDuplexClientBase()
+    {
```
roncain: Nit: Close is not in contract, but oddly Abort is.  Recommend rename DuplexClientBase_Ctor_Initializes_State and just validate the properties and any other reachable state. Then make a separate test DuplexClientBase_Aborts.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexClientBaseTest.cs
```diff
@@ -0,0 +1,79 @@
...
+        Assert.Equal<CommunicationState>(CommunicationState.Closed, duplexClientBase.State);
+    }
+
+    [Fact]
+    public static void CreateDuplexClientBase_NullContext_Throws()
+    {
+        Binding binding = new NetTcpBinding();
+        EndpointAddress endpoint = new EndpointAddress(Endpoints.Tcp_NoSecurity_Callback_Address);
+        Assert.Throws<ArgumentNullException>(() => { MyDuplexClientBase<IWcfDuplexService> duplexClientBase = new MyDuplexClientBase<IWcfDuplexService>(null, binding, endpoint); }); 
+    }
```
roncain: Same comment as before and I'll stop pointing out each one -- I think ArgumentException tests should validate ParamName.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
iamjasonp: Fixed.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
iamjasonp: Admittedly, there would probably be problems should the state be inconsistent, but I see your point. I tend to think that a scenario test really should verify an entire scenario and be as comprehensive as we can make it, but I'll concede it bloats things and can cause failures where there may not be one for that scenario.

I'll copy out the state checking into another test.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
iamjasonp: Changed.

iamjasonp: I will, however, leave `DuplexClientBaseOfT_OverNetTcp_Call` to fulfill the purpose of the "normal path" test, including the state.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
iamjasonp: Done.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexClientBaseTest.cs
iamjasonp: Done.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
iamjasonp: Makes sense; this particular test was revised pretty heavily as I realized that the test shouldn't continue anyway after the `InvalidOperationException`

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -125,31 +125,92 @@ public static void ClientMessageInspector_Verify_Invoke()
...
-            CustomBinding customBinding = new CustomBinding();
-            customBinding.Elements.Add(new TextMessageEncodingBindingElement());
-            customBinding.Elements.Add(new HttpTransportBindingElement());
+            string result = serviceProxy.Echo("Hello");
+            Assert.Equal(CommunicationState.Opened, client.State);
 
-            MyClientBase client = new MyClientBase(customBinding, new EndpointAddress(BaseAddress.HttpBaseAddress));
-            IWcfServiceGenerated serviceProxy = client.ChannelFactory.CreateChannel();
+            // Not in profile for CoreCLR / .NET Native
+            // client.Close();
```
roncain: Nit: remove commented out code.  If you really want to Close(), you can still do ((ICommunicationObject)client).Close()

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -125,31 +125,92 @@ public static void ClientMessageInspector_Verify_Invoke()
 
     [Fact]
     [OuterLoop]
-    public static void ClientBaseOfT_Call()
+    public static void ClientBaseOfT_Sync_RoundTrip_Check_CommunicationState()
```
roncain: Okay, I give up about checking communication state :smile: .  But then why doesn't every other e2e test do the same?  I think it is good practice to make each test verify exactly one thing.  @StephenBonikowsky and I had a conversation and agreed scenario tests should test only one thing.  Pre-conditions related to setting up for the test and post-conditions tangential to the main test should be assumed to be correct and not explicitly tested. They belong in dedicated tests of their own.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -125,31 +125,92 @@ public static void ClientMessageInspector_Verify_Invoke()
```
roncain: Incidentally, what I forgot to say last review is that this is great to see this level of unit tests for these types.  Nice coverage.

src/System.ServiceModel.Primitives/tests/ServiceModel/ChannelFactoryTest.cs
```diff
@@ -127,6 +127,53 @@ public static void CreateChannel_Of_IRequestChannel_Using_BasicHttpBinding()
     }
 
     [Fact]
+    public static void ChannelFactory_Verify_CommunicationStates()
+    {
```
roncain: Nice!

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexClientBaseTest.cs
```diff
@@ -0,0 +1,88 @@
...
+        InstanceContext context = new InstanceContext(new WcfDuplexServiceCallback());
+        Binding binding = new NetTcpBinding();
+        EndpointAddress endpoint = new EndpointAddress(Endpoints.Tcp_NoSecurity_Callback_Address);
+        MyDuplexClientBase<IWcfDuplexService> duplexClientBase = new MyDuplexClientBase<IWcfDuplexService>(context, binding, endpoint);
+
+        Assert.Equal<EndpointAddress>(endpoint, duplexClientBase.Endpoint.Address);
+        Assert.Equal<CommunicationState>(CommunicationState.Created, duplexClientBase.State);
+
+        duplexClientBase.Abort();
+    }
```
roncain: Question only: do you recommend we Abort or Close every channel, channelFactory or ClientBase we create?  And if so, should this be in a finally block so it executes even if the asserts fire?  I would be inclined to put them in using statements, but I believe WCF has the ability to throw from the Dispose.  Though it might be reasonable to let the Dispose throw and let xunit report it.  Thoughts @iamjasonp ?

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
```diff
@@ -15,10 +13,45 @@ public static class DuplexClientBaseTests
 
     [Fact]
     [OuterLoop]
-    public static void DuplexClientBaseOfT_OverNetTcp_Call()
+    public static void DuplexClientBaseOfT_OverHttp_Call_Throws_InvalidOperation()
+    {
+        DuplexClientBase<IWcfDuplexService> duplexService = null;
+        Guid guid = Guid.NewGuid();
+
```
roncain: Aside: this test failed for me running against our TFS CoreCLR/NETNative test environment.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
iamjasonp: I actually got rid of most of them; if you see any scenario tests other than this one with CommunicationState checks, it was probably a bad push. I agree that for most of the rest of the tests, they are overkill and hence are gone. 

In this case, I will be a little rigid and keep this one. We do not have a test that tests the CommunicationObject, and CommunicationObject itself can't be tested unless we want to mock up our own derived CommunicationObject and test that (it might not be a bad idea, I'm just not seeing it as a high pri right now) 

I'm being a little lazy in this case, and having a round-trippable ICommunicationObject (in this case, ClientBase) where you can actually get to the `Opened` state is valuable.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexClientBaseTest.cs
iamjasonp: The prevailing wisdom has been to Close() every channel, thus freeing up resources ASAP. In the case of testing, it probably doesn't matter that we don't close and abort everything since the server is short-lived and the client probably isn't long-lived. However, I'm working on the assumption that *every piece of code that we put into test is a potential code sample* . 

Thus, if we think it's good practice for production, we should, to the best of our ability, include this in our own best practices, even if it's "just" test code.

iamjasonp: Also, we as a rule don't use `using` when dealing with WCF service proxies - it's well documented (see [http://blogs.msdn.com/b/jjameson/archive/2010/03/18/avoiding-problems-with-the-using-statement-and-wcf-service-proxies.aspx](example)). 

While in CoreCLR we can probably (finally) change this behaviour, we have to be mindful that in-contract code working in CoreCLR must work in Desktop... so if we start using `using` statements here... well, see above about "sample code" :smile:

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
iamjasonp: I was on the fence about this one. I *know* that I'll forget that I thought about this already the next time I see this code :smile: 

I don't necessarily care to Close(), I just want an indication that it's been thought about...

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
iamjasonp: #167 opened so we don't forget.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexClientBaseTest.cs
roncain: We should probably discuss how "sample worthy" the test code should be.  I think we recommend Close() followed by Abort() if an exception occurs (http://www.codeproject.com/Articles/622989/WCF-and-the-Try-Catch-Abort-Pattern ).  If we agree all scenario tests are potential samples, I think we should agree what that acceptable sample pattern is regarding closing channels, channel factories, ClientBase<T> etc.   Don't alter this PR for this -- I just want consistency across our tests.

iamjasonp: Definitely worth a discussion :smile: 
Incidentally, the test you commented on doesn't happen to fall into this case where `try - finally` would be needed, since if the ctor throws, there's nothing to Abort.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
iamjasonp: Looking into this right now.

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexClientBaseTest.cs
roncain: Assert.Equal can throw.  And that ties in with our discussion about sample-quality try/catch/finally.  I'll open a new issue to carry on the discussion of sample-quality tests and the proper close patterns that account for asserts being the throwers,  No need to change anything in this PR for this.

iamjasonp: Ergh... right. I guess I'm thinking about the happy path too much here.

roncain:
LGTM

StephenBonikowsky:
:shipit:

iamjasonp:
Prior to merging, let me squash down the commits some.

roncain:
Why don't you do the merge after you have squashed?

src/System.ServiceModel.Primitives/tests/ServiceModel/ChannelFactoryTest.cs
```diff
@@ -127,6 +127,53 @@ public static void CreateChannel_Of_IRequestChannel_Using_BasicHttpBinding()
...
+                String.Format("Channel type '{0}' was not assignable to '{1}'", channel.GetType(), typeof(IRequestChannel)));
+            Assert.Equal(CommunicationState.Opened, factory.State);
+
+            // Validate ToString()
+            string toStringResult = channel.ToString();
+            string toStringExpected = "System.ServiceModel.Channels.IRequestChannel";
+            Assert.Equal<string>(toStringExpected, toStringResult);
+
+        }
+        finally
```
SajayAntony: Why should we do this in the finally?

iamjasonp:
@roncain said: 
> Why don't you do the merge after you have squashed?

... eh? Confused by that statement. How else would you do it?

PR #165, opened 2015-07-14
Fixing Callback contracts with a Task<T> return type

mconnew:
Callback contracts with a return type of simply Task were working, but Task<T> were causing a null ref exception.


roncain:
LGTM

iamjasonp:
LGTM

PR #173, opened 2015-07-15
Integrate recent CoreFx changes to props and targets files

roncain:
CoreFx made some changes to alter how the DnuRestore worked
and to enable warnings as errors (which is how the projects
are built in the TFS mirrored location).  This PR brings those
changes to WCF.

It also fixes the one warning we had that caused the build to fail.


hongdai:
LGTM.

iamjasonp:
LGTM

PR #176, opened 2015-07-16
Fix async channel open so that continuations don't use the current sync context

### LINKED ISSUE #78: Aync APIs genereated by ASR do not work for NetTcpBinding endpoint
In a UWP app, use "Add Service Reference" to generate a service client. All the operation APIs will be async. They work fine for http endpoint, but hangs for NetTcpBinding endpoint. This repros for both debug and release mode, so I suspect this is not specific to .NET Native.

mconnew:
This was causing deadlocks when calling a Task based client synchronously, e.g. FooAsync().Result.
Fixes #78


iamjasonp:
LGTM

SajayAntony:
:shipit:

Issue #178, opened 2015-07-16
Make Duplex unit tests use BaseAddress or move to scenario tests


roncain:
The new Duplex unit tests use the Endpoints class which is reserved for scenario tests.  This prevents the adoption of the Bridge which must be running for the new Endpoints implementation to work.

All Duplex unit tests using Endpoints have been marked [ActiveIssue] to prevent their running.

This task is to change to the use of BaseAddress or move the tests that require Endpoints to scenario tests and mark them [OuterLoop]

SajayAntony:
@iamjasonp I'm still not sure what made these tests pass.

iamjasonp:
This was way too confusing. Thanks for fixing, @roncain

PR #179, opened 2015-07-16
Disable Duplex unit tests using Endpoints

roncain:
The Endpoints class is reserved for [OuterLoop] tests and
requires a running WCF Service.

This PR adds [ActiveIssue(178)] to all the Duplex unit tests
using Endpoints.  They need to be converted to use BaseAddress only.


SajayAntony:
Reference #178

SajayAntony:
:shipit:

PR #180, opened 2015-07-16
Re-enables Duplex unit tests previously disabled

### LINKED ISSUE #178: Make Duplex unit tests use BaseAddress or move to scenario tests

roncain:
The Duplex unit tests previously disabled with [ActiveIssue(178)]
have been re-enabled and modified to use static BaseAddress URL's

Fixes #178


SajayAntony:
:shipit:

iamjasonp:
LGTM.

PR #183, opened 2015-07-16
Fake address

### LINKED ISSUE #182: Refactor BaseAddress and Endpoints classes for better isolation
The BaseAddress class should be renamed to FakeAddress, and it should contain only fake URL's that cannot possibly resolve.  Find a well-known IP that will never resolve.  If possible, collapse to as few as possible (http, https, tcp).

The Endpoints class should move into the scenarios common project (tests\Scenarios\ServiceModel.Scenarios.Common) so that innerloop tests cannot use it.

SajayAntony:
This does not fully fix #182 but fixes just the refactoring part. 
The endpoint move is still pending.


SajayAntony:
@roncain Hope this reduces the overall set of issues with endpoints and BaseAddress

src/System.Private.ServiceModel/tests/Common/src/FakeAddress.cs
```diff
@@ -0,0 +1,11 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Net;
+
+public static class FakeAddress
+{
+    public static string HttpAddress = "http://" + IPAddress.None + ":0";
+    
+    public static string TcpAddress = "net.tcp://" + IPAddress.None + ":0";
+}
```
roncain: Nit: newline needed

roncain:
It looks like a merge is needed due to the changes in the duplex tests, but I've buddy-built (after manually merging) and all innerloop and outerloop tests pass.  LGTM after the merge.

SajayAntony:
Ok will rebase and submit.

iamjasonp:
Thanks for the head's up. Please merge, but otherwise LGTM.

Issue #189, opened 2015-07-17
CreateChannel_Using_Http_NoSecurity test fails in NET Native


roncain:
The assert implies the exception message is wrong.  When resolving this issue, please make a better Assert message to show what the wrong message was.  

Running Test: DuplexChannelFactoryTest.CreateChannel_Using_Http_NoSecurity
  Assert.True:InvalidCastException exception string should contain 'IRequestChannel'
  Xunit.AssertTestException: Assert.True: InvalidCastException exception string should contain 'IRequestChannel'
     at Xunit.Assert.HandleFail(String assertionName, String message) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 592
     at Xunit.Assert.True(Boolean condition, String message) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 147
     at DuplexChannelFactoryTest.CreateChannel_Using_Http_NoSecurity() in e:\OSS\wcf\src\System.ServiceModel.Primitives\tests\ServiceModel\DuplexChannelFactoryTest.cs:line 174
     at EntryPointMain.b__1a() in e:\ProjectK\src\QA\ToF\PN\x86\dbg\IL\FX\Conformance\System.Private.ServiceModel\4.0.0.0\WcfOpen\unit\System.ServiceModel.Primitives\System.ServiceModel.Primitives.Tests.main.cs:line 41
     at System.Action.InvokeOpenStaticThunk()
     at System.Action.Invoke()
     at CoreFXTestLibrary.Internal.Runner.RunTestMethod$catch$0() in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Internal\Runner.cs:line 160
  ---- Test FAILED ---------------

PR #197, opened 2015-07-20
Improve exception message reporting from Assert.True checks

iamjasonp:
When some checks for Assert.True were done to verify the contents of an exception message, the Assert.True statement did not report the full failing exception message. 

Note - for #189 
DuplexChannelFactoryTest.CreateChannel_Using_Http_NoSecurity attempts to check
for the string 'IRequestChannel' inside the exception string of the InvalidCastException.
While CoreCLR reports this string, .NET Native does not report anything meaningful inside
the exception string - only "Arg_InvalidCastException" with no further details.

As such, we need to remove this check as it will never pass in .NET Native - even
though this check will work fine in CoreCLR.

### DIFF
--- a/src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
+++ b/src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
@@ -172,8 +172,8 @@ public class DuplexChannelFactoryTest
             factory.CreateChannel();
         });
 
-        Assert.True(exception.Message.Contains("IRequestChannel"), 
-            string.Format("InvalidCastException exception string should contain 'IRequestChannel'. Actual message:\r\n" + exception.ToString()));
+        // Can't check that the InvalidCastException message as .NET Native only reports this message for all InvalidCastExceptions:
+        // "System.InvalidCastException: Arg_InvalidCastException"
     }
 
     [Fact]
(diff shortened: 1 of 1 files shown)


roncain:
LGTM

src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
```diff
@@ -101,7 +101,8 @@ public static void CreateChannel_BasicHttpBinding_Throws_InvalidOperation()
             factory.CreateChannel();
         });
 
-        Assert.True(exception.Message.Contains("BasicHttpBinding"), "InvalidOperationException exception string should contain 'BasicHttpBinding'");
+        Assert.True(exception.Message.Contains("BasicHttpBinding"), 
+            string.Format("InvalidOperationException exception string should contain 'BasicHttpBinding'. Actual message:\r\n" + exception.ToString()));
     }
```
SajayAntony: At this point this is inconsistent since one place we check and the other we don't. I'm ok if we decide not to check the actual message string if we validate the exception type or atleast document the default and the anomaly.

iamjasonp: Fair enough - I documented the anomaly in line 175 now. In general we always want to check for the exception message if possible.

Issue #168, opened 2015-07-14
Define proper try/catch/finally pattern for tests to close communication objects correctly


roncain:
Our tests currently are inconsistent in whether they close channels, factories or ClientBase's.  And those that do close are inconsistent.  At the same time we choose the pattern, we want to decide whether tests are also samples and could be copy&pasted.  If so, the Close pattern should follow our public guidelines.

This issue is to serve as a discussion topic for the right way to handle this and will then be farmed out to a new set of issues to repair the tests.

roncain:
Not required for beta7

SajayAntony:
I believe this has always been the pattern we have recommended. Its not ideal and is not in accordance with the dispose pattern. But for those who use WCF it has always been consistent. 
Try/Catch/Abort - http://blogs.msdn.com/b/drnick/archive/2007/05/04/the-try-catch-abort-pattern.aspx

roncain:
Thanks @SajayAntony .  Now the broader questions for everyone -- is it our intent to make all tests close the channels, channel factories, and ClientBases they open during a test?   And is it our intent that any scenario test could be copy/pasted and used as a sample?  If the answers are "yes" and "yes", then most scenario tests will need to be updated.

My personal opinion is that all tests should explicitly clean-up any resources they acquire.  I'm undecided about how copy/pastable scenario tests should be for samples.

StephenBonikowsky:
I agree with @iamjasonp 's statement elsewhere that we should view our scenario tests as samples, therefore; we should use best practices or established/recommended patterns when writing them.

iamjasonp:
I think the answer is "yes" we want to clean up as best as possible after a test - as a best practice. It's not such a big deal now that we've moved away from a central server for open sourced tests. 

Whether or not a test is going to be copy-and-pastable is debatable, but we've had best practice stating that we don't use `using` in WCF. I think it's best not to confuse the community that it *might* be all right in WCF-CoreFx.

roncain:
Time to get concrete about this proposal.  I propose that every scenario test uses a try/finally pattern, and in the finally it closes all ICommunicationObjects it has acquired following the pattern described at http://blogs.msdn.com/b/drnick/archive/2007/05/04/the-try-catch-abort-pattern.aspx.  That means a Close() in a Try/Catch, and then an Abort if a TimeoutException or CommunicationException occurs.

[...]

PR #200, opened 2015-07-20
Adding FaultContract negative test for Duplex scenario.

### LINKED ISSUE #169: Test required for duplex callback contract using FaultContractAttribute
When using duplex callback contracts, the callback contract could have a FaultContract defined on an operation. We need a test to make sure this works end to end.

StephenBonikowsky:
* Test currently fails, opened Issue 194 and added the ActiveIssue attribute to the test.
Fixes #169

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/src/TestTypes.cs
+++ b/src/System.Private.ServiceModel/tests/Common/src/TestTypes.cs
@@ -688,3 +688,20 @@ public class XmlCompositeType
         set { _stringValue = value; }
     }
 }
+
+public class DuplexTaskReturnServiceCallback : IWcfDuplexTaskReturnCallback
+{
+    public Task<Guid> ServicePingCallback(Guid guid)
+    {
+        // This returns the guid to the service which called this callback.
+        // We could return Task.FromResult(guid) but that means we could execute the 
+        // completion on the same thread. But if someone is using a task it means they 
+        // would potentially have the completion on another thread.
+        return Task.Run<Guid>(() => guid);
+    }
+
+    public Task<Guid> ServicePingFaultCallback(Guid guid)
+    {
+        throw new FaultException<FaultDetail>(new FaultDetail("Throwing a Fault Exception from the Callback method."), "Reason: Testing FaultException returned from Duplex Callback");
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -2,10 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.ServiceModel;
 using System.Text;
+using System.Threading.Tasks;
 using Xunit;
 
 public static class ExpectedExceptionTests
@@ -185,7 +185,7 @@ public static class ExpectedExceptionTests
         }
         finally
         {
-            watch.Stop(); 
+            watch.Stop();
         }
 
         // want to assert that this completed in < 2 s as an upper bound since the SendTimeout is 0 sec
@@ -300,7 +300,7 @@ public static class ExpectedExceptionTests
     [Fact]
     [OuterLoop]
     public static void UnknownUrl_Throws_ProtocolException()
-    { 
+    {
         string protocolExceptionUri = Endpoints.HttpProtocolError_Address;
 
         BasicHttpBinding binding = new BasicHttpBinding();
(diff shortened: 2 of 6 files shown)


src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfDuplexService.cs
```diff
@@ -37,5 +37,27 @@ public Task<Guid> Ping(Guid guid)
...
+            catch (FaultException<FaultDetail> ex)
+            {
+                // Need to throw a new instance of FaultException so that certain fields such as the Action match what the client expects
+                // Otherwise the client will just get a plain FaultException instead of the expected FaultException<FaultDetail>
+                throw new FaultException<FaultDetail>(ex.Detail, ex.Message);
+            }
+            catch (Exception exception)
+            {
+                throw exception;
+            }
```
roncain: Better to just say "throw" rather than "throw exception" so the stack trace is preserved.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -318,4 +318,43 @@ public static void UnknownUrl_Throws_ProtocolException()
...
+        try
+        {
+            var exception = Assert.Throws<FaultException<FaultDetail>>(() =>
+            {
+                using (factory = new DuplexChannelFactory<IWcfDuplexTaskReturnService>(context, binding, new EndpointAddress(Endpoints.Tcp_NoSecurity_TaskReturn_Address)))
+                {
+                    IWcfDuplexTaskReturnService serviceProxy = factory.CreateChannel();
+
+                    Task<Guid> task = serviceProxy.FaultPing(guid);
+                    if ((task as IAsyncResult).AsyncWaitHandle.WaitOne(new TimeSpan(0, 0, 10)))
```
iamjasonp: NIT: use `TimeSpan.FromSeconds(10)` instead of `new TimeSpan(0, 0, 10)` - just more readable.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -318,4 +318,43 @@ public static void UnknownUrl_Throws_ProtocolException()
...
+        binding.Security.Mode = SecurityMode.None;
+
+        DuplexTaskReturnServiceCallback callbackService = new DuplexTaskReturnServiceCallback();
+        InstanceContext context = new InstanceContext(callbackService);
+
+        try
+        {
+            var exception = Assert.Throws<FaultException<FaultDetail>>(() =>
+            {
+                using (factory = new DuplexChannelFactory<IWcfDuplexTaskReturnService>(context, binding, new EndpointAddress(Endpoints.Tcp_NoSecurity_TaskReturn_Address)))
```
iamjasonp: ChannelFactory suffers the same `using` pattern problem as ClientBase - see [ChannelFactory.cs](https://github.com/dotnet/wcf/blob/a891cd38db0fdc2cbd6282a727a23884e7b42472/src/System.Private.ServiceModel/src/System/ServiceModel/ChannelFactory.cs#L171) - so best practice is to avoid using the `using` pattern, see #168.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
mconnew: You've used a magic number for the timeout. Please replace with a named constant.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -318,4 +318,43 @@ public static void UnknownUrl_Throws_ProtocolException()
...
+                {
+                    IWcfDuplexTaskReturnService serviceProxy = factory.CreateChannel();
+
+                    Task<Guid> task = serviceProxy.FaultPing(guid);
+                    if ((task as IAsyncResult).AsyncWaitHandle.WaitOne(new TimeSpan(0, 0, 10)))
+                    {
+                        Guid returnedGuid = task.GetAwaiter().GetResult();
```
mconnew: If the operation times out, it won't throw any exception. Do we want to make the distinction between the test timing out vs the operation completing successfully for some reason?

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfDuplexService.cs
mconnew: I believe that was there to enable setting a breakpoint to test some behavior during test development. The second catch can be removed altogether.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
StephenBonikowsky: Good idea.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfDuplexService.cs
StephenBonikowsky: Thanks for catching that.

StephenBonikowsky:
I believe I addressed everybody's feedback, squashed and updated the PR, if there is no additional feedback it can be merged after a couple shipits.

src/System.Private.ServiceModel/tests/Common/src/TestTypes.cs
```diff
@@ -688,3 +688,20 @@ public string StringValue
...
+        // This returns the guid to the service which called this callback.
+        // We could return Task.FromResult(guid) but that means we could execute the 
+        // completion on the same thread. But if someone is using a task it means they 
+        // would potentially have the completion on another thread.
+        return Task.Run<Guid>(() => guid);
+    }
+
+    public Task<Guid> ServicePingFaultCallback(Guid guid)
+    {
+        throw new FaultException<FaultDetail>(new FaultDetail("Throwing a Fault Exception from the Callback method."), "Reason: Testing FaultException returned from Duplex Callback");
```
mconnew: From my testing and fixing of bugs based on this new test, we need another slightly different variation. This callback throws as soon as it's called which means it doesn't return a faulted Task. A second variation where it uses `Task.FromException` to return a faulted Task would be beneficial as they behave differently in the product.

mconnew: Btw, this isn't a blocker, just an additional test variation that needs to be written.

mconnew:
:shipit:

iamjasonp:
LGTM

Issue #198, opened 2015-07-20
Bridge should close gracefully after configurable max idle time


roncain:
We should define a new TestProperty (BridgeMaxIdleTimeSeconds?) that defines how long the Bridge will stay alive after it has received a request.  No requests for that amount of time mean the Bridge should close gracefully.  The special value of 0 should mean it never times out.

Once #185 is resolved, this property will flow naturally to the Bridge during configuration.  So this task involves defining the property and then using that property in Bridge.exe.

SajayAntony:
Given that bridge times out in 10 minutes if there are no put/get requests. I think all that is pending is to pass in this timeout values.

roncain:
I'm working on passing in the name/value pairs now and can add this to TestProperties to pass to the config POST.

PR #201, opened 2015-07-20
Enable IdleTimeout for Bridge.exe . Part of #198

SajayAntony:
Default timeout is 10 minutes to avoid lingering processes. 
Ideally build.cmd should be calling ensurebridge just before executing the tests.

### DIFF
--- a/tools/Bridge/Bridge/IdleTimeoutManager.cs
+++ b/tools/Bridge/Bridge/IdleTimeoutManager.cs
@@ -1,0 +1,136 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Web.Http;
+
+namespace Bridge
+{
+    public class IdleTimeoutHandler : DelegatingHandler
+    {
+        IdleTimeoutManager _timeoutManager;
+
+        private IdleTimeoutHandler(TimeSpan idleTimeout)
+        {
+            _timeoutManager = new IdleTimeoutManager(idleTimeout); ;
+            _timeoutManager.OnTimeOut += (s, e) =>
+            {
+                Console.WriteLine("Timed out as there were no messages to the bridge for {0} seconds", (int)e.TotalSeconds);
+                Environment.Exit(-1);
+            };
+        }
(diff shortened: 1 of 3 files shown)


tools/Bridge/Bridge/IdleTimeoutManager.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        IdleTimeoutManager _timeoutManager;
+
+        private IdleTimeoutHandler(TimeSpan idleTimeout)
+        {
+            _timeoutManager = new IdleTimeoutManager(idleTimeout); ;
+            _timeoutManager.OnTimeOut += (s, e) =>
+            {
+                Console.WriteLine("Timed out as there were no messages to the bridge for {0} seconds", (int)e.TotalSeconds);
+                Environment.Exit(-1);
```
iamjasonp: NIT: Consider defining exit codes as an enum or some similar mechanism, e.g. `Success = 0, TimedOut = -1`, so we're not confused about why we're exiting. Also future-proofing in case we start having other exit codes.

tools/Bridge/Bridge/IdleTimeoutManager.cs
```diff
@@ -0,0 +1,136 @@
...
+                _waitTimeout = idleTimeout;
+                _timer = new Timer(this.TimeOutCallback);
+                this.Restart();
+            }
+
+            public IDisposable Start()
+            {
+                if (Update(Timeout.Infinite))
+                {
+                    return new OperationScope(this);
```
roncain: Just confirming my understanding -- once an OperationScope starts, it has inifinite time to complete the operation, but after it is disposed, it restarts the timer for the configured 10 minute wait.

roncain:
LGTM

tools/Bridge/Bridge/IdleTimeoutManager.cs
SajayAntony: Yes. We don't monitor operations themselves. I didn't want a full sever implementation. So if one request to bridge starts then we avoid it being torn down for about 10 mins.

tools/Bridge/Bridge/IdleTimeoutManager.cs
SajayAntony: Hmm thought about it a bit more and decided to avoid adding a file for a single constant. 
I think we will have to consolidate exit codes when we have more than one.

PR #202, opened 2015-07-20
Make Bridge use configurable location for resources

### LINKED ISSUE #186: Make Bridge use TestProperties.BridgeResourceFolder_PropertyName
There is a reserved property called TestProperties.BridgeResourceFolder_PropertyName which is intended to be where the resources for the Bridge live.

This task has to do these things:
1. Ensure wcf.targets uses the correct value for $(BridgeResourceFolder)
2. Ensure the projects that build Bridge resources set their $(OutputPath) to $(BridgeResourceFolder) or do some post-build step to copy them there.
2. Ensure BridgeClient uses TestProperties.GetProperty( TestProperties.BridgeResourceFolder_PropertyName ) as the resources folder it passes to the POST to configure the Bridge.

roncain:
TestProperties contains a configurable property for the
location of the Bridge's resources.  This PR makes the
WcfService compile to that location, and the Bridge client
configures the Bridge using this location.

Fixes #186

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
+++ b/src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
@@ -50,8 +50,10 @@ namespace System.ServiceModel.Tests.Common
             using (HttpClient httpClient = new HttpClient())
             {
                 httpClient.BaseAddress = new Uri(BridgeBaseAddress);
+                string resourceFolder = TestProperties.GetProperty(TestProperties.BridgeResourceFolder_PropertyName);
+                string contentPayload = "{ resourcesDirectory : \"" + resourceFolder + "\" }"; 
                 var content = new StringContent(
-                    @"{ resourcesDirectory : ""WcfService"" }",
+                    contentPayload,
                     Encoding.UTF8,
                     "application/json");
                 try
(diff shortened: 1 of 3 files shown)


src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
```diff
@@ -50,8 +50,10 @@ private static void MakeConfigRequest()
             using (HttpClient httpClient = new HttpClient())
             {
                 httpClient.BaseAddress = new Uri(BridgeBaseAddress);
+                string resourceFolder = TestProperties.GetProperty(TestProperties.BridgeResourceFolder_PropertyName);
+                string contentPayload = "{ resourcesDirectory : \"" + resourceFolder + "\" }";
```
roncain: I use string addition here because String.Format throws FormatException for the squiggly braces.  In a coming PR, this payload will be a set of name value pairs and will replace this concat statementt.

hongdai:
LGTM

src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
dmetzgar: To put squigglies in string.Format you just need to double them up: string.Format("{{ resources: \"{0}\" }}"

dmetzgar:
LGTM

SajayAntony:
:shipit:

iamjasonp:
LGTM

PR #177, opened 2015-07-16
Modify csproj and project.json files to use Contract references

iamjasonp:
Changes all our test projects to use contract references rather than P2P references. 

Not converted

````
src\System.Private.ServiceModel\tests\Scenarios\SelfHostWcfService\WcfService.csproj
````

Converted: 

````
src\System.Private.ServiceModel\tests\Common\src\System.ServiceModel.Tests.Common.csproj
src\System.Private.ServiceModel\tests\Common\tests\System.ServiceModel.Common.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Binding\Custom\Binding.Custom.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Binding\Http\Binding.Http.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Client\ChannelLayer\Client.ChannelLayer.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Client\ClientBase\Client.ClientBase.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Client\ExpectedExceptions\Client.ExpectedExceptions.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Client\TypedClient\Client.TypedClient.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Contract\Data\Contract.Data.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Contract\Fault\Contract.Fault.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Contract\Message\Contract.Message.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Contract\XmlSerializer\Contract.XmlSerializer.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Encoding\Encoders\Encoding.Encoders.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Encoding\MessageVersion\Encoding.MessageVersion.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\Security\TransportSecurity\Security.TransportSecurity.Tests.csproj
src\System.Private.ServiceModel\tests\Scenarios\ServiceModel.Scenarios.Common\ServiceModel.Scenarios.Common.csproj
src\System.Private.ServiceModel\tests\Unit\System.Private.ServiceModel.Tests.csproj
[...]

PR #203, opened 2015-07-21
Fixup p2p references and project.json files after #177

iamjasonp:
5fe8647 (#200) was merged around the same time as #177, but 5fe8647 was using a FaultException ctor that doesn't exist in the contract. When #177 was merged, this caused a build break.

After #177, `ExpectedExceptions/project.json` omitted a couple of project references. For some reason though, this issue didn't show up in our previous buddy builds or even in CI. 

All this should be caught during development time now that #177 is fixed.


iamjasonp:
cc: @roncain @StephenBonikowsky - please review 
Merging even without LGTMs, as this is breaking builds.

PR #204, opened 2015-07-21
Trigger TestProperties code-gen from CoreCompile

roncain:
The TestProperties code-gen target was previously listed
as a DefaultTarget and executed when the project was built.

However the build tools Coverage targets execute only the
Build target and bypass this code-gen target.

This change moves this target into $(CoreCompileDependsOn) to
ensure it is called by any form of build.


iamjasonp:
LGTM.

PR #205, opened 2015-07-21
Make self-hosted test service works for Kerbero

### LINKED ISSUE #193: First step for Issue #59:Create WCF test service capable of testing Kerberos
The first step is making self-hosted test service works for Kerberos and  integrate with the Bridge work being done by @dmetzgar to create a resource that will launch the necessary WCF service on demand, and then to add the necessary scenario tests to invoke it.

hongdai:
* Make self-hosted test service works for Kerberos
* Integrate with Bridge
* Make the existing test work on single machine

Fix #193

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
@@ -102,8 +102,9 @@ public static class Https_ClientCredentialTypeTests
         Assert.True(errorBuilder.Length == 0, String.Format("Test Case: NtlmAuthentication FAILED with the following errors: {0}", errorBuilder));
     }
 
+    // Issue 53 will enable us optional run this test on domain machines.
     [Fact]
-    [ActiveIssue(6)]
+    [ActiveIssue(53)]
     [OuterLoop]
     public static void WindowsAuthentication_RoundTrips_Echo()
     {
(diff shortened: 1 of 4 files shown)


src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/TestResources/EndpointResource.cs
```diff
@@ -32,10 +32,22 @@ public object Put()
                     {
                         host = new ServiceHost(typeof(ServiceType));
                         // URI assumes localhost but that instead should be passed in
+                        // For Kerberos, localhost cannot be used as localhost will fall back to NTLM
+                        // this is a workaround until the URI is passed in
+                        Binding binding = GetBinding();
+                        string baseAddress = "localhost";
+                        if (binding is BasicHttpBinding)
```
roncain: I don't think this logic belongs here in the base class.  We need a new subclass of EndpointResource (HttpWindowsResource?) that knows to use the MachineName.  If we need to add to the EndpointResource API to provide a virtual method to build the URI, then I'd go ahead and do that at the same time.

SajayAntony: +1  having and overriable method that builds the service host would be ideal and Put just calls it and have the Httpkerb method implement and know about these types. Checking for binding and special casing in the base class makes this fragile in the long run.

hongdai: This logic is temporary workaround and should be removed once URI is passed in. I'd prefer to keep it here for easier removal.

roncain: I think it should be fixed now. The issue for passing in the URL should not also be responsible for refactoring this code.  It means we need a virtual method to compose the URI -- and that is what the URI-passing issue expects to fix.

SajayAntony: With #207 you should be able to override just the host part. Was there anything else besides that?
IN `HttpWindowsResource` you should be able to just do 

```csharp
override string Host {get { return Environment.MachineName; }}
```

dmetzgar:
LGTM, I like the ability to change the host.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/TestResources/EndpointResource.cs
hongdai: I have rebased the PR with #207 and override Host.

SajayAntony:
:shipit:

PR #206, opened 2015-07-21
Update build tools to version 61

roncain:
Code coverage builds were broken with the previous
version of build-tools.  Version 61 fixes that break.


SajayAntony:
:shipit:

PR #207, opened 2015-07-21
Test fix to expose all parts of the host listener uri for #205

SajayAntony:
@roncain  @hongdai  does this allow you to override the host part? for #205


roncain:
If this meets what @hongdai needs, it LGTM

SajayAntony:
Will wait for @hongdai's ack.

hongdai:
Ack. I can just override the host part.

Issue #199, opened 2015-07-20
Create test for streamed mode


roncain:
Issue #37 requires a scenario test to verify the fix.  This will require creating a new endpoint that sets the TransferMode to streamed, meaning the Bridge code will need a new endpoint, and the Endpoints.cs class will need to expose it

SajayAntony:
@shmao  @dmetzgar need readme to add a new endpoint.

shmao:
Fixed with #209.

Issue #37, opened 2015-05-27
Support StreamFormatter.SerializeAsyncResult


roncain:
StreamFormatter.SerializeAsyncResult currently throws PlatformNotSupportedException because it required API not available via public contracts.  The external implementation it required is small enough that the implementation can be duplicated locally in StreamFormatter

Issue https://github.com/dotnet/wcf/issues/7 (Support TransferMode.Streamed) depends on this issue being done first.

shmao:
Fixed with #209

PR #209, opened 2015-07-21
Support TransferMode.Streamed.

### LINKED ISSUE #7: Support TransferMode.Streamed in Http and Tcp
The only supported transfer mode on any binding is TransferMode.Buffered.
Not supported are Streamed, StreamedRequest or StreamedResponse

shmao:
The fix is to enable Streamed transfer mode. 

Fixes #7, #37, and #199.


shmao:
@roncain @iamjasonp

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TextMessageEncoder.cs
```diff
@@ -546,7 +546,9 @@ public override Task<ArraySegment<byte>> WriteMessageAsync(Message message, int
                     await message.WriteMessageAsync(xmlWriter);
                     xmlWriter.WriteEndDocument();
                 }
-                await xmlWriter.FlushAsync();
+
+                // The underlying type of xmlWriter, XmlUTF8TextWriter, has not implemented FlushAsync() yet.
+                xmlWriter.Flush();
```
iamjasonp: This is really crappy if we suddenly go sync while in a supposedly async path. Can you please open up a bug to track this implementation in the corefx repo?

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +175,163 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
...
+
+            if (!success)
+            {
+                errorBuilder.AppendLine(String.Format("    Error: expected response from service: '{0}' Actual was: '{1}'", true, success));
+            }
+        }
+        catch (Exception ex)
+        {
+            errorBuilder.AppendLine(String.Format("    Error: Unexpected exception was caught while doing the basic echo test for variation...\n'{0}'\nException: {1}", variationDetails, ex.ToString()));
+        }
```
iamjasonp: Avoid this try/catch format - look at f2fd28a to see how code was changed to avoid the try - catchall pattern. 

We are standardizing on using something like 

```
try {
    // code
} finally {
    if (factory != null && factory.State != CommunicationState.Closed)
    {
        factory.Abort();
    }
}
... 
See dotnet/wcf#168 for details.

File: src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs [hunk 3]
```diff
@@ -94,7 +103,10 @@ internal class SerializeAsyncResult : AsyncResult
             {
                 _streamFormatter = streamFormatter;
                 _writer = writer;
-                throw ExceptionHelper.PlatformNotSupported("StreamFormatter.SerializeAsyncResult is not supported.");
+
+                // As we use the Task-returning method for async operation,
+                // we shouldn't get to this point. Throw exception just in case.
+                throw new NotImplementedException();
```
iamjasonp: `throw ExceptionHelper.AsError(NotImplemented.ByDesign);` if we'll never get here. Consider getting rid of the rest of the implementation if the method shouldn't be implemented at all.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
```diff
@@ -348,5 +360,81 @@ public void ReleaseStream(Stream stream)
...
+                int blockSize = 256;
+                int bytesRead = 0;
+                byte[] block = new byte[blockSize];
+                while (true)
+                {
+                    bytesRead = await stream.ReadAsync(block, 0, blockSize);
+                    if (bytesRead > 0)
+                    {
+                        // XmlDictionaryWriter has not implemented WriteBase64Async() yet.
+                        writer.WriteBase64(block, 0, bytesRead);
```
iamjasonp: Please open a bug in the corefx repo to track implementation of this.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +175,163 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
         }
     }
 
+    [Fact]
```
iamjasonp: @StephenBonikowsky, can you comment on the suitability of putting the test in this part of the tree?

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +175,163 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
...
+
+            if (!success)
+            {
+                errorBuilder.AppendLine(String.Format("    Error: expected response from service: '{0}' Actual was: '{1}'", true, success));
+            }
+        }
+        catch (Exception ex)
+        {
+            errorBuilder.AppendLine(String.Format("    Error: Unexpected exception was caught while doing the basic echo test for variation...\n'{0}'\nException: {1}", variationDetails, ex.ToString()));
+        }
```
iamjasonp: As above: avoid catch-all blocks - I'll stop commenting on the rest of them.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +175,163 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
...
+    private static string StreamToString(Stream stream)
+    {
+        var reader = new StreamReader(stream, Encoding.Unicode);
+        return reader.ReadToEnd();
+    }
+
+    private static Stream StringToStream(string str)
+    {
+        var ms = new MemoryStream();
+        var sw = new StreamWriter(ms, Encoding.Unicode);
```
iamjasonp: Any particular reason we're using Unicode here rather than, say, Encoding.UTF8?

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/OperationFormatter.cs
```diff
@@ -342,6 +343,22 @@ private void SerializeBodyContents(XmlDictionaryWriter writer, MessageVersion ve
...
+
+            SetupStreamAndMessageDescription(isRequest, out streamFormatter, out messageDescription);
+
+            if (streamFormatter != null)
+            {
+                await streamFormatter.SerializeAsync(writer, parameters, returnValue);
+                return;
+            }
+
+            SerializeBody(writer, version, RequestAction, messageDescription, returnValue, parameters, isRequest);
```
iamjasonp: This may be something to take on later... serialization can take a while, we might need to have a virtual SerializeBodyAsync that delegates to the sync SerializeBody, but allows derived OperationFormatters to implement a SerializeBodyAsync.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BodyWriter.cs
```diff
@@ -80,6 +81,10 @@ internal BodyWriter OnCreateBufferedCopy(int maxBufferSize, XmlDictionaryReaderQ
         }
 
         protected abstract void OnWriteBodyContents(XmlDictionaryWriter writer);
+        protected virtual Task OnWriteBodyContentsAsync(XmlDictionaryWriter writer)
+        {
+            throw new NotImplementedException();
```
iamjasonp: This should be delegated in the default case to the sync OnWriteBodyContents, rather than throwing NotImplemented. 

We will probably need to plumb this async change in through our derived BodyWriters too. 

`BodyWriter` is public, so we need to be sure that people don't get nasty surprises if someone ends up calling this through some other async method. 

(Note: `throw ExceptionUtility.ThrowHelperError(NotImplemented.ByDesign)` rather than throwing a NIE)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TextMessageEncoder.cs
roncain: Question for @mconnew -- would it be better to await a Task.Run invoking the xmlWriter.Flush()?

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
roncain: +1 to use the finally pattern but don't agree we have the correct pattern yet.  See #168 for discussion.  When we decide, we'll revisit all the finally blocks.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
roncain: We're also trying to settle on the xunit asserts directly, so Assert.Equal(result, testString) should be good enough.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/OperationFormatter.cs
roncain: I like @iamjasonp 's proposal.  I recommend checking in as-is and opening a new issue to make the appropriate virtual methods on OperationFormatter so the derived classes can do different things.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BodyWriter.cs
roncain: Recommend making @iamjasonp 's fixes and opening a new issue to consider updating the other derived BodyWriters to use the new virtual for async

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
```diff
@@ -186,7 +186,7 @@ protected override Task<Stream> CreateContentReadStreamAsync()
 
         protected override Task SerializeToStreamAsync(Stream stream, TransportContext context)
         {
-            return _messageEncoder.WriteMessageAsync(_message, _stream);
+            return _messageEncoder.WriteMessageAsync(_message, stream);
```
roncain: The code being replaced looks odd.  It was serializing the _stream instance member, not the passed in stream.  If it turns out they are the same stream, I recommend adding a Contract.Assert(Object.ReferenceEquals(stream, _stream)).

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
```diff
@@ -348,5 +360,81 @@ public void ReleaseStream(Stream stream)
                 //Noop
             }
         }
+
+        internal class StreamFormatterHelper
```
roncain: Recommend adding comments whether this code was duplicated from somewhere else and why (i.e. not available in public contract).

src/System.Private.ServiceModel/tests/Common/src/ServiceInterfaces.cs
```diff
@@ -47,6 +48,18 @@ public interface IWcfService
 
     [OperationContract(Action = "http://tempuri.org/IWcfService/GetRequestCustomHeader", ReplyAction = "*")]
     string GetRequestCustomHeader(string customHeaderName, string customHeaderNamespace);
+
+    [OperationContract]
+    Stream GetStream(string data);
+
+    [OperationContract]
+    bool UploadStream(Stream stream);
+
```
roncain: A 'bool' return isn't very informative.  I recommend it return the stream as a string so the test can verify it was received correctly.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +175,163 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
...
+            errorBuilder.AppendLine(String.Format("    Error: Unexpected exception was caught while doing the basic echo test for variation...\n'{0}'\nException: {1}", variationDetails, ex.ToString()));
+        }
+
+        Assert.True(errorBuilder.Length == 0, "Test case FAILED with errors: " + errorBuilder.ToString());
+    }
+
+    [Fact]
+    [OuterLoop]
+    public static void DefaultSettings_Echo_RoundTrips_String_Streamed()
+    {
```
roncain: Nit: DefaultSettings_Echo_RoundTrips_Streamed_Request would probably be better.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +175,163 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
...
+        {
+            ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
+
+            IWcfService serviceProxy = factory.CreateChannel();
+
+            Stream stream = StringToStream(testString);
+            var returnStream = serviceProxy.EchoStream(stream);
+            var result = StreamToString(returnStream);
+            success = string.Equals(result, testString);
+
```
roncain: Same comment about using Assert.Equal(result, testString) directly and using the finally block.  I'll stop pointing those out.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +175,163 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
...
+            errorBuilder.AppendLine(String.Format("    Error: Unexpected exception was caught while doing the basic echo test for variation...\n'{0}'\nException: {1}", variationDetails, ex.ToString()));
+        }
+
+        Assert.True(errorBuilder.Length == 0, "Test case FAILED with errors: " + errorBuilder.ToString());
+    }
+
+    [Fact]
+    [OuterLoop]
+    public static void DefaultSettings_Echo_RoundTrips_String_Streamed_Async()
+    {
```
roncain: Nit: DefaultSettings_Echo_RoundTrips_Streamed_Request_And_Response_Async might be better

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.cs
```diff
@@ -234,5 +236,46 @@ public XmlCompositeType GetDataUsingXmlSerializer(XmlCompositeType composite)
...
+            return StringToStream(data);
+        }
+
+        public bool UploadStream(Stream stream)
+        {
+            string data = StreamToString(stream);
+
+            if (!string.IsNullOrWhiteSpace(data))
+            {
+                return true;
```
iamjasonp: We should be testing that the stream, with some content, arrived properly, right?

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.cs
```diff
@@ -234,5 +236,46 @@ public XmlCompositeType GetDataUsingXmlSerializer(XmlCompositeType composite)
...
+        public Stream GetStream(string data)
+        {
+            return StringToStream(data);
+        }
+
+        public bool UploadStream(Stream stream)
+        {
+            string data = StreamToString(stream);
+
+            if (!string.IsNullOrWhiteSpace(data))
```
roncain: See other comment -- I recommend we just return 'data' directly and let the test deal with testing whether it was right.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.cs
iamjasonp: ... and on that note, maybe you need to define some magic stream contents, or return the stream back with the original data to test the StreamedResponse.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.cs
shmao: Agree. I will make it return 'data'.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.cs
shmao: Yes, I will return the 'data' back.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
shmao: No, just make the client/server use the same encoding. Do we want to change it to UTF8?

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
shmao: 'Streamed' here means that the TransferMode is Streamed, not StreamedRequest or StreamedResponse. Or I can change the name to DefaultSettings_Echo_RoundTrips_String_TransferModeStreamed

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
iamjasonp: UTF8 is what I've generally used in the past and what's out in the wild a lot; I'm not too concerned about it either way though.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
shmao: The old code seems to be a bug. I checked this with Matt and he confirmed this was a bug. I found this because I got a NullRef exception here. Those two variables are not pointing to the same object.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
roncain: Sorry I was unclear -- I only meant that a Stream goes out as a request, and a Stream comes back as a response.  That is different than the tests that only send the Stream in the request or accept the Stream in the response.  I wanted to make it clear when a Stream was being round-tripped entirely as a Stream and not being converted to or from a string by the server.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.cs
iamjasonp: Careful... if you return `data` back as a string, you're not going to be testing StreamedResponse. Depends what your intention for the test is :smile:

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
shmao: Thanks for clarifying this. I thought it would be better to make the name indicate the TransferMode being used because we can roundtrip streams even with TransferMode.Buffered.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.cs
shmao: Thanks for the heads-up. This method is used for testing StreamedRequest only.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
shmao: I will throw the exception in the standard way. I thought it would be better to make the method throw NotImplemented.ByDesign for now just in case on any path we call this method.

src/System.Private.ServiceModel/tests/Common/src/ServiceInterfaces.cs
shmao: Will make it return the 'data'.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.cs
roncain: @iamjasonp raises a great point.  Regardless of the kind of data being sent or returned, we want to test each permutation of TransferMode is being used properly.  The case where Stream "auto chooses" a transfer mode is a special case (I think). After this is checked in, we should articulate which scenarios require a server endpoint that itself sets TransferMode.  I think we may need a new HttpResource devoted to the different TransferModes (and verify simple and complex types stream both ways properly.  I think we will also need TCP flavors of all the streamed test, but let's make that a new issue and not tackle it here.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
roncain: Just a question -- does one of these test a Stream round-tripping using TransferMode.Buffered?  If not we can open a new test issue

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: There are two ways that HttpClient can call into an HttpContent class. One is to call CreateContentReadStreamAsync, and the other is to call SerializeToStreamAsync. The first way expects a stream to be returned that the content can be read from. The second passes in a stream that the content should be written to so doesn't use the _stream field. The .Net Native implementation of HttpClient uses the first method while the coreclr implementation based on WinHTTP uses the second method.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TextMessageEncoder.cs
mconnew: @roncain, I would prefer to keep the call to xmlWriter.FlushAsync and let the xmlWriter decide what to do. If the xmlWriter is doing nothing right now, then that's a bug that needs to be fixed. But the fix could simply be for xmlWriter to either directly call the synchronous Flush method or to call Task.Run(). xmlWriter is the correct place to decide what to do as it is likely easy to check if there's anything to flush and if there isn't, return, otherwise call Task.Run to invoke Flush(). This wouldn't require converting everything to async and would allow a fast path for the no-op situation avoiding starting a new task on another thread just to no-op on that thread and mark the Task complete.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/OperationFormatter.cs
mconnew: :+1:

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/OperationFormatter.cs
```diff
@@ -698,6 +715,11 @@ protected override void OnWriteBodyContents(XmlDictionaryWriter writer)
                     }
                 }
 
+                protected override async Task OnWriteBodyContentsAsync(XmlDictionaryWriter writer)
+                {
+                    await _operationFormatter.SerializeBodyContentsAsync(writer, _version, _parameters, _returnValue, _isRequest);
```
mconnew: When the method body solely consists of awaiting another method, just return the Task from that method and remove the async keyword. E.g.:  
```csharp
protected override Task OnWriteBodyContentsAsync(XmlDictionaryWriter writer)
{
    return _operationFormatter.SerializeBodyContentsAsync(writer, _version, _parameters, _returnValue, _isRequest);
}
```

shmao: Good point. I will make this change.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
```diff
@@ -49,7 +49,16 @@ private StreamFormatter(MessageDescription messageDescription, MessagePartDescri
...
             Stream streamValue = GetStreamAndWriteStartWrapperIfNecessary(writer, parameters, returnValue);
-            writer.WriteValue(new OperationStreamProvider(streamValue));
+            var streamProvider = new OperationStreamProvider(streamValue);
+            StreamFormatterHelper.WriteValue(writer, streamProvider);
+            WriteEndWrapperIfNecessary(writer);
+        }
+
+        internal async Task SerializeAsync(XmlDictionaryWriter writer, object[] parameters, object returnValue)
+        {
+            Stream streamValue = GetStreamAndWriteStartWrapperIfNecessary(writer, parameters, returnValue);
```
mconnew: Should we have an async variation of GetStreamAndWriteStartWrapperIfNecessary? Not for this change, but if we should (it looks like we're writing to the stream in that method), then make sure this is recorded an a relevant bug somewhere.

shmao: I don't think an async version of this method is necessary: 1. we writes only the start wrapper, which should be very small, into the stream in the method; 2. the desktop version of the BeginSerialize method uses this same method.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/WcfService.cs
roncain: To respond to an earlier comment -- the server should *never* participate in the testing, such as verifying it got some magic content.  Server side code should instead be written to return something to the client so the test can verify what the server received was correct.  The operation can be written to control what the server does with its inputs, but the server should never directly know anything about how the calling test behaves.

iamjasonp:
Also, please revise your commit comment prior to merging the PR - omit the "To"

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
StephenBonikowsky: Based on following the general same tree structure as is used by the published WCF Samples, the Streaming sample is located 'WF_WCF_Samples\WCF\Basic\Contract\Service'. Service in this case refers to "Service Contract". So we would need to add a new directory under 'System.Private.ServiceModel\tests\Scenarios\Contract'.

shmao: Thanks @StephenBonikowsky . I will put the tests under Contract\Service.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BodyWriter.cs
shmao: Changed the virtual async method to call the sync method. Opened #217.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/OperationFormatter.cs
shmao: Opened #218 per Ron's suggestion.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
shmao: Added comments.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
shmao: I just added a test case covering Stream round-tripping using TransferMode.Buffered.

shmao:
Hi Folks, I've updated the PR. Can you please have a look? Thanks!

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TextMessageEncoder.cs
iamjasonp: :+1: can we get this calling FlushAsync() instead?

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/Contract.Service.Tests.csproj
```diff
@@ -0,0 +1,40 @@
...
+    <Compile Include="ServiceContractTests.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="project.json" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include='$(WcfSourceProj)'>
+      <Project>{9e50e7bf-cd6e-4269-a6dd-59fd0bd6c0fd}</Project>
+      <Name>System.Private.ServiceModel</Name>
+    </ProjectReference>
```
iamjasonp: Should read: 

```
    <!-- Compile tests against the contract, but copy our local-built implementation for testing -->
    <ProjectReference Include='$(WcfSourceProj)'>
      <Project>{9e50e7bf-cd6e-4269-a6dd-59fd0bd6c0fd}</Project>
      <Name>System.Private.ServiceModel</Name>
      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
      <OutputItemType>Content</OutputItemType>
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
      <Targets>Build;DebugSymbolsProjectOutputGroup</Targets>
    </ProjectReference>
```

unless there's a really good reason for using the P2P references. Using the above ensures that the tests are testing only the surface area that is available in the public contract.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/project.json
```diff
@@ -0,0 +1,64 @@
...
+    "xunit": "2.0.0-beta5-build2785",
+    "xunit.abstractions.netcore": "1.0.0-prerelease",
+    "xunit.assert": "2.0.0-beta5-build2785",
+    "xunit.core.netcore": "1.0.1-prerelease",
+    "xunit.netcore.extensions": "1.0.0-prerelease-*"
+  },
+  "frameworks": {
+    "dnxcore50": {}
+  }
+}
```
iamjasonp: There was a huge exercise to reduce the dependency list. See [this](https://github.com/dotnet/wcf/blob/master/src/System.ServiceModel.Http/tests/project.json) for an example

iamjasonp: Please also update the project.lock.json after reducing the dependency list in project.json

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/Contract.Service.Tests.csproj
```diff
@@ -0,0 +1,40 @@
...
+      <Project>{e896294a-ab4a-4af5-a01c-a19e3972eff9}</Project>
+      <Name>System.ServiceModel.Tests.Common</Name>
+    </ProjectReference>
+    <ProjectReference Include="..\..\ServiceModel.Scenarios.Common\ServiceModel.Scenarios.Common.csproj">
+      <Project>{c2210035-aab0-4dbb-9a42-6a23c0bd763b}</Project>
+      <Name>ServiceModel.Scenarios.Common</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+</Project>
```
iamjasonp: NIT: need newline.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/IWcfService.cs
```diff
@@ -65,5 +66,14 @@ internal interface IWcfService
 
         [OperationContract(Action = "http://tempuri.org/IWcfService/GetDataUsingXmlSerializer"), XmlSerializerFormat]
         XmlCompositeType GetDataUsingXmlSerializer(XmlCompositeType composite);
+
+        [OperationContract]
+        Stream GetStream(string data);
```
iamjasonp: NIT: Consider naming this `GetStreamFromString`, as the meaning of this endpoint will eventually become lost as we add more tests in. Same with the `UploadStream` - consider naming something like `GetStringFromStream`

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TextMessageEncoder.cs
shmao: We can't at this point as the API is not implemented yet.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/Contract.Service.Tests.csproj
shmao: Thanks. Added this.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/project.json
shmao: Updated.

src/System.Private.ServiceModel/tests/Scenarios/SelfHostWcfService/IWcfService.cs
shmao: Sounds good. Changed the names.

iamjasonp:
:shipit: says: LGTM

StephenBonikowsky:
:shipit:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BodyWriter.cs
```diff
@@ -80,6 +81,11 @@ internal BodyWriter OnCreateBufferedCopy(int maxBufferSize, XmlDictionaryReaderQ
         }
 
         protected abstract void OnWriteBodyContents(XmlDictionaryWriter writer);
+        protected virtual Task OnWriteBodyContentsAsync(XmlDictionaryWriter writer)
```
SajayAntony: NIT: new line

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TextMessageEncoder.cs
SajayAntony: The api is not exposed by the XmlWriter contract?

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
SajayAntony: IMHO. if you are async any blocking write is a potential issue also with the await model if should complete synchronously if the write happens fast. I would not assume that this is short in any IO path.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TextMessageEncoder.cs
shmao: The API is in the contract, but it’s not implemented yet.

SajayAntony: In that case lets open a bug and give a dummy sync implementation which just returns
Task.FromResult();

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
shmao: Good point. Added GetStreamAndWriteStartWrapperIfNecessaryAsync and WriteEndWrapperIfNecessaryAsync

shmao: However, XmlDictionaryWriter has not implemented WriteStartElementAsync and WriteEndElementAsync yet. Opened #221 for revisiting in the future.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TextMessageEncoder.cs
shmao: Opened #222 for tracking this.

roncain:
LGTM

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
iamjasonp: :+1:

PR #212, opened 2015-07-22
This is the initial commit for the stress tests.

KKhurin:
These stress tests focus on using shared instances of WCF channel factories and channels by multiple threads. 

We group those instances in pools, and we group the scenarios that use them in static classes (with public static methods for easy consumption without any preliminary setup from various stress frameworks).

Some scenarios focus on just making WCF calls using a pool of already created channels while other scenarios open/close/recycle/abort channels and factories on some threads while simultaneously invoking service calls over other threads. 

We use a number of helper methods and classes to wrap the actual calls to WCF factories and channels and service calls in order to be able to verify invariants (e.g. events do get fired only once, things do get closed, etc - this list can be greatly extended to provide a safety net for all kinds of scenarios).

The names of the scenarios should be self-explanatory (e.g. RecyclablePooledFactoriesAndChannelsAsync). Some tests ('*openonce') contain workarounds for stress bugs that we already found but haven't fixed yet.

Finally, there is a small sample on how to run the tests (program.cs).

@iamjasonp 
@roncain
@hongdai
@SajayAntony
@zhenlan
@mconnew
@StephenBonikowsky

### DIFF
--- a/src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
+++ b/src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
@@ -1,0 +1,205 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Diagnostics;
+
+namespace WCFClientStressTests
+{
+    /// <summary>
+    /// This class provides a few Call*AndRecordStats() utility methods which will store and aggregate the timing of the delegates they invoke 
+    /// 
+    /// Note: all Call*AndRecordStats() methods catch all exceptions coming from the delegates
+    /// However we don't want to miss the case when an unexpected exception happens for the calls that should never have exceptions
+    /// </summary>
+    public class CallStats
+    {
+        public CallTimingStatsCollector SunnyDay { get; set; }
+        public CallTimingStatsCollector RainyDay { get; set; }
+
+        public static int SunnyDaySamples = 1000000;
+        public static int RainyDaySamples = 10000;
+
+        public CallStats()
+            : this(SunnyDaySamples, RainyDaySamples, CallTimingStatsCollector.CalculateSomePercentiles, CallTimingStatsCollector.CalculateSomePercentiles)
+        {
+        }
+
+        public CallStats(int samples, int errorSamples)
+            : this(samples, errorSamples, CallTimingStatsCollector.CalculateSomePercentiles, CallTimingStatsCollector.CalculateSomePercentiles)
+        {
+        }
+
+        public CallStats(int samples, int errorSamples, Func<long[], TimingPercentileStats> statsCalculator, Func<long[], TimingPercentileStats> errorStatsCalculator)
+        {
+            SunnyDay = new CallTimingStatsCollector(samples, statsCalculator);
+            RainyDay = new CallTimingStatsCollector(errorSamples, errorStatsCalculator);
+        }
+
+        public void CallActionAndRecordStats(Action action)
+        {
+            Stopwatch sw = new Stopwatch();
+            try
+            {
+                sw.Restart();
+                action();
+                SunnyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+            catch (Exception e)
+            {
+                RainyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+        }
+
+        public T CallFuncAndRecordStats<T>(Func<T> func)
+        {
+            T result = default(T);
+            Stopwatch sw = new Stopwatch();
+            try
+            {
+                sw.Restart();
+                result = func();
+                SunnyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+            catch (Exception e)
+            {
+                RainyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+            return result;
+        }
+
+        public T CallFuncAndRecordStats<T, P>(Func<P, T> func, P param)
+        {
+            T result = default(T);
+            Stopwatch sw = new Stopwatch();
+            try
+            {
+                sw.Restart();
+                result = func(param);
+                SunnyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+            catch (Exception e)
+            {
+                RainyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+            return result;
+        }
+
+        public async Task CallAsyncFuncAndRecordStatsAsync(Func<Task> func)
+        {
+            Stopwatch sw = new Stopwatch();
+            try
+            {
+                sw.Restart();
+                await func();
+                SunnyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+            catch (Exception e)
+            {
+                RainyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+        }
+
+        public async Task CallAsyncFuncAndRecordStatsAsync<P>(Func<P, Task> func, P param)
+        {
+            Stopwatch sw = new Stopwatch();
+            try
+            {
+                sw.Restart();
+                await func(param);
+                SunnyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+            catch (Exception e)
+            {
+                RainyDay.AddCallTiming(sw.ElapsedTicks);
+            }
+        }
+    }
+
+    public class TimingPercentileStats
+    {
+        public DateTime Time;       // Time when the stats were calculated
+        public int[] Centile;       // negative % means take bottom part, positive % means take top part
+        public long[] AvgTiming;    // within this time
+    }
+
+    public class CallTimingStatsCollector
+    {
+        private int _samples;
+        private int _numCalls;
+        private long[] _timings;
+        private Object _lock = new Object();
+        private Func<long[], TimingPercentileStats> _statsCalculator;
+        public ConcurrentQueue<TimingPercentileStats> PercentileStats { get; set; }
+
+        public CallTimingStatsCollector(int samples, Func<long[], TimingPercentileStats> statsCalculator)
+        {
+            _samples = samples;
+            _timings = new long[_samples * 2];
+            _statsCalculator = statsCalculator;
+            PercentileStats = new ConcurrentQueue<TimingPercentileStats>();
+        }
+
+        // For the most part lock-free and isn't particularly thread safe
+        //
+        // However here we use it without any additional synchronization 
+        // since we use a large number of samples and 
+        public void AddCallTiming(long ticks)
+        {
+            int index = Interlocked.Increment(ref _numCalls) % (_samples * 2);
+            _timings[index] = ticks;
+
+            if (index % _samples == 0)
+            {
+                // if we get a context switch here then the data in _timings may get overwritten by other threads calling AddCallTiming()
+                // however no data will be lost if there were fewer than _samples calls while we make a copy of our half of _timings
+                // the copy may land into LOH - before trying to optimize it we'll need to optimize _statsCalculator as well
+                long[] copy = new long[_samples];
+                int timingsIndex = index == 0 ? _samples + 1 : 1;
+                // this almost useless lock will only stop the first thread that wrapped up _samples calls while we copy the data
+                lock (_lock)
+                {
+                    for (int i = 0; i < _samples; i++)
+                    {
+                        int ti = (timingsIndex + i) % (_samples * 2);
+                        copy[i] = _timings[ti];
+                    }
+                }
+                //calculate percentiles
+                PercentileStats.Enqueue(_statsCalculator(copy));
+            }
+        }
+
+        public static TimingPercentileStats CalculateSomePercentiles(long[] timingArr)
+        {
+            var stats = new TimingPercentileStats()
+            {
+                Time = DateTime.Now,
+                Centile = new int[] { -50, 1 },
+                AvgTiming = new long[2]
+            };
+
+            for (int i = 0; i < stats.Centile.Length; i++)
+            {
+                if (stats.Centile[i] < 0)
+                {
+                    int lowestXPctNum = (timingArr.Length * -1 * stats.Centile[i]) / 100;
+                    stats.AvgTiming[i] = timingArr.OrderBy(t1 => t1).Take(lowestXPctNum).Sum(t2 => t2) / lowestXPctNum;
+                }
+                else
+                {
+                    int topXPctNum = (timingArr.Length * stats.Centile[i]) / 100;
+                    stats.AvgTiming[i] = timingArr.OrderByDescending(t1 => t1).Take(topXPctNum).Sum(t2 => t2) / topXPctNum;
+                }
+
+                //Console.WriteLine(String.Format("Percentile {0}, time {1}", stats.Centile[i], stats.AvgTiming[i]));
+            }
+
+            return stats;
+        }
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/IService1.cs
+++ b/src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/IService1.cs
@@ -1,0 +1,47 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+using System.ServiceModel;
+using System.Runtime.Serialization;
+
+
+namespace WcfService1
+{
+    // NOTE: You can use the "Rename" command on the "Refactor" menu to change the interface name "IService1" in both code and config file together.
+    [ServiceContract]
+    public interface IService1
+    {
+        [OperationContract]
+        string GetData(int value);
+
+        [OperationContractAttribute(Name = "GetData")]
+        Task<string> GetDataAsync(int value);
+
+        [OperationContract]
+        CompositeType GetDataUsingDataContract(CompositeType composite);
+    }
+
+
+    // Use a data contract as illustrated in the sample below to add composite types to service operations.
+    [DataContract]
+    public class CompositeType
+    {
+        bool boolValue = true;
+        string stringValue = "Hello ";
+
+        [DataMember]
+        public bool BoolValue
+        {
+            get { return boolValue; }
+            set { boolValue = value; }
+        }
+
+        [DataMember]
+        public string StringValue
+        {
+            get { return stringValue; }
+            set { stringValue = value; }
+        }
+    }
+}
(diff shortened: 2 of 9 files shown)


src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/project.json
```diff
@@ -0,0 +1,46 @@
...
+
+  "commands": {
+    "WCFClientStressTestsCore" : "WCFClientStressTestsCore"
+  },
+
+  "frameworks" : {
+    "dnxcore50" : {
+      "dependencies": {
+        "Microsoft.Test.Stress":  "1.0.0",
+        "System.Console": "4.0.0-beta-*",
```
roncain: My build fails trying to locate Microsoft.Test.Stress

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/project.json
```diff
@@ -0,0 +1,46 @@
...
+        "System.Net.Security": "4.0.0-beta-*",
+        "System.Net.WebHeaderCollection": "4.0.0-beta-*",
+        "System.Reflection.DispatchProxy": "4.0.0-beta-*",
+        "System.Runtime.Serialization.Xml": "4.0.10-beta-*",
+        "System.Runtime.Serialization.Primitives": "4.0.10-beta-*",
+        "System.Security.Cryptography.Hashing": "4.0.0-beta-*",
+        "System.Xml.XmlDocument": "4.0.0-beta-*",
+        "System.ServiceModel.Http": "4.0.10-beta-*",
+        "System.ServiceModel.NetTcp": "4.0.0-beta-*",
+        "System.ServiceModel.Primitives": "4.0.0-beta-*"
```
roncain: Jason has recently updated all the project.json files to use only the dependencies actually used.  Please look at the other test project.json files and see if you can reduce yours to match.  Incidentally, once you build, I expect it to generate a project.lock.json file that will need to be checked in.  Make sure the "locked" property inside that file is set to "true" like the others

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
```diff
@@ -0,0 +1,197 @@
...
+    /// 
+    /// Note: all Call*AndRecordStats() methods catch all exceptions coming from the delegates
+    /// However we don't want to miss the case when an unexpected exception happens for the calls that should never have exceptions
+    /// </summary>
+    public class CallStats
+    {
+        public CallTimingStatsCollector SunnyDay { get; set; }
+        public CallTimingStatsCollector RainyDay { get; set; }
+        public CallStats(int samples = 1000000, int errorSamples = 10000)
+            : this(samples, errorSamples, CallTimingStatsCollector.CalculateSomePercentiles, CallTimingStatsCollector.CalculateSomePercentiles)
```
roncain: These integer values should probably be turned into public static fields at the top file.  As we get further along (and if you can take a dependency on TestProperties), we can find a way to make them configurable from the command line.

KKhurin:
Re-formatted with CodeFormatter and updated with Ron's feedback.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
```diff
@@ -0,0 +1,205 @@
...
+        private long[] _timings;
+        private Object _lock = new Object();
+        private Func<long[], TimingPercentileStats> _statsCalculator;
+        public ConcurrentQueue<TimingPercentileStats> PercentileStats { get; set; }
+
+        public CallTimingStatsCollector(int samples, Func<long[], TimingPercentileStats> statsCalculator)
+        {
+            _samples = samples;
+            _timings = new long[_samples * 2];
+            _statsCalculator = statsCalculator;
```
roncain: No changed needed in this PR, but I'm a little confused by the "_samples * 2" length for timings and the index modulo computations below.  Consider for a future PR comments saying why _timings is twice as long as _samples and why the modulo computations.  I presume the _timings[] array is just a circular buffer that captures the most recent timings, not all timings.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
```diff
@@ -0,0 +1,205 @@
...
+                PercentileStats.Enqueue(_statsCalculator(copy));
+            }
+        }
+
+        public static TimingPercentileStats CalculateSomePercentiles(long[] timingArr)
+        {
+            var stats = new TimingPercentileStats()
+            {
+                Time = DateTime.Now,
+                Centile = new int[] { -50, 1 },
```
roncain: No change needed for this PR, but consider comments explaining these magic numbers (-50 and 1), and why AvgTiming has 2 elements.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/IService1.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System.Threading.Tasks;
+using System.ServiceModel;
+using System.Runtime.Serialization;
+
+
+namespace WcfService1
+{
+    // NOTE: You can use the "Rename" command on the "Refactor" menu to change the interface name "IService1" in both code and config file together.
+    [ServiceContract]
```
roncain: Nit: not for this PR, but this comment can be deleted.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/PoolOfThings.cs
```diff
@@ -0,0 +1,374 @@
...
+            }
+            else
+            {
+                // we are the one who replaced the old _thePool - we're responsible for cleaning it up
+                DestroyAllPooledInstanciesImpl(oldPool, _destroyInstance);
+            }
+        }
+
+        public IEnumerable<T> GetAllPooledInstances()
+        {
```
roncain: Not for this PR but future consideration.  Returning an IEnumerable<T> means the values might be different every time someone evaluates it (e.g 2 Linq lambdas in a row might have different results).  If that is desired behavior, okay, but it might be better to do a .ToArray() at some point to give callers an IEnumerable that does not change

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/PoolOfThings.cs
```diff
@@ -0,0 +1,374 @@
...
+                    {
+                        _destroyInstance(instance);
+                        instance = _thePool[i];
+                        // Somebody stored the new instance before us - dispose ours and re-read theirs
+                        // Since the whole pool can also be replaced we can still get a null from _thePool[index]
+                        // so we retry while (instance == null)
+                        // Note that _destroyInstance(instance) should not throw since nobody uses it yet
+
+                        //Interlocked.Increment(ref _numPoolSetRetries);
+                    }
```
roncain: Not for this code review but future -- can that commented out Interlocked be removed?  Or is it something that should be uncommented?  One style rule for OSS is to avoid "dead code" in comments.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/PoolOfThings.cs
```diff
@@ -0,0 +1,374 @@
...
+                }
+                //catch (Exception)
+                {
```
roncain: For future consideration: either remove the commented out catch (and its curly braces) or uncomment, whichever is correct.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/PoolOfThings.cs
```diff
@@ -0,0 +1,374 @@
...
+        private void DestroyAllPooledInstanciesImpl(T[] instancesPool, Action<T> destroyInstance)
+        {
+            foreach (var instance in instancesPool)
+            {
+                if (instance != null)
+                {
+                    try
+                    {
+                        //Interlocked.Increment(ref _stats.FactoriesClosed);
+                        destroyInstance(instance);
```
roncain: For future consideration -- and I'll stop commenting on it.  These commented out Interlocked statements should either be removed or placed in some sort of conditionally compiled code (for debugging?)

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/PoolOfThings.cs
```diff
@@ -0,0 +1,374 @@
...
+                destroyInstance: destroyObject);
+        }
+
+        public void Destroy()
+        {
+            // should never be recycled more than once
+            if (Factory == null)
+            {
+                System.Diagnostics.Debugger.Break();
+            }
```
roncain: For future consideration -- breaking to the debugger should probably be avoided in stress runs unless you are debugging some problem.  It could stop the test run for what is arguably not a huge issue (double destroy).

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+namespace WCFClientStressTests
+{
+    public class Program
+    {
+        public void Main(string[] args)
+        {
+            ProcessRunOptions(args);
+
+            DoTheRun(3000000);
+
```
roncain: For future consideration -- explain the magic 3000000 or parse it from args.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+        private void ProcessRunOptions(string[] args)
+        {
+            // All hardcoded right now, but we need to extract the following from the parameters:
+            //  -   HostName
+            //  -   AppName
+            //  -   TCP/HTTP
+            //  -   duration and/or # of iterations to run
+            //  -   # factories to pool
+            //  -   # channels to pool
+            //  -   etc
```
roncain: Optional -- open an issue to make the changes described here.  We try to avoid any kind of "TODO" statements in OSS code.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/SharedPoolsOfWCFObjects.cs
```diff
@@ -0,0 +1,612 @@
...
+        private static CallStats s_CloseChannelStats = new CallStats();
+        private static CallStats s_CloseFactoryStats = new CallStats();
+
+        static CreateAndCloseFactoryAndChannelFullCycleTest()
+        {
+        }
+        public static void CreateFactoriesAndChannelsUseAllOnceCloseAll()
+        {
+            using (var theOneTimeThing = new PoolOfThings<FactoryAndPoolOfItsObjects<ChannelFactory<WcfService1.IService1>, WcfService1.IService1>>(
+                    maxSize: 3, // # of pooled FactoryAndPoolOfItsObjects
```
roncain: For future consideration -- make this "3" configurable or a static readonly int somewhere.  It's not clear to the reader why 3 is used.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/SharedPoolsOfWCFObjects.cs
```diff
@@ -0,0 +1,612 @@
...
+        public static void UseChannelsInPooledFactoriesAndChannels()
+        {
+            foreach (var factoryAndPoolOfItsChannels in s_pooledFactoriesAndChannels.GetAllPooledInstances())
+            {
+                foreach (var channel in factoryAndPoolOfItsChannels.ObjectsPool.GetAllPooledInstances())
+                {
+                    s_CallChannelStats.CallActionAndRecordStats(() =>
+                    {
+                        channel.GetData(44);
+                    });
```
roncain: For future consideration -- make '44' some visible const or static readonly field somewhere.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/SharedPoolsOfWCFObjects.cs
```diff
@@ -0,0 +1,612 @@
...
+                }
+            }
+        }
+    }
+
+    public class RecyclablePooledFactoriesAndChannels_OpenOnce
+    {
+        private static int s_iteration = 0;
+        private static CallStats s_CreateChannelStats = new CallStats(samples: 1000, errorSamples: 1000);
+        private static CallStats s_CallChannelStats = new CallStats(samples: 1000000, errorSamples: 10000);
```
roncain: For future consideration -- make these magic numbers visible as const's or static readonly fields

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/SharedPoolsOfWCFObjects.cs
```diff
@@ -0,0 +1,612 @@
...
+                RecyclePooledChannels();
+            }
+            else if (seed < useWeight + recycleChannelsWeight + recycleFactoriesWeight)
+            {
+                RecyclePooledFactories();
+            }
+
+            // move this outside of the test - an external code should query the test for its call stats
+            if (Interlocked.Increment(ref s_iteration) % 1000000 == 0)
+            {
```
roncain: For future consideration -- and I'll stop commenting on them.  Magic literals like 1000000 should be some visible const or static field.  Also, the comment counts as a secret TODO, so it should be turned into an issue if you really intend to do it.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/SharedPoolsOfWCFObjects.xproj
```diff
@@ -0,0 +1,20 @@
...
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+  </PropertyGroup>
+
+  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.Props" Condition="'$(VSToolsPath)' != ''" />
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>4904c84b-3ea5-425b-aa82-c444c24b88c9</ProjectGuid>
+    <RootNamespace>WCFClientStressTestsCore</RootNamespace>
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">..\..\artifacts\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\bin\$(MSBuildProjectName)\</OutputPath>
+  </PropertyGroup>
```
roncain: For future consideration: we can discuss this separately but will want a convention where generated artifacts go to ensure they don't land in random places in the enlistment.  Later, not now.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/project.json
```diff
@@ -0,0 +1,35 @@
...
+    "System.ServiceModel.NetTcp": "4.0.0-beta-*",
+    "System.ServiceModel.Http": "4.0.10-beta-*",
+    "System.ServiceModel.Primitives": "4.0.0-beta-*",
+    "System.ServiceModel.Security": "4.0.0-beta-*",
+    "System.Console": "4.0.0-beta-*",
+    "System.Collections": "4.0.10-beta-*",
+    "System.Linq": "4.0.0-beta-*",
+    "System.Threading": "4.0.10-beta-*",
+    "System.Diagnostics.Debug": "4.0.10-beta-*",
+    "System.Collections.Specialized": "4.0.0-beta-*",
```
roncain: Are you able to remove any of these dependencies?  We have reduced the other project.json files quite a bit.  Not now, later.

roncain:
@dotnet-bot test this please

iamjasonp:

roncain:
LGTM.  I recommend we merge and deal with the feedback later.
@KKhurin our convention is normally to get 2 LGTM's and then let the author do the merge.  This lets them do any last fixups or squashes.  I recommend @mconnew or @SajayAntony or @zhenlan give it the 2nd approval if they are okay.  I have  build and run this PR to verify it doesn't break OuterLoop, but I haven't tried running the stress tests themselves.

zhenlan:
LGTM

SajayAntony:
Do these build in someway that we can validate it using build.cmd?
I would be cautious of getting stress tests or perf tests that have to constantly maintained by dev's when we need to run perf or stress runs. So basically safeguard the source tree from having stale tests. 
That said we could have that a part of this checkin to somehow ensure that build also builds the stress or there is a way to ensure that others can run this. 

@roncain - The fact that you haven't been able to run this kind of means putting into another branch would be a better place and then merge into master once we have confidence to pull it in someway that anyone can build and run this. 

That said :shipit: and maybe put it in another branch for now.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
KKhurin: CallTimingStatsCollector was "invented" as a short-term substitution of a much more thorough and hopefully well thought out solution for collecting and analyzing the timings that David/Jorge are working on in their stress infrastructure. I wanted this temporary solution to be:
- small / compact and not over engineered (and also not a lot of time investment)
- have no dependencies on anything else (ETW, ILogger, etc)
- fast (preferably lock-free) so that we won't have additional contentions / serialization of execution in our stress tests
- low memory & CPU footprint (e.g. not many allocations so we can focus on the product instead)
- to give a fairly good representation of the collected data

Now, to your question: with the number of samples typically over 10K we can simply overcommit the amount of memory we allocated for those samples and *assume* that we won't collect more than _samples while we process the already collected _samples. When this assumption is broken (_samples is small or the rate of adding new time samples is very high) we will loose some data and the stats won't be precise. The No1 reason for this approach is its code simplicity, although it is not that difficult to design a loss-less solution similar to the one in SignalR with a circular array of buckets to store messages. 

I have some more concerns over this code too:
- the allocation of long[] copy may land to LOH - we might consider making it a the stack allocation (if the size isn't too large comparing to the remaining stack size).
- the Func<long[], TimingPercentileStats> statsCalculator should be generic towards the result it produces - e.g. different stats calculators should be able to produce totally different representation of the data it collects in real-time. Re-working the stats calculation would also eliminate the non test-specific magic numbers like (-50 and 1) that you mentioned below.
- Since we're collecting the data real-time we may also benefit from catching (and breaking into debugger ) unusually long-running operations (which I actually do see in stress based on the stats that I collected). 

In this initial check-in I tried to keep the code related to the stats to the minimum as there are many possible directions to improve it.

PR #216, opened 2015-07-22
Fixing callback exception handling and adding tests

mconnew:
Three commits breaking up the changes.  

1. Fixing the product code handling exceptions
2. Fixing the test code and added another test variation
3. Fixed build references for System.Net.Sockets (see commit message for more details)


src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/TaskMethodInvoker.cs
```diff
@@ -74,16 +74,32 @@ public object InvokeEnd(object instance, out object[] outputs, IAsyncResult resu
...
+                    ae = task.Exception;
+                }
+            }
 
+            if (ae != null)
+            {
                 // If FaultException is thrown, we will get 'callFaulted' behavior below.
                 // Any other exception will retain 'callFailed' behavior.
-                throw FxTrace.Exception.AsError<FaultException>(task.Exception);
+                throw FxTrace.Exception.AsError<FaultException>(ae);
```
iamjasonp: Does this change the exception type that's surfaced outwards? If yes, would anyone key off the exception type to do any processing?

src/System.Private.ServiceModel/tests/Common/src/TestTypes.cs
```diff
@@ -702,10 +709,16 @@ public Task<Guid> ServicePingCallback(Guid guid)
...
-          new FaultCode("ServicePingFaultCallback"),
-          "http://tempuri.org/IWcfDuplexTaskReturnCallback/ServicePingFaultCallback");
+        var fault = new FaultException<FaultDetail>(
+            new FaultDetail("Throwing a Fault Exception from the Callback method."),
+            new FaultReason("Reason: Testing FaultException returned from Duplex Callback"),
+            new FaultCode("ServicePingFaultCallback"),
+            "http://tempuri.org/IWcfDuplexTaskReturnCallback/ServicePingFaultCallbackFaultDetailFault");
+        if (_wrapExceptionInTask)
+        {
+            return Task.FromException<Guid>(fault);
```
iamjasonp: The internal build for CoreCLR and NET Native is breaking at this line for me - would be good to make sure that this isn't an environmental issue just for me.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -320,9 +320,53 @@ public static void UnknownUrl_Throws_ProtocolException()
...
+                else
+                {
+                    throw new TimeoutException(String.Format("The call to the Service did not complete within the alloted time of: {0}", ScenarioTestHelpers.TestTimeout));
+                }
+
+                // Not closing the factory as an exception will always be thrown prior to this point.
+            });
+
+            Assert.Equal("ServicePingFaultCallback", exception.Code.Name);
+            Assert.Equal("Reason: Testing FaultException returned from Duplex Callback", exception.Reason.GetMatchingTranslation().Text);
```
iamjasonp: Would the .NET Native test harness be able to view this exception reason? We've discovered in the past that a lot of exception information gets stripped.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/DataContractSerializerOperationFormatter.cs
```diff
@@ -35,10 +35,11 @@ internal static DataContractSerializer CreateSerializer(Type type, string rootNa
 
         internal static DataContractSerializer CreateSerializer(Type type, IList<Type> knownTypes, string rootName, string rootNs, int maxItems)
         {
+            XmlDictionary dictionary = new XmlDictionary(2);
             return new DataContractSerializer(
                 type,
-                rootName,
-                rootNs,
+                dictionary.Add(rootName),
+                dictionary.Add(rootNs),
```
iamjasonp: What's the advantage in using this overload of DCS..ctor?

iamjasonp: Ah, I see this in the commit message 

> Also, there is a bug introduced in the DataContractSerializer where it creates XmlDictionaryStrings with null IXmlDictionary's when passed the root name and namespace as strings. We need to use a different overload or wait for the bug to be fixed.

Makes sense

iamjasonp:
LGTM - though would have preferred that the `System.Net.Sockets` dependency fix have been a different PR to minimize the noise in the main set of changes you were doing :smile:

roncain:
Minor Nit -- these are logically different fixes (Socket dependencies and exceptions) and belong in separate PR's.  No need to change this now.

src/System.Private.ServiceModel/tests/Common/src/TestTypes.cs
roncain: It breaks for me too.  FromException<T> is not in public contract in 4.0.0 of Threading.Tasks but is in the contracdt of 4.0.10.  I see some of the .lock.json files take a dependency on 4.0.0, but I don't see that in the project.json files.  Something in the dependency resolution is bringing in 4.0.0 for this test.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/TaskMethodInvoker.cs
```diff
@@ -74,16 +74,32 @@ public object InvokeEnd(object instance, out object[] outputs, IAsyncResult resu
...
+                Fx.Assert(invokeTask.Exception != null, "Task.IsFaulted guarantees non-null exception.");
+                ae = invokeTask.Exception;
+            }
+            else
             {
-                Fx.Assert(task.Exception != null, "Task.IsFaulted guarantees non-null exception.");
+                Fx.Assert(invokeTask.IsCompleted, "Task.Result is expected to be completed");
+                tuple = invokeTask.Result;
+                task = tuple.Item1 as Task;
+                if (task.IsFaulted)
```
roncain: Maybe Contract.Assert tuple.Item1 is not null --optional

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
roncain: The build failure in NET Native prevents us from verifying how this behaves at runtime.  Let's be sure to buddy-test this again in NET Native after the Threading.Tasks dependency is fixed.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/project.json
```diff
@@ -12,6 +12,7 @@
     "xunit.netcore.extensions": "1.0.0-prerelease-*"
   },
   "frameworks": {
-    "dnxcore50": {}
+    "dnxcore50": {
+    }
   }
 }
```
roncain: nit: missing newline

src/System.Private.ServiceModel/tests/Unit/project.json
```diff
@@ -51,6 +50,11 @@
     "xunit.netcore.extensions":  "1.0.0-prerelease-*"
   },
   "frameworks": {
-    "dnxcore50": {}
+    "dnxcore50": {
+      "dependencies": {
+        "System.Net.Sockets": "4.0.10-beta-*",
+        "System.Reflection.Emit.Lightweight": "4.0.0-beta-*",
+      }
```
roncain: Did you fix the Reflection.Emit.Lightweight dependency everywhere needed -- or is this a left-over from experimentation?

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/TaskMethodInvoker.cs
mconnew: task.Exception is an AggregateException. The use of the ae variable is to allow either Task to be faulted. That is the Invoked delegate Task or the operations returned Task. The two scenarios are equivalent to when using APM whether the Begin call throws or if the End call throws.

src/System.Private.ServiceModel/tests/Common/src/TestTypes.cs
mconnew: I'll change this to use a TaskCompletionSource instead. Task.FromException only appeared (externally anyway) in 4.6 on Desktop so is new.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
mconnew: The .NET Native issues are related to string resources not being fully evaluated. It doesn't strip the exception message itself, just provides resource string names instead of full values.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/TaskMethodInvoker.cs
mconnew: Actually, this is a good point that item1 might be null. An operation that is expected to return a Task could in theory return a null value. I need to verify what happens on Desktop and make this do the same.

src/System.Private.ServiceModel/tests/Unit/project.json
mconnew: Left over from experimentation. I'll remove it.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/TaskMethodInvoker.cs
mconnew: I checked desktop and it turns out I needed an extra line in the InvokeBegin code to detect a null return value and throw the correct exception there. This should now guarantee a non-null here so I'm adding the assert.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/DataContractSerializerOperationFormatter.cs
zhenlan: @mconnew  can you please open an issue in corefx against DCS? No need to change code here.

mconnew: I already opened the issue. Because of the delay in the issue getting fixed and the fix being available from nuget, waiting for the fix will leave our scenario broken for a while. Customers would get a null reference exception and our tests would fail.

Sent from my Windows Phone
________________________________

src/System.Private.ServiceModel/tests/Common/src/TestTypes.cs
```diff
@@ -691,6 +691,13 @@ public string StringValue
 
 public class DuplexTaskReturnServiceCallback : IWcfDuplexTaskReturnCallback
 {
+    private bool _wrapExceptionInTask;
+
+    public DuplexTaskReturnServiceCallback(bool wrapExceptionInTask = false)
```
StephenBonikowsky: I like how you did this variation.

StephenBonikowsky:
:ship: pending an updated PR

roncain:

roncain:
LGTM

PR #220, opened 2015-07-24
Modify the stress test project.*.json files

roncain:
The prior versions of the .json files in
the stress folder were considered invalid
and caused a 'dnu restore' which updated
multiple project.lock.json files during the
build.

This change changes the project.json file to
resemble the other project.json's and also
adds the project.lock.json file produced by
the dnu restore.

This prevents dnu restore from running during
the build due to what it considered an invalid
lock file.


SajayAntony:
@roncain thanks for taking care of this and @KKhurin :person_frowning: this should have been a bug on your plate 
:shipit: LGTM

PR #226, opened 2015-07-27
Move Bridge and setupfiles under tools folder

### LINKED ISSUE #181: Move Bridge and setupFiles under the System.Private.ServiceModel tests folder
Create a tools folder under System.Private.ServiceModel\tests.
Move all the Bridge and setupFiles files under this new folder.
Fix any scripts necessary to use the new locations

roncain:
This PR moves the Bridge project under a new
tools\test folder, and it moves the setupfiles
folder under the new tools\setupfiles.

It also fixes all relative project and file
paths to reflect the new locations.

Fixes #181


roncain:
Reviewers: the original intent was to move Bridge under the tests folder, but it turns out the build-tools make assumptions about the projects under the tests folder, and they fail to compile.  So instead, this PR creates a new tools folder parallel to tests and moves the Bridge into a test folder under that.  It also moves the scripts.

build.cmd
```diff
@@ -18,7 +18,9 @@ if %errorlevel% equ 0 (
 )
 
 if "%outloop%" equ "true" (
+        pushd src\System.Private.ServiceModel\tools\setupfiles
 	start /wait BuildWCFTestService.cmd
+        popd
 )
```
mconnew: It looks like there might be a tab/spaces difference here as they aren't lining up in the PR review

dmetzgar:
When I moved the bridge to the tools folder initially, it was because the build that was happening on the CI machine would do a search for every packages.config file under the src directory and try to restore all of those packages. Since the bridge is using nuget.org packages it would fail. Since it says this PR passed all checks is it safe to say that issue has been fixed?

build.cmd
```diff
@@ -73,7 +75,7 @@ findstr /ir /c:".*Warning(s)" /c:".*Error(s)" /c:"Time Elapsed.*" "%_buildlog%"
 echo Build Exit Code = %BUILDERRORLEVEL%
 
 if "%outloop%" equ "true"  (
-	pushd setupfiles
+	pushd src\System.Private.ServiceModel\tools\setupfiles
```
mconnew: The same path is used twice in the one script. Would it be worth putting in a variable? I'm of two minds. If I want to run things by hand, I usually end up looking at the script and having a variable means my "in my head" cmd script parser needs to do more work, but it feels wrong to duplicate it.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ensureBridge.ps1
```diff
@@ -24,7 +24,9 @@ $result= checkBridge;
 if(!$result)
 {
 	Write-Host Launching bridge.exe.
-	$bridgePath = Join-Path $PSScriptRoot bridge.exe
+
+	$bridgePath = Join-Path $PSScriptRoot ..\bin\bridge.exe
+
 	Start-Process $bridgePath $portNumber -WorkingDirectory $PSScriptRoot
```
mconnew: Shouldn't the WorkingDirectory also change to be $PSScriptRoot\..\bin?

build.cmd
```diff
@@ -18,7 +18,9 @@ if %errorlevel% equ 0 (
 )
 
 if "%outloop%" equ "true" (
+        pushd src\System.Private.ServiceModel\tools\setupfiles
```
dmetzgar: Alternative to pushd/popd is to use the /D option to pass in the working directory:
    start /D src\System.Private.ServiceModel\tools\setupfiles /wait BuildWCFTestService.cmd

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ensureBridge.ps1
roncain: @dmetzgar -- do you have thoughts on what the WorkingDirectory should be here?

roncain: The root error was invoking ensureBridge.ps1 from its source location, not the bin location where it lives after a build.  Reverting the change to ensureBridge.ps1 and invoking it from its bin location

StephenBonikowsky:
Buddy build passed fine.
:shipit:

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ensureBridge.ps1
dmetzgar: Sorry about the delay. I believe Matt is correct. The working directory was set here so that if you passed a relative path into the config call it would look in the right place.

build.cmd
mconnew: :+1:

SajayAntony:
:shipit: 
Discussed some wonderful points with @dmetzgar about how we get numb over time.

roncain:

roncain:
Merging after addressing issues with verbal LGTM

PR #231, opened 2015-07-28
Improve BridgeClient error reporting

### LINKED ISSUE #227: Improve BridgeClient checking and error reporting
The BridgeClient should validate the resources folder it passes to the Bridge exists.  It should also capture the content in any non-200 response from the Bridge and include that information in any exception it throws to fail the client tests.

roncain:
Adds valiation of some of the configuration
properties before they are sent to the Bridge
so failure is detected earlier.

Also captures any content body returned from
the Bridge for any non-200 response so it can
be part of the exception thrown.  This makes it
easier to see what the Bridge detected was wrong.

Fixes #227

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
+++ b/src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
@@ -21,7 +21,7 @@ namespace System.ServiceModel.Tests.Common
             get { return TestProperties.GetProperty(TestProperties.BridgeUrl_PropertyName); }
         }
 
-        static BridgeClient()
+        private static void EnsureBridgeIsRunning()
         {
             if (_BridgeStatus == BridgeState.NotStarted)
             {
@@ -32,9 +32,10 @@ namespace System.ServiceModel.Tests.Common
                 throw new Exception("Bridge is not running");
             }
         }
-
         public static string GetResourceAddress(string resourceName)
         {
+            EnsureBridgeIsRunning();
+
             string resourceAddress = null;
             if (_BridgeStatus == BridgeState.Started && !_Resources.TryGetValue(resourceName, out resourceAddress))
             {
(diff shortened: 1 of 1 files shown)


roncain:
This PR is not ready for public review yet -- I'm verifying behavior in NET Native

StephenBonikowsky:
Looks like there is a merge conflict.
Other than that, :shipit:

src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
```diff
@@ -21,20 +22,26 @@ private static string BridgeBaseAddress
             get { return TestProperties.GetProperty(TestProperties.BridgeUrl_PropertyName); }
         }
 
-        static BridgeClient()
+        private static void EnsureBridgeIsRunning()
         {
```
mconnew: What was the reason for moving from a static constructor to a static method?

src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
```diff
@@ -58,7 +65,18 @@ private static void MakeConfigRequest()
...
                     var response = httpClient.PostAsync("/config/", content).Result;
                     if (!response.IsSuccessStatusCode)
-                        throw new Exception("Unexpected status code: " + response.StatusCode);
+                    {
+                        string reason = String.Empty;
+                        if (response.Content != null)
+                        {
+                            string contentAsString = response.Content.ReadAsStringAsync().Result;
+                            reason = String.Format("{0}Bridge returned content:{0}{1}",
+                                                    Environment.NewLine, contentAsString);
```
mconnew: You've dropped the status code. We should have the status code and the status reason included.

src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
```diff
@@ -58,7 +65,18 @@ private static void MakeConfigRequest()
...
+                        }
+                        throw new Exception(
+                            String.Format("Bridge returned unexpected status code {0}{1}", 
+                                            response.StatusCode, reason));
+                    }
```
mconnew: There's also a status reason which is a textual description which might be helpful to include in the description.

src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
```diff
@@ -58,7 +65,18 @@ private static void MakeConfigRequest()
                 {
                     var response = httpClient.PostAsync("/config/", content).Result;
```
mconnew: GetAwaiter().GetResult() to avoid AggregateException if not using await

mconnew:
Pending a couple of minor changes, the overall change LGTM

src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
roncain: Better behavior for error handling.  As a static ctor, we got TypeInitialization exceptions for the BridgeClient that sort of masked something as simple as the bridge not running.  With an explicit failure from EnsureBridgeIsRunning, the exception stack trace is more informative.

src/System.Private.ServiceModel/tests/Common/src/BridgeClient.cs
roncain: Good catch -- added.

roncain:
I merged this PR into #238 because I needed these improvements to debug it.  So I am closing this PR because it is now redundant.  The comments made in this PR were applied to the BridgeClient in #238

Issue #224, opened 2015-07-24
Basic Authentication: empty user name should throw ArgumentException


hongdai:
For the same empty user name scenario:
On desktop, it throws ArgumentException, message: "The username cannot be empty."
On ProjectK, it throws MessageSecurityException, message:"The HTTP request was forbidden with client authentication scheme 'Basic'."

The exception and the message does not help on ProjectK. It needs to be fixed to match desktop.

hongdai:
Exception callstack from desktop:
System.ArgumentException: The username cannot be empty.

Server stack trace:
   at System.IdentityModel.Tokens.UserNameSecurityToken..ctor(String userName, S
tring password, String id)
   at System.IdentityModel.Tokens.UserNameSecurityToken..ctor(String userName, S
tring password)
   at System.IdentityModel.Selectors.UserNameSecurityTokenProvider..ctor(String
userName, String password)
   at System.ServiceModel.ClientCredentialsSecurityTokenManager.CreateSecurityTo
kenProvider(SecurityTokenRequirement tokenRequirement, Boolean disableInfoCard)
   at System.ServiceModel.ClientCredentialsSecurityTokenManager.CreateSecurityTo
kenProvider(SecurityTokenRequirement tokenRequirement)
   at System.ServiceModel.Channels.TransportSecurityHelpers.GetUserNameTokenProv
ider(SecurityTokenManager tokenManager, EndpointAddress target, Uri via, String
transportScheme, AuthenticationSchemes authenticationScheme, ChannelParameterCol
lection channelParameters)
   at System.ServiceModel.Channels.HttpChannelFactory`1.CreateAndOpenTokenProvid
er(TimeSpan timeout, AuthenticationSchemes authenticationScheme, EndpointAddress
 target, Uri via, ChannelParameterCollection channelParameters)
   at System.ServiceModel.Channels.HttpChannelFactory`1.CreateAndOpenTokenProvid
ersCore(EndpointAddress to, Uri via, ChannelParameterCollection channelParameter
s, TimeSpan timeout, SecurityTokenProviderContainer& tokenProvider, SecurityToke
nProviderContainer& proxyTokenProvider)
   at System.ServiceModel.Channels.HttpChannelFactory`1.HttpRequestChannel.Creat
eAndOpenTokenProviders(TimeSpan timeout)
   at System.ServiceModel.Channels.HttpsChannelFactory`1.HttpsRequestChannel.OnO
pen(TimeSpan timeout)
[...]

PR #233, opened 2015-07-28
Add Test Coverage

### LINKED ISSUE #88: Need a negative test for invalid user name or password for Basic authentication
we need a negative test that catches exactly the kind of exception we expect for a service that throws SecurityTokenException for an invalid user name or password. A negative test that validates the ArgumentNullException path would also be good to include.

hongdai:
* Add a test to cover incorrect password.
* Add a test to cover empty user name

Fix #88

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.ServiceModel;
+using System.ServiceModel.Security;
 using System.Text;
 using Xunit;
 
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
```diff
@@ -54,6 +55,59 @@ public static void BasicAuthentication_RoundTrips_Echo()
     }
 
     [Fact]
+    [OuterLoop]
+    public static void BasicAuthenticationInvalidPwd_throw_MessageSecurityException()
+    {
+        StringBuilder errorBuilder = new StringBuilder();
+        //Will need to use localized string once it is available
+        string message = "The HTTP request was forbidden with client authentication scheme 'Basic'.";
```
hongdai: It turns out the exception message has stripped in .Net Native retail build. I will update it to take it into consideration.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
```diff
@@ -54,6 +55,62 @@ public static void BasicAuthentication_RoundTrips_Echo()
...
+
+    [Fact]
+    [ActiveIssue(224)]
+    [OuterLoop]
+    public static void BasicAuthenticationEmptyUser_throw_ArgumentException()
+    {
+        StringBuilder errorBuilder = new StringBuilder();
+        //Will need to use localized string once it is available
+        string message = "The username cannot be empty.";
+
```
roncain: I think we will run into trouble with NET Native retail changing the message.
We should verify the ArgumentException.ParamName is correct for any kind of ArgumentException.
There is a special overload of Assert.Throws that takes this name -- look in CreateChannel_ExpectedNetTcpScheme_HttpScheme_ThrowsUriFormat

hongdai: I've just verified the ParamName in this scenario s NULL on Desktop.

I agree we will run into with NET Native retail changing the message. However it seems we do not have better option right now.

roncain: My point here was that this test is guaranteed to fail on NET Native because you are comparing the full message.  You need to find a way to make this test pass on NET Native (when it is enabled), either by finding some substring to match or not testing the message.

hongdai: Point taken. When issue #224 get fixed, I will find some substring and update the test.

roncain:
LGTM.  For tests, I think we need only one approver, so go ahead and merge when you want.

PR #236, opened 2015-07-29
Make Bridge service URLs configurable

### LINKED ISSUE #187: Make the Bridge's returned service URL's reflect where the Bridge is running
The Bridge currently returns URL's as localhost.  These URL's need to reflect where the service is actually running for the scenario when the Bridge + services are running on a separate machine.

Special considerations:
1) Once issue #185 is resolved, the configuration name/value pairs will contain an entry "BridgeURL" which reflects where the BridgeClient believes the Bridge is running.  It can be used.
2) @hongdai observed that localhost is preferrable to the machine name when the Bridge really is running on the same machine as the tests
3) After issue #185 is resolved, the configuration name/value pairs will contain an entry for "UseFiddlerUrl". If set to 'true' any URL that would have used localhost should substitute "localhost.fiddler" for "localhost". Example: http://localhost.fiddler:8081/ instead of http://localhost:8081/.  This will allow Fiddler to capture localhost traffic.

roncain:
This PR makes the Bridge no longer hard-code
localhost as the host for all services but instead
forms the URL based on the configuration information
passed to the Bridge POST config.

To allow this, all IResources PUT verb now accepts
a context object providing access to the Bridge
configuration.

Fixes #187


dmetzgar:
:shipit:

hongdai:
LGTM

PR #235, opened 2015-07-29
Pull request for Issue 140

StephenBonikowsky:
I tried to have each commit do a similar thing.
I have not yet created the TFS variation that needs to happen in tandem.
All the tests were running for me.

PR #238, opened 2015-07-30
Make Bridge idle timeout configurable

### LINKED ISSUE #198: Bridge should close gracefully after configurable max idle time

roncain:
This PR allows the Bridge to accept an
idle timeout as one of the properties in
the POST config name/value pairs.  And it
sets the idle timeout manager to use that
as its new idle timeout between requests.

To avoid making the Bridge aware of all its
configuration sensitive types, this PR also adds
a new EventHandler for configuration changes.
Types that care about configuration changes register
to be notified of the change.

The ConfigurationExtensions class was updated to
use this same notification mechanism.  And it was
renamed to reflect its true purpose -- AppDomainManager.

Fixes #198

### DIFF
--- a/src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ConfigController.cs
+++ b/src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ConfigController.cs
@@ -16,14 +16,25 @@ namespace Bridge
 {
     public class ConfigController : ApiController
     {
-        private const int Default_BridgeIdleTimeoutMinutes = 20;
+        public static EventHandler<Tuple<BridgeConfiguration, BridgeConfiguration>> BridgeConfigurationChanged;
 
         private static BridgeConfiguration s_bridgeConfiguration = new BridgeConfiguration
         {
-            BridgeIdleTimeoutMinutes = Default_BridgeIdleTimeoutMinutes   // initialize so is accessible during app startup
+            BridgeIdleTimeoutMinutes = IdleTimeoutHandler.Default_Timeout_Minutes
         };
 
-        internal static string CurrentAppDomain;
+        internal static string CurrentAppDomainName;
+
+        static ConfigController()
+        {
+            // Register to manage AppDomains in response to changes to configuration
+            BridgeConfigurationChanged += (object s, Tuple<BridgeConfiguration, BridgeConfiguration> tuple) =>
+            {
+                BridgeConfiguration oldConfig = tuple.Item1;
+                BridgeConfiguration newConfig = tuple.Item2;
+                CurrentAppDomainName = AppDomainManager.UpdateApp(oldConfig, newConfig);
+            };
+        }
 
         public static BridgeConfiguration BridgeConfiguration
         {
@@ -44,16 +55,24 @@ namespace Bridge
                 string nameValuePairs = request.Content.ReadAsStringAsync().Result;
                 Dictionary<string, string> configInfo = JsonSerializer.DeserializeDictionary(nameValuePairs);
 
+                Trace.WriteLine("POST config content: " + Environment.NewLine + nameValuePairs);
+
                 // Create a new configuration combining the existing one with any provided properties.
                 BridgeConfiguration newConfiguration = new BridgeConfiguration(BridgeConfiguration, configInfo);
+                Trace.WriteLine("POST config properties: " + Environment.NewLine + newConfiguration);
 
-                Trace.WriteLine("POST config request: " + Environment.NewLine + newConfiguration);
+                // Take the new configuration and notify listeners of the change.
+                BridgeConfiguration oldConfiguration = BridgeConfiguration;
+                BridgeConfiguration = newConfiguration;
 
-                string friendlyName = newConfiguration.UpdateApp();
-                CurrentAppDomain = friendlyName;
+                if (BridgeConfigurationChanged != null)
+                {
+                    Tuple<BridgeConfiguration, BridgeConfiguration> eventArgs = new Tuple<BridgeConfiguration, BridgeConfiguration>(oldConfiguration, newConfiguration);
+                    BridgeConfigurationChanged(this, eventArgs);
+                }
 
-                string configResponse = PrepareConfigResponse(TypeCache.Cache[friendlyName]);
-                Trace.WriteLine("POST config response: " + Environment.NewLine + configResponse);
+                string configResponse = PrepareConfigResponse(TypeCache.Cache[CurrentAppDomainName]);
+                Trace.WriteLine("POST config response content: " + Environment.NewLine + configResponse);
 
                 // Directly return a json string to avoid use of MediaTypeFormatters
                 HttpResponseMessage response = Request.CreateResponse(HttpStatusCode.OK);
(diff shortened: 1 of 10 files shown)


roncain:
Reviewers: this PR is ready for review but will have one final commit to add the idle timeout to TestProperties. I'm waiting for PR #235 to be merged first to avoid conflicts.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ConfigController.cs
```diff
@@ -16,14 +16,25 @@ namespace Bridge
 {
     public class ConfigController : ApiController
     {
-        private const int Default_BridgeIdleTimeoutMinutes = 20;
+        public static EventHandler<Tuple<BridgeConfiguration, BridgeConfiguration>> BridgeConfigurationChanged;
```
SajayAntony: I would recommend just defining this. I personally don't like Tuples since they Item1 and Item2 is just :scream: 
Its your call and don't block on this comment. 

``` 
struct ChangedEventArgs<T> : EventArgs
{ 
    public T Previous {get; set;}
    public T Updated {get; set;}
}
```

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/IdleTimeoutManager.cs
```diff
@@ -55,6 +58,26 @@ public IdleTimeoutManager(TimeSpan idleTimeout)
                 _waitTimeout = idleTimeout;
                 _timer = new Timer(this.TimeOutCallback);
                 this.Restart();
+
+                // Register to be notified when the Bridge configuration changes
+                ConfigController.BridgeConfigurationChanged += (object s, Tuple<BridgeConfiguration, BridgeConfiguration> tuple) =>
```
SajayAntony: Interestingly this bleeds the configController into the idleTimeout manager. I understand this this test code but from a design principle its concepts spilling over  and we could think of the config instance class to be like viewcontroller and have an event like  `OnIdleTimeOutChanged(Timespan)` and the handler only registers to that one event. Just a bit more decoupling rather than have the entire config passed in here. 
Again don't block on these comments.

SajayAntony:
:shipit:

StephenBonikowsky:
:shipit:

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ConfigController.cs
mconnew: You don't need to declare your own struct. `EventHandler<TEventArgs>` is just a predefined convenience delegate. It's declared as this:
```csharp
public delegate void EventHandler<TEventArgs>(Object sender, TEventArgs e); 
```
So just define your own delegate type for the events and don't worry about `EventHandler<TEventArgs>`. Something like this:
```csharp
public delegate void ConfigChangeEventHandler(Object sender, BridgeConfiguration previous, BridgeConfiguration updated);
```
As you aren't even using the event handler type and just using anonymous delegates, you could dispense with any custom types altogether and declare the event using `Action<Object, BridgeConfiguration, BridgeConfiguration>`.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ConfigController.cs
```diff
@@ -44,16 +55,24 @@ public HttpResponseMessage POST(HttpRequestMessage request)
                 string nameValuePairs = request.Content.ReadAsStringAsync().Result;
                 Dictionary<string, string> configInfo = JsonSerializer.DeserializeDictionary(nameValuePairs);
```
mconnew: GetAwaiter().GetResult()

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ConfigController.cs
```diff
@@ -44,16 +55,24 @@ public HttpResponseMessage POST(HttpRequestMessage request)
...
+                if (BridgeConfigurationChanged != null)
+                {
+                    Tuple<BridgeConfiguration, BridgeConfiguration> eventArgs = new Tuple<BridgeConfiguration, BridgeConfiguration>(oldConfiguration, newConfiguration);
+                    BridgeConfigurationChanged(this, eventArgs);
+                }
 
-                string configResponse = PrepareConfigResponse(TypeCache.Cache[friendlyName]);
-                Trace.WriteLine("POST config response: " + Environment.NewLine + configResponse);
+                string configResponse = PrepareConfigResponse(TypeCache.Cache[CurrentAppDomainName]);
+                Trace.WriteLine("POST config response content: " + Environment.NewLine + configResponse);
```
mconnew: Is there any benefit to using the Category parameter in tracing? I could see some benefit in having the class name provided as the category.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/IdleTimeoutManager.cs
```diff
@@ -55,6 +58,26 @@ public IdleTimeoutManager(TimeSpan idleTimeout)
...
+                {
+                    BridgeConfiguration newConfig = tuple.Item2;
+
+                    int minutes = newConfig.BridgeIdleTimeoutMinutes;
+                    TimeSpan newTimeout = (minutes <= 0)
```
mconnew: You were looking for some mechanism to use timeout for immediate shutdown. Normally a value of -1 means infinite. If you specify 0 minutes idle, then that would seem to me to mean shutdown immediately and not infinite.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ConfigController.cs
roncain: I liked @SajayAntony 's suggestion of a ChangedEventArgs<T> because it allowed a better factoring of his 2nd suggestion, to remove knowledge of BridgeConfiguration from the IdleTimeoutManager.  Now, the Bridge has separate events for changed resource folder and changed idle timespan, and clients sign up for only the changes they care about.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ConfigController.cs
roncain: Agree -- added type name as category

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/IdleTimeoutManager.cs
roncain: I changed the timeout to be a TimeSpan to eliminate any special values that receive different treatment.  You can now shut down the Bridge in 5 seconds by sending it a timeout of "00:00:05" or ask it to run for days.

Issue #166, opened 2015-07-14
Fix intermittent failure in UnexpectedExceptionTests


roncain:
This test fails intermittently in our CI machine as well as nightly builds.
Stack trace from our CI machine is:
MESSAGE:
Assert.Throws() Failure Expected: typeof(System.ServiceModel.EndpointNotFoundException) Actual: typeof(System.ServiceModel.CommunicationException): Error while copying content to a stream.
+++++++++++++++++++
STACK TRACE:
[...]

PR #241, opened 2015-07-31
Add additional logging to help diagnose Issue #166

mconnew:
Issue #166 looks like it's a race condition which I can't duplicate on my local machine. There are inner exceptions with call stacks that are needed in order to debug this problem. This adds additional information to the exception message so that when this test fails on our CI machine, I will have the required call stacks to understand what's causing this.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Net.Http;
 using System.ServiceModel;
 using System.Text;
 using System.Threading.Tasks;
@@ -286,10 +287,37 @@ public static class ExpectedExceptionTests
 
         EndpointNotFoundException exception = Assert.Throws<EndpointNotFoundException>(() =>
         {
-            using (ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(notFoundUrl)))
+            try
             {
-                IWcfService serviceProxy = factory.CreateChannel();
-                string response = serviceProxy.Echo("Hello");
+                using (
+                    ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding,
+                        new EndpointAddress(notFoundUrl)))
+                {
+                    IWcfService serviceProxy = factory.CreateChannel();
+                    string response = serviceProxy.Echo("Hello");
+                }
+            }
+            catch (EndpointNotFoundException)
+            {
+                throw;
+            }
+            catch (CommunicationException ce)
+            {
+                if (ce.InnerException == null)
+                    throw;
+
+                if (ce.InnerException.GetType() == typeof (HttpRequestException))
+                {
+                    var httpReqExcep = ce.InnerException as HttpRequestException;
+                    StringBuilder sb = new StringBuilder();
+                    sb.Append("Received HttpRequestException with unknown error code ")
+                        .AppendLine(ce.InnerException.HResult.ToString())
+                        .AppendLine("Full details for HttpRequestException:")
+                        .AppendLine(httpReqExcep.ToString());
+                    throw new CommunicationException(sb.ToString());
+                }
+
+                throw;
             }
         });
 


src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -286,10 +287,37 @@ public static void UnknownUrl_Throws_EndpointNotFoundException()
...
+                        new EndpointAddress(notFoundUrl)))
+                {
+                    IWcfService serviceProxy = factory.CreateChannel();
+                    string response = serviceProxy.Echo("Hello");
+                }
+            }
+            catch (EndpointNotFoundException)
+            {
+                throw;
+            }
```
iamjasonp: Is this `catch (EndpointNotFoundException)` necessary if all you're doing is throwing?

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -286,10 +287,37 @@ public static void UnknownUrl_Throws_EndpointNotFoundException()
...
             {
-                IWcfService serviceProxy = factory.CreateChannel();
-                string response = serviceProxy.Echo("Hello");
+                using (
+                    ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding,
+                        new EndpointAddress(notFoundUrl)))
+                {
+                    IWcfService serviceProxy = factory.CreateChannel();
+                    string response = serviceProxy.Echo("Hello");
+                }
```
iamjasonp: I think we should avoid using `using` (though pattern isn't settled entirely yet - see #168)

iamjasonp:
Small nits, otherwise LGTM.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -286,10 +287,37 @@ public static void UnknownUrl_Throws_EndpointNotFoundException()
...
+            {
+                if (ce.InnerException == null)
+                    throw;
+
+                if (ce.InnerException.GetType() == typeof (HttpRequestException))
+                {
+                    var httpReqExcep = ce.InnerException as HttpRequestException;
+                    StringBuilder sb = new StringBuilder();
+                    sb.Append("Received HttpRequestException with unknown error code ")
+                        .Append(ce.InnerException.HResult).AppendLine();
```
roncain: Nit: formatting a bit odd.  It might look better with something like:
   sb.Append(...)
       .Append(...)
       .Append(...)
Don't block on this suggestion.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
roncain: I believe this catch/throw is needed to avoid the lower catch but still preserve the Assert.Throws<T> usage

roncain:
LGTM

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
mconnew: This is existing text code. I wasn't looking to tidy up in this change.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
mconnew: @iamjasonp , @roncain is correct. EndpointNotFoundException derives from CommunicationException.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
mconnew: Fixed this in an update PR. I'll merge after.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: Ah I see.

Issue #12, opened 2015-05-13
Support DnsEndpointIdentity


iamjasonp:
DnsEndpointIdentity currently throws PlatformNotSupportedException.

zhenlan:
This is currently pending the availability of TokenManager

zhenlan:
@hongdai will work with @mconnew and add tests to see what are involved to enable this feature.

hongdai:
After discuss with Matt, I will:
1 add a scenario test to test DnsEndpointIdentity E2E 
2. Create a feature item to track DnsEndpointIdentity can be created, serialized and deserialized correctly. This combines #13 should enable the feature.

hongdai:
There are two major scenarios, both are blocked because of dependent feature is not available yet.
1) Verify Server identity for WS* binding. This is blocked by WS*binding is not in.
2) Verify Server identity for NetTcpBinding when SecurityMode is on. SecurityMode on is not supported yet.

hongdai:
I have implemented a positive E2E Scenario 2. Once the feature is fully enabled. I will add negative scenario tests.

iamjasonp:
This may start coming through as part of my large change to turn on SslStream. Stay tuned on this.

hongdai:
E2E tests are passing in the lab.

PR #247, opened 2015-08-03
Verify Server Dns for NetTcp binding

hongdai:
* This is one of the key scenarios for DnsEndpointIdentity
* Note that it will fail until all dependent features mentioned in the test resolved.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.ServiceModel;
+using System.Text;
+using Xunit;
+
+namespace Security.TransportSecurity.Tests.Tcp
+{
+    public static class IdentityTests
+    {
+        [Fact]
+        [ActiveIssue(12)]
+        [OuterLoop]
+        // The product code will check the Dns identity from the server and throw if it does not match what is specified in DnsEndpointIdentity
+        public static void VerifyServiceIdentityMatchDnsEndpointIdentity()
+        {
+            string testString = "Hello";
+
+            NetTcpBinding binding = new NetTcpBinding();
+            //SecurityMode.Transport is not supported yet, we will get an exception here, tracked by issue #81
+            binding.Security.Mode = SecurityMode.Transport;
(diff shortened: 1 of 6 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
```diff
@@ -0,0 +1,44 @@
...
+using Xunit;
+
+namespace Security.TransportSecurity.Tests.Tcp
+{
+    public static class IdentityTests
+    {
+        [Fact]
+        [ActiveIssue(12)]
+        [OuterLoop]
+        //Once this feature is enabled, we will need to add negative tests
```
iamjasonp: NIT: consider adding this comment to the issue itself rather than putting it in code here.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        public static void VerifyServiceDNS()
```
iamjasonp: NIT: Consider naming something like Verify_ServerValidationMode_ServerCertificate or something like that, so we have the binding type in the name. Otherwise, it's not super obvious what's being tested because the test just looks like any other echo test, and we don't explicitly validate the DnsEndpointIdentity aside from waiting to see if an exception is thrown.

iamjasonp:
Aside from minor nits, LGTM.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
hongdai: Agree it's hard to tell what's being tested from the test itself. I will add below comment:
"The product code will check the Dns identity from the server and throw if it does not match what is specified in DnsEndpointIdentity".

I will rename the test to VerifyServiceIdentityMatchDnsEndpointIdentity

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
hongdai: Added to the issue #12 .

PR #252, opened 2015-08-05
Removing dead code associated with ClientBase<TChannel>

### LINKED ISSUE #240: Cleanup ClientBase<TChannel> constructors.
All the InstanceContext overrides are not in contract.
The ServiceEndpoint override is not in contract.
The three endpointConfigurationName overrides are in contract but not supported.
Fixing these overrides allows internal sealed class ConfigurationEndpointTrait to be removed as dead code.

This task improves CodeCoverage.

StephenBonikowsky:
* Some constructors in DuplexClientBase<TChannel> and ClientBase<TChannel> are either not in contract or are in contract but are intended to be used in connection with Configuration files which is not a supported feature.
* Cleaned up a lot more code in ClientBase<TChannel> related to caching of ChannelFactorys which is not used.
* With the cleanup of DuplexClientBase<TChannel> and ClientBase<TChannel> several files were able to be completely removed.
* Fixes #240


iamjasonp:
Are we sure we want to get rid of the caching code in ChannelFactory entirely? I don't know if this might be needed down the line for performance reasons. I suppose we can get rid of this since it's not being used right now, but consider making two commits (one for code cleanup, one for getting rid of the ChannelFactory caching) in case we want to back out the removal. 

Else, LGTM.

mconnew:
@iamjasonp, if we introduce caching in the future, it would require a contract change to expose turning it on so would be quite a bit in the future. If/when that happens I would rather grab the code from the current desktop .Net version than have to dig through past commit's to try and find the relevant code.

src/System.Private.ServiceModel/src/System/ServiceModel/ClientBase.cs
```diff
@@ -17,124 +15,30 @@ public abstract class ClientBase<TChannel> : ICommunicationObject, IDisposable
...
-                _channelFactoryRef = new ChannelFactoryRef<TChannel>(new ChannelFactory<TChannel>("*"));
-                _channelFactoryRef.ChannelFactory.TraceOpenAndClose = false;
-                TryDisableSharing();
-            }
-            else
-            {
-                _endpointTrait = new ConfigurationEndpointTrait<TChannel>("*", null, null);
-                InitializeChannelFactoryRef();
-            }
+            throw new NotSupportedException(SR.ConfigurationFilesNotSupported);
```
mconnew: PlatformNotSupportedException

src/System.Private.ServiceModel/src/System/ServiceModel/ClientBase.cs
```diff
@@ -464,28 +158,7 @@ public void Abort()
...
-                    }
-                }
-            }
-
-            // Abort the ChannelFactory if we released the last one. We should be able to abort it when another thread is closing it.
-            if (_releasedLastRef)
-            {
-                _channelFactoryRef.Abort();
-            }
+            _channelFactory.Abort();
```
mconnew: _channelFactory might be null.

src/System.Private.ServiceModel/src/System/ServiceModel/ClientBase.cs
```diff
@@ -493,58 +166,14 @@ public void Close()
...
-            }
-            finally
-            {
-                if (!_sharingFinalized && s_cacheSetting == CacheSetting.Default)
-                {
-                    // this.CreateChannel() is not called. For safety, we disable sharing.
-                    TryDisableSharing();
-                }
-            }
+            _channel = this.CreateChannel();
```
mconnew: don't use `this.`

src/System.Private.ServiceModel/src/System/ServiceModel/ClientBase.cs
```diff
@@ -17,124 +15,30 @@ public abstract class ClientBase<TChannel> : ICommunicationObject, IDisposable
         where TChannel : class
     {
         private TChannel _channel;
-        private ChannelFactoryRef<TChannel> _channelFactoryRef;
-        private EndpointTrait<TChannel> _endpointTrait;
```
mconnew: I don't believe EndpointTrait and any of the derived classes are used any more. Can you check if it's safe to remove them?

mconnew:
There's inconsistent usage of GetChannelFactory() vs _channelFactory. The method was there before as it had to deal with the ChannelFactoryRef class and didn't have a direct reference. I would suggest removing the GetChannelFactory() method and replacing usages with directly accessing the field.

roncain:
LGTM, subject to any of Matt's suggestions.  It appears to pass ApiCompat verification.

mconnew:
LGTM :shipit:

PR #253, opened 2015-08-05
Cleaning up ServiceModel string resources.

StephenBonikowsky:
* Removed resource strings related to configuration files.
* Removed resource strings related to reliable messaging and reliable sessions.

### DIFF
--- a/src/System.Private.ServiceModel/src/Resources/Strings.resx
+++ b/src/System.Private.ServiceModel/src/Resources/Strings.resx
@@ -732,9 +732,6 @@
   <data name="CannotPerformImpersonationOnUsernameToken" xml:space="preserve">
     <value>Impersonation using the client token is not possible. The binding ('{0}', '{1}') for contract ('{2}', '{3}') uses the Username Security Token for client authentication with a Membership Provider registered. Use a different type of security token for the client.</value>
   </data>
-  <data name="SecureConversationRequiredByReliableSession" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


iamjasonp:
Good first pass - how'd you go through these to determine what was used/not? Did you use some kind of tool, or manually? 

It seems like there are a number of strings (e.g. SecurityBindingElementCannotBeExpressedInConfig, DeliveryAssuranceRequired) that weren't taken out. 

No objections to merging what you have now and then we can continue to prune the list as time goes on.

src/System.Private.ServiceModel/src/Resources/Strings.resx
```diff
@@ -3108,51 +2964,24 @@
   <data name="UnknownSequenceFaultReason" xml:space="preserve">
     <value>The value of wsrm:Identifier is not a known Sequence identifier.</value>
   </data>
```
roncain: The "Sequence" and "RM" messages above can probably also be removed at some point

roncain:
LGTM

StephenBonikowsky:
I had to touch this file to add a resource for my previous PR and that's when I saw a bunch of the resources for config, so I thought any cleanup done to this file would be positive, @mconnew mentioned that we don't want to remove any resource strings that we might eventually pull in so then I hunted for Reliable Messaging and Reliable Sessions strings which I don't think we will ever pull in. This isn't a complete cleanup by any means, just some of the more obvious ones.

PR #254, opened 2015-08-06
Enable duplex callback usage in NET Native

### LINKED ISSUE #110: CallbackContract unit test fails when run with NET Native
Using the experimental tools to run the GitHub tests against the NET Native toolchain, the CallbackContract unit test fails:

  Running Test: ContractDescriptionTest.Duplex_ContractDescription_Builds_From_ServiceContract
  Assert.Equal:Expected: [2]. Actual:[1]. 
  Xunit.AssertTestException: Assert.Equal: Expected: [2]. Actual:[1]. 

Early analysis shows the NET Native environment does not naturally allow reflection over the type declared in the CallbackContractg

roncain:
Prior to this change, using Reflection on the concrete duplex
callback type failed in NET Native because its metadata was not
available.

The modification to the rd.xml file allows the NET Native version
of System.Private.ServiceModel to use Reflection on any interface
type specified in [ServiceContract(CallbackContractType)]. This is
necessary to build OperationDescriptions for the callback type.

But this alone does not enable Reflection for the concrete type.
And TypeLoader analyzes the concrete type to locate custom
attributes that implement IOperationBehavior to add them to the
operation's behaviors.

The change to TypeLoader allows it to tolerate missing metadata
when it analyzes these custom attributes. Because the
[OperationBehavior] attribute is not available in a public contract,
these would occur only if the developers create their own custom
attributes that implement IOperationBehavior and apply them to the
concrete type for the callback contract. In that edge-case, these
attributes will not be found in NET Native unless the developer
enables Reflection for that concrete type via an rd.xml file in the app.

Fixes #110

### DIFF
--- a/src/System.Private.ServiceModel/src/Resources/System.Private.ServiceModel.rd.xml
+++ b/src/System.Private.ServiceModel/src/Resources/System.Private.ServiceModel.rd.xml
@@ -19,6 +19,12 @@
             <TypeParameter Name="detailType" DataContractSerializer="Public" DataContractJsonSerializer="Public" />
           </Method>
         </Type>
+         <!-- Objects passed to the InstanceContext ctor require reflection for duplex contracts -->
+        <Type Name="InstanceContext" Dynamic="Required All" >  
+          <Method Name=".ctor">  
+            <Parameter Name="implementation" Dynamic="Required All" />  
+          </Method>  
+        </Type>  
         <Type Name="MessageBodyMemberAttribute" Dynamic="Required All" />
         <Type Name="MessageContractMemberAttribute" Dynamic="Required All" />
         <Type Name="MessageParameterAttribute" Dynamic="Required All" />
@@ -29,6 +35,10 @@
         <!-- ServiceContractAttribute on a type makes it available for Reflection -->
         <Type Name="ServiceContractAttribute" Dynamic="Required All" >
           <AttributeImplies Dynamic="Required All" />
+          <!-- Type in CallbackContract named property requires reflection -->
+          <Method Name=".ctor">
+             <TypeParameter Name="CallbackContract" Dynamic="Required All" />
+          </Method>
         </Type>
         <!-- ExceptionDetail requires serialization for FaultException -->
         <Type Name="ExceptionDetail" Dynamic="Required All" DataContractJsonSerializer="Public" DataContractSerializer="Public" />


roncain:
Reviewers -- I'm looking for feedback on how impactful it is to not locate IOperationBehavior custom attributes on the concrete type for the callback contract (emphasis on "concrete type").  The fact [OperationBehavior] is not available in NET Native or CoreCLR suggests the presence of those attributes is now an edge case.  If it is critical we support that, it represents significant feature work for the NET Native toolchain.  The app would still be able to include an rd.xml to cover that case.

src/System.Private.ServiceModel/src/Resources/System.Private.ServiceModel.rd.xml
```diff
@@ -29,6 +29,10 @@
         <!-- ServiceContractAttribute on a type makes it available for Reflection -->
         <Type Name="ServiceContractAttribute" Dynamic="Required All" >
           <AttributeImplies Dynamic="Required All" />
+          <!-- Type in CallbackContract named property requires reflection -->
+          <Method Name=".ctor">
+             <TypeParameter Name="CallbackContract" Dynamic="Required All" />
+          </Method>
         </Type>
         <!-- ExceptionDetail requires serialization for FaultException -->
```
mconnew: Would something like the following work?
```xml
<Type Name="InstanceContext" Dynamic="Required All" >  
    <Method Name=".ctor">  
        <Parameter Name="implementation" Dynamic="Required All" />  
    </Method>  
</Type>  
```
As any callback implementation type is passed to WCF wrapped in an `InstanceContext`, we might be able to get the compiler to generate the metadata based on the `InstanceContext` constructor

src/System.Private.ServiceModel/src/System/ServiceModel/Description/TypeLoader.cs
```diff
@@ -432,7 +432,31 @@ private KeyedByTypeCollection<IOperationBehavior> GetIOperationBehaviorAttribute
...
-                    MethodInfo implMethod = GetCorrespondingMethodFromType(implType, opMethod);
+                    MethodInfo implMethod = null;
+                    try
+                    {
+                        GetCorrespondingMethodFromType(implType, opMethod);
+                    }
+
+                    catch (TypeAccessException)
+                    {
+#if FEATURE_NETNATIVE
```
mconnew: Having a catch/rethrow for .Net Core because of some .NET Native code doesn't feel right. Try/Catch blocks reduce the ability for the JIT to do some optimizations too so I would rather only have the try/catch for .NET Native

StephenBonikowsky:
:shipit:

src/System.Private.ServiceModel/src/Resources/System.Private.ServiceModel.rd.xml
roncain: @mconnew -- good suggestion.  It does work.  I also removed the try/catch because no exception occurs in our test usage.  If we find any usage where an exception occurs, we'll get PNSE and can deal with that as a new issue.

PR #255, opened 2015-08-06
Fixing OuterLoop tests broken by PR #244

StephenBonikowsky:
Fixing two issues,
One, the new version of buildtools relies on xunit 2.1, we can't use this yet because the buildtools packages have not yet been published. See https://github.com/dotnet/corefx/pull/2609

Two, because of a dependency on SecureString in System.Private.Networking we need to continue using the beta version of our contracts. For more details see the notes for this commit: https://github.com/dotnet-bot/corefx/commit/5d9f9d4c6ea450366c7b70af6e6006a401284d14


roncain:
LGTM.  I did a buddy-build of this PR and it passes all innerloop and outerloop tests

PR #256, opened 2015-08-06
Moving to xunit 2.1

mconnew:
Update to build tools version 69 which uses xunit 2.1.
Updating package dependencies to using xunit 2.1.
Moved System.Net.NameResolution to the dnxcore50 dependencies as it's not available and not needed on .Net Native

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -10,7 +10,7 @@
 
   <!-- Build Tools Versions -->
   <PropertyGroup>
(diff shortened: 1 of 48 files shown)


roncain:
LGTM.  I pulled and built this PR, and it passes innerloop, outerloop and code coverage runs.

StephenBonikowsky:
Built and ran all tests as well.
:shipit:

PR #260, opened 2015-08-10
Check null/empty user name

hongdai:
* Throw the same exception type and message as desktop
in case of null/empty user name

Fix issue #224

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
@@ -178,10 +178,16 @@ namespace System.ServiceModel.Channels
                         switch (_authenticationScheme)
                         {
                             case AuthenticationSchemes.Basic:
-                                if (credentials.UserName.UserName != string.Empty)
+                                if (credentials.UserName.UserName == null)
                                 {
-                                    creds = new NetworkCredential(credentials.UserName.UserName, credentials.UserName.Password);
+                                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("userName");
                                 }
+                                if (credentials.UserName.UserName == string.Empty)
+                                {
+                                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgument(SR.UserNameCannotBeEmpty);
+                                }
+
+                                creds = new NetworkCredential(credentials.UserName.UserName, credentials.UserName.Password);
                                 break;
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -178,10 +178,16 @@ internal HttpClient GetHttpClient()
                         switch (_authenticationScheme)
                         {
                             case AuthenticationSchemes.Basic:
-                                if (credentials.UserName.UserName != string.Empty)
+                                if (credentials.UserName.UserName == null)
                                 {
-                                    creds = new NetworkCredential(credentials.UserName.UserName, credentials.UserName.Password);
+                                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("userName");
                                 }
+                                if (credentials.UserName.UserName == string.Empty)
```
dmetzgar: Would a username with only whitespace produce the same error as empty?

hongdai: Nope. This is the same behavior as the desktop. We would like to be consistent with desktop.

roncain: Just confirming my understanding of your answer to @dmetzgar -- what does happen on the desktop of you send an empty string or a string containing only whitespace?  And I notice the other AuthenticationSchemes cases below this code are inconsistent both in how they check and what they do if they find string.Empty.  Can you confirm whether we are handling all of these case statements identically to desktop?

hongdai: On desktop,
1) Empty string, It will throw ArgumentException.
2) String containing only whitespace. It's treated as non empty string and will pass validation.

As for all other case statements, I am not clear about the usage scenarios and we do not have test cases failing because of them. I would suggest we solve them separately.

roncain:
Okay, LGTM.  Please open an issue to investigate negative tests for the other Auth schemes.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
hongdai: I have opened an issue #263 to track investigating negative scenarios for other authentication schemas to ensure we will cover them.

SajayAntony:
:shipit:

PR #261, opened 2015-08-10
Cleaning up XD class of static xml dictionaries not being used.

StephenBonikowsky:
* In the past an xd.xml file was used to generate this cs file but that is no longer used.


iamjasonp:
LGTM

roncain:
LGTM

PR #274, opened 2015-08-13
Enable NTLM test and Basic Auentication negative test

hongdai:
* The related product bug are fixed
* I have verified both of the tests are passing now.


StephenBonikowsky:
:shipit:

PR #278, opened 2015-08-14
Scenario test helper for proper cleanup of all communication objects …

### LINKED ISSUE #168: Define proper try/catch/finally pattern for tests to close communication objects correctly

StephenBonikowsky:
…used in tests.

* We really need to close on Issue 168, this is my attempt at a proposal that takes into account everybodies feedback.
* We should still attempt to Close communication objects in the Try block.
* We should Close after doing validation, if both the scenario being tested and the call to Close Throw we would want to fix the scenario Throw first as it could be the cause for the Close throwing.
* The helper method ensures all com objects have been cleaned up regardless of the test results.
* Fixes #168

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
@@ -153,6 +153,44 @@ public static class ScenarioTestHelpers
 
         return compositeObject;
     }
+
+    /// <summary>
+    /// Closes com objects in the order passed in if not already closed.
+    /// If Close fails for Timeout or Communication exception then Aborts.
+    /// </summary>
+    /// <param name="objects">Any communication objects that need to be cleaned up.
+    /// In the order in which they need to be cleaned up.</param>
+    public static void CloseCommunicationObjects(params ICommunicationObject[] objects)
+    {
+        foreach(ICommunicationObject comObj in objects)
+        {
+            try
+            {
+                if(comObj == null)
+                {
+                    continue;
+                }
+                // Only want to call Close if it is in the Opened state
+                if(comObj.State == CommunicationState.Opened)
+                {
+                    comObj.Close();
+                }
+                // Anything not closed by this point should be aborted
+                if(comObj.State != CommunicationState.Closed)
+                {
+                    comObj.Abort();
+                }
+            }
+            catch(TimeoutException)
+            {
+                comObj.Abort();
+            }
+            catch(CommunicationException)
+            {
+                comObj.Abort();
+            }
+        }
+    }
 }
 
 
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
@@ -16,32 +16,32 @@ public static class Binding_Http_BasicHttpBindingTests
     [OuterLoop]
     public static void DefaultSettings_Echo_RoundTrips_String()
     {
-        string variationDetails = "Client:: BasicHttpBinding/DefaultValues\nServer:: BasicHttpBinding/DefaultValues";
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
         string testString = "Hello";
-        StringBuilder errorBuilder = new StringBuilder();
-        bool success = false;
-
-        BasicHttpBinding binding = new BasicHttpBinding(BasicHttpSecurityMode.None);
+        Binding binding = null;
 
         try
         {
-            ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
+            // *** SETUP *** \\
+            binding = new BasicHttpBinding(BasicHttpSecurityMode.None);
+            factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
+            serviceProxy = factory.CreateChannel();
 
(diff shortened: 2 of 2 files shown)


StephenBonikowsky:
Please note, this PR does not address how to do the negative tests that expect an exception. I would like to compartmentalize a little bit and work on that as part of #74 

I'm not sure if I like the comments I threw into the test such as // *** ASSEMBLE *** \\ etc,.. please provide feedback.

I am proposing that in every test at the beginning we declare types first that we will subsequently use in the try block and that we always close any communication objects we use while still in the try block.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
```diff
@@ -16,32 +16,32 @@ public static class Binding_Http_BasicHttpBindingTests
     [OuterLoop]
     public static void DefaultSettings_Echo_RoundTrips_String()
     {
-        string variationDetails = "Client:: BasicHttpBinding/DefaultValues\nServer:: BasicHttpBinding/DefaultValues";
+        // *** ASSEMBLE *** \\
+        ChannelFactory<IWcfService> factory = null;
```
roncain: Nit: the terms the WebAPI uses here are "Arrange", "Act" and "Assert".  We are free to choose our own of course but I recommend we follow some convention.  Personally I think I like "Setup", "Act", "Validate" and "Cleanup".  I have no strong opinion for the "Force Cleanup".

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
```diff
@@ -16,32 +16,32 @@ public static class Binding_Http_BasicHttpBindingTests
...
+            factory.Close();
+            ((ICommunicationObject)serviceProxy).Close();
         }
-        catch (Exception ex)
+        // *** FORCE CLEANUP *** \\
+        finally
         {
-            errorBuilder.AppendLine(String.Format("    Error: Unexpected exception was caught while doing the basic echo test for variation...\n'{0}'\nException: {1}", variationDetails, ex.ToString()));
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, (ICommunicationObject)factory);
         }
```
roncain: Were you really forced to cast these explicitly to ICommunicationObject like this/  I would have expected the compiler to just let you pass them in without the cast (because they all implement ICommunicationObject). I'd prefer the shortest possible form.

roncain:
LGTM.  This answers most people's concerns in #168.  If a Close() throws on an otherwise successful test, you will see the exception and the test will fail.  But if an Assert bypasses the normal Close(), the object is still (silently) closed and the original Assert is what the user sees.  I like this approach.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.cs
StephenBonikowsky: I was getting compiler errors if I didn't cast them first.

StephenBonikowsky:
Updated with some one-on-one feedback from Ron.

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
```diff
@@ -153,6 +153,42 @@ public static ComplexCompositeType GetInitializedComplexCompositeType()
...
+    /// </summary>
+    /// <param name="objects">Any communication objects that need to be cleaned up.
+    /// In the order in which they need to be cleaned up.</param>
+    public static void CloseCommunicationObjects(params ICommunicationObject[] objects)
+    {
+        foreach(ICommunicationObject comObj in objects)
+        {
+            try
+            {
+                if(comObj != null)
```
iamjasonp: do you mean `if (comObj == null)` here?

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
```diff
@@ -153,6 +153,42 @@ public static ComplexCompositeType GetInitializedComplexCompositeType()
...
+                {
+                    continue;
+                }
+                if(comObj.State == CommunicationState.Opened)
+                {
+                    comObj.Close();
+                }
+                else if (comObj.State != CommunicationState.Closed)
```
iamjasonp: Should this be an `if` rather than `else if`? After the call to Close(), we're expecting the channel to be Closed regardless of whether or not the channel was in an Opened state when we entered the method.

roncain: +1.  Good point @iamjasonp .  Agree the "else" should be removed

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
StephenBonikowsky: woops

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
StephenBonikowsky: Seems like a stylistic issue? I can see how it makes more sense to not use the else if.

PR #279, opened 2015-08-14
Improve Bridge firewall support

roncain:
These changes allow the firewall rules created by the Bridge
to be scoped to specific IP's to allow them to be secured against
unauthorized access.

Remotely starting the Bridge has been made easier with a new script.
and a new markdown file describes how to run the Bridge remotely.

### DIFF
--- a/Documentation/cross-machine-test-guide.md
+++ b/Documentation/cross-machine-test-guide.md
@@ -1,0 +1,132 @@
+The Bridge and cross-machine testing
+====================================
+
+Bridge overview
+---------------
+The basic OuterLoop scenario tests normally run against locally hosted
+WCF test services.  But some scenarios require running the WCF test
+service(s) on a machine other than the machine running the test
+(example: cross-platform testing).
+
+For this we have developed what we call the "Bridge." This Bridge is
+a WebAPI application meant to run on a Windows OS using the full
+.NET framework.  It offers a REST API that the scenario tests can
+invoke through normal HTTP requests to launch WCF services on other
+machines.
+
+The Bridge itself is agnostic of WCF services. Instead, it is aware
+of types that implement IResource, and it uses reflection to analyze
+a collection of assemblies to discover them.  It then exposes each of these
+as a "resource" to the Bridge REST API.  When a test requests a particular
+resource, the Bridge invokes the appropriate IResource.  In this case, the
+WCF tests provide IResources that know how to create and host WCF services.
+
+When the tests start up, they inform the Bridge where to find these
+resource assemblies.  This folder is known as the "Bridge Resource Folder".
+
+There are several Bridge-specific properties that are meaningful to
+both the Bridge and the tests.  They can be defined either as Environment
+variables or passed into build.cmd as MSBuild properties.  They are:
+
+  - **BuildHost**: the name of the machine on which the Bridge runs (default localhost)
+  - **BridgePort**: the port to use to communicate with the Bridge (default 44283)
+  - **BridgeResourceFolder**: the full path to the folder that contains the Bridge IResource assemblies (default bin\Wcf\Bridge\Resources)
+  - **BridgeAllowRemote**: indicates whether the Bridge will accept requests from machines other than itself (default: false)
+
+Starting the Bridge Locally
+---------------------------
+By default, the Bridge will self-start locally during OuterLoop tests and accept only
+requests from the machine running the tests.  It will shutdown when all the OuterLoop
+tests have completed.
+
+If you want to run the Bridge on a different machine, you must follow the steps below.
+
+Starting the Bridge on a different machine
+------------------------------------------
+On the Bridge where the machine will run, do these things:
+
+cd to the root folder of a local WCF Git repository and run this command:
+
+```
+    startBridge.cmd {options}
+```
+
+The options can be any of these:
+
+    -portNumber NNN
+    -allowRemote true/false
+    -remoteAddresses comma-separated-list
+
+The 'portNumber' option allows you to choose a port other than
+the default 44283.
+
+The 'allowRemote' option tells the Bridge it is allowed to accept
+requests from other machines.  If unspecified, it will accept only
+from localhost.  It must be set to true if the Bridge is running
+on a machine other than where the tests run.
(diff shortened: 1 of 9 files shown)


Documentation/cross-machine-test-guide.md
```diff
@@ -0,0 +1,132 @@
...
+The 'allowRemote' option tells the Bridge it is allowed to accept
+requests from other machines.  If unspecified, it will accept only
+from localhost.  It must be set to true if the Bridge is running
+on a machine other than where the tests run.
+
+The 'remoteAddresses' is a comma-separated list of IP addresses,
+a range of IP's, or one of the supported predefined terms supported
+for the Scope properties page in the Windows Firewall.
+See https://technet.microsoft.com/en-us/library/dd759059.aspx .
+
```
roncain: This statement about 'remoteAddresses' is a partial lie.  I can set values like LocalSubnet using this mechanism, but I have not found a way in the API to set other predefined names like "intranet".  I will continue to investigate.

src/System.Private.ServiceModel/tools/setupfiles/SetupWCFTestService.cmd
```diff
@@ -21,7 +21,7 @@ if '%BridgeAllowRemote%' neq '' (
    set _bridgeAllowRemoteArg=-allowRemote %BridgeAllowRemote%
 )
 
-echo Executing: start powershell -ExecutionPolicy Bypass -File ..\test\Bridge\bin\ensureBridge.ps1 %_bridgeHostArg% %_bridgePortArg% %_bridgeAllowRemoteArg%
+echo Executing: start powershell -ExecutionPolicy Bypass -File ..\test\Bridge\bin\ensureBridge.ps1 %_bridgeHostArg% %_bridgePortArg% %_bridgeAllowRemoteArg% %2 %3 %4 %5 %6 %7
 
-start powershell -ExecutionPolicy Bypass -File ..\test\Bridge\bin\ensureBridge.ps1 %_bridgeHostArg% %_bridgePortArg% %_bridgeAllowRemoteArg%
+start powershell -ExecutionPolicy Bypass -File ..\test\Bridge\bin\ensureBridge.ps1 %_bridgeHostArg% %_bridgePortArg% %_bridgeAllowRemoteArg% %2 %3 %4 %5 %6 %7
```
iamjasonp: NIT: can we do something like `%*` rather than `%2 %3 %4 %5 %6 %7` ?

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/PortManager.cs
```diff
@@ -26,82 +60,111 @@ private static void RegisterForProcessExit()
...
-                Type netFwPolicy2Type = Type.GetTypeFromProgID("HNetCfg.FwPolicy2", false);
-                INetFwPolicy2 fwPolicy2 = (INetFwPolicy2)Activator.CreateInstance(netFwPolicy2Type);
-
-                // Get the Rules object
-                INetFwRules rulesObject = fwPolicy2.Rules;
+                // If we already created this rule, we don't create it again.
+                string ruleName = String.Format("{0} {1}", s_RuleNamePrefix, port);
+                if (FindRule(ruleName, port) != null)
+                {
+                    return;
```
iamjasonp: Is there a potential for conflict if two rules are added to open port 44283 with different names?

iamjasonp: Or maybe you found a rule that blocks, in which case, the block rule might supersede the allow rule. Consider displaying a warning if a potential conflict situation exists. 

Suggestion only, the chances are probably really slim that there would be an issue arising here.

src/System.Private.ServiceModel/tools/setupfiles/SetupWCFTestService.cmd
roncain: Not unless we can find a way to exclude %1 which is used already.  I intend to nuke ensureBridge.ps1 in favor of code on a subsequent checkin

mconnew: Does using the shift command work?

roncain: I think I will just remove the need to pass the working folder through %1. But remember, I intend to remove as many of these CMD --> VBS --> PS --> CMD --> EXE transitions as I can shortly.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/PortManager.cs
roncain: Unless someone pushes back, I don't think reverse engineering firewall logic is a good place to spend resources.  Think about all the permutations after you consider other firewall properties like localports, remoteports, grouping, lists of IP's, etc.  For example, maybe there are 10 rules for this single port, each with different IP's.  If our rule enables for "LocalSubNet" do I then need to understand whether those blocked IP's would affect our rule.  I could detect and report if I find other rules for the same port, but I suspect it will just produce noise we will learn to ignore.

iamjasonp: I suggested just in case we wanted to give some debuggability to those who are using this script on their own. For us, I think I'm reasonably certain that our own environment wouldn't run into this. 

Yeah, I agree - forget this, we have better things to spend our time on.

iamjasonp:
Seems fine and LGTM. I pulled this and ran it on another box based on the instructions in cross-machine-test-guide.md, and it looks fine. 

Be sure to include a link to cross-machine-test-guide.md from readme.md

src/System.Private.ServiceModel/tools/setupfiles/SetupWCFTestService.cmd
roncain: 2nd commit eliminates need to pass working directory on command line, allowing %*

PR #283, opened 2015-08-17
Fixes EndpointNotFound test to work when Bridge is remote

roncain:
The test was assuming localhost for the host rather than where the
Bridge was actually running.  It was also possible to remove the
custom IResource for the endpoint-not-found case and replace it
with use of a normal running service with a bogus endpoint suffix.


StephenBonikowsky:
:shipit:

iamjasonp:
Good catch - LGTM

PR #284, opened 2015-08-18
Bridge port configure

roncain:


roncain:
Reviewers: this PR is preparation for getting https working when the Bridge is on another machine.  I initially started on that effort but found too many hard-coded dependencies on localhost, hard-coded port numbers, and some misc factoring issues.  This PR addresses those first so that enabling http x-machine is not muddied with this cleanup.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/App.config
```diff
@@ -2,180 +2,6 @@
...
-                    <userNameAuthentication userNamePasswordValidationMode="Custom" customUserNamePasswordValidatorType="WcfService.CustomUserNameValidator, WcfService"/>
-                  </serviceCredentials>
-                </behavior>
-            </serviceBehaviors>
-        </behaviors>
-        <protocolMapping>
-            <add binding="basicHttpBinding" scheme="http" />
-        </protocolMapping>
-    </system.serviceModel>
     <system.diagnostics>
```
StephenBonikowsky: Thanks for cleaning this up, I had forgotten about it.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
```diff
@@ -66,6 +70,71 @@ public static string GetResourceAddress(string resourceName)
...
+                    _BridgeStatus = BridgeState.Faulted;
+                    throw new Exception("Bridge is not running", exc);
+                }
+            }
+        }
+
+        // Returns true if the Bridge will respond or has responded to localhost.
+        // This is used to determine whether local file paths are acceptable for the BridgeResourceFolder.
+        private static bool DoesBridgeRespondToLocalHost()
+        {
```
mconnew: I've had the build scripts launch the bridge locally even though I specified a remote host. The tests still ran against the remote host but I had a local instance still running. This code would return true in that situation.

roncain: Are you saying it should be a supported scenario that it is running in both machines?  Or that there is an error in the scripts that launched it remotely when you had it running remote?

StephenBonikowsky:
LGTM
All OuterLoop tests passed as well.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
mconnew: Not a supported scenario, this is what happened for me when running the build script. It launched Bridge.exe locally but the tests (correctly) ran against the remote host.

PR #288, opened 2015-08-20
Enable Bridge certificate configuration

roncain:
This PR allows the Bridge to install and uninstall certificates
based on the resources that need them.

Specific changes are:
  - Hard-coded knowledge of specific thumbprints removed
  - New .cer and .pfx files generated for localhost
  - New TestProperties and BridgeConfiguration properties
    to make these certificate names configurable
  - New CertificateManager class to manage installing and uninstalling


roncain:
Reviewers: getting this to work across machines will happen in multiple passes.  This PR is scoped to getting the current self-hosted scenario to pass all tests.  After this is merged, I will issue another PR to generate certs on-demand for cross-machine scenarios (and try to remove netsh process invokes).

All feedback welcome, but please help me by indicating whether your feedback needs to be addressed for this PR to be accepted vs. a suggestion for how to improve something in the next PR.

dmetzgar:
LGTM

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateManager.cs
```diff
@@ -0,0 +1,236 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace WcfTestBridgeCommon
+{
+    public class CertificateManager
```
iamjasonp: NIT: Could be made a static class

iamjasonp:
Ah, dang, my comment didn't go through the first time. 

For whatever it's worth now, LGTM :smile:

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateManager.cs
```diff
@@ -0,0 +1,236 @@
...
+                store.Add(cert);
+                store.Close();
+
+                s_myCertificates[cert.Thumbprint] = certificateName;
+                return cert.Thumbprint;
+            }
+        }
+
+        private static void UninstallAllRootCertificates()
+        {
```
mconnew: Although not a blocker for this PR, I would like to see it in a follow up one. We should have some mechanism to identify left behind certificates from a previous run to be able to clean them up. Similar to how the firewall rules are done.

PR #293, opened 2015-08-26
Removed unused events

dmetzgar:
Removed EventSource events that are not currently being used. This should address the performance problems we're seeing in the ToF tests.


roncain:
LGTM -- question @dmetzgar , is it the intent to solve whatever the reentrancy (or whatever) issue was causing the issue and to check this back in at some future time?

dmetzgar:
@roncain The original PR ported the entire WCF/WF ETW provider. It included WF, WF tracking, events related to ServiceHost, things like that. Many of the events won't apply to WCF on Core. As features like discovery are added, those events could be added back in. For right now, I'm just looking at what is actually being referenced.

There is a possibility that this may not completely solve the performance problem. It may not be the number of events but the max event id (>62K). If that's the case, I'll also need to make a fix to the EventSource code in coreclr.

roncain:
@dmetzgar and @mconnew -- this LGTM.  I recommend it be merged asap because we are seeing these test failures in nightly test runs of the CoreCLR tests now.

dmetzgar:
@roncain Is it possible to kick off one of the nightly test runs?

Issue #292, opened 2015-08-26
XmlSerializer does not work on NetTcp/Duplex MEP


StephenBonikowsky:
The System.ServiceModel.Dispatcher.DispatchOperation field (IDispatchMessageFormatter) _formatter is not getting set.

System.InvalidOperationException occurred
Message=The DispatchOperation 'OnXmlPingCallback' requires Formatter, since DeserializeRequest and SerializeReply are not both false.
  
  StackTrace:
system.private.servicemodel.dll!System.ServiceModel.Dispatcher.DispatchOperationRuntime.DispatchOperationRuntime(System.ServiceModel.Dispatcher.DispatchOperation operation, System.ServiceModel.Dispatcher.ImmutableDispatchRuntime parent) Line 66
system.private.servicemodel.dll!System.ServiceModel.Dispatcher.ImmutableDispatchRuntime.ImmutableDispatchRuntime(System.ServiceModel.Dispatcher.DispatchRuntime dispatch) Line 66
system.private.servicemodel.dll!System.ServiceModel.Dispatcher.DispatchRuntime.GetRuntimeCore() Line 322
system.private.servicemodel.dll!System.ServiceModel.Dispatcher.DispatchRuntime.GetRuntime() Line 312
system.private.servicemodel.dll!System.ServiceModel.Dispatcher.DispatchRuntime.GetOperation(ref System.ServiceModel.Channels.Message message) Line 299
system.private.servicemodel.dll!System.ServiceModel.Dispatcher.ChannelHandler.DispatchAndReleasePump(System.ServiceModel.Channels.RequestContext request, bool cleanThread, System.ServiceModel.OperationContext currentOperationContext) Line 264
system.private.servicemodel.dll!System.ServiceModel.Dispatcher.ChannelHandler.HandleRequest(System.ServiceModel.Channels.RequestContext request, System.ServiceModel.OperationContext currentOperationContext) Line 711
system.private.servicemodel.dll!System.ServiceModel.Dispatcher.ChannelHandler.AsyncMessagePump(System.IAsyncResult result) Line 214
system.private.servicemodel.dll!System.ServiceModel.Dispatcher.ChannelHandler.OnAsyncReceiveComplete(System.IAsyncResult result) Line 752
system.private.servicemodel.dll!System.Runtime.Fx.AsyncThunk.UnhandledExceptionFrame(System.IAsyncResult result) Line 990
[...]

PR #294, opened 2015-08-26
Adding test cases to resolve Issues 250 and 251

### LINKED ISSUE #250: Create a duplex scenario test to pass a complex type in callback not used in service contract
We need a test where the duplex "Ping" callback passes a complex type that is not used in any service contract.  This will verify whether a serializer has been generated for it in NET Native.  It is critical this type not appear in any service contract's operations, or the serializer will have been generated from that usage.  It is also critical that no test explicitly serializes or deserializes this type, or the toolchain will create a serializer based on that usage.

### LINKED ISSUE #251: Create a duplex scenario test to use Xml in the callback contract
First, determine whether it is a supported scenario for a duplex callback contract to specify an [XmlSerializerFormat] attribute or an XmlSerializerOperationBehavior so that its "Ping" callback passes the value as xml rather than data contract.  And for the type, use a complex type not used in any service contract.

If supported, create a scenario test that does this.  This is to verify 3 things:
1. An xml serializer will be generated in NET Native for this type
2. The ContractDescription creation correctly recognizes this operation behavior on a callback contract
3. The e2e serialization+deserialization works correctly

StephenBonikowsky:
* Fixes #250 and Fixes #251

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
@@ -688,6 +688,61 @@ public class XmlCompositeType
     }
 }
 
+// This type should only be used by test Contract.DataContractTests.NetTcpBinding_DuplexCallback_ReturnsDataContractComplexType
+// It tests a narrow scenario that returns a DataContract attributed type in the callback method that is not known by the ServiceContract attributed interface
+// This test is designed to make sure the NET Native toolchain creates the needed serializer
+[DataContract(Namespace = "http://www.contoso.com/wcfnamespace")]
+public class ComplexCompositeTypeDuplexCallbackOnly : IEquatable<ComplexCompositeTypeDuplexCallbackOnly>
+{
+    private Guid _guidValue;
+
+    [DataMember]
+    public Guid GuidValue
+    {
+        get { return _guidValue; }
+        set { _guidValue = value; }
+    }
+
+    public bool Equals(ComplexCompositeTypeDuplexCallbackOnly other)
+    {
+        if (other == null) { return false; }
+        if (object.ReferenceEquals(this, other)) { return true; }
+
+        if (_guidValue != other._guidValue) { return false; }
+
+        return true;
+    }
+
+    public override string ToString()
+    {
+        StringBuilder sb = new StringBuilder();
+        sb.AppendLine("GuidValue: " + _guidValue);
+
+        return sb.ToString();
+    }
+}
+
+// This type should only be used by test Contract.DataContractTests.NetTcpBinding_DuplexCallback_ReturnsXmlComplexType
+// It tests a narrow scenario that returns an Xml attributed type in the callback method that is not known by the ServiceContract attributed interface
+// This test is designed to make sure the NET Native toolchain creates the needed serializer
+public class XmlCompositeTypeDuplexCallbackOnly
+{
+    private bool _boolValue = true;
+    private string _stringValue = "Hello ";
+
+    public bool BoolValue
+    {
+        get { return _boolValue; }
+        set { _boolValue = value; }
+    }
+
+    public string StringValue
+    {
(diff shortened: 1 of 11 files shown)


StephenBonikowsky:
These tests have not been run against NET Native yet.
The Xml test fails, see Issue #292

StephenBonikowsky:

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
```diff
@@ -803,4 +831,26 @@ public void OnPingCallback(Guid guid)
...
     }
+
+    public void OnXmlPingCallback(XmlCompositeType xmlCompositeType)
+    {
+        // Set the result in an async task with a 100ms delay to prevent a race condition
+        // where the OnPingCallback hasn't sent the reply to the server before the channel is closed.
+        Task.Run(async () =>
+        {
+            await Task.Delay(100);
+            _xml_tcs.SetResult(xmlCompositeType);
```
roncain: Just a question -- why 100ms?  Don't block on this, I was just wondering if there is some other way to detect when it's okay to set the TCS.

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
```diff
@@ -793,6 +797,30 @@ public Guid CallbackGuid
...
+            if (_xml_tcs.Task.Wait(ScenarioTestHelpers.TestTimeout))
+            {
+                return _xml_tcs.Task.Result;
+            }
+            throw new TimeoutException(string.Format("Not completed within the alloted time of {0}", ScenarioTestHelpers.TestTimeout));
+        }
+    }
+
+    public ComplexCompositeType DataContractCallbackGuid
+    {
```
roncain: Possibly add a comment that no service contract is permitted to use this type.  It is reserved for the scenario that is testing that its appearance in a callback contract generates a serializer.  Probably same for the one above.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/DuplexResources.cs
```diff
@@ -57,4 +57,38 @@ protected override Binding GetBinding()
...
+    }
+
+    internal class DuplexCallbackXmlComplexTypeResource : EndpointResource<WcfDuplexService, IWcfDuplexService_Xml>
+    {
+        protected override string Protocol { get { return BaseAddressResource.Tcp; } }
+
+        protected override int GetPort(ResourceRequestContext context)
+        {
+            return context.BridgeConfiguration.BridgeTcpPort + 5;
+        }
```
roncain: I see you found my new way of choosing TCP ports :smile: .  Not now, but let's consider making some enum somewhere that gives them names.

roncain:
LGTM -- all my comments are optional before merging.

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
StephenBonikowsky: Not sure, I think either Matt or Jason added this functionality to the existing 'OnPingCallback' which I then adopted., I can confirm the race condition does occur.
I'm not sure how else we could do it, I can't see a way to make the cleanup code smarter for this particular scenario, but there may be a solution there.

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
StephenBonikowsky: Done, I also created unique versions as I was re-using these types. I agree that having unique/use once types is better/safer than trying to figure out how they are safe to use by other tests without messing up these tests.

roncain:
Still LGTM @StephenBonikowsky and it passes all InnerLoop and OuterLoop tests.

PR #299, opened 2015-08-31
Adding IOperationBehavior unit test

### LINKED ISSUE #280: Create a scenario test that uses IOperationBehavior on a service contract
Steps:
1. Create a custom attribute that implements IOperationBehavior
2, Find a way to make this attribute a "test spy" that can report that its methods get called.
3. Add this attribute to at least one service operation in the service interface
4. Use a ClientBase<T> test to invoke that operation and verify that custom attribute's code was executed.

We have evidence that some Windows Store apps (Arkadium.SharedEntities) do this and need to verify that it works on a service interface.  It will also increase our code coverage and verify whether we ever need to reflect over the concrete implementation of that service interface (look at code coverage of TypeLoader.ProcessOpMethod.

StephenBonikowsky:
* Cleanup a few interfaces no longer needed in unit common code
* Moved WcfDuplexServiceCallback to common code so it could be re-used
* Fixes #280

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Unit/TestHelpers.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Unit/TestHelpers.cs
@@ -9,12 +9,14 @@ using System.ServiceModel.Description;
 using System.ServiceModel.Dispatcher;
 using System.Text;
 using System.Threading;
+using System.Threading.Tasks;
 using System.Xml;
 
 public static class TestHelpers
 {
     private const string testString = "Hello";
     public static readonly TimeSpan TestTimeout = TimeSpan.FromSeconds(20);
+    public static ManualResetEvent mre;
 
     public static bool XmlDictionaryReaderQuotasAreEqual(XmlDictionaryReaderQuotas a, XmlDictionaryReaderQuotas b)
     {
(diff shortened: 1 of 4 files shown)


src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
```diff
@@ -0,0 +1,41 @@
...
+        binding.Security.Mode = SecurityMode.None;
+
+        WcfDuplexServiceCallback callbackService = new WcfDuplexServiceCallback();
+        InstanceContext context = new InstanceContext(callbackService);
+
+        duplexService = new MyDuplexClientBase<ICustomOperationBehaviorDuplexService>(context, binding, new EndpointAddress(FakeAddress.TcpAddress));
+        proxy = duplexService.ChannelFactory.CreateChannel();
+
+        // Allow work to be done before validation
+        Task t = Task.Delay(1000);
```
StephenBonikowsky: If there is a better way to do this let me know.

roncain: Question only: Why does it need to wait at all?  Is there something asynchronous happening that is not done when CreateChannel returns?

src/System.Private.ServiceModel/tests/Common/Unit/TestHelpers.cs
```diff
@@ -276,3 +277,69 @@ protected override void OnWriteBodyContents(XmlDictionaryWriter writer)
...
+        {
+            errorBuilder.AppendLine(String.Format("A parameter passed into the ApplyClientBehavior method was null/nThe null parameter is: {0}", typeof(ClientOperation).ToString()));
+        }
+    }
+
+    // Implements a modification or extension of the service across an operation
+    public void ApplyDispatchBehavior(OperationDescription operationDescription, DispatchOperation dispatchOperation)
+    {
+        // This method does not get called client side.
+    }
```
roncain: Optional: it's not *supposed* to get called, but you could make the unit test also handle that negative case by writing to the errorBuilder "ApplyDispatchBehavior should not have been called".

src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
```diff
@@ -0,0 +1,41 @@
...
+using System;
+using System.ServiceModel;
+using System.Threading.Tasks;
+using Xunit;
+
+public static class OperationBehaviorTest
+{
+    [Fact]
+    [ActiveIssue(280)]
+    public static void IOperationBehavior_Methods_AreCalled()
```
roncain: I know you mentioned this doesn't work and have opened #280.  Does this code work on the desktop?

src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
StephenBonikowsky: Its just a race after the call to CreateChannel between the IOperationMethods getting called and setting the collection versus checking the collection count for validation.

src/System.Private.ServiceModel/tests/Common/Unit/TestHelpers.cs
StephenBonikowsky: true

src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
StephenBonikowsky: yes

src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
roncain: Can you loop and check that the collection count is the expected amount, and if not, sleep a few milliseconds and retry (up to some limit)?   It might get the test to run faster

StephenBonikowsky: Thread.Sleep is in NET Core CLR but not in NET Native.

iamjasonp: Can we do something like: 

```
int iterations = 10; 
while (iterations > 0)
{
    Task.Delay(100).Wait(); 
    if (...collection.Count > 3) { break; } 
    iterations--;
}
```

as a tight loop or something? 

I agree with Ron though, we can worry about optimizing later for runtime. LGTM

iamjasonp:
If we can't think of a good way to deal with the `Task.Delay(1000); ` for now, let's not block on this

LGTM

StephenBonikowsky:
My apologies for amending the initial commit with my latest changes, should have just done a second commit so the changes were easier to see.
Since the test fails now due to Issue #315 I tested the logic I used to set the manualresetevent on desktop.

src/System.Private.ServiceModel/tests/Common/Unit/TestHelpers.cs
```diff
@@ -276,3 +279,89 @@ protected override void OnWriteBodyContents(XmlDictionaryWriter writer)
...
+    {
+        applyClientBehaviorMethod = true;
+        if ((validateMethod == false) || (addBindingParametersMethod == false))
+        {
+            errorBuilder.AppendLine(String.Format("ApplyClientBehavior {1}", errorMessage));
+        }
+        else
+        {
+            mre.Set();
+        }
```
roncain: I'd recommend doing the other error checking before the Set, otherwise your code to test this might be able to execute before you finished with the code below the Set.  In other words, I'd do all the errorBuilder stuff first.

src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        WcfDuplexServiceCallback callbackService = new WcfDuplexServiceCallback();
+        InstanceContext context = new InstanceContext(callbackService);
+
+        duplexService = new MyDuplexClientBase<ICustomOperationBehaviorDuplexService>(context, binding, new EndpointAddress(FakeAddress.TcpAddress));
+        proxy = duplexService.ChannelFactory.CreateChannel();
+
+        // Wait to validate until the process has been given a reasonable time to complete.
+        MyOperationBehavior.mre.WaitOne(250);
+
```
roncain: We should add a test about WaitOne's return value.  If it returns 'true' the mre.Set was called.  If it return 'false' it means it timed out.  So we should Assert.True(waitOneResult, "IOperationBehavior methods were not called").  It might also be useful that in the 'false' case it shows which of the methods *were* called.

StephenBonikowsky:
When @SajayAntony asked me why I didn't use Task.WaitAll, I learned about it from @mconnew and couldn't resist using it.
Please review one more time.

src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
```diff
@@ -0,0 +1,36 @@
...
+        WcfDuplexServiceCallback callbackService = new WcfDuplexServiceCallback();
+        InstanceContext context = new InstanceContext(callbackService);
+
+        duplexService = new MyDuplexClientBase<ICustomOperationBehaviorDuplexService>(context, binding, new EndpointAddress(FakeAddress.TcpAddress));
+        proxy = duplexService.ChannelFactory.CreateChannel();
+
+        // Wait to validate until the process has been given a reasonable time to complete.
+        Task[] taskCollection = { MyOperationBehavior.validateMethodTcs.Task, MyOperationBehavior.addBindingParametersMethodTcs.Task, MyOperationBehavior.applyClientBehaviorMethodTcs.Task };
+        bool waitAll = Task.WaitAll(taskCollection, 250);
```
SajayAntony: Nice :+1:

SajayAntony:
LGTM :shipit:

iamjasonp:
Nice! LGTM

PR #307, opened 2015-09-01
Changed event ids for the WCF EventSource to fix issue with ToF chk tests

dmetzgar:


mconnew:
This makes the events incompatible with desktop and breaks the events showing up in event viewer. Any existing tooling to parse events will break. For example svclog viewer won't understand these events.

roncain:
@mconnew yes, the change in event ID's was discussed briefly.  Do you have a different proposal to resolve https://github.com/dotnet/coreclr/issues/1461?  As is, running CHK builds fail in our TFS mirror site because the static ctor causes the TestHost.exe(s) to use 100% cpu for several minutes.

dmetzgar:
I've posted an update to the discussion at dotnet/coreclr#1461. Basically there are two bugs in the EventSource code and I proposed a fix there that would alleviate our problem. If I make a PR to coreclr, how long would the turnaround time be before you had an updated System.Diagnostics.Tracing package? The WCF PR would get us past the issue until the other fix gets around to us. Maybe we could also #if it for chk builds?

roncain:
Thanks for investigating so thoroughly @dmetzgar .  I think you have identified issues with EventSource that should be fixed (for more than just WCF) while we figure out our story.  As I understand it though, doesn't this issue occur only when dealing with a CHK EventSource, which is why we don't see it running OSS tests (because the NuGet packages for EventSource contain RET binaries)?  So an #if inside WCF would only coincidentally work in the ToF world where both WCF and EventSource are CHK, but would probably give us an odd debugging experience in OSS where WCF is CHK but EventSource is RET (we would use the CHK event ID's against a RET EventSource that doesn't have the issue).
Here is another idea.  I'm not thrilled about it but it would work for us.  Make an #if that conditions the new event ID numbering -- call it something like RENUMBERED_EVENT_IDSm but don't set that define anywhere in OSS code.  Then we can change the "wrapper" .csproj in TFS to set this define only in CHK builds.  This means the new ids would be used only when TFS creates a CHK build of WCF, which really only happens in ToF.  The pure OSS version would continue to use the old event ids.  Thoughts?

dmetzgar:
@mconnew and I were talking this morning about finding whatever bit is being set to generate the manifest. Unfortunately it's buried somewhere where I can't adjust it. Like Vance said, there's nothing obvious in the code that makes it take a different path if you're using a chk build. Perhaps the machine running the ToF tests doesn't have the .Net framework installed and it thinks it has to generate the manifest? I thought the chk and ret tests would happen on the same machines though so I doubt it.

I'm reading the docs to figure out how to build and test a fix to EventSource. But that's still going to take time. Is it possible to leave the ToF tests failing with a known issue?

roncain:
Thanks @dmetzgar .  I think it's okay to temporarily allow CHK builds of CoreCLR to fail with this while we investigate.  Note that it happens on my VM's too. I think any dev machine attempting to run the ToF tests with a CHK build will fail, so it is not an anomaly of nightly build machines.  I can try experiments on any of my VM's if it will help. 

There is another RET/CHK issue with NET Native which is conceivably related.  Nightly test runs of CHK builds in NET Native fail because resource lookup from CHK ServiceModel builds appears to give different behavior when consuming RET NuGet packages.  This may be a NET Native only issue right now, but worth mentioning.

I still think we could temporarily #if the ID renumbering and define that particular DefineConstant in the wrapper projects that exist only in TFS.  It would not pollute the GitHub code with what appears to be a ToF specific issue.

dmetzgar:
Came up with a better PR, #329

PR #317, opened 2015-09-04
Fix and re-enable disabled tests due to Issue #301

### LINKED ISSUE #301: DuplexChannelFactoryTests will fail due to lack of WindowsStreamSecurity
There are numerous tests that fail in DuplexChannelFactory testing because of a dependency on NetTcpBinding's use of WindowsStreamSecurity - #300. 

Upon calling `CreateChannel`, the factory will throw an exception as 
`System.ServiceModel.Channels.WindowsStreamSecurityBindingElement.BuildClientStreamUpgradeProvider(BindingContext context)` is not yet implemented. 

These tests would previously pass due to the binding entirely omitting WindowsStreamSecurity, but as we are turning on more features (and hence stubbing others), we have run into this. 

Fix by either implementing #300, or by changing the tests not to depend on the default ctor of NetTcpBinding.

StephenBonikowsky:
See Issue 301 for my analysis of every test that was disabled and my proposed action for each.
The three tests in this PR are the only ones I suggest re-enabling at this point.

Fixes #301

### DIFF
--- a/src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
+++ b/src/System.ServiceModel.Primitives/tests/ServiceModel/DuplexChannelFactoryTest.cs
@@ -10,12 +10,12 @@ using Xunit;
 public class DuplexChannelFactoryTest
 {
     [Fact]
-    [ActiveIssue(301)]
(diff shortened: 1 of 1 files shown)


iamjasonp:
LGTM

PR #318, opened 2015-09-04
Fix for Issue #315

### LINKED ISSUE #315: Incorrect logic in GetCustomAttributes breaks IOperationBehavior
System\ServiceModel\Description\ServiceReflector.GetCustomAttributes expects that all CustomAttributes it gets derive from Attribute. This is incorrect as CustomAttributes can also derive from interfaces such as IOperationBehavior.

The logic needs to be changed to get all CustomAttributes on the Operation and add to the collection any that match the Type being looked for.

StephenBonikowsky:
Fixes #315

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Description/ServiceReflector.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Description/ServiceReflector.cs
@@ -457,7 +457,17 @@ namespace System.ServiceModel.Description
                 }
                 else
                 {
-                    return Array.Empty<object>();
+                    List<object> attrTypeAttributes = new List<object>();
+                    object[] allCustomAttributes = attrProvider.GetCustomAttributes(inherit);
+                    foreach (object customAttribute in allCustomAttributes)
+                    {
+                        if (attrType.IsAssignableFrom(customAttribute.GetType()))
+                        {
+                            attrTypeAttributes.Add(customAttribute);
+                        }
+                    }
+
+                    return attrTypeAttributes.ToArray();
                 }
             }
             catch (Exception e)
--- a/src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
+++ b/src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
@@ -8,7 +8,6 @@ using Xunit;
 public static class OperationBehaviorTest
 {
     [Fact]
-    [ActiveIssue(315)]
     public static void IOperationBehavior_Methods_AreCalled()
     {
         DuplexClientBase<ICustomOperationBehaviorDuplexService> duplexService = null;


StephenBonikowsky:
This is not intended for a Merge, I'm putting up what I got done as of today because I will be gone next week, I would like to complete it but I don't want fixing this to be blocked on me.

I left the original code commented out so it is easier to see what I'm changing.

With the fix as it currently is I am getting three other Scenario test failures but I haven't been able to investigate yet.
System.Private.ServiceModel\tests\Scenarios\Security\TransportSecurity\Https\ClientCredentialTypeTests.cs
- Https_ClientCredentialTypeTests.BasicAuthentication_RoundTrips_Echo
- Https_ClientCredentialTypeTests.BasicAuthenticationInvalidPwd_throw_MessageSecurityException

and 

System.Private.ServiceModel\tests\Scenarios\Client\ExpectedExceptions\ExpectedExceptionTests.cs
- ExpectedExceptionTests.UnknownUrl_Throws_EndpointNotFoundException

src/System.Private.ServiceModel/src/System/ServiceModel/Description/CustomAttributeProvider.cs
```diff
@@ -77,13 +77,17 @@ public object[] GetCustomAttributes(Type attributeType, bool inherit)
             switch (this.ProviderType)
             {
                 case AttributeProviderType.Type:
-                    return this.Type.GetTypeInfo().GetCustomAttributes(attributeType, inherit).ToArray();
+                    //return this.Type.GetTypeInfo().GetCustomAttributes(attributeType, inherit).ToArray();
+                    return this.Type.GetTypeInfo().GetCustomAttributes().ToArray();
```
roncain: Thanks for going after the root cause @StephenBonikowsky .  This PR will need a couple tweaks before merging.

We should not modify CustomAttributeProvider like this because that changes its general behavior for our specific failure case.  I believe there is an alternate form of GetCustomAttributes which does not take an 'attributeType' parameter, and we should be calling that one instead.

src/System.Private.ServiceModel/src/System/ServiceModel/Description/ServiceReflector.cs
```diff
@@ -451,14 +451,25 @@ static internal object[] GetCustomAttributes(CustomAttributeProvider attrProvide
         {
             try
             {
-                if (typeof(Attribute).IsAssignableFrom(attrType))
+                List<object> attrTypeAttributes = new List<object>();
```
roncain: I recommend this code to make a new List and filter the attributes should really happen only in the case attributeType is not assignable to Attribute.  In other words, uncomment the original code below, but move this List-generating code into the "else" part of the code where an empty array was being returned.  And rather than calling the modified version of GetCustomAttributes that takes "attrType", it should call the version of GetCustomAttributes that does not take an attrType.

The original "if" block commented out below should be uncommented exactly as it was.  We want the #if part of the code to be unchanged, and only modify the #else part.

roncain:
LGTM.  IT is probably worth a buddy-build to test behavior in NET Native, but I recommend you go ahead and merge now.  If it fails in NET Native we can open a new issue.

Issue #313, opened 2015-09-02
Solve poor performance of WcfEventSource in CHK builds


roncain:
In CHK builds the tests run on TFS machines cause TestHost.exe to take 100% cpu for several minutes. See issue https://github.com/dotnet/coreclr/issues/1461 for a discussion and analysis.  Conclusion appears to be that DebugCheckEvent is doing excessive work due to the relatively large (62k) array it has constructed.

Renumbering the event ID's solves the issue but breaks existing tooling.

The purpose of this issue is to understand the root problem and find a solution.  This may involve a fix in this repo or as a PR to the CoreCLR repo.

zhenlan:
@roncain has verified and this is not a problem after @dmetzgar's fix is merged.

PR #329, opened 2015-09-08
Fix to the WCF event source that will alleviate the problems with ToF CHK tests

dmetzgar:
This fixed the issue in my testing and doesn't require changes to EventSource or renumbering the event ids.

### DIFF
--- a/src/System.Private.ServiceModel/src/Internals/WcfEventSource.cs
+++ b/src/System.Private.ServiceModel/src/Internals/WcfEventSource.cs
@@ -64,7 +64,7 @@ namespace System.Runtime
             return base.IsEnabled(EventLevel.Informational, Keywords.Infrastructure, EventChannel.Debug);
         }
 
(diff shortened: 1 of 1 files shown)


dmetzgar:
Fixes #313

roncain:
Questions only @dmetzgar -- what is the affect of not setting this property in the attribute?  Does it affect how the event appears to tooling?  It passes innerloop and outerloop tests on my machine, but I'd like to do a full CHK build in TFS to see how it behaves in ToF, but that will take a couple hours.  Were you able to try this fix in a CHK build against ToF already?

dmetzgar:
@roncain I used my repro of the issue here: https://github.com/dmetzgar/eventsourcetest
The original manifest takes **2.88 days** to complete. With this change it takes **0.02 secs**. Opcode.Info is the default opcode for an event. But explicitly setting it appears to cause problems with the current EventSource code. That code still needs to be fixed. At least with this PR we do not have to revisit this later from WCF side.

dmetzgar:

roncain:
LGTM.  I'll confirm with a CHK ToF run tomorrow, but recommend you go ahead and merge now.

PR #330, opened 2015-09-08
Adding back System.Net.Primitives dependency

mconnew:
This broke the internal .Net native build so adding the dependency back.


roncain:
Question only: did you try building for NET Native in TFS with this change?  I'm trying now but it looks like I will have to scorch to get the right build tools restored.

roncain:
LGTM.  I was able to buddy build and run OuterLoop, and I was able to copy into TFS and rebuild for CoreCLR and NET Native successfully.  @mconnew I recommend you merge asap to unbreak TFS build.

iamjasonp:
Thanks for fixing this, @mconnew

PR #337, opened 2015-09-15
Set BridgeResourceFolder at Bridge startup

### LINKED ISSUE #334: Refactor Bridge so BridgeResourceFolder is not provided by BridgeClient
Today the BridgeResourceFolder is sent to the Bridge by the BridgeClient, and this is how the Bridge accesses its IResources.  But this solution cannot work for x-plat because the Bridge resources cannot be built by Linux or OSX.  Moreover, it means we would need a general solution to have Linux and Windows machines share a folder, just for this scenario. The solution is to let Bridge.exe default to a BridgeResourceFolder it has already built and to remove the code from BridgeClient to specify it.  Therefore, the BridgeResourceFolder is declared wherever the Bridge is started, not where it is used by the client.

It also means the DELETE verb to config should not alter the BridgeResourceFolder, but instead should simply release all the resources and AppDomains

roncain:
Prior to this change, the BridgeResourceFolder was specified by
BridgeClient when tests ran.  However, this was not a valid
strategy for x-platform use, otherwise both client and Bridge
would need to share a x-OS folder.

With this change, Bridge.exe is required to have a BridgeResourceFolder
when it is asked to start, and the BridgeClient no longer alters that.

Also adds a 'reset' option to Bridge.exe so that when tests complete
the Bridge can be asked to release all remote resources.  They will
be reacquired the next time Bridge requests are made.

Also move the Http DELETE handling to the ResourceController where it
really belonged.  An Http DELETE to the /resource endpoint releases all
resources.

Fixes #334

### DIFF
--- a/src/System.Private.ServiceModel/tools/setupfiles/CleanupWCFTestService.cmd
+++ b/src/System.Private.ServiceModel/tools/setupfiles/CleanupWCFTestService.cmd
@@ -5,9 +5,13 @@ echo BridgeKeepRunning=%BridgeKeepRunning%
 if '%BridgeKeepRunning%' neq 'true' (
     echo Stopping the Bridge...
     pushd %~dp0..\..\..\..\bin\wcf\tools\Bridge
-    call Bridge.exe -stopIfLocal %*
+    call Bridge.exe -stopIfLocal -reset %*
     popd
 ) else (
+    echo Releasing Bridge resources but leaving it running...
+    pushd %~dp0..\..\..\..\bin\wcf\tools\Bridge
+    call Bridge.exe -reset %*
+    popd
     echo The Bridge was left running because BridgeKeepRunning is true
 )
 
--- a/src/System.Private.ServiceModel/tools/setupfiles/CleanupWCFTestService.cmd
+++ b/src/System.Private.ServiceModel/tools/setupfiles/CleanupWCFTestService.cmd
@@ -3,13 +3,15 @@ setlocal
 
 echo BridgeKeepRunning=%BridgeKeepRunning%
 if '%BridgeKeepRunning%' neq 'true' (
-    echo Stopping the Bridge...
+    echo Releasing Bridge resources and stopping it if running locally.
     pushd %~dp0..\..\..\..\bin\wcf\tools\Bridge
+    echo Invoking Bridge.exe -stopIfLocal -reset %* ...
     call Bridge.exe -stopIfLocal -reset %*
     popd
 ) else (
-    echo Releasing Bridge resources but leaving it running...
+    echo Releasing Bridge resources but leaving it running.
     pushd %~dp0..\..\..\..\bin\wcf\tools\Bridge
+    echo Invoking Bridge.exe -reset %* ...
     call Bridge.exe -reset %*
     popd
     echo The Bridge was left running because BridgeKeepRunning is true
--- a/src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
+++ b/src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
+using System.Text;
 using WcfTestBridgeCommon;
 
 namespace Bridge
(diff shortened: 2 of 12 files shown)


src/System.Private.ServiceModel/tools/setupfiles/CleanupWCFTestService.cmd
```diff
@@ -3,11 +3,17 @@ setlocal
...
+    echo Invoking Bridge.exe -stopIfLocal -reset %* ...
+    call Bridge.exe -stopIfLocal -reset %*
     popd
 ) else (
+    echo Releasing Bridge resources but leaving it running.
+    pushd %~dp0..\..\..\..\bin\wcf\tools\Bridge
+    echo Invoking Bridge.exe -reset %* ...
+    call Bridge.exe -reset %*
+    popd
     echo The Bridge was left running because BridgeKeepRunning is true
```
iamjasonp: IIRC, CleanupWcfTestService.cmd is called immediately after a run happens. Do we necessarily want to get rid of all resources that the bridge is holding? 

My flow of work is typically the following: 
- run startBridge.cmd
- run build.cmd to look at what's wrong
- make changes
- rebuild. 

Sometimes, in the "rebuild" step I run build.cmd, other times I build the project and test directly. 
My suspicion is that if someone has left the bridge running, we should probably leave any resources opened alone, and rely on the developer to either `exit` the Bridge or call `reset` manually

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -174,6 +180,68 @@ private static void StopBridge(CommandLineArguments commandLineArgs)
...
+                    {
+                        problem = String.Format("{0}Bridge returned unexpected status code='{1}', reason='{2}'",
+                                                    Environment.NewLine, response.StatusCode, response.ReasonPhrase);
+                        if (response.Content != null)
+                        {
+                            string contentAsString = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+                            if (contentAsString.Length > 1000)
+                            {
+                                contentAsString = contentAsString.Substring(0, 999) + "...";
+                            }
```
iamjasonp: Wondering if it's useful to have the Bridge also log things to a file concurrently with spewing debug info to the log.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -195,6 +263,22 @@ private static void StartBridge(CommandLineArguments commandLineArgs)
                 Environment.Exit(1);
             }
 
+            string resourceFolder = commandLineArgs.BridgeConfiguration.BridgeResourceFolder;
+            if (String.IsNullOrWhiteSpace(resourceFolder))
+            {
+                Console.WriteLine("Starting the Bridge requires the BridgeResourceFolder to be specified.");
+                Console.WriteLine("Use either -BridgeResourceFolder:folderName or set it as an environment variable.");
+                Environment.Exit(1);
```
iamjasonp: Have we defined what our exit codes are somewhere? Maybe it'd be good to start sticking them into an enum if we anticipate more than just 0 and 1

iamjasonp:
I don't see any issues with this at a glance, so LGTM. 
Might be worth merging so that we can see the effects of it running live. When merged, please also keep an eye on the [OuterLoop] for unexpected failures.

src/System.Private.ServiceModel/tools/setupfiles/CleanupWCFTestService.cmd
roncain: CleanupWcfTestService.cmd is the script called to clean up all resources -- that's its purpose.  In this case, cleanup means remove all certs and firewall rules created, shutdown the AppDomains, and continue running.  The Bridge will reacquire resources the next time it receives requests.

I think it is fairly critical that after a run all DLL's loaded from the BridgeResourceFolder be releaesd, otherwise the build will fail because it cannot write the new DLL's into the BridgeResourceFolder.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
roncain: Yes, I think we want the Bridge to capture both its own logs as well as access the WCF service's traces.  But that's another feature for another day.  I'll open an issue to track it.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
roncain: Good idea.  How about doing this as soon as we return anything other than 0 or 1?  I don't think it needs a separate issue for tracking but will do it as soon as we need another exit code.

PR #342, opened 2015-09-18
Use x509Certificates for Net Native.

### LINKED ISSUE #306: Use x509Certificates for both CoreCLR and NET Native when available
The current security work underway is adding a dependency on x509Certificates for only the CoreCLR version.  This current issue is to track the deliver of NET Native x509Certificates and to update the code where it is currently #if FEATURE_CORECLR.

shmao:
The changes are to use x509Certificates for Net Native.

Fixes #306.

### DIFF
--- a/src/System.Private.ServiceModel/src/project.lock.json
+++ b/src/System.Private.ServiceModel/src/project.lock.json
@@ -1,5 +1,5 @@
 {
-  "locked": true,
+  "locked": false,
   "version": -9996,
   "targets": {
     "DNXCore,Version=v5.0": {
@@ -333,9 +333,9 @@
           "lib/DNXCore50/System.Net.Http.dll": {}
         }
       },
-      "System.Net.NameResolution/4.0.0-beta-23308": {
+      "System.Net.NameResolution/4.0.0-beta-23316": {
         "dependencies": {
-          "System.Private.Networking": "4.0.1-beta-23308"
+          "System.Private.Networking": "4.0.1-beta-23316"
         },
         "compile": {
           "ref/dotnet/System.Net.NameResolution.dll": {}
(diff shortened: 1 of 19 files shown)


iamjasonp:
#306 implies that we are trying to convert from X509 to RT certificates - does this change already include that change, or is that a further work item for #306? If there's further work, please re-open #306 once you've committed. 

LGTM, pending merge of the project.json files. 

Additionally, please make sure that we've done a run through with the .NET Native test suite and it doesn't break :smile:

src/System.Private.ServiceModel/src/project.lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "locked": true,
+  "locked": false,
   "version": -9996,
```
roncain: I believe the value of "locked" must remain set to true or it will break TFS.

src/System.Private.ServiceModel/src/project.lock.json
```diff
@@ -677,50 +677,51 @@
...
-          "System.IO": "4.0.10"
+          "System.Text.Encoding": "4.0.0"
         },
         "compile": {
           "ref/dotnet/System.Security.Cryptography.Algorithms.dll": {}
         },
         "runtime": {
-          "lib/DNXCore50/System.Security.Cryptography.Algorithms.dll": {}
+          "lib/dotnet/System.Security.Cryptography.Algorithms.dll": {}
         }
```
roncain: The change to "dotnet" is interesting.  Did that happen automatically when updating the project.lock.json files?

iamjasonp: This happens as a result of moving the project.json dependencies out of the `dnxcore50` to the general dependencies section.

iamjasonp:
#347 replaces this PR due to the merge conflict and @shmao is on vacation at the moment. Thanks!

Issue #333, opened 2015-09-14
Negative encoding tests usingTestData.InvalidEncodings fail on Linux


roncain:
There are 2 unit tests that use TestData.InvalidEncodings as a negative test to ensure ArgumentException is thrown for encodings that are not supported.  On Linux, no exception is thrown.  Possible cause might be a difference in which encodings are supported on Linux.

The tests are:
  TextMessageEncodingBindingElementTest. WriteEncoding_Property_Set_Throws_For_Invalid_Encodings

BasicHttpBindingTests.TextEncoding_Property_Set_Invalid_Value_Throws

roncain:
This behavior is due to CoreFx issue https://github.com/dotnet/corefx/issues/2774.
When we compile a list of known unsupported Encodings, Encoding.GetEncoding(name) always returns the utf-8 encoding.  Therefore, the negative tests that believe they are testing unsupported Encodings are in fact testing utf-8, which is supported.

roncain:
This was fixed as described in https://github.com/dotnet/corefx/issues/2774.  Reactivating these unit tests and will verify they pass in Linux before merge.

roncain:
The issue was fixed in CoreFx.  The deactivated tests were reactivated with PR #421, and they now pass on Linux.

PR #348, opened 2015-09-21
Disable failing negative Encoding tests on Linux

roncain:
Two negative tests using known-unsupported Encodings were
failing due to issue https://github.com/dotnet/corefx/issues/2774
which returned UTF-8 for every unrecognized encoding.

Adds [ActiveIssue] to these tests using issue #333 which tracks
https://github.com/dotnet/corefx/issues/2774.  The [ActiveIssue]
disables these tests only on Linux.  They still execute on Windows.

Also modifies run-test.sh to exclude the TestRuntime folder from
CoreFx binaries when preparing the tests. The binaries in the WCF
test folders already contain the correct dependent binaries, and
allowing the TestRuntime folder to overwrite them gives out-of-date
or platform-incorrect binaries.

With these 2 changes, all unit tests not marked with [ActiveIssue]
pass on Linux.

### DIFF
--- a/run-test.sh
+++ b/run-test.sh
@@ -145,7 +145,7 @@ create_test_overlay()
 	echo "Corefx binaries not found at $CoreFxBins"
 	exit 1
   fi
-  find $CoreFxBins -name '*.dll' -and -not -name "*Test*" -exec cp '{}' "$OverlayDir" ";"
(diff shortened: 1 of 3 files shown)


iamjasonp:
LGTM

PR #350, opened 2015-09-22
Update README.md badges

iamjasonp:
* Switch to shields.io so we can customize labels on the build shields
* Change the code coverage shield to display data 
* Added outerloop test shield


roncain:
LGTM

PR #354, opened 2015-09-23
Enable IResources to receive and return name/value pairs

### LINKED ISSUE #353: Make the Bridge copy its name/value pairs into ResourceRequestContext
We need the resource GET and PUT handlers to be able to access the name/value pairs passed to the Bridge GET/PUT.

Update ResourceRequestContext to serialize these name/value pairs across AppDomain boundaries, and provide friendly wrappers so the resources can consume them

roncain:
Prior to these changes, invoking the IResource interface
on the Bridge did not pass any part of the request content
to the IResource.  After this change, the ResourceRequestContext
contains name/value pairs deserialized from the request content.

This change also solidifies the IResource contract so that the
Get and Put methods return a ResourceResponse rather than object.
This permits resources to return name/value pairs as well as to
receive them.

The Bridge Client was updated to expect a response containing
name/value pairs as json instead of the regex match.

The ResourceController was hardened to handle missing or
incomplete request content and to trace the incoming and
outgoing name/value pairs.

Fixes #353

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
@@ -6,17 +6,18 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
 using System.Text;
-using System.Text.RegularExpressions;
 
 namespace Infrastructure.Common
 {
     public static class BridgeClient 
     {
+        private const string EndpointResourceResponseUriKeyName = "uri";
+        private const string EndpointResourceRequestNameKeyName = "name";
+
         private static object s_thisLock = new object();
         private static Dictionary<string, string> _Resources = new Dictionary<string, string>();
         private static BridgeState s_BridgeStatus = BridgeState.NotStarted;
         private static string s_BridgeFaultReason = String.Empty;
-        private static readonly Regex regexResource = new Regex(@"details\s*:\s*""([^""]+)""", RegexOptions.IgnoreCase);
 
         private static string BridgeBaseAddress
         {
@@ -205,11 +206,14 @@ namespace Infrastructure.Common
                     EnsureSuccessfulResponse(response);
 
                     var responseContent = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
-                    var match = regexResource.Match(responseContent);
-                    if (!match.Success || match.Groups.Count != 2)
+                    Dictionary<string, string> properties = JsonSerializer.DeserializeDictionary(responseContent);
+                    string uri = null;
+                    if (!properties.TryGetValue(EndpointResourceResponseUriKeyName, out uri) || String.IsNullOrWhiteSpace(uri))
+                    {
                         throw new Exception("Invalid response from bridge: " + responseContent);
+                    }
 
-                    return match.Groups[1].Value;
+                    return uri;
                 }
                 catch (Exception exc)
                 {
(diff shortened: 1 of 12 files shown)


src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
```diff
@@ -6,17 +6,18 @@
...
     public static class BridgeClient 
     {
+        private const string EndpointResourceResponseUriKeyName = "uri";
+        private const string EndpointResourceRequestNameKeyName = "name";
+
         private static object s_thisLock = new object();
         private static Dictionary<string, string> _Resources = new Dictionary<string, string>();
         private static BridgeState s_BridgeStatus = BridgeState.NotStarted;
         private static string s_BridgeFaultReason = String.Empty;
-        private static readonly Regex regexResource = new Regex(@"details\s*:\s*""([^""]+)""", RegexOptions.IgnoreCase);
```
dmetzgar: Happy to see this get replaced with JSON.

dmetzgar:
LGTM

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ResourceController.cs
```diff
@@ -67,39 +82,62 @@ public HttpResponseMessage Put(HttpRequestMessage request)
...
+                Trace.WriteLine(String.Format("{0:T} - {1}",
+                                              DateTime.Now,
+                                              badRequestMessage),
                                 this.GetType().Name);
-                return Request.CreateErrorResponse(HttpStatusCode.BadRequest, "name not specified.");
+                return request.CreateResponse(HttpStatusCode.BadRequest, badRequestMessage);
             }
 
-            try
+            string nameValuePairs = request.Content.ReadAsStringAsync().GetAwaiter().GetResult();
```
iamjasonp: Do GETs have bodies? 

I don't know that the HTTP spec forbids it, but I don't think I see it in common use (for example, if you use Fiddler, you'll see that you can't actually send anything in the body). Typically for a GET, all the name/value pairs are specified on the URI itself

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/ResourceRequestContext.cs
```diff
@@ -11,7 +11,15 @@ namespace WcfTestBridgeCommon
     [Serializable]
     public class ResourceRequestContext
     {
+        //public ResourceRequestContext(Dictionary<string, string> properties)
+        //{
+        //    if (properties != null)
+        //    {
+        //        PropertyKeys = properties.Keys
+        //    }
+        //}
```
iamjasonp: NIT: remove if unused

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ResourceController.cs
iamjasonp: https://groups.yahoo.com/neo/groups/rest-discuss/conversations/topics/9962

> ... Server semantics for GET, however, are restricted such that a body, if any, has no semantic meaning to the request. The requirements on parsing are separate from the requirements on method semantics.
> 
> So, yes, you can send a body with GET, and no, it is never useful to do so....

iamjasonp:
I believe that the use of GET with a message body isn't correct - can we double check on this before potentially introducing a use of the GET verb that breaks convention? 

No further issues on the rest of the code.

dmetzgar:
Regarding a GET with a message body. You're right that we shouldn't be doing that. The alternatives would be to change it to a POST or stick stuff in the headers.

iamjasonp:
In theory, GETs should be fairly simple - the hard work should have been done by the PUT or POST, and all we're doing is GETting the resource using a known ID. Probably overkill to define a POST for this, IMO, at least for now. 

We can probably just use something like [ParseQueryString](https://msdn.microsoft.com/en-us/library/ms150046(v=vs.110).aspx) to handle the incoming query string and present the name-value pairs?

roncain:
WebAPI already supplies query parameter parsing, so I now use that.  GET allows name/value pairs only in query parameters, and PUT allows them in either the content or query parameters.

roncain:

iamjasonp:
LGTM. 
Probably would be good to squash 67d2843 and 97a4acd prior to merge.

PR #362, opened 2015-09-25
Adding ActiveIssue for SendTimeout negative test.

### LINKED ISSUE #273: Investigate and enhance SendTimeout_For_Long_Running_Operation_Throws_TimeoutException
This scenario test recently failed with this exception, indicating possible product failure or incorrect test assumptions.

  Running Test: ExpectedExceptionTests.SendTimeout_For_Long_Running_Operation_Throws_TimeoutException
  Assert.InRange:Expect value between 5000 and 10000 inclusive. Actual: 4994
  Xunit.AssertTestException: Assert.InRange: Expect value between 5000 and 10000 inclusive. Actual: 4994
     at Xunit.Assert.HandleFail(String assertionName, String message) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 592
     at Xunit.Assert.InRange(Int64 actual, Int64 low, Int64 high) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 576
     at ExpectedExceptionTests.SendTimeout_For_Long_Running_Operation_Throws_TimeoutException() in e:\oss\wcf\src\System.Private.ServiceModel\tests\Scenarios\Client\ExpectedExceptions\ExpectedExceptionTests.cs:line 165
     at EntryPointMain.b__3() in e:\ProjectK\src\QA\ToF\PN\x86\dbg\IL\FX\Conformance\System.Private.ServiceModel\4.0.0.0\WcfOpen\Scenarios\Client\ExpectedExceptions\Client.ExpectedExceptions.Tests.main.cs:line 16
     at System.Action.InvokeOpenStaticThunk()
     at System.Action.Invoke()
     at CoreFXTestLibrary.Internal.Runner.RunTestMethod$catch$0() in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Internal\Runner.cs:line 170
  ---- Test FAILED ---------------

I recommend 3 things:
1. Investigate how the test might timeout before its SendTimeout
2. Hard-coded expectations of test execution time usually make tests that fail sporadically.  Find a better way to ensure the timeout exception happened due to the timeout mechanisms.  Don't assume a busy lab machine will behave the same as your dev machine.
3. Remove from the server code any hard coded time delay.  That information should flow from the client test to the server.  There should never be magic constants or behavior on the server that the test somehow knows about.

StephenBonikowsky:
* Test case name: SendTimeout_For_Long_Running_Operation_Throws_TimeoutException
* Updated it to provide the service operation with the timeout value to use on the server.
* Checking the elapsed time against an expected range still seems like the right thing to do.
* Fixes #273

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -139,7 +139,7 @@ public static class ExpectedExceptionTests
     // The client should throw a TimeoutException
     [Fact]
     [OuterLoop]
-    [ActiveIssue(3464)]// Issue is in the corefx repo
+    [ActiveIssue(273)]
     public static void SendTimeout_For_Long_Running_Operation_Throws_TimeoutException()
     {
         TimeSpan serviceOperationTimeout = TimeSpan.FromMilliseconds(10000);
@@ -164,7 +164,7 @@ public static class ExpectedExceptionTests
 
         // want to assert that this completed in > 5 s as an upper bound since the SendTimeout is 5 sec
         // (usual case is around 5001-5005 ms) 
-        Assert.InRange<long>(watch.ElapsedMilliseconds, 5000, 10000);
+        Assert.InRange<long>(watch.ElapsedMilliseconds, 4985, 6000);
     }
 
     // SendTimeout is set to 0, this should trigger a TimeoutException before even attempting to call the service.
@@ -172,7 +172,7 @@ public static class ExpectedExceptionTests
     [OuterLoop]
     public static void SendTimeout_Zero_Throws_TimeoutException_Immediately()
     {
-        TimeSpan serviceOperationTimeout = TimeSpan.FromMilliseconds(10000);
+        TimeSpan serviceOperationTimeout = TimeSpan.FromMilliseconds(5000);
         BasicHttpBinding binding = new BasicHttpBinding();
         binding.SendTimeout = TimeSpan.FromMilliseconds(0);
         ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -139,8 +139,10 @@ public static class ExpectedExceptionTests
     // The client should throw a TimeoutException
     [Fact]
     [OuterLoop]
+    [ActiveIssue(3464)]// Issue is in the corefx repo
     public static void SendTimeout_For_Long_Running_Operation_Throws_TimeoutException()
     {
+        TimeSpan serviceOperationTimeout = TimeSpan.FromMilliseconds(10000);
         BasicHttpBinding binding = new BasicHttpBinding();
         binding.SendTimeout = TimeSpan.FromMilliseconds(5000);
         ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
@@ -152,7 +154,7 @@ public static class ExpectedExceptionTests
             {
                 IWcfService proxy = factory.CreateChannel();
                 watch.Start();
-                proxy.EchoWithTimeout("Hello");
+                proxy.EchoWithTimeout("Hello", serviceOperationTimeout);
             });
         }
         finally
@@ -170,6 +172,7 @@ public static class ExpectedExceptionTests
     [OuterLoop]
     public static void SendTimeout_Zero_Throws_TimeoutException_Immediately()
     {
+        TimeSpan serviceOperationTimeout = TimeSpan.FromMilliseconds(10000);
         BasicHttpBinding binding = new BasicHttpBinding();
         binding.SendTimeout = TimeSpan.FromMilliseconds(0);
         ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
@@ -181,7 +184,7 @@ public static class ExpectedExceptionTests
             {
                 IWcfService proxy = factory.CreateChannel();
                 watch.Start();
-                proxy.EchoWithTimeout("Hello");
+                proxy.EchoWithTimeout("Hello", serviceOperationTimeout);
             });
         }
         finally
--- a/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/IWcfService.cs
+++ b/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/IWcfService.cs
@@ -13,7 +13,7 @@ namespace WcfService
     internal interface IWcfService
     {
         [OperationContract]
-        String EchoWithTimeout(String message);
+        String EchoWithTimeout(String message, TimeSpan timeout);
 
         [OperationContract]
         Message MessageRequestReply(Message message);
--- a/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/IWcfService.cs
+++ b/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/IWcfService.cs
@@ -13,7 +13,7 @@ namespace WcfService
     internal interface IWcfService
     {
         [OperationContract]
-        String EchoWithTimeout(String message, TimeSpan timeout);
+        String EchoWithTimeout(String message, TimeSpan serviceOperationTimeout);
 
         [OperationContract]
         Message MessageRequestReply(Message message);
--- a/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WcfService.cs
+++ b/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WcfService.cs
@@ -13,9 +13,9 @@ namespace WcfService
 {
     internal class WcfService : IWcfService
     {
-        public String EchoWithTimeout(String message, TimeSpan timeout)
+        public String EchoWithTimeout(String message, TimeSpan serviceOperationTimeout)
         {
-            System.Threading.Thread.Sleep(timeout);
+            System.Threading.Thread.Sleep(serviceOperationTimeout);
             return message;
         }
 
--- a/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WcfService.cs
+++ b/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WcfService.cs
@@ -13,9 +13,9 @@ namespace WcfService
 {
     internal class WcfService : IWcfService
     {
-        public String EchoWithTimeout(String message)
+        public String EchoWithTimeout(String message, TimeSpan timeout)
         {
-            System.Threading.Thread.Sleep(10000);
+            System.Threading.Thread.Sleep(timeout);
             return message;
         }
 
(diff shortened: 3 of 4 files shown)


StephenBonikowsky:
Even though the Issue is in corefx the ID of the issue doesn't seem to have any impact to the ActiveIssue attribute. The test still gets skipped.

@roncain I could not come up with any way that made sense to me to " Find a better way to ensure the timeout exception happened due to the timeout mechanisms." If we hadn't had this range check we would have missed this bug. WCF just passes this value on down to the HttpClient layer, so we are not actually doing anything with this timeout, this test mostly just tests what HttpClient does with the value and that it acts as we might expect.

Different machine performance specs should have almost no effect on this particular test especially as it is not dependent on any network traffic. I would argue that narrowing the range would be of value except that from WCF perspective this is mainly a sanity check that our timeout actually gets used by HttpClient. Deeper testing should be done by corefx.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -139,8 +139,10 @@ public static void NonExistentAction_Throws_ActionNotSupportedException()
     // The client should throw a TimeoutException
     [Fact]
     [OuterLoop]
+    [ActiveIssue(3464)]// Issue is in the corefx repo
     public static void SendTimeout_For_Long_Running_Operation_Throws_TimeoutException()
```
roncain: I think we can make this issue #273 because it is in our repo.  #273 should probably be updated to refer to the CoreFx issue.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -170,6 +172,7 @@ public static void SendTimeout_For_Long_Running_Operation_Throws_TimeoutExceptio
     [OuterLoop]
     public static void SendTimeout_Zero_Throws_TimeoutException_Immediately()
     {
+        TimeSpan serviceOperationTimeout = TimeSpan.FromMilliseconds(10000);
         BasicHttpBinding binding = new BasicHttpBinding();
```
roncain: If we expect this to timeout immediately, I recommend we reduce how long we ask the server to stall (5 seconds?) just so the server side isn't tied up so long.  Your call.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
StephenBonikowsky: I wanted to close #273 since there is nothing more for us to do and I didn't like the idea of having an issue hanging around just to track another issue in someone else's repo?

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -152,7 +154,7 @@ public static void SendTimeout_For_Long_Running_Operation_Throws_TimeoutExceptio
             {
                 IWcfService proxy = factory.CreateChannel();
                 watch.Start();
-                proxy.EchoWithTimeout("Hello");
+                proxy.EchoWithTimeout("Hello", serviceOperationTimeout);
             });
         }
         finally
```
roncain: BTW, this test fails sporadically on NET Native with a time **less** than 5 seconds (like 4992 ms).  I don't know whether we should consider that a product bug, but we could make the test more reliable by allowing some grace period (like the elapsed time should be in the range 4000 ms to 8000 ms).

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
StephenBonikowsky: We are expecting it to timeout after 5 seconds, it "should" just take a few more milliseconds to get back to test code. The high-end of the range we check for should be equal to or less than the timeout we use for the service operation. When the bug is fixed and the test working again the service won't remain tied-up any longer than the SendTimeout because all the connections get shut-down by HttpClient.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
StephenBonikowsky: That actually doesn't make any sense, if HttpClient is handling the timeout correctly this should be a mathematical impossibility, in which case this should be a bug and something we would want to fail for.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
roncain: Thanks for checking on the different approaches. Yeah, I can't think of a good way to double check that we timed out for the right reason.

I don't think we should close #273 till we've verified the fix, but I don't feel strongly about it.  I've used issues in our repo to track issues in other repo's.  Those links show up when they look at the bug, and seeing ours is Active is a good way to say we're waiting on it.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
SajayAntony: Could you reference the corefx bug in the timeout here?

StephenBonikowsky: You mean as a comment?

roncain:
@StephenBonikowsky -- even when we're ready to merge this, I recommend we wait till Monday to merge.  TFS is doing rolling builds in preparation for beta8, and we don't want to trigger an OTG build for this merge while that is happening.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
SajayAntony: @StephenBonikowsky Just open a issue on our side to cross reference a bug in corefx.having a comment in code

roncain: If we feel that timing that out at 4992 ms is actually an HttpClient bug in NET Native, I recommend we open another bug for it.  Possible or no, I see this failure on N about once a week.  I doubt we can make a reliable repro.  It is conceivable some rounding error in the way they setup their timeout that might give different behavior depending exactly when the timer starts.  I've never seen it more than about 10ms too short.

StephenBonikowsky:
I'll hold till next week.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
roncain: Sorry, I was talking about the test that sets a zero timeout and expects immediate timeout.  I think when everything works properly, the service operation will probably not even get called.  But if it did, I think it would stall the full 10 seconds before noticing the client connection was lost.  I just wanted to avoid sending the service operation to sleep for an unnecessarily long time.  Not a blocking issue for me, just a suggestion.

roncain:
LGTM for Monday.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
StephenBonikowsky: My bad, yes you are correct.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/IWcfService.cs
```diff
@@ -13,7 +13,7 @@ namespace WcfService
     internal interface IWcfService
     {
         [OperationContract]
-        String EchoWithTimeout(String message);
+        String EchoWithTimeout(String message, TimeSpan timeout);
```
iamjasonp: the `timeout` parameter should be more verbose about which timeout we're talking about - e.g., `serviceOperationTimeout`.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
mconnew: If we're worried about occupying a thread on the server side, then we should make the service code be task based and do an await Task.Delay(timeout). That way we don't occupy a thread while waiting.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
mconnew: The HttpClient doesn't get passed a timeout, it get's passed a CancellationToken in the SendAsync api call. A callback is then registered against the CT which gets executed when the timeout occurs. This won't be an issue in HttpClient, it might be an issue with CancellationTokenSource or Timer. I suspect they are doing some kind of timer coalescence where some other timer has fired and then they check if there's anything else due to fire soon. Considering the standard Timer resolution is 15ms, being 8ms early is within the range of Timer accuracy.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WcfService.cs
```diff
@@ -13,9 +13,9 @@ namespace WcfService
 {
     internal class WcfService : IWcfService
     {
-        public String EchoWithTimeout(String message)
+        public String EchoWithTimeout(String message, TimeSpan timeout)
```
mconnew: If there's a concern about having a thread blocking for the timeout amount of time, change the code (and interface) to this:
```csharp
        public async Task<String> EchoWithTimeout(String message, TimeSpan timeout)
        {
            await Task.Delay(timeout);
            return message;
        }

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: @mconnew - if this is the case, do you think that we should allow for `timeout - 15 ms` when we check for ranges?

iamjasonp:
LGTM

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
StephenBonikowsky: Synced up with Matt, short version @iamjasonp yes when checking ranges we need to accommodate for a +/- 15 ms. Apparently that is the interval used when checking the cpu clock to get elapsed time.
@roncain there isn't a bug in NET Native, different implementation there, the sporadic failure is just this clock validation thing.

iamjasonp: Is there a better way to check for this? I think both @roncain  and I hate that we are checking against a range of values (prone to issues like this), but I haven't been able to come up with a nicer way to detect that this timeout fired. Thoughts?

StephenBonikowsky: I couldn't come up with another way, neither could Ron. Also, if you hadn't put this range check in we would have missed this regression.

roncain:
@StephenBonikowsky we should merge this asap so we can stop failing our OuterLoops.  Beta8 has been branched, so there is no danger to the rolling builds any longer.

iamjasonp:

PR #377, opened 2015-09-29
Allow addressable resource URIs

SajayAntony:
Part of a change which will enable listing all available resource of the bridge. 

```
$ curl -X  PUT http://localhost:44283/resource/WcfService.TestResources.BasicHttpResource -d .
{
   uri : "http://localhost:8081/BridgeAppDomain0/Basic"
}
```

PR #380, opened 2015-09-30
Fix read bug in content properties in BridgeClient

iamjasonp:
Change a260b68 (PR #377) introduced a change to BridgeClient, which changed the way resources were addressed. 

This change allows us to put more content into the PUT request other than just putting to the resource only.

This needs to be introduced either prior to, or as part of the dynamic cert generation change

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
 using System.Text;
 
@@ -74,6 +75,7 @@ namespace Infrastructure.Common
                 }
             }
         }
+
         public static string GetResourceAddress(string resourceName)
         {
             EnsureBridgeIsRunning();
@@ -83,7 +85,7 @@ namespace Infrastructure.Common
             {
                 if (s_BridgeStatus == BridgeState.Started && !_Resources.TryGetValue(resourceName, out resourceAddress))
                 {
-                    resourceAddress = MakeResourcePutRequest(resourceName);
+                    resourceAddress = MakeEndpointResourcePutRequest(resourceName);
                     _Resources.Add(resourceName, resourceAddress);
                 }
             }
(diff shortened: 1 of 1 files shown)


iamjasonp:
Probably no need to merge this yet - there are more changes I can bundle into this one for the Bridge server side too.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
```diff
@@ -83,7 +85,7 @@ public static string GetResourceAddress(string resourceName)
             {
```
roncain: Nit: consider a lock around the code block to prevent concurrent requests to Bridge.  Not critical but would reduce Bridge noise.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
```diff
@@ -190,28 +192,64 @@ private static string CreateConfigRequestContentAsJson()
...
+        private static string MakeEndpointResourcePutRequest(string resourceName)
+        {
+            var responseContent = MakeResourcePutRequest(resourceName, null);
+
+            string uri = null;
+            if (!responseContent.TryGetValue(EndpointResourceResponseUriKeyName, out uri) || String.IsNullOrWhiteSpace(uri))
+            {
+                StringBuilder sb = new StringBuilder();
+                foreach (var pair in responseContent)
+                {
```
roncain: Nit: consider a helper that does this StringBuilder over a Dictionary<string, string>.  Not critical.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
```diff
@@ -222,30 +260,48 @@ private static string MakeResourcePutRequest(string resourceName)
...
+                throw new ArgumentException(
+                    string.Format("A GET request to the bridge included parameter 'name' = '{0}' in its requestParameters, but the 'resourceName' = '{1}' specified was different", requestParametersResourceName, resourceName),
+                    "requestParameters");
+            }
+
             using (HttpClient httpClient = new HttpClient())
             {
+                StringBuilder uriBuilder = new StringBuilder(BridgeResourceEndpointAddress + resourceName);
+                if (requestParameters != null)
+                {
```
roncain: Nit: consider a helper to compose query string parameters from name/value pairs.  Not critical.

roncain:
LGTM.  The few nits are not critical but might be useful if you touch this file again.

SajayAntony:
Sounds reasonable to bundle this into the certs PR. Basically have something end to end. This commit looks good.

iamjasonp:
You're right @SajayAntony, the only person that was likely regressed was me :smile: 

Let's hold off on this particular change as it stands. 

However, I may tease out the Bridge changes from the dynamic cert changes and submit the Bridge changes first so we don't block any further changes for the Bridge. Admittedly, this is mainly for selfish reasons - I simply don't want to do another rebase on this code and find my changes stomped over by the merge :cry:

iamjasonp:
Closing - this feature isn't needed yet until we put in the certificate generation logic, so we'll bake this a bit further an include it in that PR in one go

PR #382, opened 2015-10-01
Fix ws vs http uri schema for WebSockets and fix timeout bug

mconnew:


StephenBonikowsky:
Tests don't pass as is. Matt mentioned that he had to use a more recent build of corefx, he's investigating.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -296,11 +296,14 @@ protected virtual void ValidateCreateChannelParameters(EndpointAddress remoteAdd
...
-            //EndpointAddress httpRemoteAddress = remoteAddress != null && WebSocketHelper.IsWebSocketUri(remoteAddress.Uri) ?
-            //    new EndpointAddress(WebSocketHelper.NormalizeWsSchemeWithHttpScheme(remoteAddress.Uri), remoteAddress) :
-            //    remoteAddress;
+            if (typeof (TChannel) != typeof (IRequestChannel))
+            {
+                remoteAddress = remoteAddress != null && !WebSocketHelper.IsWebSocketUri(remoteAddress.Uri) ?
+                    new EndpointAddress(WebSocketHelper.NormalizeHttpSchemeWithWsScheme(remoteAddress.Uri), remoteAddress) :
+                    remoteAddress;
+                via = !WebSocketHelper.IsWebSocketUri(via) ? WebSocketHelper.NormalizeHttpSchemeWithWsScheme(via) : via;
+            }
```
roncain: As far as I can see, this code is ported verbatim from the desktop, correct?

mconnew: Not quite. On desktop we convert ws:// to http://. This is because on desktop we use HttpWebRequest to do the HTTP handshake and that requires an http:// scheme. ClientWebSocket requires a url using the scheme ws:// or wss:// so we actually need to do the conversion the other way here.

mconnew:
@StephenBonikowsky, tests should now pass as I've updated the build tools version.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
roncain: Does this difference have any affect on existing Windows Store apps (e.g. in the URL's they use)?  Or it is strictly an internal implementation difference (existing) customer apps would never see?

mconnew: We accept either scheme from customer code on all platforms. The difference is internally our dependencies only accept one scheme, and which one scheme that  is differs from desktop.
________________________________

roncain:
@mconnew -- have you tried building this against NET Native yet?  I'm getting setup to do that but could save time if you have.

roncain:
I get build errors building this in TFS.  I am investigating but recommend @mconnew also tries this.  My build failed using an incremental build.  I'll try a full scorch rebuild next.

StephenBonikowsky:
OSS OuterLoops look good.

roncain:
LGTM.  It built under CoreCLR and NET Native after a partial scorch/rebuild.  I'd say go ahead and merge but monitor the rolling TFS builds.

Issue #308, opened 2015-09-02
Add new functionality to WCF contracts - 2015.10 edition


iamjasonp:
As we start turning on new features, we will start encountering issues where we would like to add new contract. This issue will keep track of all the issues that require contract additions. 

Please open an issue and add to the list below.

New APIs to be added:
* X509 Certificates: expose client cert setters - #302
* ChannelFactory.Credentials.ServiceCertificate.Authentication - #243
* Add support to specify SSL/TLS version for NetTcp with certificate security - #39

hongdai:
* Service Certificate Authentication mode: client can configure Service Certificate Authentication mode( X509ServiceCertificateAuthentication) from ChannelFactory.Credentials.ServiceCertificate.Authentication - #243

zhenlan:
Add support to specify SSL/TLS version for NetTcp with certificate security - #39

iamjasonp:
All resolved now. 

Part3: #440
Part2: #436
Part1:#429 (#411)
Master PR: #385

Issue #302, opened 2015-09-01
Expose client certificate setters in contract


iamjasonp:
At the moment, it's not possible to set a client certificate on the connection; we should expose this (and any other necessary other API) to allow the setting of client certificates on the channel factory.

This should be an exercise done after X509Certificate support comes into UAP, so we save one round of work. 

APIs / classes (add to this as needed):
* System.ServiceModel.Description.ClientCredentials.ClientCertificate 
* X509CertificateInitiatorClientCredential

iamjasonp:
Resolved as part of #308

Issue #243, opened 2015-07-31
Need to support ChannelFactory.Credentials.ServiceCertificate


hongdai:
Without this feature, we will not be able to use test certificate for some of the security scenarios, such as  DnsEndpointIdentity NetTcpBinding server identify verification scenario.

hongdai:
There are valid custom scenarios to support ChannelFactory.Credentials.ServiceCertificate.Authentication.
* Client needs to be able to specify Service cert authentication mode.

iamjasonp:
Resolved as part of #308

Issue #39, opened 2015-05-27
Add support to specify SSL/TLS version for NetTcp with certificate security


mconnew:
On .Net 4.5.x and earlier, NetTcp with certificate security was only able to use SSL3.0 and TLS1.0 when upgrading the transport to use certificate security. Support was added to .Net 4.6 to allow specifying the version of SSL/TLS to use. The following properties need to be added (both of type System.Security.Authentication.SslProtocols):

System.ServiceModel.TcpTransportSecurity.SslProtocols
System.ServiceModel.Channels.SslStreamSecurityBindingElement.SslProtocols

This issue is blocked on Certificate support being added to NetTcp.

iamjasonp:
Resolved as part of #308

PR #385, opened 2015-10-05
Adding APIs to contracts for certificate validation

iamjasonp:
**Please do not merge this PR - it needs to be done in steps as mentioned below** 

Adding APIs to contracts per https://github.com/dotnet/wcf/issues/308

Also for:
https://github.com/dotnet/wcf/issues/39, https://github.com/dotnet/wcf/issues/302, https://github.com/dotnet/wcf/issues/243

These APIs were previously present on desktop and we want to turn these on to enable some certificate (non-)validation scenarios in .NET Core. 

Specifically, these APIs allow for 
* specification of the SslProtocols that are allowed by ssl stream security
* ditto for the NetTcp transport
* specification of the X509 client credential type
* support the specification of the client and server certificate

Note that this change will be split into the following steps over three evenings
1. Bump version of System.Private.ServiceModel
2. Change contract for System.ServiceModel.Primitives 
3. Change contract for System.ServiceModel.NetTcp
... and bump the project.lock.json files for each of the above changes too
Each step will mean we wait for the next packages to be published

cc: @roncain, @mconnew, @weshaggard, @ericstj, @terrajobst

### DIFF
--- a/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
+++ b/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
@@ -5,7 +5,7 @@
     <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
     <Platform Condition="'$(Platform)' == ''">AnyCPU</Platform>
     <AssemblyName>System.Private.ServiceModel</AssemblyName>
-    <AssemblyVersion>4.0.1.0</AssemblyVersion>
+    <AssemblyVersion>4.1.0.0</AssemblyVersion>
     <RootNamespace>System.ServiceModel</RootNamespace>
     <OutputType>Library</OutputType>
     <NoWarn>$(NoWarn);1634;1691;649</NoWarn>
--- a/src/System.ServiceModel.NetTcp/ref/System.ServiceModel.NetTcp.cs
+++ b/src/System.ServiceModel.NetTcp/ref/System.ServiceModel.NetTcp.cs
@@ -51,6 +51,8 @@ namespace System.ServiceModel
         public TcpTransportSecurity() { }
         [System.ComponentModel.DefaultValueAttribute((System.ServiceModel.TcpClientCredentialType)(1))]
         public System.ServiceModel.TcpClientCredentialType ClientCredentialType { get { return default(System.ServiceModel.TcpClientCredentialType); } set { } }
+        [System.ComponentModel.DefaultValueAttribute((System.Security.Authentication.SslProtocols)(4080))]
+        public System.Security.Authentication.SslProtocols SslProtocols { get { return default(System.Security.Authentication.SslProtocols); } set { } }
     }
 }
 namespace System.ServiceModel.Channels
@@ -70,6 +72,8 @@ namespace System.ServiceModel.Channels
     public partial class SslStreamSecurityBindingElement : System.ServiceModel.Channels.BindingElement
     {
         public SslStreamSecurityBindingElement() { }
+        [System.ComponentModel.DefaultValueAttribute((System.Security.Authentication.SslProtocols)(4080))]
+        public System.Security.Authentication.SslProtocols SslProtocols { get { return default(System.Security.Authentication.SslProtocols); } set { } }
         public override System.ServiceModel.Channels.IChannelFactory<TChannel> BuildChannelFactory<TChannel>(System.ServiceModel.Channels.BindingContext context) { return default(System.ServiceModel.Channels.IChannelFactory<TChannel>); }
         public override bool CanBuildChannelFactory<TChannel>(System.ServiceModel.Channels.BindingContext context) { return default(bool); }
         public override System.ServiceModel.Channels.BindingElement Clone() { return default(System.ServiceModel.Channels.BindingElement); }
--- a/src/System.ServiceModel.Primitives/ref/project.json
+++ b/src/System.ServiceModel.Primitives/ref/project.json
@@ -13,6 +13,8 @@
     "System.Reflection": "4.0.0",
     "System.Net.Primitives": "4.0.0",
     "System.Security.Principal": "4.0.0",
+    "System.Security.Cryptography.X509Certificates" : "4.0.0-beta-*",
+    "System.Collections" : "4.0.0", 
     "System.Runtime.Serialization.Primitives": "4.0.0"
   },
   "frameworks": {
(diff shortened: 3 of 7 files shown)


src/System.ServiceModel.NetTcp/ref/System.ServiceModel.NetTcp.cs
```diff
@@ -51,6 +51,8 @@ public sealed partial class TcpTransportSecurity
         public TcpTransportSecurity() { }
         [System.ComponentModel.DefaultValueAttribute((System.ServiceModel.TcpClientCredentialType)(1))]
         public System.ServiceModel.TcpClientCredentialType ClientCredentialType { get { return default(System.ServiceModel.TcpClientCredentialType); } set { } }
+        [System.ComponentModel.DefaultValueAttribute((System.Security.Authentication.SslProtocols)(4080))]
```
iamjasonp: Enum value is `SslProtocols.Ssl3 | SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12`
However, the tool that generates this file uses integer values

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
```diff
@@ -5,7 +5,7 @@
     <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
     <Platform Condition="'$(Platform)' == ''">AnyCPU</Platform>
     <AssemblyName>System.Private.ServiceModel</AssemblyName>
-    <AssemblyVersion>4.0.1.0</AssemblyVersion>
+    <AssemblyVersion>4.1.0.0</AssemblyVersion>
     <RootNamespace>System.ServiceModel</RootNamespace>
```
roncain: I recommend we get sign-off on just this one change and get it merged asap.  I know the 3 stages are clearly articulated, but don't see any advantage to holding up step 1 for either steps 2 or 3.  We can do buddy-builds and TFS builds to ensure a smooth transition.
So step 1 (this file only) LGTM

iamjasonp: Done - #386. (I've been learning to cherry pick changes from one local branch to another, so that's why this took a little long :smile:) 

This is the all-up issue (PR) that I want the API review to happen on. Once the API review is done for a particular set of changes, I'll split it out and have it in a separate PR. The reason for this is that each individual change doesn't make sense on its own, but makes sense when viewed as a whole.

roncain:
@iamjasonp it looks like the 4.1.0 package was finally published, so I think we can proceed to step #2 to add the APIs

iamjasonp:
Yep, I saw these published. I'm trying to turn on a Tcp x-machine test now :)

iamjasonp:
Final call for comments? If none, I will proceed to Step #2

src/System.ServiceModel.Primitives/ref/project.json
```diff
@@ -13,6 +13,8 @@
     "System.Reflection": "4.0.0",
     "System.Net.Primitives": "4.0.0",
     "System.Security.Principal": "4.0.0",
+    "System.Security.Cryptography.X509Certificates" : "4.0.0-beta-*",
```
mconnew: You are adding a -beta-* dependency which is an implementation version to a contract. Shouldn't this be 4.0.0? Otherwise we will need to bump our contract version when we move out of beta simply to update this reference to remove the -beta-*..

iamjasonp: Unfortunately, we don't have a 4.0.0 version of this contract published, per http://www.nuget.org/packages/System.Security.Cryptography.X509Certificates. 

So yes, we will probably need to bump the patch version (i think... we shouldn't need to bump the minor) of this contract. Any concerns with that? Otherwise we'd have to wait until the 4.0.0 stable release comes out, whenever that is, before a dependency can be taken.

zhenlan: This means we cannot have a stable release until X509Certificates has one. We need to change ourselves to be beta if not yet.

iamjasonp: I think the assembly and contract version is correct - we have to make sure the packages are being generated with the right version.

weshaggard: @zhenlan you are correct you will not be able to ship a stable version of this package until all your dependencies are stable. Currently this package is "beta" so no worries there but you should be sure to track the beta dependencies closely if you plan to ship a stable package for SM.

mconnew: @weshaggard how does a package get the -beta-* suffix? Is there a file we need to modify to mark our packages as beta? Or is it all automagically done?

weshaggard: The "-beta" suffix is added by default by our internal package builds. You actually have to do work to remove it, which is to add your version to the stable package list.

roncain:
@iamjasonp -- is there anything blocking taking the next step?  I believe the version bump to System.ServiceModel.Primitives must go in as a separate PR+merge so that the nightly builds create a package for it.  The project.json's cannot take a dependency on it until it has been published to myget.

A buddy build of this PR fails saying there is no S.SM.Primitives >= 4.1.0.

Given the RC1 deadline, I'd say getting in this version bump is the highest priority task we have to enable the new API's.

iamjasonp:
@roncain: Nothing blocking except an explicit LGTM from API Review. #411 is opened for step 2, and I think zhenlan already did an effective "LGTM" there.

roncain:
@iamjasonp -- if you can update this PR to progress to the next step, I can buddy build on TFS and run the package generation phase, just to verify it will produce the right results.  So I *think* the right next step is getting S.SM.Primitives up to 4.1.0 so NetTcp can take that dependency.  S.Private.SM is of course already showing up as 4.1.0-beta-23413. 5 days till RC1 code freeze limits our nightly builds, so each one night counts now.

iamjasonp:
@roncain, can you buddy and package gen #411? That PR is Step 2 of this work item

iamjasonp:
We are waiting for a package build at https://www.myget.org/gallery/dotnet-core for System.ServiceModel.Primitives 4.1.0.

iamjasonp:
Woohoo! We did it! Thanks all for your help. 

This change is resolved via the following pull requests:
* Part3: #440 
* Part2: #436 
* Part1:#429 (#411)

PR #390, opened 2015-10-06
Correcting the word spelling

Kagamine:


iamjasonp:
@mconnew, just confirming that we don't depend on the spelling of these constants anywhere in the event sources? 

LGTM. Thanks @Kagamine for catching this

roncain:
LGTM.  I don't see any use of this constant in our current sources.  I do see this same misspelling in Microsoft.Diagnostics.Tracing.TraceEvent, but that is not OSS. I guess the question is whether these const's are needed any longer.  For the record, "public const" is generally a bad idea because it gets baked in at compile time, which can lead to DLL's with different values if it changes over time. @mconnew -- thoughts on removing these or impact of renaming?

roncain:
Passed InnerLoops and OuterLoops.  I will merge it.  Thanks @Kagamine

Issue #365, opened 2015-09-28
Clean up certificates installed by BridgeClient 


iamjasonp:
For x-machine testing, BridgeClient will necessarily install certificates on the client machine for HTTPS/TCP+SSL testing. 

However, the cleanup logic currently does not exist in the BridgeClient. 

Investigate and implement a way to clean up certificates on a BridgeClient. For example, use an executable that runs as part of test cleanup

roncain:
Done in PR #499 and #501

Issue #335, opened 2015-09-15
The Bridge should release certificates based on common naming pattern


roncain:
Today the Bridge uninstalls certificates it knows it has installed.  But this solution is not good for the scenario where it crashes, and also does not work well x-machine.

Solution is to add code in both Bridge.exe and the Bridge ResourceController to uninstall all certificates based on a fixed naming pattern we use when we create them.

iamjasonp:
We currently generate with `CN = <subject> O = DO_NOT_TRUST, OU = Created by https://github.com/dotnet/wcf`, so it's probably all right to key off the OU. 

Otherwise, we could include something in the FriendlyName field.

zhenlan:
@iamjasonp please proceed with the plan you proposed.

zhenlan:
This is done.

PR #394, opened 2015-10-07
Implement dynamic X509 client cert generation for Bridge

### LINKED ISSUE #310: Investigate Bridge certificate and certificate revocation list issues with TCP
`Tcp_ClientCredentialTypeTests.SameBinding_SecurityModeTransport_ClientCredentialTypeCertificate_EchoString` fails due to errors relating to certificate revocation lists being inaccessible and chain trust not validating properly. 

Investigate: 
* chain trust on self-signed certs (fixed - #316)
* certificate revocation lists 
* possibly: DnsIdentity checks when cert doesn't contain the DNS or CN name (fixed - #316)

### LINKED ISSUE #320: Generated X509 Certificates should contain a Subject Alternative Name
We need to generate certificates that contain a **Subject Alternative Name (SAN)** field. This is to support testing X509Certificate claims. 

According to this: http://blogs.technet.com/b/uday/archive/2012/06/21/makecert-exe-san-and-wildcard-certificate.aspx and http://stackoverflow.com/questions/6383054/add-or-create-subject-alternative-name-field-to-self-signed-certificate-using 

`makecert.exe` is now deprecated and we can't expect a fix for this in the tool.

### LINKED ISSUE #324: Investigate dynamic generation of test X509 certificates in Bridge
Our current test X509 certificates are generated using `makecert.exe` and committed to the repo. This poses a few problems: 

1. Certs expire after a set period of time (currently, two months after generation of the certificate) 
2. Certs are tied to localhost only and we do not generate certs for the fully qualified name of a machine
3. Certificates are not generated with a Subject Alternative Name (#320) or Certificate Revocation List (#310) reference

We want to investigate a solution for this issue, and to also allow the Bridge to supply certificates to a caller (script, perhaps?) when asked for them. 

To investigate
---

Methods to generate cert: 
1. marshal to Windows Crypto APIs 
2. Bouncy Castle
3. OpenSSL

How to interrogate the Bridge for certs: 
> e.g., as soon as the Bridge knows the BridgeResourceFolder via the /config enpoint, it could just generate the certs, and the scripts could pull them back out.

The assumption is that there is no need for Bridge.exe to be cross-plat at the moment.

### LINKED ISSUE #336: Implement a certificate IResource for the Bridge
Today the Bridge supports the abstraction of IResources, but the only implementation is for WCF endpoints.

I propose we create a new IResource capable of creating and installing a certificate for the POST verb, and returning the certificate content via the GET response.  Then we should make the Endpoints.cs test helper that accesses WCF endpoint IResources that require certificates first acquire the necessary certificate IResource.

The BridgeClient needs to be modified to install these certificates after it has acquired them.

### LINKED ISSUE #373: Bridge should be able to return raw data on a GET request for Resource
Currently, the Bridge returns all data as a JSON string. In order to support certificate revocation lists being dynamically served by the Bridge, we need to be able to return the raw CRL to the caller, and (optionally) specify a content-type other than `application/json`.

iamjasonp:
~~only @iamjasonp should merge, as the thirdpartynotice legal notice needs to be reviewed by legal prior to merge~~ **LCA approval received**

The bridge currently relies on static certificates for a certificate authority
cert and localhost certificate signed by the authority.
As such, cross-machine HTTP tests are dependent on a the well-known authority
certificate for server cert validation, and cross-machine TCP tests will
not work as the self-signed certs are generated for the host "localhost".

This change allows the Bridge to accept requests to generate X509 Certificates
dynamically and also serves up certificate revocation lists.

Certificates are designed to be short-lived (24h by default) but this can
be changed in the testproperties.props file along with other Bridge
configuration.

This commit addresses the following:
* X509 certificates are now generated dynamically (Fixes #324)
* Certificates are now generated via an IResource on the Bridge (Resolves #336)
* Certs generated have a Subject Alternative Name (Fixes #320)
* Support for Certificate Revocation Lists (Resolves #310)

This also addresses the following issues:
* Bridge can return raw data on a GET request (Fixes #373)
* Lay the foundation to be able to release certificates based on a
  well-known name (#335, #365 not yet fixed entirely)

This set of changes also unblocks scenarios relying on chain validation
as chain validation will now succeed - this means that tests for (say)
TCP will now work as expected, while we are waiting for new APIs to be put
in place (such as #302)

Also merges changes from PR #377 for allowing addressable Resource URIs
Also addresses issues identified in PR #380 which was closed and incorporated here

e86285b, d4e0644, c5656f3 to be squashed together upon LGTMs


iamjasonp:
Also deals with #395 (issue was opened for tracking purposes, as it wasn't part of the original scope of work)

iamjasonp:

roncain:
@iamjasonp -- FWIW, I took this PR and ran the OuterLoops.  The first run saw all https fail, like the CI machine.  The 2nd run saw only one https test failure.  3rd run, all passed.  I presume this means the cleanup logic was somehow mopping up some stale certs or firewall rules from prior runs.  Re-running all OuterLoops using master saw all tests pass. Re-running this PR after master had run saw all pass.

Conclusion: some machines, including CI might have some stale state that take 1 or 2 OuterLoop runs to cleanup

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -0,0 +1,394 @@
...
+
+        private static string LocalFqdn
+        {
+            get
+            {
+                // Get an FQDN for the local machine and cache it
+                if (string.IsNullOrEmpty(s_LocalFqdn))
+                {
+                    s_LocalFqdn = Dns.GetHostEntryAsync("127.0.0.1").GetAwaiter().GetResult().HostName;
+                }
```
roncain: Is GetHostEntryAsync available in NET Native yet?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        // Installs the root certificate from the bridge 
+        // Needed for all tests so we trust the incoming certificate coming from the service/bridge
+        // We do our best to detect if we're running on the same box as the bridge; if so, we don't try to install the cert
+        // as that operation requires admin privileges
+        public static void InstallRootCertificateFromBridge()
+        {
+            // PUT the Authority to the Bridge (returns thumbprint)
+            var response = BridgeClient.MakeResourcePutRequest(CertificateAuthorityResourceName, null);
```
roncain: Can this method go re-entrant, and if so is that a problem making multiple PUT requests to this resource?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -0,0 +1,394 @@
...
+
```
roncain: Can this method be re-entered, and if so is there an issue making multiple concurrent PUT requests?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -0,0 +1,394 @@
...
+                    if (response.TryGetValue(CertificateKeyResourceString, out certificateAsBase64))
+                    {
+                        certificateToInstall = new X509Certificate2(Convert.FromBase64String(certificateAsBase64));
+                    }
+                    else
+                    {
+                        StringBuilder sb = new StringBuilder();
+                        foreach (var pair in response)
+                        {
+                            sb.AppendFormat("{0}  {1} : {2}", Environment.NewLine, pair.Key, pair.Value);
```
roncain: Nit: we should use the helper method to format a Dictionary<string, string> as a string for exceptions or tracing.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -0,0 +1,394 @@
...
+                                s_LocalFqdn,
+                                CertificateKeyResourceString,
+                                Environment.NewLine,
+                                sb.ToString()));
+                    }
+                }
+            }
+
+            InstallCertificateToMyStore(certificateToInstall);
+
```
roncain: These 2 calls are outside the lock.  Any issues with concurrently adding the same cert to the store?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
iamjasonp: Ah yes, this does need fixing. We had discussed it earlier but I forgot to make the fix here.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
iamjasonp: Yes, this might be reentrant. There's no issue making multiple PUTs. We just use the PUT as a mechanism to ensure that the root cert is around, and retrieve the thumbprint of the root cert. The return value (assuming the Bridge hasn't reset in the meantime) will be the same.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -0,0 +1,394 @@
...
+        // already present.  Returns 'true' if the certificate was added.
+        private static bool AddToStoreIfNeeded(StoreName storeName,
+                                               StoreLocation storeLocation,
+                                               X509Certificate2 certificate)
+        {
+            X509Store store = null;
+            X509Certificate2 existingCert = null;
+            try
+            {
+                store = new X509Store(storeName, storeLocation);
```
roncain: Should this code be inside a lock?  Can it be re-entered, and if so are there any issues with opening a Store ReadWrite multiple times concurrently?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -0,0 +1,394 @@
...
+        // all certificates used by this process, whether they already were
+        // added by this process or not.
+        private static void UninstallCertificates(StoreName storeName,
+                                                  StoreLocation storeLocation,
+                                                  Dictionary<string, CertificateCacheEntry> cache,
+                                                  bool force)
+        {
+            lock (s_certificateLock)
+            {
+                if (cache.Count == 0)
```
roncain: I think we decided the bridge client would not do any uninstalls.  So I guess I'm just double checking -- this code isn't currently being called, right?  Can it be removed or is that planned for the future?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
```diff
@@ -43,13 +43,13 @@
...
+      <TestProperty Include="BridgeCertificatePassword">
+        <Value>$(BridgeCertificatePassword)</Value>
+        <DefaultValue>test</DefaultValue>
       </TestProperty>
-      <TestProperty Include="BridgeHttpsCertificate">
-        <Value>$(BridgeHttpsCertificate)</Value>
-        <DefaultValue>WcfBridge_localhost.pfx</DefaultValue>
+      <TestProperty Include="BridgeCertificateValidityPeriod">
+        <Value>$(BridgeCertificateValidityPeriod)</Value>
+        <DefaultValue>1.00:00:00</DefaultValue>
```
roncain: Elsewhere I commented the cert generator should have a configurable validity period.  Here it is!  Can the cert generator use this config property rather than a constant?  Is it intended for the same purpose?

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ResourceController.cs
```diff
@@ -182,30 +195,39 @@ public HttpResponseMessage Delete(HttpRequestMessage request)
...
 
-            // PUT allows name/value pairs in Uri query parameters or in content.
-            // Give precedence to query parameters.
-            // If there were no query parameters, allow the content to provide it.
-            if (properties.Count == 0 && request.Content != null)
+            if (!properties.ContainsKey("name"))
             {
-                properties = GetNameValuePairsFromContent(request);
+                properties["name"] = resourceName;
             }
```
roncain: Nit: don't we have a const for the "name" key?  Can we use it here?

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
```diff
@@ -0,0 +1,451 @@
...
+
+        // We need to hang onto the _authorityKeyPair and _authorityCertificate - all certificates generated 
+        // by this instance will be signed by this Authority certificate and private key
+        private AsymmetricCipherKeyPair _authorityKeyPair;
+        private X509CertificateContainer _authorityCertificate;
+
+        public void Initialize()
+        {
+            if (!_isInitialized)
+            {
```
roncain: Can Initialize() or Reset() be re-entered?  i.e. do they need a lock?

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateManager.cs
```diff
@@ -100,64 +95,97 @@ private static X509Certificate2 CertificateFromThumbprint(X509Store store, strin
...
+
+            lock (s_certificateLock)
+            {
+                if (s_localCertificate != null)
+                {
+                    return s_localCertificate;
+                }
+
+                Trace.WriteLine("[CertificateManager] Installing Root and Machine certificates to machine store.");
+
```
roncain: Nit: could use the more common trace pattern that includes a time stamp and category.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateManager.cs
```diff
@@ -195,12 +228,15 @@ public static void UninstallAllMyCertificates(bool force)
...
                         if (force || pair.Value.AddedToStore)
                         {
-                            X509Certificate2 cert = CertificateFromThumbprint(store, pair.Value.ThumbPrint);
+                            X509Certificate2 cert = CertificateFromThumbprint(store, pair.Value.Thumbprint);
                             if (cert != null)
                             {
                                 store.Remove(cert);
-                                Console.WriteLine("Uninstalled from store '{0}', location '{1}', cert '{2}'",
-                                                    storeName, storeLocation, cert.SubjectName.Name);
+                                Trace.WriteLine(string.Format("[CertificateManager] Removed certificate from store: "));
```
roncain: Nit: it would be better to StringBuilder() this and Trace once, using the pattern that includes a timestamp and category

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/WcfTestBridgeCommon.csproj
```diff
@@ -54,5 +59,12 @@
...
   </ItemGroup>
+  <ItemGroup>
+    <None Include="packages.config" />
+  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+    <Target Name="BeforeBuild">
+    <Message Text="$$$ OutputPath is $(OutputPath)" />
+  </Target>
 </Project>
```
roncain: Nit: missing newlines.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/HttpsResource.cs
```diff
@@ -19,20 +21,10 @@ protected override int GetPort(ResourceRequestContext context)
...
-
-        internal static string EnsureHttpsCertificateInstalled(BridgeConfiguration configuration)
-        {
             // Ensure the https certificate is installed before this endpoint resource is used
-            string certThumbPrint = CertificateManager.InstallMyCertificate(configuration,
-                                                                            configuration.BridgeHttpsCertificate);
-
-            // Ensure http.sys has been told to use this certificate on the https port
-            CertificateManager.InstallSSLPortCertificate(certThumbPrint, configuration.BridgeHttpsPort);
+            CertificateResourceHelpers.EnsureSslPortCertificateInstalled(context.BridgeConfiguration);
```
roncain: Not now, but this call to ensure the SSL port cert is installed feels like it should be the responsibility of one of the certificate resources.  It will be an easy thing to mess up adding new endpoint resources.

roncain:
LGTM.
I'm fine with merging now and making any changes I called out in a mop up PR.  Your choice.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
iamjasonp: This should be removable from the BridgeClient - this class was initially a clone from the Bridge server.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
iamjasonp: It's already done :smile: The reason why it's weird is because the usage pattern is weird. The earlier field you commented on in CertificateGenerator was a default value. Once the Bridge configuration is pushed over, Initialize() is called - and that's what determines the real initialized values in the CertificateGenerator.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/ResourceController.cs
iamjasonp: Added a const to ResourceController.

We had this keyName for another resource, but we haven't made the set of keyname consts shared across all Controllers and IResources yet.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/HttpsResource.cs
iamjasonp: I kept this here because the IResource has no knowledge of the serviceHost, and I feel like shouldn't need to...

iamjasonp: Actually, the MachineCertificateResource already does something about installing certs if it detects the requested cert is for a machine that is local. Maybe we can get rid of this.

iamjasonp:
Pushed a new update. Since some comments went into the individual commits instead of the "Files changed", I've saved the comments from those individual commits to PDF and attached the comments here for posterity. 

[PR394-comments-20151008.pdf](https://github.com/dotnet/wcf/files/9503/PR394-comments-20151008.pdf)
[PR394-comments-597db89.pdf](https://github.com/dotnet/wcf/files/9500/PR394-comments-597db89.pdf)
[PR394-comments-c5656f3.pdf](https://github.com/dotnet/wcf/files/9502/PR394-comments-c5656f3.pdf)
[PR394-comments-d4e0644.pdf](https://github.com/dotnet/wcf/files/9501/PR394-comments-d4e0644.pdf)
[PR394-comments-9cbe1a1.pdf](https://github.com/dotnet/wcf/files/9504/PR394-comments-9cbe1a1.pdf)

iamjasonp:

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
iamjasonp: The first lines in each of `InstallCertificateToMyStore(certificateToInstall);` and `InstallRootCertificateFromBridge();` are locked, and all checks are dealt with inside. Shouldn't be a problem here.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
hongdai: Just a recommendation, in general, use ConcurrentDictionary instead of Dictionary + lock is less error prone and more efficient.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
iamjasonp: I threw a lock around this - I believe it's possible to re-enter here on multiple threads. 

I don't know the behaviour of opening a Store and writing to it concurrently.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
iamjasonp: This was nominally a clone from the Bridge's CertificateManager (though it's been changed to the point where it's not really similar anymore). 

We're locking left and right for other reasons anyway though, so that's why I consciously didn't convert to use ConcurrentDictionary. We can certainly revisit it if we start seeing problems with the way this locking works.

iamjasonp:
LCA approval received - there are no further procedural things keeping us from merging. I have squashed to a smaller set of commits already as promised. 

Assuming no further major comments, OK to merge pending buddy builds.

roncain:
LGTM after latest changes and buddy building

iamjasonp:
Merging now. @roncain, could you help to monitor the rolling builds? Thanks! 

**Others:** if you have further comments, please continue to comment here, and I'll roll up all the comments into another PR/issue(s) as needed.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
iamjasonp: We expect users of CertificateGenerator to take any necessary locks

PR #401, opened 2015-10-08
Move System.Private.ServiceModel to use new S.Security.Claims version

### LINKED ISSUE #387: Fix TCP scenarios blocked by behaviour difference in Claim ctor
There is a behaviour difference in `System.Security.Claims.Claim...ctor` that got introduced between Desktop and CoreCLR that causes an exception to be thrown when trying to create an [AnonymousIdentity]( https://github.com/dotnet/wcf/blob/fb9eedb3bbc464bdf5819ef454a4e21f31f45571/src/System.Private.ServiceModel/src/System/IdentityModel/SecurityUtils.cs#L36). 

Repro: 

    public static void Main()
    {
        System.Security.Principal.GenericIdentity identity = new System.Security.Principal.GenericIdentity(""); 
    }

The difference is a change in the way the constructor works - the empty string (valid value in desktop) is now treated the same as null (an invalid value), and therefore the call fails with the exception below

```
System.ArgumentNullException : Value cannot be null.
Parameter name: value
Stack Trace:
    at System.Security.Claims.Claim..ctor(String type, String value, String valueType, String issuer, String originalIssuer, ClaimsIdentity subject, String propertyKey, String propertyValue)
    at System.Security.Principal.GenericIdentity.AddNameClaim()
```

This issue tracks figuring out the way forward for this issue: either the fix happens in System.Security.Claims or we work around it somehow in WCF Core code on the Identity team's advice.

iamjasonp:
There was a bug in System.Security.Claims that prevented our Ssl stream scenarios
from working. Moving the dependency for S.Security.Claims to depend on the -beta
versions so we can consume this change

Resolves #387

### DIFF
--- a/src/System.Private.ServiceModel/src/project.json
+++ b/src/System.Private.ServiceModel/src/project.json
@@ -33,7 +33,7 @@
     "System.Runtime.Handles": "4.0.0",
     "System.Runtime.InteropServices": "4.0.20-beta-*",
     "System.Runtime.Serialization.Xml": "4.0.10",
-    "System.Security.Claims": "4.0.0",
+    "System.Security.Claims": "4.0.1-beta-*",
     "System.Security.Cryptography.X509Certificates": "4.0.0-beta-*",
     "System.Security.Principal": "4.0.0",
     "System.Security.Principal.Windows": "4.0.0-beta-*",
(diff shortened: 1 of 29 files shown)


iamjasonp:
As this will touch project(.lock).json files, waiting for #402 to be resolved before merge.

roncain:
LGTM, but I would appreciate letting https://github.com/dotnet/wcf/pull/405 be merged first, just because it's been tested on several machines to prepare for checkin,  I'm aiming for Monday.

iamjasonp:
Sounds good. Let's wait for #405

roncain:
LGTM.  #405 has been merged so may need a rebase

iamjasonp:

iamjasonp:
Merging so we can catch this evening's build

PR #403, opened 2015-10-09
Remove dependency on S.Net.NameResolution for UWP in BridgeClient

### LINKED ISSUE #397: Remove dependency on System.Net.NameResolution from BridgeClientCertificateManager
BridgeClientCertificateManager is using Dns.GetHostEntryAsync() which requires System.Net.NameResolution.  But N does not support this contract.

Refactor the test not to require this dependency, and also fix the TFS ToF test wrapper that added it to allow the tests to pass in CoreCLR

iamjasonp:
BridgeClientCertificateManager took a dependency on S.Net.NameResolution, but
.NET Native doesn't support the contract

We can get away with a non-machine named cert on the client side for now

Fixes #397

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Net;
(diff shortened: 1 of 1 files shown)


iamjasonp:
A change will need to be submitted to the internal test suite to get rid of the reference to System.Net.NameResolution simultaneously

roncain:
Did a buddy build, buddy test, and even a cross-machine buddy test. LGTM so I will merge to make headway.

PR #405, opened 2015-10-09
Remove code dependencies on System.Net.Security for NET Native

roncain:
The System.Net.Security package is not available for UWP apps,
but WCF currently lists it as a dependency and uses it,
regardless of platform.  This causes UWP apps to fail in beta8.

The fix is to remove use of it when building for Net Native.

This PR is the first of 2.  The first adds conditional compilation
directives around code using System.Net.Security to exclude usage
when building for NET Native.

The 2nd PR will modify the project.json and project.lock.json files
to remove the dependency.

It is being done in 2 stages like this so that the code changes
can be integrated into the release branch for beta8 independently.

The change to the project*.json files will require careful planning
so that when integrated to the release branch it does not carry with
it the new (post-beta8) dependencies currently in WCF.

### DIFF
--- a/src/System.Private.ServiceModel/src/Mockups/ExtendedProtectionPolicy.cs
+++ b/src/System.Private.ServiceModel/src/Mockups/ExtendedProtectionPolicy.cs
@@ -1,0 +1,55 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information
+
+#if FEATURE_NETNATIVE
+using System.Collections;
+using System.ServiceModel;
+using System.Text;
+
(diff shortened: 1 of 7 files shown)


roncain:
Please don't merge yet.  Without part 2 this fails to build in TFS.  But I am testing it as-is against the release branch before adding part 2

iamjasonp:
LGTM 
Buddy built in local repo - tests pass 
Buddy buitl on TFS CoreCLR - tests pass 
Buddy built on TFS .NET Native - didn't run tests

roncain:
Did final buddy build in TFS for CoreCLR and NET Native and re-ran Inner and OuterLoops in CoreCLR and NET Native

PR #410, opened 2015-10-09
Used CoalescedTimeoutTokens to cut down thread contention

shhsu:
Whenever a CancelationToken is created with timeout, a timer object is created. Disposing a massive amount of timer object would cause thread contention. This change borrow the concept from @stephentoub's idea of coalescing timeout tokens (http://blogs.msdn.com/b/pfxteam/archive/2011/12/03/10244004.aspx) and ends up resolves huge amount of thread contention.

I have observed throughput improvement of 80% in my experiment.

### DIFF
--- a/src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
+++ b/src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
@@ -1,9 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Concurrent;
 using System.Diagnostics.Contracts;
-using System;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace System.Runtime
 {
@@ -17,6 +18,7 @@ namespace System.Runtime
         private DateTime _deadline;
         private TimeSpan _originalTimeout;
         public static readonly TimeSpan MaxWait = TimeSpan.FromMilliseconds(Int32.MaxValue);
+        private static Action<object> s_cancelOnTimeout = state => ((TimeoutHelper)state)._cts.Cancel();
 
         public TimeoutHelper(TimeSpan timeout)
         {
@@ -40,8 +42,9 @@ namespace System.Runtime
             }
             else if (timeout > TimeSpan.Zero)
             {
-                _cts = new CancellationTokenSource(timeout);
+                _cts = new CancellationTokenSource();
                 _cancellationToken = _cts.Token;
+                TimeoutTokenSource.FromTimeout((int)timeout.TotalMilliseconds).Register(s_cancelOnTimeout, this);
             }
             else
             {
@@ -246,6 +249,75 @@ namespace System.Runtime
         internal static TimeoutException CreateEnterTimedOutException(TimeSpan timeout)
         {
             return new TimeoutException(SR.Format(SR.LockTimeoutExceptionMessage, timeout));
+        }
+    }
+
+    /// <summary>
+    /// This class coalesces timeout tokens because cancelation tokens with timeouts are more expensive to expose.
+    /// Disposing too many such tokens will cause thread contentions in high throughput scenario.
+    ///
+    /// Tokens with target cancelation time 15ms apart would resolve to the same instance.
+    /// </summary>
+    internal static class TimeoutTokenSource
+    {
+        private const int COALESCING_SPAN_MS = 15;
+        private static readonly ConcurrentDictionary<long, Task<CancellationToken>> s_tokenCache =
+            new ConcurrentDictionary<long, Task<CancellationToken>>();
+
+        public static CancellationToken FromTimeout(int millisecondsTimeout)
+        {
+            return FromTimeoutAsync(millisecondsTimeout).Result;
+        }
+
+        public static Task<CancellationToken> FromTimeoutAsync(int millisecondsTimeout)
+        {
+            // Note that CancellationTokenSource constructor requires input to be >= -1,
+            // restricting millisecondsTimeout to be >= -1 would enforce that
+            if (millisecondsTimeout < -1)
+            {
+                throw new ArgumentOutOfRangeException("Invalid millisecondsTimeout value " + millisecondsTimeout);
+            }
+
+            uint currentTime = (uint)Environment.TickCount;
+            long targetTime = millisecondsTimeout + currentTime;
+            // round the targetTime up to the next closest 15ms
+            targetTime = ((targetTime + (COALESCING_SPAN_MS - 1)) / COALESCING_SPAN_MS) * COALESCING_SPAN_MS;
+
+            Task<CancellationToken> tokenTask;
+
+            if (!s_tokenCache.TryGetValue(targetTime, out tokenTask))
+            {
+                var tcs = new TaskCompletionSource<CancellationToken>();
+
+                // only a single thread may succeed adding its task into the cache
+                if (s_tokenCache.TryAdd(targetTime, tcs.Task))
+                {
+                    // Since this thread was successful reserving a spot in the cache, it would be the only thread
+                    // that construct the CancellationTokenSource
+                    var token = new CancellationTokenSource((int)(targetTime - currentTime)).Token;
+
+                    // Clean up cache when Token is canceled
+                    token.Register(t => {
+                        Task<CancellationToken> ignored;
+                        s_tokenCache.TryRemove((long)t, out ignored);
+                    }, targetTime);
+
+                    // set the result so other thread may observe the token, and return
+                    tcs.TrySetResult(token);
+                    tokenTask = tcs.Task;
+                }
+                else
+                {
+                    // for threads that failed when calling TryAdd, there should be one already in the cache
+                    if (!s_tokenCache.TryGetValue(targetTime, out tokenTask))
+                    {
+                        // In unlikely scenario the token was already cancelled and timed out, we would not find it in cache.
+                        // In this case we would simply create a non-coalsed token
+                        tokenTask = Task.FromResult(new CancellationTokenSource(millisecondsTimeout).Token);
+                    }
+                }
+            }
+            return tokenTask;
         }
     }
 }


shhsu:
@SajayAntony @mconnew

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,69 @@
...
+using System.Collections.Concurrent;
 using System.Diagnostics.Contracts;
 using System;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace System.Runtime
 {
+    class CoalescedTimeoutTokens
+
```
SajayAntony: Have you run code formatter?

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,69 @@
...
+        {
+            _coalescingSpanMs = (uint)coalescingSpanMs;
+        }
+
+        public CancellationToken FromTimeout(int millisecondsTimeout)
+        {
+            return FromTimeoutAsync(millisecondsTimeout).Result;
+        }
+
+        public Task<CancellationToken> FromTimeoutAsync(int millisecondsTimeout)
```
stephentoub: Why have tasks involved?  It looks like the only consumer of this always immediately gets the Result to extract the CancellationToken.  It'd be cheaper to not allocate new tasks on each call if those tasks aren't necessary.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,69 @@
...
+                var tcs = new TaskCompletionSource<CancellationToken>();
+                if (_timeToToken.TryAdd(targetTime, tcs.Task))
+                {
+                    var token = new CancellationTokenSource((int)(targetTime - currentTime)).Token;
+                    token.Register(t => {
+                        Task<CancellationToken> ignored;
+                        _timeToToken.TryRemove((long)t, out ignored);
+                    }, targetTime);
+                    tcs.TrySetResult(token);
+                    tokenTask = tcs.Task;
```
stephentoub: Why use a TaskCompletionSource? It's only ever set here synchronously, so why not just use Task.FromResult?

And then since everything here related to tasks is synchronous, I'd suggest removing tasks from it entirely.  If you really need to get a ```Task<CancellationToken>``` for some other future consumer, that consumer can then just use Task.FromResult with the CancellationToken this gives back.

stephentoub:
> This change borrow the concept from @stephentoub's idea of coalescing timeout tokens 

Glad to hear it was helpful :smile:

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,69 @@
...
+        public Task<CancellationToken> FromTimeoutAsync(int millisecondsTimeout)
+        {
+            if (millisecondsTimeout <= 0)
+            {
+                return Task.FromResult(new CancellationToken(true));
+            }
+
+            uint currentTime = (uint)Environment.TickCount;
+            long targetTime = millisecondsTimeout + currentTime;
+            targetTime = ((targetTime + (_coalescingSpanMs - 1)) / _coalescingSpanMs) * _coalescingSpanMs;
```
SajayAntony: Need to document the fact that that you have a ceil of 15ms.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,69 @@
...
 namespace System.Runtime
 {
+    class CoalescedTimeoutTokens
+
+    {
+        private readonly uint _coalescingSpanMs;
+        private readonly ConcurrentDictionary<long, Task<CancellationToken>> _timeToToken =
+            new ConcurrentDictionary<long, Task<CancellationToken>>();
+
+        public CoalescedTimeoutTokens(int coalescingSpanMs = 15)
```
SajayAntony: Const this to avoid confusion.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: We considered the alternative implementation as followed:

Line 1: if (!s_tokenCache.TryGetValue(targetTime, out tokenTask))
Line 2:    var token = new CancellationTokenSource((int)(targetTime - currentTime)).Token;
Line 3:    if (s_tokenCache.TryAdd(targetTime, token)) .....

What we have observed is that the above implementation still causes enough contention to slow down our high throughput test. There are still a lot of CancellationTokenSource being created at line 2 every 15ms and disposed at the same time.

What we are doing here is using the Task as a placeholder. Note that if multiple threads come in and get false at the line "!s_tokenCache.TryGetValue(targetTime, out tokenTask)", they would all create a tcs, try and insert it to the cache and only 1 will succeed and get to call new CancellationTokenSource(...).

stephentoub: I see.  Ok, that makes sense.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,80 @@
...
+        private static readonly ConcurrentDictionary<long, Task<CancellationToken>> s_tokenCache =
+            new ConcurrentDictionary<long, Task<CancellationToken>>();
+
+        public static CancellationToken FromTimeout(int millisecondsTimeout)
+        {
+            return FromTimeoutAsync(millisecondsTimeout).Result;
+        }
+
+        public static Task<CancellationToken> FromTimeoutAsync(int millisecondsTimeout)
+        {
```
iamjasonp: Is this ever expected to be < 0? Should we assert here?

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,80 @@
...
+            if (millisecondsTimeout <= 0)
+            {
+                return Task.FromResult(new CancellationToken(true));
+            }
+
+            uint currentTime = (uint)Environment.TickCount;
+            long targetTime = millisecondsTimeout + currentTime;
+            // round the targetTime up to the next closest 15ms
+            targetTime = ((targetTime + (COALESCING_SPAN_MS - 1)) / COALESCING_SPAN_MS) * COALESCING_SPAN_MS;
```
iamjasonp: Will (COALESCING_SPAN_MS - 1) be optimized automagically, or is this an extra instruction that happens on every call here?

iamjasonp:
@shhsu, please don't forget to sign the CLA!
All, just a reminder that the cla must be signed before we start reviewing PRs, even if @shhsu is a Microsoftie :smile:

I don't have much feedback here, the comments explain the changes fairly well. LGTM.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
stephentoub: The C# compiler will do constant folding and will generate the literal ```14``` in the IL.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,80 @@
...
 namespace System.Runtime
 {
+    /// <summary>
+    /// This class coalesces timeout tokens because cancelation tokens with timeouts are more expensive to expose.
+    /// Disposing too many such tokens will cause thread contentions in high throughput scenario.
+    ///
+    /// Tokens with target cancelation time 15ms apart would resolve to the same instance.
+    /// </summary>
+    static class TimeoutTokenSource
+    {
```
roncain: Please add explicit "internal" qualifier (one of the styles required by dotnet repo's).

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,80 @@
...
+
+            uint currentTime = (uint)Environment.TickCount;
+            long targetTime = millisecondsTimeout + currentTime;
+            // round the targetTime up to the next closest 15ms
+            targetTime = ((targetTime + (COALESCING_SPAN_MS - 1)) / COALESCING_SPAN_MS) * COALESCING_SPAN_MS;
+
+            Task<CancellationToken> tokenTask;
+
+            if (!s_tokenCache.TryGetValue(targetTime, out tokenTask))
+            {
```
roncain: With ConcurrentDictionary, I believe it would be better to use GetOrAdd rather than separate TryGet and TryAdd.  If there is a reason why not, would you please add some comments?

roncain:
@mconnew I know you have been looking at the CT cleanup. Any comments on this PR?

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
stephentoub: With the overloads currently exposed on ConcurrentDictionary, using GetOrAdd is likely to result in needing to allocate on each call, due to the state necessary to construct the CancellationTokenSource if TryGetValue returns false.  New overloads are being added to ConcurrentDictionary to help address that, but they're still in the corefx future branch.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,80 @@
...
+        public static CancellationToken FromTimeout(int millisecondsTimeout)
+        {
+            return FromTimeoutAsync(millisecondsTimeout).Result;
+        }
+
+        public static Task<CancellationToken> FromTimeoutAsync(int millisecondsTimeout)
+        {
+            if (millisecondsTimeout <= 0)
+            {
+                return Task.FromResult(new CancellationToken(true));
```
stephentoub: Is it common for millisecondsTimeout to be 0?  If so, I'd suggest caching this task into a static, e.g.
```C#
private static readonly s_canceledTokenTask = Task.FromResult(new CancellationToken(true));
...
if (millisecondsTimeout <= 0)
{
    return s_canceledTokenTask;
}
```

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -40,8 +108,8 @@ private void InitializeCancellationToken(TimeSpan timeout)
             }
             else if (timeout > TimeSpan.Zero)
             {
-                _cts = new CancellationTokenSource(timeout);
-                _cancellationToken = _cts.Token;
+                _cts = new CancellationTokenSource();
+                _cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(_cts.Token, TimeoutTokenSource.FromTimeout((int)timeout.TotalMilliseconds)).Token;
```
stephentoub: Since you're not cleaning this up anyway, consider trying out:
```C#
_cts = new CancellationTokenSource();
_cancellationToken = _cts.Token;
TimeoutTokenSource.FromTimeout((int)timeout.TotalMilliseconds)).Token.Register(s => ((CancellationTokenSource)s).Cancel(), _cts);
```
and seeing whether that helps with performance.  It should have the same behavior but should save a couple of allocations.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Done

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Sure, it seems disallowing negatives would be more in line of the behavior of CancellationTokenSource.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: No longer allow negatives

stephentoub: What about 0?

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: I was able to further cut down another allocation by making the registered delegate static. There seem to be a throughput improvement of around 3%.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -17,6 +87,7 @@ public struct TimeoutHelper : IDisposable
         private DateTime _deadline;
         private TimeSpan _originalTimeout;
         public static readonly TimeSpan MaxWait = TimeSpan.FromMilliseconds(Int32.MaxValue);
+        private static Action<object> CancelOnTimeout = state => ((TimeoutHelper)state)._cts.Cancel();
```
SajayAntony: what is the coding convention for static delegate?

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Note that targetTime - currentTime >= millisecondsTimeout, so restricting it to > -1 would set a valid upper bound for the parameter of CancellationTokenSource.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Fixing in next revision

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Also, note that the timeout equals zero use case is addressed here.
https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs#L48
So I won't be having a pre-timedouted stub in this method

shhsu:
It seems possible to eliminate the synchronous code path here as caller of TimeoutHelper.CancellationToken seem to be all async. I will be evaluating this improvement and create a new PR for it once this change is checked in

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,81 @@
...
+                    tcs.TrySetResult(token);
+                    tokenTask = tcs.Task;
+                }
+                else
+                {
+                    // for threads that failed when calling TryAdd, there should be one already in the cache
+                    // Unless in unlikely scenario the token was already cancelled and timed out
+                    if (!s_tokenCache.TryGetValue(targetTime, out tokenTask))
+                    {
+                        tokenTask = Task.FromResult(new CancellationTokenSource((int)(targetTime - currentTime)).Token);
```
mconnew: Change to
```csharp
    tokenTask = Task.FromResult(new CancellationTokenSource(millisecondsTimeout).Token);
```
Because we're creating a non-coalesced CT there's no need to round it and we might as well give an accurate Token.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,81 @@
...
+                    // set the result so other thread may observe the token, and return
+                    tcs.TrySetResult(token);
+                    tokenTask = tcs.Task;
+                }
+                else
+                {
+                    // for threads that failed when calling TryAdd, there should be one already in the cache
+                    // Unless in unlikely scenario the token was already cancelled and timed out
+                    if (!s_tokenCache.TryGetValue(targetTime, out tokenTask))
+                    {
```
mconnew: Can you add a comment that explicitly mentions that rather than spend the cycles trying to resolve a rare/infrequent race condition, the caller will be given a non-coalesced CT.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -1,12 +1,81 @@
...
 
 namespace System.Runtime
 {
+    /// <summary>
+    /// This class coalesces timeout tokens because cancelation tokens with timeouts are more expensive to expose.
+    /// Disposing too many such tokens will cause thread contentions in high throughput scenario.
+    ///
+    /// Tokens with target cancelation time 15ms apart would resolve to the same instance.
+    /// </summary>
+    internal static class TimeoutTokenSource
```
mconnew: As a final step (otherwise it will make reviewing harder), can you move TimeoutHelper to be the first class in this file. It's a personal style thing, but I like to have the first class in a file match the filename.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Fixed

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Fixed

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Fixed

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -40,8 +42,9 @@ private void InitializeCancellationToken(TimeSpan timeout)
             }
             else if (timeout > TimeSpan.Zero)
             {
-                _cts = new CancellationTokenSource(timeout);
+                _cts = new CancellationTokenSource();
```
mconnew: I don't like having this extra CancellationTokenSource allocation. It's only needed to support the CancelCancellationToken method which I can only find a single usage of in HttpChannelFactory. This functionality should be either placed behind an overload or embedded into HttpChannelFactory itself. Otherwise we have an extra unnecessary CTS and CT being created every time. If you want to measure the perf impact, you can do a temporary experiment by removing the creation of the extra CTS and CT, remove the call to _cts.Dispose() in the Dispose method and make the CancellationTokenSource() method throw. HttpChannelFactory only calls this method in the Abort/Faulting case so you should be able to run your perf tests unaffected by the temporary change. This will give us numbers to show how much of an impact (and how important to fix) this is.

stephentoub: Definitely agree.  If this isn't needed in most cases, it should definitely be opt-in and/or simply done at the call site where needed, since the implementation here isn't doing anything a caller couldn't do as well.

mconnew:
As this improves performance and is functional, I'm okay with this being checked in as it is currently with later incremental improvements being made.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Will investigate after merging this PR

PR #386, opened 2015-10-05
Bump System.Private.ServiceModel version to 4.1.0

iamjasonp:
Related to #385

In preparation for API changes to System.ServiceModel.Primitives and System.ServiceModel.NetTcp, we need to bump the minor version of System.Private.ServiceModel and wait for the nightly build to push out a new package

PR #411, opened 2015-10-12
Add X509Certificate-related APIs to System.ServiceModel.Primitives

iamjasonp:
Dependent on #386, where we bumped the version of System.Private.ServiceModel

Full API addition discussion is at #385

### DIFF
--- a/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
+++ b/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
@@ -4,7 +4,12 @@
 // Changes to this file must follow the http://aka.ms/api-review process.
 // ------------------------------------------------------------------------------
 
-
+namespace System.IdentityModel.Selectors {
+  public abstract partial class X509CertificateValidator {
+    protected X509CertificateValidator() { }
+    public abstract void Validate(System.Security.Cryptography.X509Certificates.X509Certificate2 certificate);
+  }
+}
 namespace System.ServiceModel
 {
     public partial class ActionNotSupportedException : System.ServiceModel.CommunicationException
@@ -1160,7 +1165,9 @@ namespace System.ServiceModel.Description
     {
         public ClientCredentials() { }
         protected ClientCredentials(System.ServiceModel.Description.ClientCredentials other) { }
+        public System.ServiceModel.Security.X509CertificateInitiatorClientCredential ClientCertificate { get { return default(System.ServiceModel.Security.X509CertificateInitiatorClientCredential); } }
         public System.ServiceModel.Security.HttpDigestClientCredential HttpDigest { get { return default(System.ServiceModel.Security.HttpDigestClientCredential); } }
+        public System.ServiceModel.Security.X509CertificateRecipientClientCredential ServiceCertificate { get { return default(System.ServiceModel.Security.X509CertificateRecipientClientCredential); } }
         public System.ServiceModel.Security.UserNamePasswordClientCredential UserName { get { return default(System.ServiceModel.Security.UserNamePasswordClientCredential); } }
         public System.ServiceModel.Security.WindowsClientCredential Windows { get { return default(System.ServiceModel.Security.WindowsClientCredential); } }
         public virtual void ApplyClientBehavior(System.ServiceModel.Description.ServiceEndpoint serviceEndpoint, System.ServiceModel.Dispatcher.ClientRuntime behavior) { }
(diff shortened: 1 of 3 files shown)


src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.csproj
```diff
@@ -2,7 +2,7 @@
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
-    <AssemblyVersion>4.0.1.0</AssemblyVersion>
+    <AssemblyVersion>4.1.0.0</AssemblyVersion>
     <OutputType>Library</OutputType>
```
roncain: Isn't this 4.1.0 change already merged?  Or am I looking at an old PR?

iamjasonp: This is the change to System.ServiceModel.Primitives. The previous change was to System.Private.ServiceModel :smile:

src/System.ServiceModel.Primitives/ref/project.json
```diff
@@ -13,6 +13,8 @@
     "System.Reflection": "4.0.0",
     "System.Net.Primitives": "4.0.0",
     "System.Security.Principal": "4.0.0",
+    "System.Security.Cryptography.X509Certificates" : "4.0.0-beta-*",
```
iamjasonp: Is depending on -beta- okay? (Main discussion, see #385)

@weshaggard, @ericstj

ericstj: Adding -beta is fine and expected for contracts that are still under development.  You won't be able to mark this stable until X509Certificates is stable, but that's expected to happen at the same time: ASP.NET v5 RTM.  Assuming you don't plan to ship a stable version of this contract before then, go ahead.

iamjasonp:
@immol, this PR should interest you for API review

zhenlan:
Everything looks consistent with desktop .NET except X509CertificateValidator. I understand why you end up with what you have in this PR (as configuration is not supported here), but it will be good if we can get confirmation from @weshaggard for X509CertificateValidator. Other than that, this LGTM.

iamjasonp:
@roncain, I was able to cleanly build and run tests on this, but I couldn't do the package gen (it failed with some obscure error). Could you assist?

weshaggard:
From an API review prospective it would be good to understand if this is just a subset of APIs that are in desktop or if this has new APIs that don't exist on desktop. If it is just a subset then we just need to ensure it is the right subset. If it is new APIs then we have to figure out our story for rationalizing them with desktop.

src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
```diff
@@ -4,7 +4,12 @@
 // Changes to this file must follow the http://aka.ms/api-review process.
 // ------------------------------------------------------------------------------
 
-
+namespace System.IdentityModel.Selectors {
+  public abstract partial class X509CertificateValidator {
+    protected X509CertificateValidator() { }
+    public abstract void Validate(System.Security.Cryptography.X509Certificates.X509Certificate2 certificate);
+  }
+}
```
roncain: Is there any issue introducing this abstract class that is a subset of the one Desktop?  The Desktop version also implements ICustomIdentityConfiguration.

weshaggard: There isn't an issue as long as it is a pure subset. You can essentially hide members or interfaces/base types but you cannot introduce new members without having to figure out a plan.

The one question I would have with this type is that is isn't part of ServiceModel. It is part of IdentityModel so I'm not sure if it bellows in ServiceModel.Primitives. How critical is it for the scenario you are trying to accomplish?

mconnew: It is very important. Without this type, customers can't do their own custom certificate validation. On desktop, outside of WCF the only place this class is used is in System.IdentityModel.dll itself.

iamjasonp: System.IdentityModel and System.ServiceModel were tightly coupled starting in 3.0, so most of the security and identity bits live in IdentityModel. We don't have an IdentityModel contract though, hence why it's being exposed in this manner.

weshaggard: If we ever needed to expose the rest of that type what do you guys think we would do? Would we add the rest of it into S.SM.Primitives contract? Or would we create another IdentityModel contract and have it be a dependency of S.SM.Primtives and type forward this type into there? I just want to make sure we think about what we would do with this in the future.

mconnew: The rest of the class that isn't included are a single instance method used by the configuration system on desktop (so we wouldn't include that), and static properties or methods to provide validators with specific implementations (e.g. None). I believe those should have their current behavior evaluated and would possibly be superseded by a newer implementation with a different behavior as certificate validation criteria has moved on a bit since the code was originally written. At that point, we would likely have more classes we would want to add to a new IdentityModel contract and we would type forward this class, likely with no changes.

weshaggard: Sounds good. Thanks @mconnew.

iamjasonp:
Closed via #429. Thanks all for your comments and feedback!

PR #412, opened 2015-10-12
Remove Contract.Requires<Type> calls from files

iamjasonp:
Replace with Contract.Requires. With the generic Contract.Requires, an exception will be thrown in CoreCLR when this codepath is hit.

This blocks the TCP with security scenario


roncain:
LGTM.

mconnew:
LGTM

PR #414, opened 2015-10-13
Add documentation for certificate generation resource

iamjasonp:
After adding the certificate generation logic to the Bridge, we need to add some documentation to explain how it's used so people other than me can write tests against this resource :smile:

### DIFF
--- a/Documentation/certificate-generation-resources.md
+++ b/Documentation/certificate-generation-resources.md
@@ -1,0 +1,67 @@
+# Dynamic test certificate generation
+
+## Background
+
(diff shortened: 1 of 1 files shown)


roncain:
Nit: I think we should reorganize slightly after this goes in and create 'building' and 'testing' folders under documentation, and move our existing documentation under them.  In the meantime, I'd say go ahead and merge and we can move it after we agree on the layout.

iamjasonp:
Since there aren't comments and these are just docs - merging this now.

PR #421, opened 2015-10-13
Reactivate negative encoding tests disabled for Linux

roncain:
These tests were deactivated on Linux because Encoding had
not yet been implemented.  But this was fixed in CoreFx via
issue https://github.com/dotnet/corefx/issues/2774.


iamjasonp:
Code LGTM. I haven't yet verified this as my Linux box is being set up.

roncain:
Buddy tested on both Windows and Linux, and these reactivated tests pass now

PR #422, opened 2015-10-14
Validate X509CertificateClaims by SAN instead of by CN

### LINKED ISSUE #321: X509CertificateClaimSet should validate based on SAN instead of the Subject Name
For versions of .NET Framework > 4.6 on Desktop, we will no longer be checking certificate Subject Names for CN=host anymore; rather, we will test based on Subject Alternative Name. 

Currently, we are adopting the old behaviour checking against the Subject Name (CN=host) as our test certs are not generated with SANs present (see #320). 

In the wild, most certs issued should already contain SANs for the names that the server can be. We should therefore make a change in WCF Core to check certificates for SANs once #320 is resolved.

iamjasonp:
X509CertificateClaimSet should validate based on Subject Alternative Name instead of the Canonical Name to be consistent with Desktop .NET >= 4.6

Today, We are checking against the Subject (e.g. CN=host) on the certificate, to determine whether or not the cert was issued for a particular server, as this was the behaviour in <= 4.5. However, this is inflexible as the host could be serving the same cert but bound to multiple endpoints.

In the old case, for example, foo-bar.test.com != foo-bar, and a generated cert can only be valid for exactly one of these. In the new case, a certificate can have a Subject with a particular CN, but  multiple Subject Alt Names for which the cert is also valid.

In the wild, most certs issued already contain SANs for the names that the server can be, so we should be consistent with current behaviour and ditch the legacy way of checking certs against servers.

Fixes #321

### DIFF
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
@@ -171,10 +171,12 @@ namespace System.IdentityModel.Claims
             if (!string.IsNullOrEmpty(value))
                 claims.Add(Claim.CreateX500DistinguishedNameClaim(_certificate.SubjectName));
 
-            // #321 - Desktop implmentation > 4.6 replaces this with a SAN check
-            value = _certificate.GetNameInfo(X509NameType.DnsName, false);
-            if (!string.IsNullOrEmpty(value))
-                claims.Add(Claim.CreateDnsClaim(value));
+            // Since a SAN can have multiple DNS entries
+            string[] entries = GetDnsFromExtensions(_certificate);
+            for (int i = 0; i < entries.Length; ++i)
+            {
+                claims.Add(Claim.CreateDnsClaim(entries[i]));
+            }
 
             value = _certificate.GetNameInfo(X509NameType.SimpleName, false);
             if (!string.IsNullOrEmpty(value))
@@ -243,11 +245,10 @@ namespace System.IdentityModel.Claims
             {
                 if (right == null || Rights.PossessProperty.Equals(right))
                 {
-                    // #321 - Desktop implmentation > 4.6 replaces this with a SAN check
-                    string value = _certificate.GetNameInfo(X509NameType.DnsName, false);
-                    if (!string.IsNullOrEmpty(value))
+                    string[] entries = GetDnsFromExtensions(_certificate);
+                    for (int i = 0; i < entries.Length; ++i)
                     {
-                        yield return Claim.CreateDnsClaim(value);
+                        yield return Claim.CreateDnsClaim(entries[i]);
                     }
                 }
             }
@@ -269,6 +270,32 @@ namespace System.IdentityModel.Claims
                     }
                 }
             }
+        }
+
+        private static string[] GetDnsFromExtensions(X509Certificate2 cert)
+        {
+            foreach (X509Extension ext in cert.Extensions)
+            {
+                // Extension is SAN or SAN2
+                if (ext.Oid.Value == "2.5.29.7" || ext.Oid.Value == "2.5.29.17")
+                {
+                    string asnString = ext.Format(true);
+                    if (string.IsNullOrEmpty(asnString))
+                    {
+                        return new string[0];
+                    }
+
+                    string[] rawDnsEntries = asnString.Split(new string[1] { "\n" }, StringSplitOptions.RemoveEmptyEntries);
+                    string[] dnsEntries = new string[rawDnsEntries.Length];
+                    for (int i = 0; i < rawDnsEntries.Length; ++i)
+                    {
+                        int equalSignIndex = rawDnsEntries[i].IndexOf('=');
+                        dnsEntries[i] = rawDnsEntries[i].Substring(equalSignIndex + 1).Trim();
+                    }
+                    return dnsEntries;
+                }
+            }
+            return new string[0];
         }
 
         public override IEnumerator<Claim> GetEnumerator()
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
@@ -362,4 +362,13 @@ public static partial class Endpoints
             return BridgeClient.GetResourceAddress("WcfService.TestResources.TcpTransportSecuritySslCustomCertValidationResource");
         }
     }
+
+    public static string Tcp_ClientCredentialType_Certificate_With_ServerAltName_Address
+    {
+        get
+        {
+            BridgeClientCertificateManager.InstallLocalCertificateFromBridge();
+            return BridgeClient.GetResourceAddress("WcfService.TestResources.TcpCertificateWithServerAltNameResource");
+        }
+    }
 }
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -514,7 +514,6 @@ public static class ExpectedExceptionTests
 
     [Fact]
     [OuterLoop]
-    [ActiveIssue(533)]
     // Verify product throws MessageSecurityException when the service cert is revoked
     public static void TCP_ServiceCertRevoked_Throw_MessageSecurityException()
     {
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
@@ -252,4 +252,43 @@ public static class Tcp_ClientCredentialTypeTests
             ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
         }
     }
+
+    [Fact]
+    [OuterLoop]
+    public static void TcpClientCredentialType_Certificate_With_ServerAltName_EchoString()
+    {
+        EndpointAddress endpointAddress = null;
+        string testString = "Hello";
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
+
+        try
+        {
+            // *** SETUP *** \\
+            NetTcpBinding binding = new NetTcpBinding(SecurityMode.Transport);
+            binding.Security.Transport.ClientCredentialType = TcpClientCredentialType.None;
+
+            endpointAddress = new EndpointAddress(new Uri(Endpoints.Tcp_ClientCredentialType_Certificate_With_ServerAltName_Address));
+
+            factory = new ChannelFactory<IWcfService>(binding, endpointAddress);
+            factory.Credentials.ServiceCertificate.Authentication.CertificateValidationMode = X509CertificateValidationMode.ChainTrust;
+            
+            serviceProxy = factory.CreateChannel();
+
+            // *** EXECUTE *** \\
+            string result = serviceProxy.Echo(testString);
+
+            // *** VALIDATE *** \\
+            Assert.Equal(testString, result);
+
+            // *** CLEANUP *** \\
+            ((ICommunicationObject)serviceProxy).Close();
+            factory.Close();
+        }
+        finally
+        {
+            // *** ENSURE CLEANUP *** \\
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+        }
+    }
 }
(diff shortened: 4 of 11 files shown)


iamjasonp:
@bartonjs / @tushargupta51 / @brentschmaltz, could someone please take a look?

iamjasonp:
After this fix, and after System.Security.Claims >= 4.0.1-beta-23408 references resolve properly, the TCP transport security scenario will light up.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
```diff
@@ -267,6 +268,32 @@ public override IEnumerable<Claim> FindClaims(string claimType, string right)
             }
         }
 
+        private static string[] GetDnsFromExtensions(X509Certificate2 cert)
+        {
+            foreach (X509Extension ext in cert.Extensions)
+            {
+                // Extension is SAN or SAN2
+                if (ext.Oid.Value == "2.5.29.7" || ext.Oid.Value == "2.5.29.17")
```
bartonjs: [OID-info](http://www.oid-info.com/get/2.5.29.7) says that 2.5.29.7 was declared Obsolete in/prior-to the year 2000; you can probably safely ignore it.  I don't think that Unix builds would even recognize it as anything to pretty print, instead just dumping the bytes in hex.  And certainly SslStream won't use that one

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
```diff
@@ -267,6 +268,32 @@ public override IEnumerable<Claim> FindClaims(string claimType, string right)
...
+                    string[] dnsEntries = new string[rawDnsEntries.Length];
+                    for (int i = 0; i < rawDnsEntries.Length; ++i)
+                    {
+                        int equalSignIndex = rawDnsEntries[i].IndexOf('=');
+                        dnsEntries[i] = rawDnsEntries[i].Substring(equalSignIndex + 1).Trim();
+                    }
+                    return dnsEntries;
+                }
+            }
+            return new string[0];
```
bartonjs: If you're trying for RFC 2818 matching logic, in the event there are no DNS values in the Subject Alternative Names list (or there's no SAN extension at all) you should fall back to the CN.  So the previous code would go here.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
```diff
@@ -267,6 +268,32 @@ public override IEnumerable<Claim> FindClaims(string claimType, string right)
...
+                    if (string.IsNullOrEmpty(asnString))
+                    {
+                        return new string[0];
+                    }
+
+                    string[] rawDnsEntries = asnString.Split(new string[1] { "\n" }, StringSplitOptions.RemoveEmptyEntries);
+                    string[] dnsEntries = new string[rawDnsEntries.Length];
+                    for (int i = 0; i < rawDnsEntries.Length; ++i)
+                    {
+                        int equalSignIndex = rawDnsEntries[i].IndexOf('=');
```
bartonjs: There are a couple of problems with this.

* This won't work on x-plat, see https://github.com/dotnet/corefx/issues/3843 (aside: I missed moving this bug to github, glad it came up, so that I could link to it, and then couldn't find it!)
    * Firstly, because it doesn't have a way to print multiline, so the boolean doesn't do anything
    * Secondly, because it is "DNS:" as a prefix, vs Windows (en-US) "DNS Name="
* This is matching on all of the SAN entries.  You really only want the dNSName entries.  Sort of maybe you also want the iPAddress values, but those need to be matched as binary, not text (particularly IPv6, since there're a lot of text representations that are valid for localhost)

If not for the x-plat newlines, a viable option would be to save a byte\[\] of a Subject Alternative Names extension containing one value (and dNSName) and figure out what its prefix is; then match only on those entries.

(SAN Entries are each of type GeneralName:
```
   GeneralName ::= CHOICE {
        otherName                       [0]     OtherName,
        rfc822Name                      [1]     IA5String,
        dNSName                         [2]     IA5String,
        x400Address                     [3]     ORAddress,
        directoryName                   [4]     Name,
        ediPartyName                    [5]     EDIPartyName,
        uniformResourceIdentifier       [6]     IA5String,
        iPAddress                       [7]     OCTET STRING,
        registeredID                    [8]     OBJECT IDENTIFIER }
```
)

iamjasonp:
@bartonjs: Aw dang. 

I took the code in green pretty much verbatim from desktop as it exists for .NET 4.6 - if we make any changes to this logic, we'll have to make the same changes into desktop WCF too :frowning:

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
iamjasonp: It looks like that for .NET <= 4.5, we use the old (red) behaviour, and for .NET >= 4.6, we use the new (green) behaviour. Anything we do here should be backported to desktop

bartonjs:
@iamjasonp In the modern average case you won't be able to really tell a difference; I don't think I've seen anything other than dNSName in any certificate I didn't create myself for testing :).  And the guideline for several years has been that if the certificate is being issued for hostname matching purposes (e.g. TLS server authentication) that it use a dNSName instead of the CN in the subject.

That makes things (in general) happy for Desktop, at least.  The X-Plat formatting problem (or, really, the lack of X509SubjectAlternativeNameExtension as a rich X509Extension) is really what's biting you here.  If you're Windows-only, it'll be fine... but x-plat's... annoyingly difficult right now (sorry).

iamjasonp:
What do you recommend then? Should we get the codebases between .NET Core and Desktop to be the same again (i.e., accept this PR), and after that we can make the appropriate fixes to both places? 

Or should we just hold off on this change until we can make this completely correct? 

I vote for the first option, as it makes life easier when testing (our Bridge cert generation logic would be able to use one cert for multiple SANs, as opposed to attempting to figure out if subject `CN=foo` or `CN=foo.bar.com` is the correct cert to generate)

iamjasonp:
> That makes things (in general) happy for Desktop, at least. The X-Plat formatting problem (or, really, the lack of X509SubjectAlternativeNameExtension as a rich X509Extension) is really what's biting you here. If you're Windows-only, it'll be fine... but x-plat's... annoyingly difficult right now (sorry).

Heh, well, nobody ever said certs were easy :smile:. 

We're aiming to be x-plat, obviously, but we'd rather skip the *nix test and have the Windows one working rather than test neither.

bartonjs:
If this code is the same was what 4.6 has, there's certainly an argument to be made that you're no worse off than 4.6 :smile:.

You'd be improving the Windows version at the expense of the Unix version.  If you're not building platform-specific, what you could do is:
* change `Format(true)` to `Format(false)`
* List-split on commas
* Key-value split on equals or colon.
    * UPNs, if present, come from Windows as "Other Name:Principal Name=subjectupn2@example.org", so maybe you want the last split segment, rather than segments[1]

The newlines becoming commas would potentially cause complication for a certificate which had a SAN entry of type directoryName, since that would format like the subject line.


So.. yeah.  For test code I'd try to make the code more compatible ([I knew that plan sounded familiar, I did it in test code](https://github.com/dotnet/corefx/blob/master/src/System.Security.Cryptography.Encoding/tests/AsnEncodedData.cs#L36-L67)).  For the main assemblies... you'll have to decide what level of breaking something you're willing to risk.  You could try hiding the x-plat compatibility code behind a RuntimeInformation check...

iamjasonp:
Crap. this puts us between a rock and a hard place. We probably want to get this CoreCLR with parity to .NET Desktop 4.6 as a priority, but if this comes at a cost to *nix, then... hmm 

Also we want to make sure that any changes we make here get ported over to Desktop to maintain parity of behaviour between the two platforms. 

RC1 is looming...

roncain:
@iamjasonp  -- what is your proposal to move forward? And what are the fewest changes we can make in RC1 that have the most benefit for both Windows and *NIX? If this isn't decided today and resolved by Wednesday, what functionality will be broken in RC1?  I'd like to turn this conversation into known action items and owners, including both Windows and *NIX.

Specifically:
1. What is broken if we do nothing (both Windows and *NIX)?
2. What if fixed if this PR is merged as-is (both on Windows and *NIX)
3. What will be (or remain) broken is this PR is merged as-is?

If we need conditional compilation or runtime checking to support both, we need to know exactly what that is and when to do the work.

We should also not take it as a given anything we decide will get permission to move to Desktop.  So I'd treat that as a future conversation post-RC1.

iamjasonp:
> 1. What is broken if we do nothing (both Windows and *NIX)?
We would be validating against Canonical Names instead of Subject Alternative Names. If we do nothing, we are likely going to face problems in the real world if customers depend on this RFC 2818 logic - which dates to May 2000. 

> 2. What if fixed if this PR is merged as-is (both on Windows and *NIX)
> 3. What will be (or remain) broken is this PR is merged as-is?
We would be validating against SANs rather than CNs. Windows would improve and validation logic would be mostly compliant with RFC 2818, but a Linux client will be busted according to @bartonjs. 

It's worth noting that I think the Desktop > 4.6 behaviour is non-compliant to the RFC as well, due to it not checking for CN if no SAN are found. 

I think at this point, we are probably best off letting this PR sit until after RC1, after which: 
* Bring the code to parity with 4.6 on Windows
* We open up new issues for all the issues that @bartonjs brought up, as these *need* to be fixed 
* We fix the major x-plat issues above and hide them behind a RuntimeInformation check. 

* Note that we will likely have to diverge the code from desktop and WCF Core; the desktop code does not need to deal with the x-plat issues. Some changes (such as checking CN if no dNSName) will have to be ported to desktop.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
```diff
@@ -171,10 +171,12 @@ IList<Claim> InitializeClaimsCore()
             if (!string.IsNullOrEmpty(value))
                 claims.Add(Claim.CreateX500DistinguishedNameClaim(_certificate.SubjectName));
 
-            // #321 - Desktop implmentation > 4.6 replaces this with a SAN check
-            value = _certificate.GetNameInfo(X509NameType.DnsName, false);
-            if (!string.IsNullOrEmpty(value))
-                claims.Add(Claim.CreateDnsClaim(value));
+            // Since a SAN can have multiple DNS entries
+            string[] entries = GetDnsFromExtensions(_certificate);
```
shmao: Nit: change the variable name to "dnsEntries" to make it more meaningful.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
```diff
@@ -267,6 +268,32 @@ public override IEnumerable<Claim> FindClaims(string claimType, string right)
...
+        {
+            foreach (X509Extension ext in cert.Extensions)
+            {
+                // Extension is SAN or SAN2
+                if (ext.Oid.Value == "2.5.29.7" || ext.Oid.Value == "2.5.29.17")
+                {
+                    string asnString = ext.Format(true);
+                    if (string.IsNullOrEmpty(asnString))
+                    {
+                        return new string[0];
```
shmao: Do we want to return Array<string>.Empty() instead?

iamjasonp: I may take this as an additional comment after this is merged; this current commit brings us into line with what Desktop looks like

roncain:

roncain:

iamjasonp:

iamjasonp:

roncain:
LGTM.  Do we have an open issue to verify this works with a test?  Or is it possible to test with our Bridge endpoints?

iamjasonp:
This may be difficult to test explicitly - verification is that the system *continue* to work as they do, as we've now changed the validation mechanism from CN= to SAN. If they stop working, then we have a problem. 

What we can do is add a negative test to ensure that certs with *only* a CN= (i.e., without the SAN specified) stop working, but @bartonjs mentioned we should probably allow for a fallback in this case too, at some future date. This would require the Bridge to generate such certs, but it's not insurmountable - I'm just wondering if it's worth it for this scenario at hand. 

All the feedback that's here I'll capture into another set of issues and we'll work on them from there. It's a nontrivial list.

bartonjs:
@iamjasonp A good test case would be one where the CN doesn't match and the first SAN entry doesn't match, but a later one does.  For example:

Target: www.example.com
Cert: CN=example.com, SAN(DNS): example.com, SAN(DNS): *.example.com

Of course, there's also SAN(IP) (and IDNA concerns in CN/SAN(DNS))

iamjasonp:
hot damn :) that works. I don't think we're ready yet to handle IDNA yet though, but that's another thing for us to handle in future.

iamjasonp:

iamjasonp:
We've had ample time to review this, so I'm going to close this so we can start work on improving the overall story. https://github.com/dotnet/wcf/pull/422#issuecomment-149645749 is our current plan.

PR #426, opened 2015-10-14
Adding three WebSocket streaming scenario tests.

StephenBonikowsky:
* Fixed one issue where a timeout in the WebSocketTransportDuplexSessionChannel class was not being set.
* Ran into another problem, opened issue 424
* These tests run and pass on desktop using the Bridge.
* These tests are a partial fix of issue 384
* Placing these tests in a new folder called "Extensibility" based on the logic used for locating previously released WCF Samples.
* Opened issue 425 to move other WebSocket tests to this location
* The new Contracts include operations not used by current set of tests, leaving as is because forthcoming tests will use these operations.
* Removed and sorted using statements wherever I saw that it needed it.


src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
```diff
@@ -1,18 +1,24 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Infrastructure.Common;
 using System;
```
roncain: Nit: we normally put all the System ones first, then all the other ones, even though that is not alphabetical.

src/System.Private.ServiceModel/tests/Common/Scenarios/project.json
```diff
@@ -6,7 +6,9 @@
     "System.ServiceModel.Primitives": "4.0.1-beta-*",
     "System.ServiceModel.Security": "4.0.1-beta-*",
     "xunit": "2.1.0",
-    "xunit.netcore.extensions": "1.0.0-prerelease-*"
+    "xunit.netcore.extensions": "1.0.0-prerelease-*",
+	"System.Threading": "4.0.10",
```
roncain: Nit: we normally keep these alphabetized

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/Extensibility.WebSockets.Tests.csproj
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
roncain: Just a reminder -- we will need an equivalent wrapper .csproj for ToF checked into TFS so this will get run in ToF

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
```diff
@@ -909,3 +910,172 @@ public void OnDataContractPingCallback(ComplexCompositeTypeDuplexCallbackOnly da
...
+    }
+
+    public override void Write(byte[] buffer, int offset, int count)
+    {
+        throw new NotImplementedException();
+    }
+}
+
+public class ClientReceiver : IPushCallback
+{
```
roncain: I recommend this class implement IDisposable and explicitly dispose its ManualResetEvents.  That means the tests would need to use "using" statements or do it in the finally block.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -0,0 +1,234 @@
...
+
+                do
+                {
+                    readResult = stream.Read(buffer, 0, buffer.Length);
+                }
+                while (readResult != 0);
+            }
+
+            uploadStream = new FlowControlledStream();
+            uploadStream.ReadThrottle = TimeSpan.FromMilliseconds(500);
```
roncain: FlowControlledStream is a Stream which is IDisposable.  So it should be put in a 'using' or disposed in the finally block

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -0,0 +1,234 @@
...
+            // *** EXECUTE *** \\
+            using (Stream stream = client.DownloadStream())
+            {
+                int readResult;
+                // Read from the stream, 1000 bytes at a time.
+                byte[] buffer = new byte[1000];
+                do
+                {
+                    readResult = stream.Read(buffer, 0, buffer.Length);
+                }
```
roncain: Question only: Can this synchronous Stream.Read block forever in any error conditions?  Just checking if we need to think about an async read with a timeout or something like that.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -0,0 +1,234 @@
...
+            client.UploadStream(uploadStream);
+            client.StartPushingStream();
+            // Wait for the callback to get invoked before telling the service to stop streaming.
+            // This ensures we can read from the stream on the callback while the NCL layer at the service
+            // is still writing the bytes from the stream to the wire.  
+            // This will deadlock if the transfer mode is buffered because the callback will wait for the
+            // stream, and the NCL layer will continue to buffer the stream until it reaches the end.
+
+            clientReceiver.ReceiveStreamInvoked.WaitOne();
+            clientReceiver.ReceiveStreamInvoked.Reset();
```
roncain: I don't we should have any infinite WaitOne() calls in case a test failure.  Recommend using ScenarioTestHelpers.MaxTestTimeout.  And also checking the return value of WaitOne, where a 'false' means it timed out.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/WebSocketHttpsResource.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using WcfService.CertificateResources;
+using WcfTestBridgeCommon;
+
+namespace WcfService.TestResources
+{
+    internal abstract class WebSocketHttpsDuplexBinaryStreamed : EndpointResource<WSDuplexService, IWSDuplexService>
+    {
```
roncain: Recommend the name of the file match the name of the class.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/WebSocketResources.cs
```diff
@@ -47,4 +42,54 @@ protected override Binding GetBinding()
             return binding;
         }
     }
+
```
roncain: Nit: I kind of prefer a file per class, though I know we already violate that practice.  Better discovery.  Not a required change, just a suggestion if you refactor things.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/WebSocketResources.cs
```diff
@@ -47,4 +42,54 @@ protected override Binding GetBinding()
...
+        protected override int GetPort(ResourceRequestContext context)
+        {
+            return context.BridgeConfiguration.BridgeWebSocketPort;
+        }
+
+        protected override string Address { get { return "WebSocketHttpDuplexBinaryStreamed"; } }
+
+        protected override Binding GetBinding()
+        {
+            int DefaultMaxReceivedMessageSize = 64 * 1024 * 1024;
```
roncain: Nit: this number keeps showing up, so maybe it's time to go into a Constants file somewhere.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/WebSocketResources.cs
```diff
@@ -47,4 +42,54 @@ protected override Binding GetBinding()
...
+    {
+        protected override string Protocol { get { return BaseAddressResource.Http; } }
+
+        protected override int GetPort(ResourceRequestContext context)
+        {
+            return context.BridgeConfiguration.BridgeWebSocketPort;
+        }
+
+        protected override string Address { get { return "WebSocketHttpRequestReplyBinaryStreamed"; } }
+
```
roncain: Suggestion only: I see several Address overrides return a literal string that matches their class name.  It might be more maintainable to return this.GetType().Name() from the base class implementation of Address.  That way, most classes would not need to override it.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WSDuplexService.cs
```diff
@@ -0,0 +1,198 @@
...
+
+namespace WcfService
+{
+    [ServiceBehavior(InstanceContextMode = InstanceContextMode.Single, ConcurrencyMode = ConcurrencyMode.Reentrant)]
+    public class WSDuplexService : IWSDuplexService
+    {
+        private string ContentToReplace = "ContentToReplace";
+        private string ResponseReplaceThisContent = "ResponseReplaceThisContent";
+        private string ReplacedContent = "ReplacedContent";
+        private string LastMessage = "LastMessage";
```
roncain: Nit: these should probably be 'const' or 'static readonly' strings.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WSDuplexService.cs
```diff
@@ -0,0 +1,198 @@
...
+        static Random rand = new Random(seed);
+
+        static FlowControlledStream localStream;
+
+        public void UploadData(string data)
+        {
+            if (data.Contains(ContentToReplace) || data.Contains(ReplacedContent) || data.Contains(ResponseReplaceThisContent))
+            {
+                log.Add(string.Format("UploadData received {0}", data));
+            }
```
roncain: Note for the future only -- I believe @shhsu was considering adding a logging layer.  When that happens, it might make sense to use that mechanism.  I also see an interesting issue here -- this service is single instance and reentrant.  I believe that implies it needs to be thread-safe when it touches the log List.  If multiple tests every hit this same endpoint at the same time, we risk fatal exceptions on the server side.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WSRequestReplyService.cs
```diff
@@ -0,0 +1,153 @@
...
+    [ServiceBehavior(InstanceContextMode = InstanceContextMode.Single)]
+    public class WSRequestReplyService : IWSRequestReplyService
+    {
+        private string ContentToReplace = "ContentToReplace";
+        private string ResponseReplaceThisContent = "ResponseReplaceThisContent";
+        private string ReplacedContent = "ReplacedContent";
+        private string RemoteEndpointMessagePropertyFailure = "RemoteEndpointMessageProperty did not contain the address of this machine.";
+
+        static List<string> log = new List<string>();
+        static int seed = DateTime.Now.Millisecond;
```
roncain: Same suggestions about 'const' or 'static readonly' strings and considering making log thread safe.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WSRequestReplyService.cs
```diff
@@ -0,0 +1,153 @@
...
+            return "Done with delay.";
+        }
+
+        public List<string> GetLog()
+        {
+            return log;
+        }
+
+        public static string CreateInterestingString(int length)
+        {
```
roncain: I see some of these constants and methods appearing in more than one place.  Can we move them to some common helper classes?

roncain:
LGTM subject to some optional cleanup in the comments.
It might be wise to wait till Monday to merge just to prevent any possibility of merge conflict or build break on the TFS machines trying to build the packages we need,

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
mconnew: Also the coding standards say when the top level of the namespace is different, there should be a blank line between them.

src/System.Private.ServiceModel/tests/Common/Scenarios/project.json
mconnew: Nit: different spacing on the last item

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -0,0 +1,234 @@
...
+using System.IO;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using Xunit;
+
+public static class WebSocketTests
+{
+    [Fact]
+    [OuterLoop]
+    [ActiveIssue(424)]
```
mconnew: If new WebSocket packages are available on Monday, verify if these tests still fail and if they don't, remove the ActiveIssue.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: Not sure what error condition would cause this and not also throw an exception. Been working fine as used in the legacy test bed.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/WebSocketHttpsResource.cs
StephenBonikowsky: I expect to add additional classes to this file all related to Web Sockets over Https, other files use this same convention otherwise we would end up with many more files.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WSDuplexService.cs
StephenBonikowsky: Added Issue #475 to do this.

PR #427, opened 2015-10-14
Allow caller to await on TimeoutHelper.CancelationToken

shhsu:
This is the first of 2 code improvement requests from the previous PR https://github.com/dotnet/wcf/pull/410. Since TimeoutHelper now create cancellation token asynchronously it make sense to allow caller to await on the creation of Cancellation token.

(Performance progression at about 1% in the Http scenario)

### DIFF
--- a/src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
+++ b/src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
@@ -8,69 +8,54 @@ using System.Threading.Tasks;
 
 namespace System.Runtime
 {
-    public struct TimeoutHelper : IDisposable
+    public struct TimeoutHelper
     {
+        public static readonly TimeSpan MaxWait = TimeSpan.FromMilliseconds(Int32.MaxValue);
+        private static readonly CancellationToken s_precancelledToken = new CancellationToken(true);
+
         private bool _cancellationTokenInitialized;
         private bool _deadlineSet;
 
         private CancellationToken _cancellationToken;
-        private CancellationTokenSource _cts;
         private DateTime _deadline;
         private TimeSpan _originalTimeout;
-        public static readonly TimeSpan MaxWait = TimeSpan.FromMilliseconds(Int32.MaxValue);
-        private static Action<object> s_cancelOnTimeout = state => ((TimeoutHelper)state)._cts.Cancel();
 
         public TimeoutHelper(TimeSpan timeout)
         {
             Contract.Assert(timeout >= TimeSpan.Zero, "timeout must be non-negative");
 
             _cancellationTokenInitialized = false;
-            _cts = null;
             _originalTimeout = timeout;
             _deadline = DateTime.MaxValue;
             _deadlineSet = (timeout == TimeSpan.MaxValue);
         }
 
-        // No locks as we expect this class to be used linearly. 
-        // If another CancellationTokenSource is created, we might have a CancellationToken outstanding
-        // that isn't cancelled if _cts.Cancel() is called. This happens only on the Abort paths, so it's not an issue. 
-        private void InitializeCancellationToken(TimeSpan timeout)
+        public CancellationToken GetCancellationToken()
         {
-            if (timeout == TimeSpan.MaxValue || timeout == Timeout.InfiniteTimeSpan)
-            {
-                _cancellationToken = CancellationToken.None;
-            }
-            else if (timeout > TimeSpan.Zero)
-            {
-                _cts = new CancellationTokenSource();
-                _cancellationToken = _cts.Token;
-                TimeoutTokenSource.FromTimeout((int)timeout.TotalMilliseconds).Register(s_cancelOnTimeout, this);
-            }
-            else
-            {
-                _cancellationToken = new CancellationToken(true);
-            }
-            _cancellationTokenInitialized = true;
+            return GetCancellationTokenAsync().Result;
         }
 
-        public CancellationToken CancellationToken
+        public async Task<CancellationToken> GetCancellationTokenAsync()
         {
-            get
+            if (!_cancellationTokenInitialized)
             {
-                if (!_cancellationTokenInitialized)
+                var timeout = RemainingTime();
+                if (timeout >= MaxWait || timeout == Timeout.InfiniteTimeSpan)
                 {
-                    InitializeCancellationToken(this.RemainingTime());
+                    _cancellationToken = CancellationToken.None;
                 }
-                return _cancellationToken;
+                else if (timeout > TimeSpan.Zero)
+                {
+                    _cancellationToken = await TimeoutTokenSource.FromTimeoutAsync((int)timeout.TotalMilliseconds);
+                }
+                else
+                {
+                    _cancellationToken = s_precancelledToken;
+                }
+                _cancellationTokenInitialized = true;
             }
-        }
 
-        public void CancelCancellationToken(bool throwOnFirstException = false)
-        {
-            if (_cts != null)
-            {
-                _cts.Cancel(throwOnFirstException);
-            }
+            return _cancellationToken;
         }
 
         public TimeSpan OriginalTimeout
@@ -194,16 +179,6 @@ namespace System.Runtime
             _deadlineSet = true;
         }
 
-        public void Dispose()
-        {
-            if (_cancellationTokenInitialized && _cts !=null)
-            {
-                _cts.Dispose();
-                _cancellationTokenInitialized = false;
-                _cancellationToken = default(CancellationToken);
-            }
-        }
-
         public static void ThrowIfNegativeArgument(TimeSpan timeout)
         {
             ThrowIfNegativeArgument(timeout, "timeout");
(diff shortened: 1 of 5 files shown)


shhsu:
@mconnew @roncain @iamjasonp

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -31,38 +43,34 @@ public TimeoutHelper(TimeSpan timeout)
...
-                _cancellationToken = _cts.Token;
-                TimeoutTokenSource.FromTimeout((int)timeout.TotalMilliseconds).Register(s_cancelOnTimeout, this);
-            }
-            else
-            {
-                _cancellationToken = new CancellationToken(true);
-            }
-            _cancellationTokenInitialized = true;
-        }
+                // No locks as we expect this class to be used linearly.
```
mconnew: Comment no longer applicable as the coalescing code will hand out the same token if called from multiple threads.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -31,38 +43,34 @@ public TimeoutHelper(TimeSpan timeout)
...
+                // If another CancellationTokenSource is created, we might have a CancellationToken outstanding
+                // that isn't cancelled if _cts.Cancel() is called. This happens only on the Abort paths, so it's not an issue.
 
-        public CancellationToken CancellationToken
-        {
-            get
-            {
-                if (!_cancellationTokenInitialized)
+                var timeout = RemainingTime();
+                if (timeout == TimeSpan.MaxValue || timeout == Timeout.InfiniteTimeSpan)
```
mconnew: Probably want to compare timeout with MaxWait too.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -31,38 +43,34 @@ public TimeoutHelper(TimeSpan timeout)
...
                 {
-                    InitializeCancellationToken(this.RemainingTime());
+                    _cancellationToken = CancellationToken.None;
                 }
-                return _cancellationToken;
+                else if (timeout > TimeSpan.Zero)
+                {
+                    _cts = new CancellationTokenSource();
```
mconnew: As discussed, don't create a CancellationTokenSource every time as it's only needed by HttpChannelFactory.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CoreClrClientWebSocketFactory.cs
```diff
@@ -24,7 +24,8 @@ class CoreClrClientWebSocketFactory : ClientWebSocketFactory
                 webSocket.Options.SetRequestHeader(header, headers[header]);
             }
 
-            await webSocket.ConnectAsync(address, timeoutHelper.CancellationToken);
+            var cancelToken = await timeoutHelper.GetCancellationTokenAsync();
+            await webSocket.ConnectAsync(address, cancelToken);
```
mconnew: Can fetch the token inline. E.g.
```csharp
            await webSocket.ConnectAsync(address, await timeoutHelper.GetCancellationTokenAsync());
```

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -752,7 +754,8 @@ public void Fault(RequestChannel channel)
                     }
                     catch (OperationCanceledException)
                     {
-                        if (_timeoutHelper.CancellationToken.IsCancellationRequested)
+                        var cancelToken = _timeoutHelper.GetCancellationTokenAsync().Result;
```
mconnew: We should have a synchronous version. Calling .Result will wrap any exceptions in an aggregate exception so you need to use .GetAwaiter().GetResult() instead. That's too clumsy and specific to expect everybody using it to remember to do it correctly, so instead make the method TimeoutHelper.GetCancellationToken();

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportDuplexSessionChannel.cs
```diff
@@ -222,7 +222,7 @@ protected override AsyncCompletionResult StartWritingBufferedMessage(Message mes
                     RemoteAddress != null ? RemoteAddress.ToString() : string.Empty);
             }
 
-            Task task = WebSocket.SendAsync(messageData, outgoingMessageType, true, helper.CancellationToken);
+            Task task = WebSocket.SendAsync(messageData, outgoingMessageType, true, helper.GetCancellationTokenAsync().Result);
```
mconnew: Create Synchronous method

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportDuplexSessionChannel.cs
```diff
@@ -1051,11 +1051,12 @@ public override void Flush()
             {
             }
 
-            public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+            public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
```
mconnew: Don't convert any non-async methods to async as that will cause an extra Task allocation on every call,

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Fixed (copy oversight)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CoreClrClientWebSocketFactory.cs
shhsu: I try to only do as little per line as possible for clarity as easier debugging

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
shhsu: Fixed.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportDuplexSessionChannel.cs
shhsu: Fixed

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportDuplexSessionChannel.cs
shhsu: Fixed.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: The commit mentioned this is just step 1 of 2, I was going to do this for the next step but I will just include it in this commit then.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Changed to >= MaxWait

shhsu:
Ping

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -294,12 +280,20 @@ public static Task<CancellationToken> FromTimeoutAsync(int millisecondsTimeout)
...
                     token.Register(t => {
                         Task<CancellationToken> ignored;
-                        s_tokenCache.TryRemove((long)t, out ignored);
+                        try
+                        {
+                            s_tokenCache.TryRemove((long)t, out ignored);
+                        }
+                        finally
+                        {
+                            tokenSource.Dispose();
```
mconnew: This will cause a closure to be created as you are capturing the `tokenSource` local variable. After the timer has fired (which will have happened when you're getting into this code block), is there any work needed to be done by `Dispose()`? I believe the timer is a fire once timer so it will no longer be registered and there's nothing left in the timer to clean up on Disposal. If there is some work which is beneficial to be done here, one option would be to change the type of `s_tokenCache`  to `ConcurrentDictionary<long, Task<CancellationTokenSource>>`. This would make the type of ignore to be `Task<CancellationTokenSource>` and it would no longer be ignored :smile:

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -8,69 +8,65 @@
 
 namespace System.Runtime
 {
-    public struct TimeoutHelper : IDisposable
+    public struct TimeoutHelper
     {
+        public static readonly TimeSpan MaxWait = TimeSpan.FromMilliseconds(Int32.MaxValue);
+        private static readonly Action<object> s_cancelOnTimeout = state =>
```
mconnew: I don't think this is used any more is it?

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
```diff
@@ -294,12 +280,20 @@ public static Task<CancellationToken> FromTimeoutAsync(int millisecondsTimeout)
                 {
                     // Since this thread was successful reserving a spot in the cache, it would be the only thread
                     // that construct the CancellationTokenSource
-                    var token = new CancellationTokenSource((int)(targetTime - currentTime)).Token;
+                    var tokenSource = new CancellationTokenSource((int)(targetTime - currentTime));
+                    var token = tokenSource.Token;
 
                     // Clean up cache when Token is canceled
                     token.Register(t => {
                         Task<CancellationToken> ignored;
```
mconnew: It might be work make this delegate a static to avoid accidental closure allocation.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TimeoutHelper.cs
shhsu: Gone

PR #432, opened 2015-10-16
Fix commit message for  "[tfs-changeset: 1538408]", PR #429

iamjasonp:
Due to infrastructure and package building reasons, we had to commit #411 via TFS and let it mirror to Git. Unfortunately, the checkin message was omitted when it was checked in via TFS - naturally the commit message in Git was blank too and the "merge" button was clicked before we could make it right.

This PR: 
* Reverts 8e06f18
* Re-adds the same change with a new commit message


iamjasonp:
Committing this now so that we can merge other PRs as well - we want to keep the commits close in time.

PR #433, opened 2015-10-16
Dispose websocket object after Close

### LINKED ISSUE #431: Websocket stress test run shows significant memory leak
We use WebSocket.CloseAync to Close the WebSocket. We does not dispose it. We need to aggressively dispose it as WebSocket is a disposable object and hold both managed and unmanaged resources.

I have verified the memory leak is gone by calling Cleanup() after CloseAync completes.

See more details below.
From ClientWebSocket  Class, we can see Dispose method is responsible to release a number of resources including innerWebSocket which holds unmanaged resources.

public override void Dispose() 
        { 
            var priorState = (InternalState)Interlocked.Exchange(ref _state, (int)InternalState.Disposed); 
            if (priorState == InternalState.Disposed) 
            { 
                // No cleanup required. 
                return; 
            } 
            _cts.Cancel(false); 
            _cts.Dispose(); 
            if (_innerWebSocket != null) 
            { 
                _innerWebSocket.Dispose(); 
            } 
        }

hongdai:
* We use WebSocket.CloseAsync to close the WebSocket object. We do not dispose it
in normal Close case. We need to aggressively dispose it as WebSocket
is a disposable object and hold unmanaged resources.

Fixes #431

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportDuplexSessionChannel.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportDuplexSessionChannel.cs
@@ -202,6 +202,18 @@ namespace System.ServiceModel.Channels
             }
         }
 
+        protected internal override async Task OnCloseAsync(TimeSpan timeout)
+        {
+            try
+            {
+                await base.OnCloseAsync(timeout);
+            }
+            finally
+            {
+                Cleanup();
+            }
+        }
+
         protected override void ReturnConnectionIfNecessary(bool abort, TimeSpan timeout)
         {
         }


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportDuplexSessionChannel.cs
```diff
@@ -134,6 +134,8 @@ protected override void CompleteClose(TimeSpan timeout)
             {
                 TD.WebSocketConnectionClosed(WebSocket.GetHashCode());
             }
+
+            Cleanup();
```
mconnew: I've looked into this a little more and I don't think this is the right fix. In the OnClose method, we call Cleanup() in a finally block. I think the problem is I didn't override the OnCloseAsync method which should call the Cleanup method. This means the base class  TransportDuplexSessionChannel.OnCloseAsync() is being called and we're missing the Cleanup() call.

hongdai: Agree override OnCloseAsync is a better fix. I will make the change.

mconnew:
LGTM

PR #443, opened 2015-10-20
Update run-test.sh to overwrite minimal CoreFx binaries

roncain:
Prior to this change, run-test.sh overwrote all the binaries
from the CoreFx packages with those built in the CoreFx repo.

After this change, it relies mostly on the binaries from the
packages except for select list of binaries that are not correct
in their package.

### DIFF
--- a/run-test.sh
+++ b/run-test.sh
@@ -15,19 +15,20 @@ wait_on_pids()
 
 usage()
 {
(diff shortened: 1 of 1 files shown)


roncain:
Merging because I need these changes to retest latest bits.  I'll react to PR review comments retroactively as needed

Issue #407, opened 2015-10-09
Running x-machine tests should not require admin privileges on client side


iamjasonp:
Currently, when running cross machine tests, the BridgeClient needs admin privileges in order to install the trusted root cert into the LOCAL MACHINE\Trusted Root Authorities store. 

However, this causes exceptions to be thrown if the launching prompt isn't already in elevated. The exceptions are fairly clear in what they say, but we'd like to avoid them. 

Investigate whether or not it would be okay to instead install into the Current User store. I haven't yet tried to see whether or not that requires privileged access though. Alternatively, we might want to catch these exceptions and fail fast for these tests when we aren't able to write to the necessary store. 

This has impact on the cleanup side too - #365

iamjasonp:
Investigated today - I think that all we need on the BridgeClient is to install certs into the CurrentUser store; the Bridge (server) needs to install into LocalMachine though.

iamjasonp:
... Grumble. The problem is that if we want to install root certs, there is a UI prompt that pops up asking if you want to install the root cert, and there doesn't seem to be a way to bypass this.

iamjasonp:
We may have to let this one slide x-machine :frowning: We may want to tag tests that require elevated permissions to prevent them from running. 

There's still an issue to be fixed here, as testing TCP on localhost needs elevation with the current BridgeClient after #403 (we have no tests running that hit this right now, but the tcp client credential tests will hit this)

We should probably still get to a state where if testing locally on a machine that is already running the Bridge, we shouldn't require admin - this helps with local testing.

roncain:
I don't understand why #403 changed the situation but agree we should prioritize this lower than RC1 work.  In the meantime, do we know whether ToF or CI machines run elevated?  Will they break due to a UI appearing?

iamjasonp:
Prior to #403: 
- The BridgeClient would look up its own hostname and send that to the Bridge as part of a certificate creation request
- If the hostname reported by BridgeClient was the same as the one as the Bridge, then the BridgeClient will not attempt to install the certs. This is because the Bridge already installed the correct certs. 
- else, retrieve the cert from the bridge and install

After #403: 
- The BridgeClient could no longer look up its own hostname as we needed to remove the dependency from System.Net.NetworkInformation. 
- Hence, we don't report this information to the Bridge
- We now simply generate a user cert titled "WCF Client Certificate"
[...]

PR #446, opened 2015-10-21
Add platform-dependent logic to BridgeClientCertificateManager

### LINKED ISSUE #434: X509Store throws NotImplemented for read/Write on Linux
Https_ClientCredentialTypeTests.BasicAuthenticationInvalidPwd_throw_MessageSecurityException [FAIL]
      Assert.Throws() Failure
      Expected: typeof(System.ServiceModel.Security.MessageSecurityException)
      Actual:   typeof(System.NotImplementedException): The method or operation is not implemented.
      Stack Trace:
            at Internal.Cryptography.Pal.StorePal.FromSystemStore(String storeName, StoreLocation storeLocation, OpenFlags openFlags)
            at System.Security.Cryptography.X509Certificates.X509Store.Open(OpenFlags flags)
            at Infrastructure.Common.BridgeClientCertificateManager.AddToStoreIfNeeded(StoreName storeName, StoreLocation storeLocation, X509Certificate2 certificate)
            at Infrastructure.Common.BridgeClientCertificateManager.InstallCertificateToRootStore(X509Certificate2 certificate)
            at Infrastructure.Common.BridgeClientCertificateManager.InstallRootCertificateFromBridge()
            at Endpoints.get_Https_BasicAuth_Address()
            at Https_ClientCredentialTypeTests.<>c.<BasicAuthenticationInvalidPwd_throw_MessageSecurityException>b__4_0()

iamjasonp:
In Windows, we read/write our certificates from/to the LocalMachine store.
On *nix, however, we only have access to the User store. The BridgeClient needs
to be aware of where it's running and switch the cert store used as necessary

Fixes #434

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
@@ -36,6 +36,51 @@ namespace Infrastructure.Common
         private static Dictionary<string, CertificateCacheEntry> s_myCertificates = new Dictionary<string, CertificateCacheEntry>(StringComparer.OrdinalIgnoreCase);
         private static Dictionary<string, CertificateCacheEntry> s_rootCertificates = new Dictionary<string, CertificateCacheEntry>(StringComparer.OrdinalIgnoreCase);
 
+        private static bool s_PlatformSpecificStoreLocationIsSet = false;
+        private static StoreLocation s_PlatformSpecificRootStoreLocation = StoreLocation.LocalMachine;
+
+        // The location from where to open StoreName.Root
+        //
+        // Since we don't have access to System.Runtime.InteropServices.IsOSPlatform API, we need to find a novel way to switch
+        // the store we use on Linux
+        //
+        // For Linux: 
+        // On Linux, opening stores from the LocalMachine store is not supported yet, so we toggle to use CurrentUser
+        // Furthermore, we don't need to sudo in Linux to install a StoreName.Root : StoreLocation.CurrentUser. So this will allow
+        // tests requiring certs to pass in Linux. 
+        // See dotnet/corefx#3690
+        //
+        // For Windows: 
+        // We don't want to use CurrentUser, as writing to StoreName.Root : StoreLocation.CurrentUser will result in a 
+        // modal dialog box that isn't dismissable if the root cert hasn't already been installed previously. 
+        // If the cert has already been installed, writing to StoreName.Root : StoreLocation.CurrentUser results in a no-op
+        // 
+        // In other words, on Windows, we can bypass the modal dialog box, but only if we install to StoreName.Root : StoreLocation.LocalMachine
+        // To do this though means that we must run certificate-based tests elevated
+        private static StoreLocation PlatformSpecificRootStoreLocation
+        {
+            get
+            {
+                if (!s_PlatformSpecificStoreLocationIsSet)
+                {
+                    try
+                    {
+                        using (var store = new X509Store(StoreName.Root, StoreLocation.LocalMachine))
+                        {
+                            store.Open(OpenFlags.ReadWrite);
+                        }
+                    }
+                    catch (PlatformNotSupportedException)
+                    {
+                        // Linux
+                        s_PlatformSpecificRootStoreLocation = StoreLocation.CurrentUser; 
+                    }
+                    s_PlatformSpecificStoreLocationIsSet = true;
+                }
+                return s_PlatformSpecificRootStoreLocation;
+            }
+        }
+
         // Installs the root certificate from the bridge 
         // Needed for all tests so we trust the incoming certificate coming from the service/bridge
         // We do our best to detect if we're running on the same box as the bridge; if so, we don't try to install the cert
@@ -63,7 +108,8 @@ namespace Infrastructure.Common
                 }
                 else
                 {
-                    using (X509Store store = new X509Store(StoreName.Root, StoreLocation.LocalMachine))
+                    // See explanation of StoreLocation selection at PlatformSpecificRootStoreLocation
+                    using (X509Store store = new X509Store(StoreName.Root, PlatformSpecificRootStoreLocation))
                     {
                         store.Open(OpenFlags.ReadOnly);
                         var collection = store.Certificates.Find(X509FindType.FindByThumbprint, thumbprint, false);
@@ -254,7 +300,8 @@ namespace Infrastructure.Common
                     return entry.Thumbprint;
                 }
 
-                bool added = AddToStoreIfNeeded(StoreName.Root, StoreLocation.LocalMachine, certificate);
+                // See explanation of StoreLocation selection at PlatformSpecificRootStoreLocation
+                bool added = AddToStoreIfNeeded(StoreName.Root, PlatformSpecificRootStoreLocation, certificate);
                 s_rootCertificates[certificate.Thumbprint] = new CertificateCacheEntry
                 {
                     Thumbprint = certificate.Thumbprint,
@@ -278,7 +325,10 @@ namespace Infrastructure.Common
                     return entry.Thumbprint;
                 }
 
-                bool added = AddToStoreIfNeeded(StoreName.My, StoreLocation.LocalMachine, certificate);
+                // Always install client certs to CurrentUser
+                // StoreLocation.CurrentUser is supported on both Linux and Windows 
+                // Furthermore, installing this cert to this location does not require sudo or admin elevation
+                bool added = AddToStoreIfNeeded(StoreName.My, StoreLocation.CurrentUser, certificate);
                 s_myCertificates[certificate.Thumbprint] = new CertificateCacheEntry
                 {
                     Thumbprint = certificate.Thumbprint,
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
@@ -156,7 +156,7 @@ public static class Tcp_ClientCredentialTypeTests
             factory = new ChannelFactory<IWcfService>(binding, endpointAddress);
             factory.Credentials.ServiceCertificate.Authentication.CertificateValidationMode = X509CertificateValidationMode.None;
             factory.Credentials.ClientCertificate.SetCertificate(
-                StoreLocation.LocalMachine,
+                StoreLocation.CurrentUser,
                 StoreName.My,
                 X509FindType.FindByThumbprint,
                 clientCertThumb);


iamjasonp:

iamjasonp:

iamjasonp:

iamjasonp:

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -32,6 +34,20 @@ public static class BridgeClientCertificateManager
         private static Dictionary<string, CertificateCacheEntry> s_myCertificates = new Dictionary<string, CertificateCacheEntry>(StringComparer.OrdinalIgnoreCase);
         private static Dictionary<string, CertificateCacheEntry> s_rootCertificates = new Dictionary<string, CertificateCacheEntry>(StringComparer.OrdinalIgnoreCase);
 
+        private static readonly StoreLocation PlatformSpecificStoreLocation; 
+
+        static BridgeClientCertificateManager()
+        {
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
```
roncain: My API Catalog shows RuntimeInformation is not available on NET Native.  Have you tried building in TFS as NET Native yet?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -199,11 +215,18 @@ private static X509Certificate2 CertificateFromThumbprint(X509Store store, strin
...
                 try
                 {
+                    // We open ReadOnly first to prevent elevation in Windows when running as a non-admin
+                    // Non-admins can read from the LocalMachine store but not write to the store
+
                     store = new X509Store(storeName, storeLocation);
-                    store.Open(OpenFlags.ReadWrite);
+                    store.Open(OpenFlags.ReadOnly);
                     existingCert = CertificateFromThumbprint(store, certificate.Thumbprint);
+                    store.Dispose();
```
roncain: Recommend 'using' statement instead of explicit Dispose (in case of failure before the Dispose).

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
iamjasonp: Yep. pwned. I was investigating what else I could do for this...

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -199,11 +215,18 @@ private static X509Certificate2 CertificateFromThumbprint(X509Store store, strin
...
+
                     if (existingCert == null)
                     {
+                        store = new X509Store(storeName, storeLocation);
+                        store.Open(OpenFlags.ReadWrite);
                         store.Add(certificate);
```
roncain: Does this use of store require a 'using' to dispose it?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/project.json
```diff
@@ -1,5 +1,6 @@
 {
   "dependencies": {
+    "System.Runtime.InteropServices.RuntimeInformation" :  "4.0.0-beta-*",
     "System.ServiceModel.Duplex": "4.0.1-beta-*",
```
roncain: Please double check -- if this package does not exist in NET Native, I believe this dependency could break.  If add #if code to make use of RuntimeInformation in CoreCLR only, would need to move the position of this dependency.

roncain:
I saw a failure in OuterLoop.  Do you think it was for this PR?

iamjasonp:
Https_ClientCredentialTypeTests.BasicAuthentication_RoundTrips_Echo is the failing test in the outerloops that I see - that test is pretty flaky for me.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/project.json
iamjasonp: No, it's not there. We'd have to conditionally compile - and this project doesn't do any checking for FEATURE_CORECLR or the like. We'd also run into the same issue we had when dealing with System.Net.DnsInformation in that the ToF runtime would have a problem distinguishing when to include the reference and when not to. 

Best to find a way to avoid this entirely...

roncain:
That test is flakey for me too.  It always succeeds the 2nd time I try to run it.  There is either some timing issue or another test is messing with shared resources at the same time.

iamjasonp:
@roncain - I took another look at this today, and I really haven't found a good way to key off the build platform without depending on build time dependencies. Thoughts on that?

roncain:
@iamjasonp -- what happens if we don't distinguish platforms and always use CurrentUser?  On my quick and dirty tests using the Bridge running on the same machine, the tests passed when I unconditionally used CurrentUser. If it is mandatory we distinguish, I see no choice but to add conditional compilation for *NIX and FEATURE_CORECLR

iamjasonp:
Did some additional thinking on this: 

* If we run all on CurrentUser, when running x-machine, we have an interactive prompt that pops up when trying to install the root authority cert to the CurrentUser Root Certification Authorities store. I believe this happens even if running elevated. (needs verification) 

* If we run all on LocalMachine, we basically need to elevate on the BridgeClient; we currently have a bunch of workarounds in place on the Bridge side to prevent this from happening. When running x-machine, we must elevate. 

The annoying thing is that we MUST touch the LocalMachine store. 

In the first scenario, if testing interactively, this isn't a problem. In CI - we don't have a way to acknowledge the dialog, and I have a hunch that this will break OuterLoop testing in CI. 

If we're okay with something like this

    try
    {
        // try reading from the LocalMachine store
    } 
        catch (NotImplementedException)
    {
        // try reading from the CurrentUser store
    }

then maybe we can avoid the runtime/compilation check. It's ugly as sin, but it might work. I am however in the business of preventing the addition of dependencies that exist only on one platform. 

My priorities when solving this problem: 

* don't break CI testing
* allow x-machine testing in Windows (elevated) 
* prevent interactive dialog boxes when running non-interactively
* allow Linux testing
* allow x-machine testing in Windows (without elevation)

iamjasonp:

iamjasonp:
@roncain, can you please buddy build and test with Linux [OuterLoop]?

iamjasonp:
For posterity: I hate this workaround, but there's no good way to deal with this. See https://github.com/dotnet/wcf/issues/407 for a further explanation...

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -36,6 +36,50 @@ public static class BridgeClientCertificateManager
...
+
+        // The location from where to open StoreName.Root
+        //
+        // Since we don't have access to System.Runtime.InteropServices.IsOSPlatform API, we need to find a novel way to switch
+        // the store we use on Linux
+        //
+        // For Linux: 
+        // On Linux, opening stores from the LocalMachine store is not supported yet, so we toggle to use CurrentUser
+        // Furthermore, we don't need to sudo in Linux to install a StoreName.Root : StoreLocation.CurrentUser. So this will allow
+        // tests requiring certs to pass in Linux.
```
roncain: Optional: include link to https://github.com/dotnet/corefx/issues/3690

roncain:
@iamjasonp -- testing in Linux.  Be aware we need to merge https://github.com/dotnet/wcf/pull/481 before this becomes easy (else I have to manually combine the PR's).  If you can buddy-build #481 on TFS, it would help get that one merged.

roncain:
I instrumented this and ran it in Linux, but if fails because it still tries to open LocalMachine read/write.
Here is console output (BCCM == BridgeClientCertificateManager).  Because it succeeded in opening the store readonly to determine which store location to use, it believe is can uses LocalMachine.  Should that attempt have been a read/write open?

BCCM: choosing store location...
  BCCM: new X509Store...
  BCCM: store.Open...
BCCM: store location is LocalMachine
BCCM: store location is LocalMachine
AddToStoreIfNeeded -- creating new store name = Root, loc = LocalMachine
  opening readonly...
  opened readonly...
  found cert
AddToStoreIfNeeded -- cert is null, creating new store name = Root, loc = LocalMachine
  opening readwrite...
AddToStore unexpected exception System.PlatformNotSupportedException: Unix LocalMachine X509Stores are read-only for all users.
   at Internal.Cryptography.Pal.StorePal.FromSystemStore(String storeName, StoreLocation storeLocation, OpenFlags openFlags)
   at System.Security.Cryptography.X509Certificates.X509Store.Open(OpenFlags flags)
   at Infrastructure.Common.BridgeClientCertificateManager.AddToStoreIfNeeded(StoreName storeName, StoreLocation storeLocation, X509Certificate2 certificate)
  adding...
AddToStoreIfNeeded -- creating new store name = My, loc = CurrentUser
  opening readonly...
  opened readonly...
  found cert
AddToStoreIfNeeded -- cert is null, creating new store name = My, loc = CurrentUser
  opening readwrite...
  opened readwrite...
  adding...
  added...

iamjasonp:
#481 buddy built for me in TFS, but tests are refusing to run. 

And argh. I wasn't aware that Linux LocalMachine stores were read-only for all users; I misread and thought that Linux LocalMachine stores were not available at all. 

This leaves us in a bit of a pickle, because I was hoping to use this as a distinction point between: 

* Windows + non-elevated vs.
* Linux 

In this code, if we open read-write in Windows + non-elevated, then the code will switch to using a CurrentUser context, which means that when we try to install the root certificate authorities, we'll get those modal pop-ups I've been trying so hard to avoid. 

The worst part is that the modal dialog box popping up doesn't fail the test - it just hangs there until the test times out.

roncain:
Just confirming that modifying this to attempt a READWRITE open allows it to use CurrentUser, and we get past the BridgeClientCM issues (as long as you catch PNSE in the store test).  I understand the pickle.  But what alternative is there?  If we run multi-machine on Windows, we *must* add that cert, correct?  And get the pop-up if not elevated?  Ignoring Linux for the moment and talking only Windows-->Windows multi-machine, is there a solution that avoids the dialog if non-elevated?

We need to move forward, so I suggest refining this test to use READWRITE, catching PNSE (as opposed to NotImplementedException) and living with the need to run elevated.  Cert testing on Linux is blocked till we get past this.

iamjasonp:
Per https://github.com/dotnet/wcf/issues/407#issuecomment-152207912: I declare that when we test on Windows x-machine, we must elevate, on pain of having to click through many dialog boxes. 

(actually it's not that bad. Someone needs to click "yes" once, and the cert gets installed, and all tests in that run will continue. However, if not elevated, we block interactive testing)

> We need to move forward, so I suggest refining this test to use READWRITE, catching PNSE (as opposed to NotImplementedException) and living with the need to run elevated. Cert testing on Linux is blocked till we get past this.

Done. PR updated.

iamjasonp:

roncain:
LGTM.  I ran this on Linux, and all failures are now down in product-level certificate validation.  I think you should go ahead and merge.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
iamjasonp: Good suggestion - added.

iamjasonp:
Sure - let's open a new issue for the Linux product-level issues.

PR #450, opened 2015-10-22
Move reference packages down a directory level so nuget restore pick up older versions

mconnew:
Latest version of nuget doesn't traverse any more directories if it finds a project.json
file in a folder. This means src\System.ServiceModel.Http\ref\project.json was stopping
src\System.ServiceModel.Http\ref\3.9.0\project.json from being found.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/project.lock.json
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/project.lock.json
@@ -329,10 +329,10 @@
           "ref/dotnet5.4/System.Net.Security.dll": {}
         }
       },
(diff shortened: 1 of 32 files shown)


mconnew:
@ericstj Can you validate that this approach won't break other things?

ericstj:
This will break packages and partial facade projects.

mconnew:
Any suggestions for a different approach? Without this change, WCF doesn't even successfully build.

ericstj:
Use an item group (**\project.json) to grab all the project.json's and pass each one to dnu in series instead of passing the root.  It'll be slow but it's a small change to limp along until we can make a better fix.

PR #451, opened 2015-10-22
Add rule to fetch packages for older reference contracts to work around nuget changes

mconnew:


iamjasonp:
buddy built successful, inner and outerloop tests pass. 
I don't know what the implications of this would be though.

mconnew:
All it does is add a couple of extra project.json files to the dnu restore command for restoring packages which are now being missed because of the upgrade to a newer version of dnu.

iamjasonp:
I think this looks fine for us to work around this current issues, but please open an issue to back out this change later. 

LGTM

PR #455, opened 2015-10-22
Update sockets

mconnew:
This removes our dependency on Begin/End connect. Updates all our dependencies to the latest that customers would be using. Also fixes test runs to use the correct packages.


mconnew:

shmao:
LGTM. Thanks for making the changes.

roncain:
To answer @shmao the change to the sync connect is to allow us to work against both old and new S.N.Sockets in the event we can't get the package in a timely manner.  It will be made ConnectAsync when S.N.Sockets is available and stable, possibly post RC1.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CoreClrSocketConnection.cs
```diff
@@ -892,15 +892,15 @@ protected override IConnection CreateConnection(IPAddress address, int port)
...
         {
             Socket socket = null;
             try
             {
                 AddressFamily addressFamily = address.AddressFamily;
                 socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
-                await Task.Factory.FromAsync<EndPoint>(socket.BeginConnect, socket.EndConnect, new IPEndPoint(address, port), null);
-                return new CoreClrSocketConnection(socket, _connectionBufferPool);
+                socket.Connect(new IPEndPoint(address, port));
+                return Task.FromResult<IConnection>(new CoreClrSocketConnection(socket, _connectionBufferPool));
```
roncain: Nit: not a blocking issue, but we should open an issue once this is checked in to come back and make it async.  Much lower pri than getting this merged.

roncain:
I am merging this PR now to unblock further testing and enable the async PR.  The TFS mirror is currently disabled, and I will manually verify this works in TFS once the ProjectK builds are stable again.  This merge will become part of RC1 only after clearance from shiproom

PR #423, opened 2015-10-14
Separate User and Machine Certificate Resources in Bridge

iamjasonp:
Currently we do not differentiate between machine and user certificates in the MachineCertificateResource. The certs are slightly different, and will fail validation when it comes time to validate a cert (at least on a Windows box)

In order to properly generate user certificates with User Principal Names, we need to add some different Subject Alt Name X509 extensions to the certificate compared to a cert for the machine name

NOTE: We fudged the "user" cert when fixing #403 due to time constraints in getting it right, but it didn't affect our existing scenarios, so we let it slide. Now's the time to fix it as we want to turn on the tcp client credential tests soon.

PR #456, opened 2015-10-23
Return certificates with private key from Bridge

iamjasonp:
* Machine and user certs returned from the Bridge previously did not contain a private key, making some scenarios fail.
* BridgeClientCertificateManager will request User rather than Machine certificates
* BridgeClientCertificateManager will now check whether or not certificates were previously installed by opening the X509Store ReadOnly before attempting to install them
* BridgeClientCertificateManager error message for when an CryptographicException "Access Denied" is thrown improved
* Change the Bridge to install all "user" certs generated (not machine)
   We need this so that localhost tests don't require elevation

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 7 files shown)


iamjasonp:

iamjasonp:
The PR is a little big because there was some fallout from #423 and #403 that needed to be addressed, mainly in running tests locally with client certs in a non-elevated prompt.

hongdai:
LGTM.

Issue #368, opened 2015-09-28
Implement NetTcpBinding Security.Transport negative E2E scenarios a)


hongdai:
This issue tracks implementation of the following test
NetTcpBinding Security.Transport negative E2E scenarios
 a) Service Identity does not match the identify specified by DnsEndpointIdentity

 in test plan #322 .

zhenlan:
This is currently blocked by issue #308 which @iamjasonp is working on.

PR #457, opened 2015-10-23
Update Identity positive test and add a negative test

hongdai:
* Update identity positve test per cert generation changes in bridge.
* Enable the positve test because it's now pass as both product and cert
functionalities are in place.
* Add a negative test.

Fixes issue #368

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
@@ -13,10 +13,12 @@ namespace Infrastructure.Common
     public static class BridgeClient 
     {
         private const string EndpointResourceResponseUriKeyName = "uri";
+        private const string EndpointResourceResponseFQHNKeyName = "hostname";
         private const string EndpointResourceRequestNameKeyName = "name";
 
         private static object s_thisLock = new object();
-        private static Dictionary<string, string> _Resources = new Dictionary<string, string>();
+        //The key is resource name, The value are resource properties such as EndpointResourceResponseUriKeyName
+        private static Dictionary<string, Dictionary<string, string>> _Resources = new Dictionary<string, Dictionary<string, string>>();
         private static BridgeState s_BridgeStatus = BridgeState.NotStarted;
         private static string s_BridgeFaultReason = String.Empty;
 
@@ -83,14 +85,45 @@ namespace Infrastructure.Common
             string resourceAddress = null;
             lock (s_thisLock)
             {
-                if (s_BridgeStatus == BridgeState.Started && !_Resources.TryGetValue(resourceName, out resourceAddress))
+                Dictionary<string, string> resources = null;
+                if (s_BridgeStatus == BridgeState.Started)
                 {
-                    resourceAddress = MakeEndpointResourcePutRequest(resourceName);
-                    _Resources.Add(resourceName, resourceAddress);
+                    if (!_Resources.TryGetValue(resourceName, out resources))
+                    {
+                        resources = MakeEndpointResourcePutRequest(resourceName);
+                      
+                        _Resources.Add(resourceName, resources);
+                    }
+                  
+                    resources.TryGetValue(EndpointResourceResponseUriKeyName, out resourceAddress);
                 }
             }
 
             return resourceAddress;
+        }
+
+        public static string GetResourceHostName(string resourceName)
+        {
+            EnsureBridgeIsRunning();
+
+            string hostName = null;
+            lock (s_thisLock)
+            {
+                Dictionary<string, string> resources = null;
+                if (s_BridgeStatus == BridgeState.Started)
+                {
+                    if (!_Resources.TryGetValue(resourceName, out resources))
+                    {
+                        resources = MakeEndpointResourcePutRequest(resourceName);
+                        _Resources.Add(resourceName, resources);
+                    }
+
+                    resources.TryGetValue(EndpointResourceResponseFQHNKeyName, out hostName);
+                    
+                }
+            }
+
+            return hostName;
         }
 
         // Ensure the given response was successful. If it was not, generate
@@ -192,7 +225,7 @@ namespace Infrastructure.Common
             return sb.ToString();
         }
 
-        private static string MakeEndpointResourcePutRequest(string resourceName)
+        private static Dictionary<string, string> MakeEndpointResourcePutRequest(string resourceName)
         {
             var responseContent = MakeResourcePutRequest(resourceName, null);
 
@@ -202,7 +235,7 @@ namespace Infrastructure.Common
                 throw new Exception("Invalid response from bridge: " + FormatKeyValuePairsAsString(responseContent));
             }
 
-            return uri;
+            return responseContent;
         }
 
         internal static Dictionary<string,string> MakeResourcePutRequest(string resourceName, Dictionary<string, string> requestParameters)
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
@@ -176,10 +176,13 @@ namespace Infrastructure.Common
                 }
             }
 
-            InstallCertificateToMyStore(certificateToInstall);
-
-            // We also need to install the root cert if we install a local cert
-            InstallRootCertificateFromBridge();
+            // certificateToInstall could be null in the case the user certification exists
+            if (certificateToInstall != null)
+            {
+                InstallCertificateToMyStore(certificateToInstall);
+                // We also need to install the root cert if we install a local cert
+                InstallRootCertificateFromBridge();
+            }
         }
 
         // Returns the certificate matching the given thumbprint from the given store.
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
@@ -145,6 +145,14 @@ public static partial class Endpoints
         }
     }
 
+    public static string Tcp_VerifyDNS_HostName
+    {
+        get
+        {
+            return BridgeClient.GetResourceHostName("WcfService.TestResources.TcpVerifyDNSResource");
+        }
+    }
+
     public static string Tcp_NoSecurity_Callback_Address
     {
         get { return BridgeClient.GetResourceAddress("WcfService.TestResources.DuplexResource"); }
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Security.TransportSecurity.Tests.csproj
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Security.TransportSecurity.Tests.csproj
@@ -39,6 +39,10 @@
             <Project>{9098B41C-9C2E-4FC6-B7D8-FC3411736A22}</Project>
             <Name>ScenarioTests.Common</Name>
         </ProjectReference>
+    	<ProjectReference Include="$(WcfInfrastructureCommonProj)">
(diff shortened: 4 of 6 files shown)


hongdai:

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
```diff
@@ -1,14 +1,17 @@
...
+using System.Collections;
 using System.ServiceModel;
+using System.ServiceModel.Security;
 using System.Text;
 using Xunit;
 
-namespace Security.TransportSecurity.Tests.Tcp
+using Infrastructure.Common;
+
+namespace Tcp_IdentityTests
```
hongdai: I will remove the namespace as it will mess up the test result report.

hongdai:

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
```diff
@@ -83,16 +84,51 @@ public static string GetResourceAddress(string resourceName)
...
+                Dictionary<string, string> resources = null;
+                if (s_BridgeStatus == BridgeState.Started)
                 {
-                    resourceAddress = MakeEndpointResourcePutRequest(resourceName);
-                    _Resources.Add(resourceName, resourceAddress);
+                    if (!_Resources.TryGetValue(resourceName, out resources))
+                    {
+                        resources = MakeEndpointResourcePutRequest(resourceName);
+                        resources.TryGetValue(EndpointResourceResponseUriKeyName, out resourceAddress);
+                        _Resources.Add(resourceName, resources);
```
roncain: Nit: do this resource.TryGetValue just in one place, after the 'if' block.  Don't need the 'else' block.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -176,10 +176,13 @@ public static void InstallLocalCertificateFromBridge()
                 }
             }
 
-            InstallCertificateToMyStore(certificateToInstall);
-
-            // We also need to install the root cert if we install a local cert
-            InstallRootCertificateFromBridge();
+            // certificateToInstall could be null if local cert has been installed in the previous tests
+            if (certificateToInstall != null)
+            {
```
roncain: This comment and test makes me nervous.  Tests can run in parallel.  Can more than one set this value?  What happens if 2 tests run concurrently and change this value?  Can one test be in the middle of this != null test and have another test change it?  If so, we need to make this thread safe with a lock, or we need tests to pass in this information as they need it.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Security.TransportSecurity.Tests.csproj
```diff
@@ -39,6 +39,10 @@
             <Project>{9098B41C-9C2E-4FC6-B7D8-FC3411736A22}</Project>
             <Name>ScenarioTests.Common</Name>
         </ProjectReference>
+    	<ProjectReference Include="$(WcfInfrastructureCommonProj)">
+      	    <Project>{AFD69665-89A3-42AE-A32E-AB2CBBE6EE7E}</Project>
+            <Name>Infrastructure.Common</Name>
+        </ProjectReference>
```
roncain: This new project reference is a hint the ToF wrapper project living in TFS will also need to be updated to do the same thing.  That would require a TFS checkin separate from this one, and it could probably be done first to without harm.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
```diff
@@ -1,43 +1,75 @@
...
+        IWcfService serviceProxy = factory.CreateChannel();
+
+        try
         {
-            string testString = "Hello";
+            var result = serviceProxy.Echo(testString);
+            Assert.Equal(testString, result);
+        }
+        finally
+        {
```
roncain: Nit: should probably use the close pattern Stephen has been using (do the close before the finally, and then have the finally call the testhelper to close any remaining CommunicationObjects

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
hongdai: certificateToInstall  is a local variable, not shared. Thus no other test can change it.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Security.TransportSecurity.Tests.csproj
hongdai: Will do.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
roncain: So maybe we should clarify the comment to say that some tests pass in null?  The phrase "previous test" makes this sound like it might be affected by test order.  NIT only.

hongdai: I will clarify it.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
hongdai: Will change.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
hongdai: Will do.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -176,10 +176,13 @@ public static void InstallLocalCertificateFromBridge()
                 }
             }
 
-            InstallCertificateToMyStore(certificateToInstall);
-
-            // We also need to install the root cert if we install a local cert
-            InstallRootCertificateFromBridge();
+            // certificateToInstall could be null in the case the user certification exists
+            if (certificateToInstall != null)
```
iamjasonp: I'd rather you check for `if (!foundUserCertificate)` here.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -176,10 +176,13 @@ public static void InstallLocalCertificateFromBridge()
                 }
             }
 
-            InstallCertificateToMyStore(certificateToInstall);
-
-            // We also need to install the root cert if we install a local cert
-            InstallRootCertificateFromBridge();
+            // certificateToInstall could be null in the case the user certification exists
```
iamjasonp: NIT: I don't quite think this is accurate. If !foundUserCertificate, certificateToInstall will not be null. The reason why these calls are here is because the Install*Certificate calls also take the same lock (s_certificateLock)

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
```diff
@@ -1,43 +1,77 @@
...
-            binding.Security.Mode = SecurityMode.Transport;
-            binding.Security.Transport.ClientCredentialType = TcpClientCredentialType.None;
+            factory.Close();
+        }
+        finally
+        {
+            if (factory != null && factory.State != CommunicationState.Closed)
+            {
+                factory.Abort();
+            }
```
iamjasonp: The new pattern we were using is: 

        finally 
        {
            ScenarioTestHelpers.CloseCommunicationObjects(factory); 
        }

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
```diff
@@ -13,10 +13,11 @@ namespace Infrastructure.Common
     public static class BridgeClient 
     {
         private const string EndpointResourceResponseUriKeyName = "uri";
+        private const string EndpointResourceResponseFQHNKeyName = "hostname";
         private const string EndpointResourceRequestNameKeyName = "name";
 
         private static object s_thisLock = new object();
-        private static Dictionary<string, string> _Resources = new Dictionary<string, string>();
+        private static Dictionary<string, Dictionary<string, string>> _Resources = new Dictionary<string, Dictionary<string, string>>();
```
iamjasonp: Please comment here on what the key and value are here

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -145,6 +145,14 @@ public static string Tcp_VerifyDNS_Address
...
     }
 
+    public static string Tcp_VerifyDNS_HostName
+    {
+        get
+        {
+            return BridgeClient.GetResourceHostName("WcfService.TestResources.TcpVerifyDNSResource");
+        }
+    }
+
```
iamjasonp: This seems like a weird of the Endpoints.cs class. Not blocking though, let's just get this in and we can talk about refactoring the BridgeClient classes

iamjasonp:
I don't have many comments other than the NIT level comments. LGTM once you're done with those; please don't forget the TFS checkin.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
roncain: Agree to discuss naming.  Endpoints offer just offer services or certificates.  They don't "verify". NIT

roncain:
LGTM, subject to the NIT level issues already mentioned

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
hongdai: certificateToInstall is null when foundUserCertificate is true, as the isLocalKeyName is not set, thus it continues to  InstallCertificateToMyStore.

The bug here might be why isLocalKeyName is not set from server. I'll open a issue to investigate it. Don't want to block this checkin.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
hongdai: See blow.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
hongdai: I'll use the new pattern.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
hongdai: Will do.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
hongdai: For this particular endpoint, it's designed for only TCP Dns identity tests. I'm open to better naming.

hongdai:

PR #435, opened 2015-10-16
Use new Implemented Async Methods in System.Runtime.Serialization.Xml

shmao:
System.Runtime.Serialization.Xml implemented some async methods per our request (see https://github.com/dotnet/corefx/issues/2502). The fix is to use those methods in our code.

Fixes #218, #221, and #222.

PR #460, opened 2015-10-23
Lower dependency on System.Runtime.Serialization.Xml

roncain:
The new dependency on 4.1.0 of System.Runtime.Serialization.Xml
caused a build break on TFS for TestNet.  Reverting to the prior
version solves the build break.

The 4.1.0 version contains fixes that WCF requires, but this lower
dependency is acceptable because the DNX will contain 4.1.0.

### DIFF
--- a/src/System.Private.ServiceModel/src/project.json
+++ b/src/System.Private.ServiceModel/src/project.json
@@ -32,7 +32,7 @@
     "System.Runtime.Extensions": "4.0.11-beta-*",
     "System.Runtime.Handles": "4.0.1-beta-*",
     "System.Runtime.InteropServices": "4.0.21-beta-*",
-    "System.Runtime.Serialization.Xml": "4.1.0-beta-*",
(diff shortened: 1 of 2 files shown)


roncain:
This is currently being tested in TFS.  Feel free to comment but please do not merge until that has completed.

hongdai:
LGTM.

hongdai:

zhenlan:
LGTM.

Just add a note: With this change, if someone tries to use the latest WCF package in a UWP app (or a pre-beta7 DNX app), she/he has to explicitly add >=4.0.11-beta-23409 version of system.runtime.serialization.xml package as well. Otherwise, this pull #435 code path will fail at runtime.

roncain:
I've done 2 clean builds of FxCore with this change, plus CrossGen test, plus ToF tests.  I'm going to merge but would still appreciate anyone running buddy builds to tell me if they see issues.

PR #461, opened 2015-10-26
Enable TCP client credential test

iamjasonp:
Test was disabled due to System.Security.Claims..ctor issue where the test-runtime
package was not up to date. Enable this test as test runtime package has now been
updated

Also change the endpoint so that we correctly obtain the DnsEndpointIdentity when
testing when the Bridge is located at an address other than the FQDN (e.g., when
testing against localhost or hostname)

Needs #457 to be merged first.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
@@ -174,13 +174,22 @@ public static partial class Endpoints
     {
         get { return BridgeClient.GetResourceAddress("WcfService.TestResources.DuplexCallbackXmlComplexTypeResource"); }
     }
-            
+
+
     public static string Tcp_CustomBinding_SslStreamSecurity_Address
     {
         get
         {
             BridgeClientCertificateManager.InstallLocalCertificateFromBridge();
             return BridgeClient.GetResourceAddress("WcfService.TestResources.TcpTransportSecurityWithSslResource");
+        }
+    }
+
+    public static string Tcp_CustomBinding_SslStreamSecurity_HostName
+    {
+        get
+        {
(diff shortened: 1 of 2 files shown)


iamjasonp:
Note: needs #457 to be merged first.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -183,4 +184,12 @@ public static string Tcp_CustomBinding_SslStreamSecurity_Address
             return BridgeClient.GetResourceAddress("WcfService.TestResources.TcpTransportSecurityWithSslResource");
         }
     }
+
+    public static string Tcp_CustomBinding_SslStreamSecurity_HostName
+    {
+        get
+        {
+            return BridgeClient.GetResourceHostName("WcfService.TestResources.TcpTransportSecurityWithSslResource");
+        }
```
roncain: We should have a chat about resource naming.  I would prefer we are consistent in using "Address" or "Resource" as the suffix and making the name of the static string match the name of the resource on the server.  Having 2 Endpoints with unique names that point to the same place is confusing to me (as well as having neither match the type name).  NIT -- not a blocker.

roncain:
LGTM.  Comments welcome about how we should make our resource names more consistent.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
iamjasonp: i agree with you. Currently I'm following how #457 does its naming, so at least it's consistent there. We can take this consistency issue on once both are merged I guess?

roncain:
LGTM, subject to #457 merge first

iamjasonp:

iamjasonp:

iamjasonp:
#457 got merged, so this can now go in.

PR #464, opened 2015-10-26
Reenable streaming tests after CoreFx fix

roncain:
The streaming tests were disabled due to issue
https://github.com/dotnet/corefx/issues/4077.

But now that the issue is fixed, these tests pass again,
so this PR re-enables them. Verified in Linux too.


iamjasonp:
Code LGTM. My Linux box isn't properly running tests right now, can't buddy build there. I call it ok to merge and I'll verify when I can get the box back up.

PR #482, opened 2015-10-28
Reduce number of binaries copied from CoreFx to run in Linux

roncain:


roncain:
The only remaining binaries needed from CoreFx appear to be Networking, and I am following up.  I would still like to merge this to chip away at using CoreFx binaries at all.  I recommend this not be merged until after #481 has been merged because it was tested with those project.lock.json files.

roncain:
Merging to continue investigating why Networking still requires copying from CoreFx

PR #474, opened 2015-10-27
Make cert parameters configurable and add TCP negative security tests

hongdai:
* Make cert parameters configurable by creating a CertificateCreationSettings.cs
and pass it around(Thanks for Jason's suggestion!). I have only added parameters
needed by the current invalid tests. I expect it will grow as we are adding more tests.
* Add expired certificate and custom validation fails tests.

Fixes #371

PR #483, opened 2015-10-28
Adding test case for X509CertificateValidator scenario

### LINKED ISSUE #367: Implement NetTcpBinding Security.Transport Positive E2E scenarios for Custom validator
This issue tracks implementation of b) X509CertificateValidationMode.Custom in test plan #322 .

StephenBonikowsky:
* Fixes #367
* Added a Constants file to WcfTestBridgeCommon as little by little I see more and more variables being used that could be put here.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
@@ -4,8 +4,10 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.IdentityModel.Selectors;
 using System.IO;
 using System.Runtime.Serialization;
+using System.Security.Cryptography.X509Certificates;
 using System.ServiceModel;
 using System.ServiceModel.Channels;
 using System.Text;
@@ -1102,5 +1104,39 @@ public class ClientReceiver : IPushCallback, IDisposable
         }
 
         disposed = true;
+    }
+}
+
+public class MyX509CertificateValidator : X509CertificateValidator
+{
+    string allowedIssuerName;
+    public bool validateMethodWasCalled = false;
+
+    public MyX509CertificateValidator(string allowedIssuerName)
+    {
+        if (string.IsNullOrEmpty(allowedIssuerName))
+        {
+            throw new ArgumentNullException("allowedIssuerName", "[MyX509CertificateValidator] The string parameter allowedIssuerName was null or empty.");
+        }
+
+        this.allowedIssuerName = allowedIssuerName;
+    }
+
+    public override void Validate(X509Certificate2 certificate)
+    {
+        validateMethodWasCalled = true;
+
+        // Check that there is a certificate.
+        if (certificate == null)
+        {
+            throw new ArgumentNullException("certificate", "[MyX509CertificateValidator] The X509Certificate2 parameter certificate was null.");
+        }
+
+        // Check that the certificate issuer matches the configured issuer.
+        if (!certificate.IssuerName.Name.Contains(allowedIssuerName))
+        {
+            throw new Exception
+                (string.Format("Certificate was not issued by a trusted issuer. Expected: {0}, Actual: {1}", allowedIssuerName, certificate.IssuerName.Name));
+        }
     }
 }
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
@@ -22,22 +22,29 @@ public static class Tcp_ClientCredentialTypeTests
     {
         string testString = "Hello";
         ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
 
         try
         {
+            // *** SETUP *** \\
             NetTcpBinding binding = new NetTcpBinding();
-
             factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.Tcp_DefaultBinding_Address));
-            IWcfService serviceProxy = factory.CreateChannel();
+            serviceProxy = factory.CreateChannel();
 
+            // *** EXECUTE *** \\
             string result = serviceProxy.Echo(testString);
+
+            // *** VALIDATE *** \\
             Assert.Equal(testString, result);
 
+            // *** CLEANUP *** \\
+            ((ICommunicationObject)serviceProxy).Close();
             factory.Close();
         }
         finally
         {
-            ScenarioTestHelpers.CloseCommunicationObjects(factory);
+            // *** ENSURE CLEANUP *** \\
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
         }
     }
 
@@ -48,21 +55,29 @@ public static class Tcp_ClientCredentialTypeTests
     {
         string testString = "Hello";
         ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
 
         try
         {
+            // *** SETUP *** \\
             NetTcpBinding binding = new NetTcpBinding(SecurityMode.None);
             factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.Tcp_NoSecurity_Address));
-            IWcfService serviceProxy = factory.CreateChannel();
+            serviceProxy = factory.CreateChannel();
 
+            // *** EXECUTE *** \\
             string result = serviceProxy.Echo(testString);
+
+            // *** VALIDATE *** \\
             Assert.Equal(testString, result);
 
+            // *** CLEANUP *** \\
+            ((ICommunicationObject)serviceProxy).Close();
             factory.Close();
         }
         finally
         {
-            ScenarioTestHelpers.CloseCommunicationObjects(factory);
+            // *** ENSURE CLEANUP *** \\
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
         }
     }
 
@@ -75,21 +90,29 @@ public static class Tcp_ClientCredentialTypeTests
     {
         string testString = "Hello";
         ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
 
         try
         {
+            // *** SETUP *** \\
             NetTcpBinding binding = new NetTcpBinding(SecurityMode.Transport);
             factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.Tcp_NoSecurity_Address));
-            IWcfService serviceProxy = factory.CreateChannel();
+            serviceProxy = factory.CreateChannel();
 
+            // *** EXECUTE *** \\
             string result = serviceProxy.Echo(testString);
+
+            // *** VALIDATE *** \\
             Assert.Equal(testString, result);
 
+            // *** CLEANUP *** \\
+            ((ICommunicationObject)serviceProxy).Close();
             factory.Close();
         }
         finally
         {
-            ScenarioTestHelpers.CloseCommunicationObjects(factory);
+            // *** ENSURE CLEANUP *** \\
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
         }
     }
 
@@ -133,20 +156,19 @@ public static class Tcp_ClientCredentialTypeTests
     }
 
     [Fact]
-    //[ActiveIssue(458)]
     [OuterLoop]
     public static void TcpClientCredentialType_Certificate_EchoString()
     {
         string clientCertThumb = null;
         EndpointAddress endpointAddress = null;
         string testString = "Hello";
-        NetTcpBinding binding = null;
         ChannelFactory<IWcfService> factory = null;
         IWcfService serviceProxy = null;
 
         try
         {
-            binding = new NetTcpBinding(SecurityMode.Transport);
+            // *** SETUP *** \\
+            NetTcpBinding binding = new NetTcpBinding(SecurityMode.Transport);
             binding.Security.Transport.ClientCredentialType = TcpClientCredentialType.Certificate;
 
             endpointAddress = new EndpointAddress(new Uri(Endpoints.Tcp_ClientCredentialType_Certificate_Address),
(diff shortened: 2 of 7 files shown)


src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
```diff
@@ -40,7 +40,7 @@ public class CertificateGenerator
         // Give the cert a grace period in case there's a time skew between machines
         private readonly TimeSpan _gracePeriod = TimeSpan.FromHours(1);
 
-        private const string _authorityCanonicalName = "DO_NOT_TRUST_WcfBridgeRootCA";
+        private const string _authorityCanonicalName = Constants.CertificateIssuerName;
         private const string _signatureAlthorithm = "SHA1WithRSAEncryption";
```
roncain: Unless @iamjasonp objects, I would prefer to rename the const in Constants something like CertificateAuthorityName and just use it everywhere (and not have one const here point to the other const in Constants.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
```diff
@@ -174,4 +173,47 @@ public static void TcpClientCredentialType_Certificate_EchoString()
             ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
         }
     }
+
+    [Fact]
+    [OuterLoop]
+    public static void TcpClientCredentialType_Certificate_CustomValidator_EchoString()
+    {
```
roncain: Is there a negative version of this?   Sorry if I missed it, but I'd expect to see 2 tests -- 1 hit that endpoint with a valid certificate and a 2nd to hit it with an invalid one.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
StephenBonikowsky: Amending my local branch with that change, will revert if @iamjasonp doesn't approve.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
StephenBonikowsky: That variation isn't in, frankly I wasn't even thinking about negative scenarios, should I add it here or do we need to put it in the 'ExpectedExceptions' area?

StephenBonikowsky: Adding the negative variation actually showed that the Validate method was not even being called, so the positive test was showing a false positive. I updated the test to confirm that the Validate method actually gets called no matter what.

StephenBonikowsky:
I added comments to existing tests based on previously discussed convention, just a tidying-up effort.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -441,4 +445,59 @@ public static void DuplexCallback_Throws_FaultException_ReturnsFaultedTask()
...
+                // *** SETUP *** \\
+                NetTcpBinding binding = new NetTcpBinding(SecurityMode.Transport);
+                binding.Security.Transport.ClientCredentialType = TcpClientCredentialType.Certificate;
+
+                endpointAddress = new EndpointAddress(new Uri(Endpoints.Tcp_ClientCredentialType_Certificate_CustomValidation_Address),
+                    new DnsEndpointIdentity(Endpoints.Tcp_VerifyDNS_HostName));
+                clientCertThumb = BridgeClientCertificateManager.LocalCertThumbprint; // ClientCert as given by the Bridge
+
+                factory = new ChannelFactory<IWcfService>(binding, endpointAddress);
+                factory.Credentials.ServiceCertificate.Authentication.CertificateValidationMode = X509CertificateValidationMode.None;
```
hongdai: I have already added a negative test in pull request https://github.com/dotnet/wcf/pull/474 for custom validator. The test passes. You need to set factory.Credentials.ServiceCertificate.Authentication.CertificateValidationMode = Custom, not None to hook up your own validator.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
```diff
@@ -163,14 +185,69 @@ public static void TcpClientCredentialType_Certificate_EchoString()
...
+            // *** SETUP *** \\
+            NetTcpBinding binding = new NetTcpBinding(SecurityMode.Transport);
+            binding.Security.Transport.ClientCredentialType = TcpClientCredentialType.Certificate;
+
+            endpointAddress = new EndpointAddress(new Uri(Endpoints.Tcp_ClientCredentialType_Certificate_CustomValidation_Address),
+                new DnsEndpointIdentity(Endpoints.Tcp_VerifyDNS_HostName));
+            clientCertThumb = BridgeClientCertificateManager.LocalCertThumbprint; // ClientCert as given by the Bridge
+
+            factory = new ChannelFactory<IWcfService>(binding, endpointAddress);
+            factory.Credentials.ServiceCertificate.Authentication.CertificateValidationMode = X509CertificateValidationMode.None;
```
hongdai: Authentication.CertificateValidationMode = Custom, not None to hook up your own validator. That's why the test fails.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/MyX509CertificateValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    public class MyX509CertificateValidator : X509CertificateValidator
+    {
+        string allowedIssuerName;
+
+        public MyX509CertificateValidator(string allowedIssuerName)
+        {
+            if (string.IsNullOrEmpty(allowedIssuerName))
+            {
+                throw new ArgumentNullException("[MyX509CertificateValidator] The string allowedIssuerName was null or empty.");
```
iamjasonp: NIT: Usually customary to add the parameter name to the exception when there's an ArgumentNullException e.g., 

```csharp
    throw new ArgumentException(message, "parameterName"); 
```

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
```diff
@@ -1104,3 +1106,35 @@ protected virtual void Dispose(bool disposing)
         disposed = true;
     }
 }
+
+public class MyX509CertificateValidator : X509CertificateValidator
+{
+    string allowedIssuerName;
+
```
iamjasonp: Can we put the `validateMethodWasCalled` here rather than have the static here? https://github.com/dotnet/wcf/pull/483/files#diff-5cc28940b9c4a251952d5274724a7f0fR25

Your test method has a reference to the instance of `MyX509CertificateValidator` used for validation, so it should be possible to keep this in an instance variable and validate the instance. 

The issue here is that if someone uses `MyX509CertificateValidator` without being aware of of its behaviour of the static in `ScenarioTestHelpers`, then you'll get some weird hard-to-debug test results.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/Client.ExpectedExceptions.Tests.csproj
```diff
@@ -36,6 +36,10 @@
             <Project>{9098B41C-9C2E-4FC6-B7D8-FC3411736A22}</Project>
             <Name>ScenarioTests.Common</Name>
         </ProjectReference>
+		<ProjectReference Include="$(WcfInfrastructureCommonProj)">
+      	    <Project>{AFD69665-89A3-42AE-A32E-AB2CBBE6EE7E}</Project>
+            <Name>Infrastructure.Common</Name>
+        </ProjectReference>
```
iamjasonp: NIT: weird spacing, check tabs vs spaces

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/project.json
```diff
@@ -5,6 +5,7 @@
     "System.ServiceModel.Http": "4.0.11-beta-*",
     "System.ServiceModel.NetTcp": "4.0.1-beta-*",
     "System.ServiceModel.Primitives": "4.1.0-beta-*",
+	"System.ServiceModel.Security": "4.0.1-beta-*",
```
iamjasonp: NIT: tabs vs spaces

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
```diff
@@ -40,7 +40,6 @@ public class CertificateGenerator
         // Give the cert a grace period in case there's a time skew between machines
         private readonly TimeSpan _gracePeriod = TimeSpan.FromHours(1);
 
-        private const string _authorityCanonicalName = "DO_NOT_TRUST_WcfBridgeRootCA";
```
iamjasonp: Revert all changes in this file referring to Constants. `CertificateGenerator` is designed to have no dependencies on the Bridge or any test classes. 

If you wish, you may add a setter to the `AuthorityCanonicalName` and make sure all instances of `CertificateGenerator` initialize the `AuthorityCanonicalName` value to what's in Constants prior to calling Initialize().

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/Constants.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace WcfTestBridgeCommon
+{
+    public static class Constants
+    {
+        public const string CertificateAuthorityName = "DO_NOT_TRUST_WcfBridgeRootCA";
+    }
+}
```
iamjasonp: Either remove this file, or ensure that you follow the guidance in https://github.com/dotnet/wcf/pull/483/files#r43360563 for usage.

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
```diff
@@ -1104,3 +1106,35 @@ protected virtual void Dispose(bool disposing)
...
+        {
+            throw new ArgumentNullException("[MyX509CertificateValidator] The string allowedIssuerName was null or empty.");
+        }
+
+        this.allowedIssuerName = allowedIssuerName;
+    }
+
+    public override void Validate(X509Certificate2 certificate)
+    {
+        ScenarioTestHelpers.validateMethodWasCalled = true;
```
iamjasonp: Are you testing if the method was merely called, or if it was successful?

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/MyX509CertificateValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+            if (certificate == null)
+            {
+                throw new ArgumentNullException("[MyX509CertificateValidator] The X509Certificate2 certificate was null.");
+            }
+
+            // Check that the certificate issuer matches the configured issuer.
+            if (!certificate.IssuerName.Name.Contains(allowedIssuerName))
+            {
+                throw new SecurityTokenValidationException
+                  ("Certificate was not issued by a trusted issuer");
```
iamjasonp: Message should include "expected '{0}', actual '{1}'"

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/MyX509CertificateValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            this.allowedIssuerName = allowedIssuerName;
+        }
+
+        public override void Validate(X509Certificate2 certificate)
+        {
+            // Check that there is a certificate.
+            if (certificate == null)
+            {
+                throw new ArgumentNullException("[MyX509CertificateValidator] The X509Certificate2 certificate was null.");
```
iamjasonp: ditto: https://github.com/dotnet/wcf/pull/483/files#r43356208

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
iamjasonp: CertificateGenerator shouldn't have any dependencies to `Constants` or the Bridge. If you want to constant-ize this, we should make `AuthorityCanonicalName` settable and ensure that every instance of `CertificateGenerator` is initialized with `CertificateGenerator.AuthorityCanonicalName = Constants.CertificateAuthorityName `

Otherwise, CertificateGenerator.AuthorityCanonicalName is effectively a const that you can use everywhere after initialization.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
roncain: I don't understand the recommendation to make this settable.  Isn't it a const?  In fact, I don't understand the null test below against a const.  Maybe before refactoring, @StephenBonikowsky could you state in these PR comments what the test needs to do with respect to CertificateGenerator?  I thought you only needed to read this value.

iamjasonp: The way it's done currently breaks encapsulation of the CertificateGenerator class. One should be initializing a certificate generator like so: 

```csharp
s_certificateGenerator = new CertificateGenerator() 
{ 
    CertificatePassword = config.BridgeCertificatePassword, 
    CrlUriBridgeHost = string.Format("http://{0}:{1}", config.BridgeHost, config.BridgePort), 
    CrlUriRelativePath = s_crlUriRelativePath, 
    ValidityPeriod = config.BridgeCertificateValidityPeriod 
}; 
```

My assertion is that we should be doing something similar here for `AuthorityCanonicalName` for initialization rather than have the `CertificateGenerator` "reach out" of the class to get it.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -441,4 +445,59 @@ public static void DuplexCallback_Throws_FaultException_ReturnsFaultedTask()
             }
         }
     }
+
+    [Fact]
+    [OuterLoop]
+    [ActiveIssue(485)]
+    public static void TcpClientCredentialType_Certifiacte_CustomValidatorFail_EchoString()
```
iamjasonp: Typo in `Certifiacte`

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/Client.ExpectedExceptions.Tests.csproj
roncain: Just checking -- the TFS ToF wrapper project will need to be updated to match this reference.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
StephenBonikowsky: Thanks Hong, when I moved the test from my desktop implementation to OSS I managed to lose that setting, thanks for catching it.
I'll remove the negative test since you already got it.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/Client.ExpectedExceptions.Tests.csproj
StephenBonikowsky: Hong already wrote this test in her current PR so I will revert this.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
StephenBonikowsky: I understand @iamjasonp 's concerns and have reverted the changes. I do only need to get the value, the problem is I need to get it from a test resource at which point I have no access to the instance of CertificateGenerator created when the client side test case makes the initial call to generate certificates and it seems to me a bit of an overkill to create a new instance of it as Jason describes just to get the value of that one variable.
I'm just going to hard-code what I need for now, if we were ever to make this particular variable configurable then we might consider doing something different.

roncain:
LGTM

StephenBonikowsky:
Rebased against latest merges.

PR #489, opened 2015-10-30
Remove makecert-generated certificates from repo

iamjasonp:
Move documentation from certificates to Documentation/archives as a reference for how we used to use makecert. The documentation might be useful in testing, but the certificates themselves are removed as they're expired and no longer work.


roncain:
LGTM -- thanks @iamjasonp

shmao:
LGTM.

PR #490, opened 2015-10-30
Add a Certificate Cleanup tool to clean up Bridge-generated certs

iamjasonp:
This is the certificate cleanup tool for cleaning up our installed certs. I've tested this on Windows, but I need to test this on Linux

EDIT: Ok to merge. We will wire up the scripts to call it in another PR, but @roncain and I have manually run this and confirmed certs get cleaned out now.

PR #496, opened 2015-11-02
run-test.sh no longer copies binaries from WCF bin folder

roncain:
Prior to this change the run-test.sh script used to run the tests
in Linux copied all the assemblies from the WCF bin folder into the
WCF test folders.  This was harmless when the assemblies were the
same but caused issues when the bin folder contains other versions
of the assemblies.

With this change, the script no longer copies assemblies from the
WCF bin folder but relies on the test folders containing all they need.

### DIFF
--- a/run-test.sh
+++ b/run-test.sh
@@ -102,7 +102,11 @@ create_test_overlay()
 	echo "WCF binaries not found at $WcfBins"
 	exit 1
   fi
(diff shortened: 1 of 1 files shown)


iamjasonp:
Thanks Ron for identifying this issue while looking at #490 with me. 
I will pull and build just to make sure nothing is busted; the change and logic LGTM

iamjasonp:
LGTM

roncain:
Merging so we can proceed to next step to add CertificateCleanup to build script

PR #501, opened 2015-11-03
Invoke CertificateCleanup.exe on Linux after all tests run

roncain:
This update the Linux script to run tests so that it invokes
the certificate cleanup logic after all tests are run.


iamjasonp:
LGTM

PR #503, opened 2015-11-04
No longer copy CoreFx binaries when running Linux tests

### LINKED ISSUE #442: Remove from run-test.sh any package overwriting from CoreFx
run-test.sh has logic to copy selected binaries from CoreFx over the top of package-restored binaries, due to the fact the binaries in the packages are Windows assemblies.

This task is to remove each of those individually copied dll's as packaging is fixed.

roncain:
Prior to this change, it was necessary to copy some binaries
from the CoreFx repo built on Linux when running tests.
This was necessary because the packages had incorrect binaries.

But that situation was resolved in preparation for RC1, and
it is no longer necessary to copy any binaries from CoreFx.

Fixes #442

### DIFF
--- a/run-test.sh
+++ b/run-test.sh
@@ -122,12 +122,12 @@ create_test_overlay()
 	exit 1
   fi
   
-# Currently need to overwrite some packaged binaries from CoreFx
-# See issue https://github.com/dotnet/wcf/issues/442
+# In the past we needed to copy some specific binaries from the CoreFx
+# repo due to packaging issues.  This is no longer necessary, but we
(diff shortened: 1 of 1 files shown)


iamjasonp:
:shipit:

mconnew:
An earlier merge disabled websockets tests which were causing the Linux test runs to not complete. Asking dotnet-bot to retest to verify this script.

mconnew:
LGTM, reran the tests more from curiosity to see if the Linux tests are passing. Don't block on the results.

PR #512, opened 2015-11-10
Fixed secure websockets setup and disable failing tests

mconnew:
There is still a failing test due to a bug in the WebSockets client code. See dotnet/corefx#4429


roncain:
LGTM -- subject only to nit-level question about Endpoints

PR #514, opened 2015-11-10
Update dependencies

roncain:


roncain:
Not yet ready to merge -- currently evaluating whether works in Linux.  This corresponds to the fixes made in CoreFx to unbreak their Linux testing

roncain:
@dotnet-bot please retest this

roncain:

src/Common/test-runtime/project.json
```diff
@@ -1,61 +1,17 @@
...
-      "System.Security.Principal": "4.0.1-beta-*",
-      "System.Text.Encoding": "4.0.11-beta-*",
-      "System.Threading": "4.0.11-beta-*",
-      "System.Threading.Tasks": "4.0.11-beta-*",
-      "System.Threading.Timer": "4.0.1-beta-*",
-      "System.Xml.ReaderWriter": "4.0.11-beta-*",
-      "System.Xml.XDocument": "4.0.11-beta-*",
-      "System.Xml.XmlDocument": "4.0.1-beta-*",
-      "System.Xml.XmlSerializer": "4.0.11-beta-*",
-
```
mconnew: Why did you remove our dependencies? Without these, the versions hard coded into some of the BuildTools packages will override what we need and we won't run our tests with the correct versions at runtime.

roncain: I actually took CoreFx's version verbatim.  But before I add these back, let me ask "which of these need to be stated explicitly, and why?"  Obviously we want to use the correct dependencies for WCF, but we also don't want to spuriously duplicate dependencies coming from the meta packages (maintenance headache).  I can analyze them one by one, but ideally we should have the smallest possible set of differences.  Do you know offhand which they are?

mconnew: I would say we should have all of them. Otherwise any of our dependent assemblies could be using a different version in our tests than what an end user would be using when referencing WCF. Without adding them all, we are relying on the various build tools dependencies to be updated in order to catch bugs in our dependencies. We had this before with HttpClient where we didn't see the break until about 3 weeks after the change was made.

roncain: Okay, I've redone this PR to make only the minimal touches CoreFx made to the runtime-tools\project.json and to retain our existing overrides

roncain:
@mconnew, @hongdai or @StephenBonikowsky -- I would like to merge this as soon as someone approves so I can continue deeper Linux testing.  See above for how I have verified this PR.

hongdai:
LGTM.

PR #516, opened 2015-11-10
Remove accidental addition of S.N.WebSockets namespace

mconnew:
This was accidentally added when debugging the failing WebSockets tests previously.


mconnew:
This is breaking building this test in our internal ToF runs.

roncain:
LGTM

PR #517, opened 2015-11-11
Implement NetTcpBinding Security.Transport Certification authority revokes a certificate

### LINKED ISSUE #369: Implement NetTcpBinding Security.Transport negative E2E scenarios:Certification authority revokes a certificate
This issue tracks implementation of the following test
 NetTcpBinding Security.Transport negative E2E scenarios
b) i  Certification authority revokes a certificate

in test plan #322 .

hongdai:
* Add a test to check client call will fail if service cert is revocated.

Fixes #369

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -314,7 +314,7 @@ public static class ExpectedExceptionTests
                 if (ce.InnerException == null)
                     throw;
 
-                if (ce.InnerException.GetType() == typeof (HttpRequestException))
+                if (ce.InnerException.GetType() == typeof(HttpRequestException))
                 {
                     var httpReqExcep = ce.InnerException as HttpRequestException;
                     StringBuilder sb = new StringBuilder();
@@ -511,6 +511,46 @@ public static class ExpectedExceptionTests
             ScenarioTestHelpers.CloseCommunicationObjects(factory);
         }
     }
+
+    [Fact]
+    [OuterLoop]
+    [ActiveIssue(533)]
+    // Verify product throws MessageSecurityException when the service cert is revoked
+    public static void TCP_ServiceCertRevoked_Throw_MessageSecurityException()
+    {
+        string testString = "Hello";
+
+        NetTcpBinding binding = new NetTcpBinding();
+        binding.Security.Mode = SecurityMode.Transport;
+        binding.Security.Transport.ClientCredentialType = TcpClientCredentialType.None;
+
+        EndpointAddress endpointAddress = new EndpointAddress(new Uri(Endpoints.Tcp_RevokedServerCertResource_Address), new DnsEndpointIdentity(Endpoints.Tcp_RevokedServerCertResource_HostName));
+        ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, endpointAddress);
+        IWcfService serviceProxy = factory.CreateChannel();
+
+        //Need to wait for 2 minutes here to ensure the CRL update are in effect as the update frequency is set to 2 minutes
+        Task.Delay(TimeSpan.FromMinutes(2)).Wait();
+        try
+        {
+            var result = serviceProxy.Echo(testString);
+            Assert.True(false, "Expected: SecurityNegotiationException, Actual: no exception");
+        }
+        catch (CommunicationException exception)
+        {
+            string exceptionType = exception.GetType().Name;
+            if (exceptionType != "SecurityNegotiationException")
+            {
+                Assert.True(false, string.Format("Expected type SecurityNegotiationException, Actual: {0}", exceptionType));
+            }
+            string exceptionMessage = exception.Message;
+            Assert.True(exceptionMessage.Contains(Endpoints.Tcp_RevokedServerCertResource_HostName), string.Format("Expected message contains {0}, actual message: {1}", Endpoints.Tcp_RevokedServerCertResource_HostName, exception.ToString()));
+        }
+        finally
+        {
+            ScenarioTestHelpers.CloseCommunicationObjects(factory);
+        }
+    }
+
 }
 
 public class MyCertificateValidator : X509CertificateValidator
--- a/src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
+++ b/src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
@@ -424,8 +424,9 @@ namespace WcfTestBridgeCommon
             
             _crlGenerator.Reset();
 
-            _crlGenerator.SetThisUpdate(_initializationDateTime);
-            _crlGenerator.SetNextUpdate(DateTime.UtcNow.Add(_crlValidityPeriod));
+            DateTime now = DateTime.UtcNow;
+            _crlGenerator.SetThisUpdate(now);
+            _crlGenerator.SetNextUpdate(now.Add(_crlValidityPeriod));
             _crlGenerator.SetIssuerDN(PrincipalUtilities.GetSubjectX509Principal(signingCertificate));
             _crlGenerator.SetSignatureAlgorithm(_signatureAlthorithm);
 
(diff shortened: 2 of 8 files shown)


hongdai:
@dotnet-bot Please test outerloop

hongdai:

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -511,6 +511,45 @@ public static void TCP_ServiceCertFailedCustomValidate_Throw_Exception()
...
+        binding.Security.Mode = SecurityMode.Transport;
+        binding.Security.Transport.ClientCredentialType = TcpClientCredentialType.None;
+
+        EndpointAddress endpointAddress = new EndpointAddress(new Uri(Endpoints.Tcp_RevokedServerCertResource_Address), new DnsEndpointIdentity(Endpoints.Tcp_RevokedServerCertResource_HostName));
+        ChannelFactory<IWcfService> factory = new ChannelFactory<IWcfService>(binding, endpointAddress);
+        IWcfService serviceProxy = factory.CreateChannel();
+
+        //Need to wait for 2 minutes here to ensure the CRL update are in effect as the update frequency is set to 2 minutes
+        Task.Delay(TimeSpan.FromMinutes(2)).Wait();
+        try
```
roncain: I understand the issue, but I don't think we want our OuterLoops to have an unconditional 2 minute delay for a single test.  I wonder if it would be better to try the test in a loop that keeps repeating as long as it doesn't get the right failure.  And only after 2 minutes of this fails.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -511,6 +511,45 @@ public static void TCP_ServiceCertFailedCustomValidate_Throw_Exception()
...
+        {
+            var result = serviceProxy.Echo(testString);
+            Assert.True(false, "Expected: SecurityNegotiationException, Actual: no exception");
+        }
+        catch (CommunicationException exception)
+        {
+            string exceptionType = exception.GetType().Name;
+            if (exceptionType != "SecurityNegotiationException")
+            {
```
roncain: Question -- is the checking of the exception type by name because that exception type is not in contract

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -511,6 +511,45 @@ public static void TCP_ServiceCertFailedCustomValidate_Throw_Exception()
...
+                Assert.True(false, string.Format("Expected type SecurityNegotiationException, Actual: {0}", exceptionType));
+            }
```
roncain: I think our convention should always be that when we can an exception we did not expect we show exception.ToString().  The name by itself is too little to diagnose what happened.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
```diff
@@ -424,8 +424,9 @@ private X509Crl CreateCrl(X509Certificate signingCertificate)
             
             _crlGenerator.Reset();
```
roncain: Maybe outside the scope of this PR, but having a single _crlGenerator that is reset and mutated makes me extremely nervous.  It raises the possibility of multiple tests intefering with each other.  I would much rather see it be rewritten so it is not necessary to call Reset or set properties like this.  Better to create some kind of object containing all the parameters that gets passed to the _crlGenerator.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateManager.cs
```diff
@@ -237,6 +237,19 @@ public static X509Certificate2 CreateAndInstallNonDefaultMachineCertificates(Cer
             }
         }
 
+        public static void RevokingACert(CertificateGenerator certificateGenerator, string serialNum)
+        {
```
roncain: Prefer name matches what it does, like "RevokeCertificate"

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
hongdai: Agreed. I'll make the change.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
hongdai: Correct. SecurityNegotiationException is not in the contract.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
hongdai: Make sense. I'll make the change.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
hongdai: We should discuss it after Jason comes back.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateManager.cs
hongdai: Ok. I'll rename it.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/CertificateResources/CertificateRevocationListResource.cs
```diff
@@ -29,13 +29,13 @@ public override ResourceResponse Put(ResourceRequestContext context)
         {
             var certGenerator = CertificateResourceHelpers.GetCertificateGeneratorInstance(context.BridgeConfiguration);
 
-            string thumbprint;
+            string serialnumber;
```
iamjasonp: NIT: `serialNumber` with a capital N.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
iamjasonp: Yeah, I see @roncain's point. We shouldn't need a single `_crlGenerator` given our changes nowadays; we should be able to simply new up a `new X509V2CrlGenerator();` to avoid the interference. 

That being said, the caller to `CrlEncoded` should be locked, as that call ultimately is going to be causing issues if > 1 test is trying to mutate the Crl at the same time.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
iamjasonp: +1 to @roncain's comment. 

I've also been tinkering with ways to deal with the CRL caching to avoid this 2 minute wait. 
However, I think that this commit should go in and these changes can go in as a later change.

roncain: Okay to check-in 2 minute delay providing an issue is open for it and someone is working on an immediate fix.  I want to avoid unconditional delays of this magnitude.

hongdai: Issue #533 is opened to discuss the solution, in case we want to unblock this check-in.

hongdai:

roncain:
LGTM -- we can reactivate this test after @iamjasonp considers alternatives to the delay.

PR #530, opened 2015-11-16
Supported scenarios table.

StephenBonikowsky:
*First draft
*Images need to be put on our repo as well, other teams added an 'Image' folder under Documentation as well.
*Need to be careful of the license for the icons used, would love to use icons already being used by other MS projects on GitHub but couldn't find the image source for some used by Jenkins.
*Click on "Display the rich diff" to see it properly, if some image didn't load just refresh.


iamjasonp:
Looking good so far! 

* For the images, can we use the Windows 10 emoji, colour them (e.g., in Paint or Word), then upload them? 
* One of the workarounds I've seen is to upload some assets (like dotnet/corefx/issues/ 911) and then directly reference them from githubusercontent.com

roncain:
Are there any GitHub emojis that would suffice?  I think we would need to get permission to make the Win10 emojis OSS.

Documentation/RC1SupportedScenarios.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+![Image](http://findicons.com/files/icons/573/must_have/16/check.png) -- Supported: Release quality
+
+![Image](http://findicons.com/files/icons/1155/diagona/16/104.png) --  Supported: Not fully tested
+
+![Image](http://findicons.com/files/icons/573/must_have/16/add.png) -- Not Yet Supported: Future release
+
+![Image](http://findicons.com/files/icons/573/must_have/16/delete.png) -- Not Supported: Will not support
+
```
roncain: Hmmm -- I think a "will not support" scenario belongs in a different table.  Otherwise we will always have tables with red X's and make the product look incomplete.  And I think something like the red X is the right icon for "Not yet supported".  Ideally the table will contain only "supported", "not yet tested", and "not yet supported".

Out of curiosity, it is possible to make the "not yet supported" items also contain a link to an issue?  Or does that blow out the size of the columns?

Documentation/RC1SupportedScenarios.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+|Scenario |UWP   |Windows |Linux |Mac OS X  |
+|---------|------|--------|------|----------|
```
roncain: I really like the shape of this.  There's a really interesting issue to solve with the Linux column, and ideas are welcome.  The issues are:
1. We need to identify which Linux distro's have been tested (e.g. ubuntu-14-04)
2. What "works" vs "doesn't work" may differ by Linux distro.

So conceptually we may need to have icons for "supported on some versions" and "tested on some versions".  And then we might include a second table with columns only for Linux, containing only the subset of scenarios affected, and they would get the standard "supported" "not yet tested" "not yet supported" icons

roncain: We'll need to start listing what scenarios should appear in the rows.  They need to be easily recognizable by WCF users and extremely terse.  We may also need a new column for "pivots" or "options" such as:

| Scenario  | Option |
| ------------- | ------------- |
| Http  | Security.None  |
|   | Security.Transport|
|   | Text encoding  |
|   | Binary encoding | 
| Https  | Security.None  |
|   | Security.Transport|
|   | Text encoding  |
|   | Binary encoding | 
| Tcp| Security.None  |
|   | Security.Transport|
|   | Text encoding  |
|   | Binary encoding | 
|   | Duplex | 
| WebSockets | etc  |
|   | etc ||

roncain:
It also occurred to me we should probably make a folder under documentation called "releases" and then have a separate folder under each one for each release (ex: documentation/releases/rc1).  We might add other md files to describe other aspects of that release.  Let me study how other repo's are handling this and come back with a proposal.  Don't block the merge on this decision -- we can move and rename anytime.

Documentation/RC1SupportedScenarios.md
StephenBonikowsky: Is it possible that something we do support in UWP and Windows might not be able to be supported in Linux or Mac, if so then that is the scenario for needing a "Will not/Cannot Support".

Links are easy, from what I saw it didn't screw up the formatting. I was thinking that if we had an issue label for each scenario then we might be able to include a link to an issue query that would only show issues related to that scenario.

roncain: You're right -- if something will never be supported in UWP or Linux, I guess we need an icon for that.  I was thinking only about having solid rows of "will not support" looking bad.  Maybe it should be something less alarming looking -- like N/A.  Thoughts?

zhenlan: Currently we don't have anything that we intentionally only support for one platform but not another, so I think we don't need the "will not support" icon for now. We can add it when we actually need it in the future.

Conceptually, anything that is not covered by any rows of the table is not supported. I agree we should avoid any solid rows of "will not support".

Documentation/RC1SupportedScenarios.md
zhenlan: @roncain you have a good point about Linux distro. I think we should explicitly list the one we tested. If we can cover more Linux distro's in the future, we can add more columns to the table. If it becomes too many columns in the future, we will need to think about a better way, but for now I hope we can keep it simple.

Documentation/RC1SupportedScenarios.md
roncain: I agree we don't need a "will not support" icon for now. I think "not supported" plus a link to an issue is sufficient (where the issue might explain why it is not supported now or that there are no plans to support it)

I also believe we have (and will have more) WCF features that we find do not work on some platform because our dependent packages don't support it.  Examples: NegotiateStream in NET Native, or using the X509Store in certain Linux distros.  I think having a single "not supported" icon is sufficient to cover this case, as long as it has a link to explain why it is not supported.

roncain: I think it would be a good convention that any icon other than "supported" should have a hyperlink the user can follow for more information.  It might be a link to an issue or a markdown file.  I think we should also adopt a simple naming convention for the hyperlink's display name.  Something simple like "?" or "why?".  As a user, I think the first thing I would want to know about a "not supported" item is "why not?" and "when?"

We can get the basic table checked in and fill in the hyperlinks afterwards (though maybe one hyperlink on the initial checkin would help show the pattern to follow.

StephenBonikowsky: We discussed links to Issues or Issue queries that correlate to each scenario, maybe we can add labels to connect Issues to the scenarios.

Documentation/RC1SupportedScenarios.md
StephenBonikowsky: Apparently emoji's do work here on GitHub (https://github.com/markdown-it/markdown-it-emoji) unfortunately I can't use them in Visual Studio Code, there may be a plug-in I don't know about, there may also be some editor GitHub/markdown-it recommends to assist during development.

StephenBonikowsky:
I'm going to update the file removing the icons so I can just check-in the simple table I started, Ron can then edit it and add scenarios and emojis.

Issue #532, opened 2015-11-17
Add scenario for WebSockets, buffered over https.


StephenBonikowsky:
Missing this scenario.

StephenBonikowsky:
Resolved via #539

PR #539, opened 2015-11-20
Adding additional Web Socket tests.

StephenBonikowsky:
* Fixes Issue #532
* Includes some minor formatting changes and one bug in other WebSocket tests in the same file.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
@@ -24,10 +24,12 @@ public static class WebSocketTests
         try
         {
             // *** SETUP *** \\
-            binding = new NetHttpBinding();
+            binding = new NetHttpBinding()
+            {
+                MaxReceivedMessageSize = ScenarioTestHelpers.SixtyFourMB,
+                MaxBufferSize = ScenarioTestHelpers.SixtyFourMB,
+            };
             binding.WebSocketSettings.TransportUsage = WebSocketTransportUsage.Always;
-            binding.MaxReceivedMessageSize = ScenarioTestHelpers.DefaultMaxReceivedMessageSize;
-            binding.MaxBufferSize = ScenarioTestHelpers.DefaultMaxReceivedMessageSize;
             binding.TransferMode = TransferMode.Streamed;
             binding.MessageEncoding = NetHttpMessageEncoding.Binary;
 
@@ -56,7 +58,8 @@ public static class WebSocketTests
             // *** VALIDATE *** \\
             foreach (string serverLogItem in client.GetLog())
             {
-                Assert.True(serverLogItem != ScenarioTestHelpers.RemoteEndpointMessagePropertyFailure, ScenarioTestHelpers.RemoteEndpointMessagePropertyFailure);
+                //Assert.True(serverLogItem != ScenarioTestHelpers.RemoteEndpointMessagePropertyFailure, ScenarioTestHelpers.RemoteEndpointMessagePropertyFailure);
+                Assert.True(!serverLogItem.Contains(ScenarioTestHelpers.RemoteEndpointMessagePropertyFailure), serverLogItem);
             }
 
             // *** CLEANUP *** \\
@@ -85,10 +88,12 @@ public static class WebSocketTests
         try
         {
             // *** SETUP *** \\
-            binding = new NetHttpBinding();
+            binding = new NetHttpBinding()
+            {
+                MaxReceivedMessageSize = ScenarioTestHelpers.SixtyFourMB,
+                MaxBufferSize = ScenarioTestHelpers.SixtyFourMB,
+            };
             binding.WebSocketSettings.TransportUsage = WebSocketTransportUsage.Always;
-            binding.MaxReceivedMessageSize = ScenarioTestHelpers.DefaultMaxReceivedMessageSize;
-            binding.MaxBufferSize = ScenarioTestHelpers.DefaultMaxReceivedMessageSize;
             binding.TransferMode = TransferMode.Streamed;
             binding.MessageEncoding = NetHttpMessageEncoding.Binary;
 
@@ -123,7 +128,8 @@ public static class WebSocketTests
             // This will deadlock if the transfer mode is buffered because the callback will wait for the
             // stream, and the NCL layer will continue to buffer the stream until it reaches the end.
 
-            clientReceiver.ReceiveStreamInvoked.WaitOne();
+            Assert.True(clientReceiver.ReceiveStreamInvoked.WaitOne(ScenarioTestHelpers.TestTimeout),
+                String.Format("Test case timeout was reached while waiting for the stream response from the Service. Timeout was: {0}", ScenarioTestHelpers.TestTimeout));
             clientReceiver.ReceiveStreamInvoked.Reset();
 
             // Upload the stream while we are downloading a different stream
@@ -134,11 +140,18 @@ public static class WebSocketTests
 
             client.StopPushingStream();
             // Waiting on ReceiveStreamCompleted from the ClientReceiver.
-            clientReceiver.ReceiveStreamCompleted.WaitOne();
+            Assert.True(clientReceiver.ReceiveStreamCompleted.WaitOne(ScenarioTestHelpers.TestTimeout),
+                String.Format("Test case timeout was reached while waiting for the stream response from the Service to be completed. Timeout was: {0}", ScenarioTestHelpers.TestTimeout));
             clientReceiver.ReceiveStreamCompleted.Reset();
 
+            // Getting results from server via callback.
+            client.GetLog();
+            Assert.True(clientReceiver.LogReceived.WaitOne(ScenarioTestHelpers.TestTimeout),
+                String.Format("Test case timeout was reached while waiting for the Logging from the Service to be received. Timeout was: {0}", ScenarioTestHelpers.TestTimeout));
+
             // *** VALIDATE *** \\
-            // Validation is based on no exceptions being thrown.
+            Assert.True(clientReceiver.ServerLog.Count > 0,
+                "The logging done by the Server was not returned via the Callback.");
 
             // *** CLEANUP *** \\
             ((ICommunicationObject)client).Close();
@@ -172,8 +185,8 @@ public static class WebSocketTests
             binaryMessageEncodingBindingElement = new BinaryMessageEncodingBindingElement();
             httpsTransportBindingElement = new HttpsTransportBindingElement()
             {
-                MaxReceivedMessageSize = ScenarioTestHelpers.DefaultMaxReceivedMessageSize,
-                MaxBufferSize = ScenarioTestHelpers.DefaultMaxReceivedMessageSize,
+                MaxReceivedMessageSize = ScenarioTestHelpers.SixtyFourMB,
+                MaxBufferSize = ScenarioTestHelpers.SixtyFourMB,
                 TransferMode = TransferMode.Streamed
             };
             httpsTransportBindingElement.WebSocketSettings.TransportUsage = WebSocketTransportUsage.Always;
@@ -211,7 +224,7 @@ public static class WebSocketTests
             // stream, and the NCL layer will continue to buffer the stream until it reaches the end.
 
             Assert.True(clientReceiver.ReceiveStreamInvoked.WaitOne(ScenarioTestHelpers.TestTimeout),
-                "Test case timeout was reached while waiting for the stream response from the Service.");
+                String.Format("Test case timeout was reached while waiting for the stream response from the Service. Timeout was: {0}", ScenarioTestHelpers.TestTimeout));
             clientReceiver.ReceiveStreamInvoked.Reset();
 
             // Upload the stream while we are downloading a different stream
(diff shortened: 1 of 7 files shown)


StephenBonikowsky:

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -123,7 +127,8 @@ public static void WebSocket_Http_Duplex_BinaryStreamed()
             // This will deadlock if the transfer mode is buffered because the callback will wait for the
```
iamjasonp: NIT: "deadlock" or "block"?

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -134,11 +139,18 @@ public static void WebSocket_Http_Duplex_BinaryStreamed()
...
             client.StopPushingStream();
             // Waiting on ReceiveStreamCompleted from the ClientReceiver.
-            clientReceiver.ReceiveStreamCompleted.WaitOne();
+            Assert.True(clientReceiver.ReceiveStreamCompleted.WaitOne(ScenarioTestHelpers.TestTimeout),
+                "Test case timeout was reached while waiting for the stream response from the Service to be completed.");
             clientReceiver.ReceiveStreamCompleted.Reset();
 
+            // Getting results from server via callback.
+            client.GetLog();
+            Assert.True(clientReceiver.LogReceived.WaitOne(ScenarioTestHelpers.TestTimeout),
```
iamjasonp: Question only: you use `ScenarioTestHelpers.TestTimeout` all the time here, but it's not really reflective of the total amount of time spent in the test. Is it necessary to use the TimeoutHelper to reduce the blocking time during each WaitOne call? 

(I don't know that it is, so it's ok with me as long as it's been thought about)

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -598,4 +632,456 @@ public static void WebSocket_WSScheme_WSTransportUsageAlways_DuplexCallback_Guid
...
+
+            channelFactory = new ChannelFactory<IWSRequestReplyService>(binding, new EndpointAddress(Endpoints.WebSocketHttpsRequestReplyBinaryBuffered_Address));
+            client = channelFactory.CreateChannel();
+
+            // *** EXECUTE *** \\
+            // Invoking DownloadData
+            string result = client.DownloadData();
+
+            // Invoking UploadData
+            client.UploadData(ScenarioTestHelpers.CreateInterestingString(123));
```
iamjasonp: Question only: is there any reason why `123`, or is it just because it's the first three numbers on the keyboard? :smile:

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -598,4 +632,456 @@ public static void WebSocket_WSScheme_WSTransportUsageAlways_DuplexCallback_Guid
...
+
+            // *** VALIDATE *** \\
+            foreach (string serverLogItem in client.GetLog())
+            {
+                Assert.True(serverLogItem != ScenarioTestHelpers.RemoteEndpointMessagePropertyFailure, ScenarioTestHelpers.RemoteEndpointMessagePropertyFailure);
```
iamjasonp: Hmm, it almost looks like you want to Assert.NotEqual here. 
If you do really mean to Assert.True, consider putting a more descriptive message as the failure message.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
iamjasonp: It also might be interesting to set a number > the buffersize to make sure that we can actually handle larger messages.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -598,4 +632,456 @@ public static void WebSocket_WSScheme_WSTransportUsageAlways_DuplexCallback_Guid
...
```
iamjasonp: ditto

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
iamjasonp: Further question: is it useful to dump `client.GetLog` upon failure?

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -598,4 +632,456 @@ public static void WebSocket_WSScheme_WSTransportUsageAlways_DuplexCallback_Guid
...
+            // Invoking StopPushingData
+            client.StopPushingData();
+            Assert.True(clientReceiver.ReceiveDataCompleted.WaitOne(ScenarioTestHelpers.TestTimeout),
+            "Test case timeout was reached while waiting for the buffered response from the Service to be completed.");
+            clientReceiver.ReceiveDataCompleted.Reset();
+
+            // Getting results from server via callback.
+            client.GetLog();
+            Assert.True(clientReceiver.LogReceived.WaitOne(ScenarioTestHelpers.TestTimeout),
+            "Test case timeout was reached while waiting for the Logging from the Service to be received.");
```
iamjasonp: NIT: would prefer if you indented this line

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -598,4 +632,456 @@ public static void WebSocket_WSScheme_WSTransportUsageAlways_DuplexCallback_Guid
...
+
+            // Invoking StartPushingData
+            client.StartPushingData();
+            Assert.True(clientReceiver.ReceiveDataInvoked.WaitOne(ScenarioTestHelpers.TestTimeout),
+            "Test case timeout was reached while waiting for the buffered response from the Service.");
+            clientReceiver.ReceiveDataInvoked.Reset();
+            // Invoking StopPushingData
+            client.StopPushingData();
+            Assert.True(clientReceiver.ReceiveDataCompleted.WaitOne(ScenarioTestHelpers.TestTimeout),
+            "Test case timeout was reached while waiting for the buffered response from the Service to be completed.");
```
iamjasonp: NIT for lines above: indentation after `Assert.True`

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
iamjasonp: Ah, I see the message is: "RemoteEndpointMessageProperty did not contain the address of this machine". I think that should be descriptive enough, though "this" is ambiguous in the context of a x-machine test

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -598,4 +632,456 @@ public static void WebSocket_WSScheme_WSTransportUsageAlways_DuplexCallback_Guid
...
+        {
+            // *** ENSURE CLEANUP *** \\  
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)client, channelFactory);
+        }
+    }
+
+    [Fact]
+    [OuterLoop]
+    public static void WebSocket_Https_Duplex_TextBuffered_KeepAlive()
+    {
```
iamjasonp: I might be missing something... how are we testing that KeepAlive is working properly here?

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -598,4 +632,456 @@ public static void WebSocket_WSScheme_WSTransportUsageAlways_DuplexCallback_Guid
...
+        {
+            // *** ENSURE CLEANUP *** \\  
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)client, channelFactory);
+        }
+    }
+
+    [Fact]
+    [OuterLoop]
+    public static void WebSocket_Http_RequestReply_BinaryBuffered_KeepAlive()
+    {
```
iamjasonp: How's KeepAlive being verified aside from being set on a property?

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/WebSocketResources.cs
```diff
@@ -200,15 +200,243 @@ protected override int GetPort(ResourceRequestContext context)
...
-            NetHttpBinding binding = new NetHttpBinding();
+            NetHttpBinding binding = new NetHttpBinding()
+            {
+                MaxReceivedMessageSize = DefaultMaxReceivedMessageSize,
+                MaxBufferSize = DefaultMaxReceivedMessageSize,
+            };
             binding.WebSocketSettings.TransportUsage = WebSocketTransportUsage.Always;
-            binding.MaxReceivedMessageSize = DefaultMaxReceivedMessageSize;
-            binding.MaxBufferSize = DefaultMaxReceivedMessageSize;
             binding.TransferMode = TransferMode.Streamed;
```
iamjasonp: For the future: would prefer the changes above (cleaning up the NetHttpBinding declarations) be in a separate commit even if it's in the same PR, but don't spend time splitting the changes now.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -226,8 +238,14 @@ public static void WebSocket_Https_Duplex_BinaryStreamed()
                 "Test case timeout was reached while waiting for the stream response from the Service to be completed.");
             clientReceiver.ReceiveStreamCompleted.Reset();
 
+            // Getting results from server via callback.
+            client.GetLog();
+            Assert.True(clientReceiver.LogReceived.WaitOne(ScenarioTestHelpers.TestTimeout),
+            "Test case timeout was reached while waiting for the Logging from the Service to be received.");
+
```
roncain: Nit: not needed to merge.  It would aid debugging to add to the message how long the TestTimeout was when the error occurred.  We might configure it differently in lab or debugging runs.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -24,10 +24,12 @@ public static void WebSocket_Http_RequestReply_BinaryStreamed()
         try
         {
             // *** SETUP *** \\
-            binding = new NetHttpBinding();
+            binding = new NetHttpBinding()
+            {
+                MaxReceivedMessageSize = ScenarioTestHelpers.DefaultMaxReceivedMessageSize,
+                MaxBufferSize = ScenarioTestHelpers.DefaultMaxReceivedMessageSize,
```
roncain: Nit: probably want a separate const for DefaultMaxBufferSize rather than reusing the max received message size.  Consider for future cleanup -- not needed for merge.

roncain:
LGTM subject to questions from @iamjasonp being considered or answered.  Nit level stuff is optional for me.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: It doesn't reduce the time for each WaitOne call, if everything is working as expected the call should complete well before the timeout, it was just recommended that we wrap the WaitOne call this way 'just in case' the call ever failed for some reason so that the test wouldn't hang here for a long time until some other timeout gets hit.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: I'm essentially porting tests from the legacy test bed that runs on desktop, there are actually many more tests I can still port that would cover additional scenarios such as the one you mentioned. They are mostly lower priority though, I tried to focus on the P0 tests first.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: Assert.NotEqual doesn't make sense to me in the current context.
On the Server there are several addresses we don't know which one is being used in the RemoteEndpointMessageProperty so if the validation fails we only know what was in the RemoteEndpointMessageProperty which is what we were expecting to find.
I can log that, then in the client this extra little bit of info could be useful unless of course the log was returned empty or the call to get the log failed for other reasons.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: Done

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: Had a long discussion about this with @mconnew will let him add to this if he wants, but the short version is that we can't.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/WebSocketResources.cs
StephenBonikowsky: Yeah, sorry, spur of the moment cleaning.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: Added, easier to do now than later.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: Done

StephenBonikowsky:

PR #542, opened 2015-11-21
Enable client certs for HTTPS plus server certificate validation

mconnew:


src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/KerberosSecurityTokenProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        public KerberosSecurityTokenProvider(string servicePrincipalName, TokenImpersonationLevel tokenImpersonationLevel, NetworkCredential networkCredential)
+        {
+            if (servicePrincipalName == null)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("servicePrincipalName");
+            if (tokenImpersonationLevel != TokenImpersonationLevel.Identification && tokenImpersonationLevel != TokenImpersonationLevel.Impersonation)
+            {
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException("tokenImpersonationLevel",
+                    "blahblahblah"/*SR.Format(SR.ImpersonationLevelNotSupported, tokenImpersonationLevel)*/));
```
iamjasonp: `blahblah` needs to be fixed :smile:

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KerberosRequestorSecurityToken.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        internal KerberosRequestorSecurityToken(string servicePrincipalName, TokenImpersonationLevel tokenImpersonationLevel, NetworkCredential networkCredential, string id)
+        {
+            if (servicePrincipalName == null)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("servicePrincipalName");
+            if (tokenImpersonationLevel != TokenImpersonationLevel.Identification && tokenImpersonationLevel != TokenImpersonationLevel.Impersonation)
+            {
+                // TODO: Add String resource
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException("tokenImpersonationLevel",
+                    "blah"/*SR.Format(SR.ImpersonationLevelNotSupported, tokenImpersonationLevel)*/));
```
iamjasonp: Fix?

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KerberosRequestorSecurityToken.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+            if (id == null)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("id");
+
+            _servicePrincipalName = servicePrincipalName;
+            if (networkCredential != null && networkCredential != CredentialCache.DefaultNetworkCredentials)
+            {
+                if (string.IsNullOrEmpty(networkCredential.UserName))
+                {
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgument("blahblah"/*SR.ProvidedNetworkCredentialsForKerberosHasInvalidUserName*/);
```
iamjasonp: ditto

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/KerberosSecurityTokenProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            this._servicePrincipalName = servicePrincipalName;
+            this._tokenImpersonationLevel = tokenImpersonationLevel;
+            this._networkCredential = networkCredential;
+        }
```
iamjasonp: NIT: `this.` should be removed

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -169,6 +188,24 @@ protected ConnectionBufferPool WebSocketBufferPool
...
+            get
+            {
+                if (_hashAlgorithm == null)
+                {
+                    _hashAlgorithm = SHA512.Create();
+                }
+                else
+                {
+                    _hashAlgorithm.Initialize();
+                }
```
iamjasonp: Could you explain the `else` behaviour here? 
Feels like `_hashAlgorithm.Initialize();` should go into `if (_hashAlgorithm == null)`, but I don't know the usage pattern of HashAlgorithm well enough to make a definitive comment.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceModelHttpMessageHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            get { return _innerHandler.AutomaticDecompression; }
+            set { _innerHandler.AutomaticDecompression = value; }
+        }
+
+        public CookieContainer CookieContainer
+        {
+            get { return _innerHandler.CookieContainer; }
+            set { _innerHandler.CookieContainer = value; }
+        }
```
iamjasonp: How are we (or are we?) going to validate that CookieContainer is specified? Or will that be taken care of when the HttpClientHandler or WinHttpHandler is used?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportSecurityHelpers.cs
```diff
@@ -5,32 +5,199 @@
...
+using System.Security.Cryptography.X509Certificates;
+using System.Security.Principal;
 using System.ServiceModel.Security;
 using System.ServiceModel.Security.Tokens;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace System.ServiceModel.Channels
 {
     internal static class TransportSecurityHelpers
```
iamjasonp: Cool! it looks pretty close to my implementation which I was re-creating in a separate change :)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportSecurityHelpers.cs
```diff
@@ -5,32 +5,199 @@
...
+                certificateTokenRequirement.RequireCryptographicToken = false;
+                certificateTokenRequirement.TransportScheme = transportScheme;
+                if (channelParameters != null)
+                {
+                    certificateTokenRequirement.Properties[ServiceModelSecurityTokenRequirement.ChannelParametersCollectionProperty] = channelParameters;
+                }
+                return tokenManager.CreateSecurityTokenProvider(certificateTokenRequirement);
+            }
+            return null;
+        }
```
iamjasonp: Don't we need to port some of the ones that create the `RecipientServiceModelSecurityTokenRequirement` as well?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
```diff
@@ -683,6 +779,54 @@ internal static string GetKeyDerivationAlgorithm(SecureConversationVersion versi
...
+                PrepareNetworkCredentialWorker();
+            }
+        }
+
+        // Since this takes a lock, it probably won't be inlined, but the typical case will be.
+        static void PrepareNetworkCredentialWorker()
+        {
+            lock (s_dummyNetworkCredentialLock)
+            {
+                s_dummyNetworkCredential = new NetworkCredential("dummy", "dummy");
```
iamjasonp: What's this dummy used for?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
```diff
@@ -683,6 +779,54 @@ internal static string GetKeyDerivationAlgorithm(SecureConversationVersion versi
...
 
+        [SecuritySafeCritical]
+        internal static void FixNetworkCredential(ref NetworkCredential credential)
+        {
+            if (credential == null)
+            {
+                return;
+            }
+            string username = NetworkCredentialHelper.UnsafeGetUsername(credential);
+            string domain = NetworkCredentialHelper.UnsafeGetDomain(credential);
```
iamjasonp: NIT: Are these really "Unsafe" anymore?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
```diff
@@ -578,6 +660,20 @@ private static NetworkCredential UnsafeGetDefaultNetworkCredentials()
             {
                 return CredentialCache.DefaultNetworkCredentials;
             }
+
+            internal static string GetCurrentUserIdAsString(NetworkCredential credential)
+            {
+#if FEATURE_NETNATIVE
+                // There's no way to retrieve the current logged in user Id in UWP apps
+                // so returning a username which is very unlikely to be a real username;
+                return "_______****currentUser****_______";
```
iamjasonp: Can we use something along the lines of this? 
https://code.msdn.microsoft.com/windowsapps/User-domain-name-sample-85ce3e49/sourcecode?fileId=51014&pathId=498305094

(Was using this in a toy app of mine, wondering if we can work it into here)

src/System.Private.ServiceModel/src/windows/project.json
```diff
@@ -0,0 +1,64 @@
...
+        "System.Security.Principal.Windows": "4.0.0-beta-*"
+      }
+    },
+    "netcore50": {
+      "dependencies": {
+        "System.Runtime.WindowsRuntime": "4.0.11-beta-*"
+      }
+    }
+  }
+}
```
iamjasonp: What's this for?

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
```diff
@@ -119,7 +125,7 @@ public static void SameBinding_Soap11_EchoString()
     }
 
     // Client and Server bindings setup exactly the same using Soap12
-    [Fact]
+    //[Fact]
```
iamjasonp: Are these comment-outs intentional?

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
```diff
@@ -22,6 +22,11 @@
   <PropertyGroup Condition="'$(IsNETNativeLibrary)' != 'true'">
     <DefineConstants>$(DefineConstants);FEATURE_CORECLR</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetsWindows)' == 'true'">
+    <DefineConstants>$(DefineConstants);TARGETS_WINDOWS</DefineConstants>
+	<ProjectJson>$(MsBuildThisFileDirectory)windows\project.json</ProjectJson>
+	<ProjectLockJson>$(MsBuildThisFileDirectory)windows\project.lock.json</ProjectLockJson>
+  </PropertyGroup>
```
iamjasonp: Wondering if

```xml
   <ItemGroup>
   </ItemGroup>
```

below (L37) should now be inside a Condition as well now

src/System.ServiceModel.Primitives/ref/project.lock.json
```diff
@@ -2209,21 +2209,21 @@
         "lib/net46/System.Security.Cryptography.Primitives.dll",
         "lib/xamarinios10/_._",
         "lib/xamarinmac20/_._",
-        "ref/dotnet5.4/System.Security.Cryptography.Primitives.dll",
+        "ref/dotnet5.1/System.Security.Cryptography.Primitives.dll",
```
iamjasonp: Why did these generations go down from 5.4 to 5.1?

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
hongdai: The same doubt. Also are there any new tests added?

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/HttpsClientCertificateResource.cs
```diff
@@ -0,0 +1,42 @@
...
+            serviceHost.Credentials.ClientCertificate.Authentication.CustomCertificateValidator = new MyX509CertificateValidator("DO_NOT_TRUST_WcfBridgeRootCA");
+            serviceHost.Credentials.ServiceCertificate.SetCertificate(StoreLocation.LocalMachine,
+                                                      StoreName.My,
+                                                      X509FindType.FindByThumbprint,
+                                                      thumbprint);
+            //serviceHost.Credentials.ClientCertificate.Authentication.CertificateValidationMode = X509CertificateValidationMode.None;
+            //serviceHost.Credentials.ServiceCertificate.SetCertificate(StoreLocation.LocalMachine,
+            //                                          StoreName.My,
+            //                                          X509FindType.FindByThumbprint,
+            //                                          thumbprint);
```
hongdai: Nit: Remove the commented out code.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/KerberosSecurityTokenProvider.cs
mconnew: Yep, that was put in there as the resources didn't exist! I forgot to fix it :smile: At least it was obvious that it was a missed item.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: The original code called Initialize before handing it out each time. With hash algorithms, you maintain a state that is modified/transformed with each new piece of data that you add to the hash. When you are done, there's a final transform done to prevent attacks based on appending data, in other words finalizes the hash such that no more data can be added. Initialize resets this state data. The docs don't say that this is needed each time, but looking at the implementation, it certainly looks like the state isn't implicitly reset each time and that you would end up with an invalid hash if this isn't done. Btw, desktop uses sha1, I switched to SHA512 so that it should hopefully be a decade before we get told we need to stop using this hash algorithm. There's no back compat risks as it's only used in-process and not persisted/transmitted.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceModelHttpMessageHandler.cs
mconnew: HttpClientHandler/WinHttpHandler will throw if things aren't right.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportSecurityHelpers.cs
mconnew: Correct, HTTPS only needs InitiatorServiceModelSecurityTokenRequirement

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
mconnew: On Desktop there is/was an issue when no NetworkCredential objects have ever been created yet and two threads tried to create a NetworkCredential object at the same time. There was some thread safety issue (I don't know exactly what the nature of the problem is). So the logic is such that we always create a dummy credential if it hasn't been created already, and do it inside a lock to ensure there's no problems. From then on, we can create a NetworkCredential object outside of a lock with no issues. This might or might not still be the case. I didn't want to spend the time to investigate if this is still an issue now as it would delay the feature.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
mconnew: This is all about partial trust code, which has gone away so regardless, I can remove the attribute. I suspect it is still technically classed as unsafe as it will do P/Invoke work on some platforms to get a security context for the currently logged in user but as we don't need to worry about partial trust, naming it unsafe doesn't gain us anything. I'll remove the attribute and rename a few things.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
mconnew: Having said this, I went and tracked down the original bug and the original issue was fixed before the code using this was released back in 2005! I'm removing the "workaround".

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
mconnew: I looked into that, but unfortunately that's no longer supported in Win10. As per the [documentation](https://msdn.microsoft.com/en-us/library/windows/apps/windows.system.userprofile.userinformation.aspx) for that api:  
The **UserInformation** class is not supported on Windows 10 or later. Use the [**User**](https://msdn.microsoft.com/en-us/library/windows/apps/windows.system.user.aspx) class instead  
Unfortunately, I wasn't able to find a way to get the User object for the currently logged in user. If you know the Win10 equivalent, I'd happily put it in there.

src/System.Private.ServiceModel/src/windows/project.json
mconnew: Unfortunately no. There is no mechanism in project.json files to specify a dependency which only exists for a particular OS. This means we need to override the ProjectJson and ProjectLockJson build properties to pick up a different project.json file. This is a real pain, but will eventually be resolved in one of two ways. Either OS specific dependency support will be added (there is a bug open for this) or HttpClientHandler will have the appropriate API's needed so we don't need to use WinHttpHandler. This will be post rtm so unavoidable for now.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
mconnew: Nope, this was a mistake. When debugging xunit tests in VS, it's nice to only run the single test. Without a working VS test runner for corefx, this is the best that can be done.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
mconnew: I wouldn't add a condition. I think I should just add the other one in the same item group. The None group simply says to show the file in the VS solution explorer. It has no functional effect. I'll add it in so it doesn't get forgotten.

src/System.ServiceModel.Primitives/ref/project.lock.json
mconnew: I have no idea. In fact, I'm not sure how we're going to be able to update this file now as a normal build on windows will only update the other file. I'll try reverting all lock files and rebuilding again.

mconnew:
I also need to open an issue for enabling client certificates on Linux. Right now we have no way to achieve that. Then I can add an active issue to this new test which is Linux specific so I can get the CI tests passing on Linux.

src/System.Private.ServiceModel/src/windows/project.json
zhenlan: Can we ask for API stubs to be added to HttpClientHandler (for *nix?) so we don't have to build for platform specific? It's my understanding this is the first one that is going to break our platform independent state.

mconnew: There's two bugs open for this. One for adding client certificates, the other for adding certificate validation. They are dotnet/corefx#4479 and dotnet/corefx#4476. I've been told the scope for this work is post rtm so we either don't have certificate validation/client certificates on any platform, or we do this code splitting. We will be able to go back to platform independent again at a later point in time.

src/System.ServiceModel.Primitives/ref/project.lock.json
mconnew: On further testing, looks like the RestorePackages target still updates this. Also moving the generation down to 5.1 means that the Crypto libraries are now available to more platforms. It doesn't mean we're now using an older version.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
iamjasonp: Got it - thanks for the explanation.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
iamjasonp: Sigh :( Got it, I do the same thing, wish there was a good way to run a single test

src/System.ServiceModel.Primitives/ref/project.lock.json
iamjasonp: @ericstj, can you comment on whether a generations rollback like this is acceptable?

src/System.Private.ServiceModel/src/Resources/Strings.resx
```diff
@@ -6750,4 +6750,10 @@
   <data name="X509ChainBuildFail" xml:space="preserve">
     <value>The X.509 certificate {0} chain building failed. The certificate that was used has a trust chain that cannot be verified. Replace the certificate or change the certificateValidationMode. {1}</value>
   </data>
+  <data name="ImpersonationLevelNotSupported" xml:space="preserve">
+    <value>The athentication modes using Kerberos do not support the impersonation level '{0}'. Specify identification or impersonation.</value>
```
iamjasonp: Typo: "athentication" should be "authentication"

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
```diff
@@ -304,11 +304,11 @@ public DefaultTaskSchedulerAwaiter GetAwaiter()
 
     // Async methods can't take an out (or ref) argument. This wrapper allows passing in place of an out argument
     // and can be used to return a value via a method argument.
-    public class OutWrapper<T> where T : class
+    public class OutWrapper<T>
```
iamjasonp: Question only: any reason why we got rid of this?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -29,14 +36,18 @@ internal class HttpChannelFactory<TChannel>
         private AuthenticationSchemes _authenticationScheme;
         private HttpCookieContainerManager _httpCookieContainerManager;
         private HttpClient _httpClient;
+        volatile MruCache<string, string> _credentialHashCache;
+        private volatile MruCache<string, HttpClient> _httpClientCache;
         private int _maxBufferSize;
         private SecurityCredentialsManager _channelCredentials;
+        SecurityTokenManager _securityTokenManager;
```
iamjasonp: NIT: access modifiers needed. 

Question only: volatile needed because we're doing the double-locked initialization, correct?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -281,11 +366,46 @@ public override int GetMaxBufferSize()
             return MaxBufferSize;
         }
 
+        SecurityTokenProviderContainer CreateAndOpenTokenProvider(TimeSpan timeout, AuthenticationSchemes authenticationScheme,
```
iamjasonp: NIT: access modifier

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -411,6 +532,17 @@ protected override void OnClosed()
             }
         }
 
+        string AppendWindowsAuthenticationInfo(string inputString, NetworkCredential credential,
```
iamjasonp: NIT: access modifier (i'll stop pointing these out)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpsChannelFactory.cs
```diff
@@ -14,14 +18,19 @@ namespace System.ServiceModel.Channels
     internal class HttpsChannelFactory<TChannel> : HttpChannelFactory<TChannel>
     {
         private bool _requireClientCertificate;
-
-
+        X509CertificateValidator _sslCertificateValidator;
+        private Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> _remoteCertificateValidationCallback;
```
iamjasonp: NIT: access modifiers in this file

src/System.Private.ServiceModel/src/System/ServiceModel/Security/ClientCredentialsSecurityTokenManager.cs
```diff
@@ -108,10 +153,53 @@ public override SecurityTokenProvider CreateSecurityTokenProvider(SecurityTokenR
...
+                //    result = CreateTlsnegoTokenProvider(initiatorRequirement, true);
+                //}
+                //else if (tokenType == ServiceModelSecurityTokenTypes.AnonymousSslnego)
+                //{
+                //    result = CreateTlsnegoTokenProvider(initiatorRequirement, false);
+                //}
+                //else if (tokenType == ServiceModelSecurityTokenTypes.SecureConversation)
+                //{
+                //    result = CreateSecureConversationSecurityTokenProvider(initiatorRequirement);
+                //}
```
iamjasonp: NIT: remove commented out code here?

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
mconnew: I needed an OutWrapper for a primitive (bool in this case)

mconnew:

mconnew:
I merged these changes as the only outstanding issues were nit's and Jason needs this code to be checked in. All tests pass, so any follow up issues can be addressed in a follow-up change if needed.

iamjasonp:
Thanks @mconnew. Changes LGTM and I too buddy built and verified that tests pass.

Issue #13, opened 2015-05-13
Support SecurityBindingElement


iamjasonp:
SecurityBindingElement is not yet supported and throws PlatformNotSupportedException

zhenlan:
@iamjasonp you may like to look into what takes to enable issue #81 first for this item.

jeffboulanger:
Hi @iamjasonp, would you be able to tell me if fixing this issue will also resolve #9.  We are at a stand still with upgrading our WIndows 8.1 app to Windows 10 because of this.

iamjasonp:
Hi @jeffboulanger - it should fix #9 as well, but we are very tightly scoping the work right now to a limited set of scenarios. Not everything will be supported out of the gate, but we'll continue to enable the scenarios as we go along. We'll probably see SSL out of the gate first, followed by other modes of transport security

iamjasonp:
I think we managed to confuse ourselves here. SecurityBindingElement mainly does with SOAP message security and mixed mode security. TransportSecurityBindingElement (#9) is what we're focusing on right now rather than this. 

@jeffboulanger, we just committed a first pass for Transport Security (We can now connect to a server successfully now with SSL security over net.tcp, but there are some issues as the client cert can't be specified yet due to us needing to enable some contracts and wait for UWP to open up some API surface). 

Can you comment further on what features you might need so we can take that into consideration when prioritizing?

jeffboulanger:
@iamjasonp Rather than list everything and possibly miss something, here is how we use WCF in our application.   We have about 10 different endpoints, all of which setup their bindings via the following code.

```csharp

    public class NoTimestampBasicHttpBinding : BasicHttpBinding
    {
        public NoTimestampBasicHttpBinding(BasicHttpSecurityMode securityMode, MessageVersion messageVersion = null)
            : base(securityMode)
        {
            CustomMessageVersion = messageVersion;
            IsNonSecure = securityMode == BasicHttpSecurityMode.TransportCredentialOnly;

            MaxBufferSize = 10485760;
            MaxBufferPoolSize = 10485760;
            MaxReceivedMessageSize = 10485760;
            ReaderQuotas = XmlDictionaryReaderQuotas.Max;
            AllowCookies = false;
            TextEncoding = Encoding.UTF8;
            TransferMode = TransferMode.Buffered;
        }

        public MessageVersion CustomMessageVersion
        {
            get;
            private set;
        }

        public bool IsNonSecure
        {
            get;
            private set;
        }

[...]

PR #545, opened 2015-11-24
Implement SecurityBindingElement.IsSetKeyDerivation

### LINKED ISSUE #540: Calling BasicHttpMessageSecurity.CreateMessageSecurity results in PNSE
In this line: 
https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/src/System/ServiceModel/BasicHttpMessageSecurity.cs#L66 

We are guaranteed that when we try to create the SecurityBindingElement, we will end up with a PNSE due to 

```csharp
    public virtual void SetKeyDerivation(bool requireDerivedKeys)
    {
        throw ExceptionHelper.PlatformNotSupported("SecurityBindingElement.SetKeyDerivation is not supported.");
    }
```

This is blocking us from the scenario described here: https://github.com/dotnet/wcf/issues/13#issuecomment-137860026

We need to implement SecurityBindingElement.SetKeyDerivation to not throw PNSE when we are calling `SecurityBindingElement.SetKeyDerivation(false)` - only when `true`. This should have no impact given we haven't implemented WS-Sec and WS-RM in WCF Core yet. 

Thanks @jeffboulanger for pointing out this issue!

iamjasonp:
Now that we have started implementing the tokens/token parameters, we can now start implementing SecurityBindingElement.IsSetKeyDerivation instead of throwing `PlatformNotSupportedException`.

Fixes #540

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityBindingElement.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityBindingElement.cs
@@ -271,12 +271,12 @@ namespace System.ServiceModel.Channels
 
         public virtual void SetKeyDerivation(bool requireDerivedKeys)
         {
-            throw ExceptionHelper.PlatformNotSupported("SecurityBindingElement.SetKeyDerivation is not supported.");
+            EndpointSupportingTokenParameters.SetKeyDerivation(requireDerivedKeys);
         }
 
         internal virtual bool IsSetKeyDerivation(bool requireDerivedKeys)
         {
-            if (!this.EndpointSupportingTokenParameters.IsSetKeyDerivation(requireDerivedKeys))
+            if (!EndpointSupportingTokenParameters.IsSetKeyDerivation(requireDerivedKeys))
                 return false;
 
             return true;


iamjasonp:
@jeffboulanger - this change should fix your scenario. I ran the code from https://github.com/dotnet/wcf/issues/13#issuecomment-137860026 under our test bed and confirmed that after this change, the PNSE you encountered is no longer thrown

roncain:
LGTM.

@iamjasonp -- can we add the test you used to verify this? Or open an issue to do it in a separate PR?

iamjasonp:
I'll extract bits of what @jeffboulanger provided us in https://github.com/dotnet/wcf/issues/13#issuecomment-137860026 as a test case (or test cases) for us to use.

jeffboulanger:
Thanks @iamjasonp, When will this fix be in a package on nuget?  Is it something that is updated often?

roncain:
Hi @jeffboulanger -- the packages get pushed to NuGet (right now) only at release milestones.  But every night the latest packages are pushed to https://www.myget.org/F/dotnet-core/ .  If you add that location as another package source, you should be able to see them, though you will have to update your project.json dependencies to the new version.  You can browse this package source at https://www.myget.org/gallery/dotnet-core.  So once this PR gets merged, you should be able to see them in the following night's package build.

jeffboulanger:
Gotcha, I currently have that location as a package source, but didn't know the release cadence for it.  Thanks for the info, I'll watch for the PR merge.

iamjasonp:
Let me merge this first to unblock @jeffboulanger, I'll get a test in shortly and open an issue to track

PR #546, opened 2015-11-25
Make WebSocketSettings.DisablePayloadMasking throw PNSE.

### LINKED ISSUE #537: WebSocketSettings.DisablePayloadMasking not supported
According to @mconnew this property is not supported but currently simply does nothing when used, it needs to throw PNSE and a unit test confirming this should be written at the same time.

shmao:
Fix #537.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportSettings.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportSettings.cs
@@ -115,7 +115,6 @@ namespace System.ServiceModel.Channels
             }
         }
 
-        [DefaultValue(WebSocketDefaults.DisablePayloadMasking)]
         public bool DisablePayloadMasking
         {
             get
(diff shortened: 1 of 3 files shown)


shmao:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportSettings.cs
```diff
@@ -21,14 +21,12 @@ public sealed class WebSocketTransportSettings : IEquatable<WebSocketTransportSe
...
         private TimeSpan _keepAliveInterval;
         private string _subProtocol;
-        private bool _disablePayloadMasking;
 
         public WebSocketTransportSettings()
         {
             _transportUsage = WebSocketDefaults.TransportUsage;
             _keepAliveInterval = WebSocketDefaults.DefaultKeepAliveInterval;
             _subProtocol = WebSocketDefaults.SubProtocol;
-            _disablePayloadMasking = WebSocketDefaults.DisablePayloadMasking;
```
iamjasonp: We should remove this too then? 

https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportDefaults.cs#L324

iamjasonp:
LGTM

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportSettings.cs
shmao: You're right. Removed it.

PR #548, opened 2015-11-26
MessageParameterAttribute for return response not working

### LINKED ISSUE #527: Customer report MessageParameterAttribute for return response not working
We expose MessageParameterAttribute in the contract. We need to investigate if this is a bug.

Customer report below does not get deserialized
[System.ServiceModel.OperationContractAttribute(Action = "", ReplyAction = "*")]
[return: System.ServiceModel.MessageParameterAttribute(Name = "result")]
Task<LoginResult> loginAsync(string username, string password);

However if the customer change the response SOAP XML, result node to loginResult, it works.

hongdai:
To enable it
* add back MessageParameterAttribute in CreateParameterPartDescription
* use methodInfo.ReturnParameter instead of methodInfo.ReturnType.
In the full framework, we use ReturnTypeCustomAttributes, it's essentially
ReturnParameter.
* Add a scenario test.

Fixes #527

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Description/CustomAttributeProvider.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Description/CustomAttributeProvider.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Reflection;
@@ -83,7 +84,10 @@ namespace System.ServiceModel.Description
                 case AttributeProviderType.MemberInfo:
                     return this.MemberInfo.GetCustomAttributes(attributeType, inherit).ToArray();
                 case AttributeProviderType.ParameterInfo:
-                    return this.ParameterInfo.GetCustomAttributes(attributeType, inherit).ToArray();
+                    //GetCustomAttributes could return null instead of empty collection for a known System.Relection issue, workaround the issue by explicitly checking the null
+                    IEnumerable<Attribute> customAttributes = null;
+                    customAttributes = this.ParameterInfo.GetCustomAttributes(attributeType, inherit);
+                    return customAttributes == null ? null: customAttributes.ToArray();
             }
             Contract.Assert(false, "This should never execute.");
             throw ExceptionHelper.PlatformNotSupported();
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/ServiceInterfaces.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/ServiceInterfaces.cs
@@ -105,6 +105,10 @@ public interface IWcfServiceGenerated
 
     [System.ServiceModel.OperationContractAttribute(Action = "http://tempuri.org/IWcfService/Echo", ReplyAction = "http://tempuri.org/IWcfService/EchoResponse")]
     System.Threading.Tasks.Task<string> EchoAsync(string message);
+
+    [System.ServiceModel.OperationContractAttribute(Action = "http://tempuri.org/IWcfService/EchoMessageParameter", ReplyAction = "http://tempuri.org/IWcfService/EchoMessageParameterResponse")]
+    [return: System.ServiceModel.MessageParameterAttribute(Name = "result")]
+    string EchoMessageParameter(string name);
 }
 
 // 
--- a/src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.Net.Http;
 using System.ServiceModel;
+using System.ServiceModel.Channels;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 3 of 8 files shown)


hongdai:
@dotnet-bot please test outerloop

hongdai:

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
```diff
@@ -189,6 +190,41 @@ public static void DefaultSettings_Echo_RoundTrips_String_Streamed_Async_WithSin
...
+                errorBuilder.AppendLine(String.Format("Expected response from Service: {0} Actual was: {1}", "Hello", result));
+            }
+
+            // *** CLEANUP *** \\
+            factory.Close();
+        }
+        catch (Exception ex)
+        {
+            errorBuilder.AppendLine(String.Format("Unexpected exception was caught: {0}", ex.ToString()));
+        }
```
iamjasonp: missing finally block with CloseAllCommunicationObjects per other tests

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
```diff
@@ -189,6 +190,41 @@ public static void DefaultSettings_Echo_RoundTrips_String_Streamed_Async_WithSin
...
+            // Note the service interface used.  It was manually generated with svcutil.
+            ChannelFactory<IWcfServiceGenerated> factory = new ChannelFactory<IWcfServiceGenerated>(customBinding, new EndpointAddress(Endpoints.DefaultCustomHttp_Address));
+            IWcfServiceGenerated serviceProxy = factory.CreateChannel();
+
+            // *** EXECUTE *** \\
+            string result = serviceProxy.EchoMessageParameter("Hello");
+            if (!string.Equals(result, "Hello Hello"))
+            {
+                errorBuilder.AppendLine(String.Format("Expected response from Service: {0} Actual was: {1}", "Hello", result));
+            }
```
iamjasonp: Confused here. we expect "Hello" back or "Hello Hello" back? 
*TINY* NIT: Consider putting the echo string into a variable so we don't have too many magic strings

Why can't we just use `Assert.True(string.Equals(result, "Hello", "Expected response from service..."));`

src/System.Private.ServiceModel/src/System/ServiceModel/Description/CustomAttributeProvider.cs
```diff
@@ -83,7 +84,10 @@ public object[] GetCustomAttributes(Type attributeType, bool inherit)
                 case AttributeProviderType.MemberInfo:
                     return this.MemberInfo.GetCustomAttributes(attributeType, inherit).ToArray();
                 case AttributeProviderType.ParameterInfo:
-                    return this.ParameterInfo.GetCustomAttributes(attributeType, inherit).ToArray();
+                    //GetCustomAttributes could return null instead of empty collection for unknown reason here, workaround the issue by explicitly checking the null
+                    IEnumerable<Attribute> customAttributes = null;
```
roncain: We need to understand why it is returning null, because an empty array may not be the appropriate thing to do. For history, look at #315.  In that case, an empty array was the wrong thing to do.

src/System.Private.ServiceModel/tests/Common/Scenarios/ServiceInterfaces.cs
```diff
@@ -105,6 +105,10 @@ public interface IWcfServiceGenerated
 
     [System.ServiceModel.OperationContractAttribute(Action = "http://tempuri.org/IWcfService/Echo", ReplyAction = "http://tempuri.org/IWcfService/EchoResponse")]
     System.Threading.Tasks.Task<string> EchoAsync(string message);
+
+    [System.ServiceModel.OperationContractAttribute(Action = "http://tempuri.org/IWcfService/EchoMessageParameter", ReplyAction = "http://tempuri.org/IWcfService/EchoMessageParameterResponse")]
+    [return: System.ServiceModel.MessageParameterAttribute(Name = "result")]
+    string EchoMessageParameter(string name);
```
roncain: Consider updating this same service interface under the unit tests or consolidating them.  We have duplicate code in several places that should probably be kept in sync or combined.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
```diff
@@ -189,6 +190,41 @@ public static void DefaultSettings_Echo_RoundTrips_String_Streamed_Async_WithSin
         Assert.True(success, "Test Scenario: DefaultSettings_Echo_RoundTrips_String_Streamed_Async_WithSingleThreadedSyncContext timed-out.");
     }
 
+    [Fact]
+    [OuterLoop]
+    public static void ServiceContract_EchoMessageParameter()
+    {
```
roncain: We need a method name that describes what the test does.  Something like "ServiceContract_Call_Operation_With_MessageParameterAttribute"

src/System.Private.ServiceModel/src/System/ServiceModel/Description/CustomAttributeProvider.cs
hongdai: In this case:
The attributeType is MessageParameterAttribute, derives from Attribute.
The Parameter Info is System.ServiceModel.Channels.Message, without any custom attributes defined.

We expect System.Reflection.RuntimeParameterInfo.GetCustomAttribute returns an empty collection per it's definition about return value:
        // Returns:
        //     A collection of the custom attributes that are applied to element and that match
        //     attributeType, or an empty collection if no such attributes exist.

It should be a System.Reflection bug. I will attempt a standalone simple repro for System.Reflection team.

hongdai: Ron found this is a known issue to system.reflection team. I'll update the comment.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
hongdai: Will add it.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
hongdai: Make sense. Will make the change.

src/System.Private.ServiceModel/tests/Common/Scenarios/ServiceInterfaces.cs
hongdai: I will update the same service interface under the unit tests.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
hongdai: Make sense. Will make the change.

hongdai:

iamjasonp:
Thanks Hong - LGTM.

Issue #413, opened 2015-10-12
Incorporate use of NegotiateStream when available.


roncain:
The Networking team plans to check-in a PNSE version of NegotiateStream into System.Net.Security soon (still under discussion whether it originates from TFS or directly in CoreFx GitHub.

When this happens, we need to add back code to use NegotiateStream and remove any remaining mock-ups.

Reactivate any tests currently deactivated due to lack of NegotiateStream, run them, and update the ActiveIssue to reflect the stackframe where the test fails.  Ensure there is a CoreFx issue for the specific PNSE encountered.

zhenlan:
@iamjasonp this is pending on the availability of networking API stubs that throw PNSE. We (WCF) would like to have this wired up whenever they are available.

iamjasonp:
Dependent on dotnet/corefx#1844

roncain:
Though lower priority than #308, I think we should take a dependency on the new NegotiateStream as soon as it becomes available.  This will flush out any mock-ups we have left and also verify whether there is a true Linux version of the stub available.  In x-plat testing, the hardest issue I face in debugging right now is finding either packages missing Linux implementations or packages containing Windows implementations where they should be for Linux.  Taking the dependency (even if not called) will verify there is a true Linux stub version available.

PR #550, opened 2015-11-30
Light up code for supporting Windows stream security

### LINKED ISSUE #300: Support Windows Stream Security
Add support for Windows Stream Security and associated binding elements. 

The default constructor of NetTcpBinding will attempt to initialize a WindowsStreamSecurityBindingElement upgrade, and as soon as we try to upgrade the connection, an exception will be thrown at `System.ServiceModel.Channels.WindowsStreamSecurityBindingElement.BuildClientStreamUpgradeProvider`

Dependencies / things to fix; 
* NegotiateStream
* In `/src/System.Private.ServiceModel/src/System/IdentityModel/Policy/UnconditionalPolicy.cs` - need to re-enable to code near line 90

(add to this list as needed)

iamjasonp:
The code still will not fully work yet as we are dependent on System.Net.Security.NegotiateStream being lit up. Once that is lit up we should be able to get to the inevitable other problems that will surface

We also yank the `includeWindowsGroups` as an option right now

The reenablement of tests and writing of them is being tracked on #301

Resolves #300


iamjasonp:

roncain:
LGTM.

Nit: if this resolves #300, is there a separate tracking issue for any follow-on work when NegotiateStream is available?

roncain:
Question: was this built and tested against NET Native?  I'm mostly curious whether the version upgrade has any issues building with version numbers for NET Native.

iamjasonp:
> Nit: if this resolves #300, is there a separate tracking issue for any follow-on work when NegotiateStream is available?
I will open another tracking issue for watching NegotiateStream

> Question: was this built and tested against NET Native? I'm mostly curious whether the version upgrade has any issues building with version numbers for NET Native.
Yes, this is a concern to me, and it's not apparent to me either that NegotiateStream-enabled packages are available for .NET Native. I had kicked off a .NET Native build but had to leave before the build finished; I'll check on it when I return.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WindowsStreamSecurityUpgradeProvider.cs
```diff
@@ -0,0 +1,425 @@
...
+                    TD.WindowsStreamSecurityOnInitiateUpgrade();
+                }
+
+                // prepare
+                InitiateUpgradePrepare(stream, out negotiateStream, out targetName, out identity);
+
+                // authenticate
+                try
+                {
+                    negotiateStream.AuthenticateAsClient(_credential, targetName, _parent.ProtectionLevel, _impersonationLevel);
```
mconnew: The NCL team are removing the sync versions of this method in this class and in SslStream. You need to switch to AuthenticateAsClientAsync method.

iamjasonp: Ooh, good catch. This was supposed to have changed but I guess I missed it.

iamjasonp:
This is mergeable pending a build on .NET Native, which I have not successfully managed yet :( @roncain, any chance you could help buddy build this?

iamjasonp:
Actually, upon reading, I think this resolves #413 not #300. #300 is dependent on #413, which is our tracking bug to ensure the NegoStream stubs were in. 

dotnet/corefx#1844 is what we will track (from #300) to actually light up support for Windows Stream Security. I'm assuming there will be more work involved once NegoStream is lit up as we're not currently able to test the implementation.

iamjasonp:

iamjasonp:
I did the #if commit separately so that it's in one change, and we can revert it quickly

iamjasonp:
Adding link to commit https://github.com/iamjasonp/wcf/commit/f84ee4de9a75341512ffb1882badeb33f9b17668 as there are some comments sitting in that commit

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
```diff
@@ -455,6 +456,168 @@ internal static string GenerateId()
...
+            Claim identityClaim;
+            Claim primaryPrincipal;
+            if (principalName.Contains("@") || principalName.Contains(@"\"))
+            {
+                identityClaim = new Claim(ClaimTypes.Upn, principalName, Rights.Identity);
+#if FEATURE_CORECLR
+                primaryPrincipal = Claim.CreateUpnClaim(principalName);
+#else 
+                throw ExceptionHelper.PlatformNotSupported("UPN claim not supported on CoreCLR"); 
+#endif // FEATURE_CORECLR
```
iamjasonp: Oops, this message needs to change. However, FEATURE_CORECLR is correct because of this:
https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/Claim.cs#L148

 so we should be consistent here.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimSet.cs
```diff
@@ -31,6 +32,26 @@ public static ClaimSet System
             }
         }
 
+        public static ClaimSet Windows
+        {
+            get
+            {
+#if FEATURE_NETNATIVE // NegotiateStream
+                throw ExceptionHelper.PlatformNotSupported("Windows Stream Security is not supported on UWP yet");
```
roncain: Question: does compiling for NET Native give a warning about not using s_windows or assigning it a value? If so I wonder if its declaration should also be conditional on FEATURE_NETNATIVE.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -1,16 +1,349 @@
...
+    public class WindowsClaimSet : ClaimSet, IIdentityInfo, IDisposable
+    {
+        internal const bool DefaultIncludeWindowsGroups = true;
+#if FEATURE_NETNATIVE //NegotiateStream
+        public override Claim this[int index]
+        {
+            get
+            {
+                throw ExceptionHelper.PlatformNotSupported("Windows Stream Security is not supported on UWP yet"); 
+            }
```
roncain: Nit: could use a const for this exception message

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -1,16 +1,349 @@
...
+            return _disposed ? base.ToString() : SecurityUtils.ClaimSetToString(this);
+        }
+
+        [Fx.Tag.SecurityNote(Critical = "Uses critical type SafeHGlobalHandle.",
+            Safe = "Performs a Demand for full trust.")]
+        [SecuritySafeCritical]
+        public static bool TryCreateWindowsSidClaim(WindowsIdentity windowsIdentity, out Claim claim)
+        {
+            throw ExceptionHelper.PlatformNotSupported();
+
```
roncain: Nit: could use an exception message for PNSE

src/System.Private.ServiceModel/src/System/IdentityModel/SecurityUtils.cs
```diff
@@ -209,6 +211,63 @@ internal static IIdentity CloneIdentityIfNecessary(IIdentity identity)
...
         }
 
+#if !FEATURE_NETNATIVE // NegotiateStream
+        /// <SecurityNote>
+        /// Critical - calls two critical methods: UnsafeGetWindowsIdentityToken and UnsafeCreateWindowsIdentityFromToken
+        /// Safe - "clone" operation is considered safe despite using WindowsIdentity IntPtr token
+        ///        must not let IntPtr token leak in or out
+        /// </SecurityNote>
+        [SecuritySafeCritical]
+        internal static WindowsIdentity CloneWindowsIdentityIfNecessary(WindowsIdentity wid)
```
roncain: Is [SecuritySafeCritical] even needed now?  I think we've been removing them over time.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimSet.cs
iamjasonp: No complaints seen on my end when I built it in my TFS environment, but I'll double check that.

roncain: I'm also buddy building on TFS now...

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -1,16 +1,349 @@
...
 using System.Security;
 using System.Security.Principal;
 using System.ServiceModel;
 
 namespace System.IdentityModel.Claims
 {
+    public class WindowsClaimSet : ClaimSet, IIdentityInfo, IDisposable
+    {
+        internal const bool DefaultIncludeWindowsGroups = true;
+#if FEATURE_NETNATIVE //NegotiateStream
```
mconnew: Could we have 2 files for this? One called WindowsClaimSet.NetNative.cs and the other WindowsClaimSet.CoreClr.cs? Having basically two compete implementations of this class in a single file doesn't seem like the right approach.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -1,16 +1,349 @@
...
+        {
+            get { return ClaimSet.Windows; }
+        }
+
+        public DateTime ExpirationTime
+        {
+            get { return _expirationTime; }
+        }
+
+        // Not sure yet if GroupSidClaimCollections are necessary in .NET Core
```
mconnew: Remove commented code. If it's necessary, it can be added.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -1,16 +1,349 @@
...
+                        if (claimType == sid.ClaimType)
+                        {
+                            yield return sid;
+                        }
+                    }
+                }
+
+                if (_includeWindowsGroups && (right == null || Rights.PossessProperty == right))
+                {
+                    // Not sure yet if GroupSidClaimCollections are necessary in .NET Core
```
mconnew: Remove commented code. If you want to catch this code path being hit, keep the if clause and throw.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -1,16 +1,349 @@
...
+        }
+
+        [Fx.Tag.SecurityNote(Critical = "Uses critical type SafeHGlobalHandle.",
+            Safe = "Performs a Demand for full trust.")]
+        [SecuritySafeCritical]
+        public static bool TryCreateWindowsSidClaim(WindowsIdentity windowsIdentity, out Claim claim)
+        {
+            throw ExceptionHelper.PlatformNotSupported();
+
+            // Not sure yet if GroupSidClaimCollections are necessary in .NET Core
```
mconnew: Remove commented code. It's obvious that there's missing implementation.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/WindowsClaimSet.cs
```diff
@@ -1,16 +1,349 @@
...
+        }
+
+        public override string ToString()
+        {
+            return _disposed ? base.ToString() : SecurityUtils.ClaimSetToString(this);
+        }
+
+        [Fx.Tag.SecurityNote(Critical = "Uses critical type SafeHGlobalHandle.",
+            Safe = "Performs a Demand for full trust.")]
+        [SecuritySafeCritical]
```
mconnew: I don't think we need SecuritySafeCritical as we're not supporting partial trust.

src/System.Private.ServiceModel/src/System/IdentityModel/Policy/UnconditionalPolicy.cs
```diff
@@ -87,22 +87,21 @@ private void Initialize(ClaimSet issuer, ClaimSet issuance, ReadOnlyCollection<C
...
+                {
+                    if (issuances[i] is WindowsClaimSet)
+                    {
+                        _disposable = true;
+                        break;
+                    }
+                }
+            }
 
             _disposable = false;
```
mconnew: If the if block above sets _disposable to true, you are just going to revert it here again. I presume this was temporary code that should have been removed.

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/WindowsSecurityToken.cs
```diff
@@ -0,0 +1,156 @@
...
+using System.Collections.ObjectModel;
+using System.Security.Principal;
+using System.ServiceModel;
+
+namespace System.IdentityModel.Tokens
+{
+
+    public class WindowsSecurityToken : SecurityToken, IDisposable
+    {
+#if FEATURE_NETNATIVE //NegotiateStream
```
mconnew: As earlier comment, I would prefer WindowsSecurityToken.CoreClr.cs and WindowsSecurityToken.NetNative.cs.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportDefaults.cs
```diff
@@ -140,8 +140,8 @@ public static class BinaryEncoderDefaults
 
     public static class TransportDefaults
     {
-        public const bool ExtractGroupsForWindowsAccounts = true;
         //public const bool ExtractGroupsForWindowsAccounts = SspiSecurityTokenProvider.DefaultExtractWindowsGroupClaims;
```
mconnew: If you are changing the default value, change the value of SspiSecurityTokenProvider.DefaultExtractWindowsGroupClaims and add a comment there that the default was different on Desktop.

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/WindowsSecurityToken.cs
iamjasonp: Hmm, the intent here is that we're stubbing this out while waiting for the S.Net.Security packages to show up. Should we do this even though the intent (hope?) is that we get rid of this in short order? 

If we end up having a substantial implementation difference "for real" then I agree we should split the files.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/WindowsClaimSet.cs
iamjasonp: Ok - will remove [SecuritySafeCritical] from new code.

src/System.Private.ServiceModel/src/System/IdentityModel/SecurityUtils.cs
iamjasonp: No, they aren't... thanks for the catch

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportDefaults.cs
iamjasonp: The default value hasn't changed. The reason why we had this is because SspiSecurityTokenProvider wasn't ported in the initial cut of code. Now that it's in we can revert back to having the correct  reference to SspiSecurityTokenProvider

src/System.Private.ServiceModel/src/System/IdentityModel/Policy/UnconditionalPolicy.cs
iamjasonp: Thanks for the catch :)

roncain:
LGTM subject to any comments @mconnew made.  We need to get this in asap because Linux tests are broken till we upgrade to the RC2 package dependencies, plus we need to upgrade build tools asap.

In future I recommend we keep version upgrade PR's independent of product change PRs so they are not blocked by code review.  No need to do it for this PR.

roncain:
@iamjasonp -- TFS just enabled the new Dev WorkFlow, so this probably means we need to rebuild against this PR in TFS again using it.  I'll try to do this but have no VPN access today.

iamjasonp:
> @iamjasonp -- TFS just enabled the new Dev WorkFlow, so this probably means we need to rebuild against this PR in TFS again using it. I'll try to do this but have no VPN access today.

Yep, that built for me safely today, and I was waiting for the internal TFS rolling build to stabilize. It has now, so merging. I'll keep an eye on it.

PR #554, opened 2015-11-30
Adding digest tests without domain controller

mconnew:
Also fixing digest authentication to allow explicit credentials. There is currently an impersonation constraint which only makes sense with message security and we don't expose the property to modify it. I'm removing that constraint so that explicit credentials can be used.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientAuthenticationManager.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientAuthenticationManager.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// 
+
+using System;
+using System.Net;
+
+namespace Infrastructure.Common
+{
+    public static class BridgeClientAuthenticationManager
+    {
+        private const string AuthenticationResourceName = "WcfService.TestResources.AuthenticationResource";
+        private const string UsernameKeyName = "authUsername";
+        private const string PasswordKeyName = "authPassword";
+        private static readonly object s_authLock = new object();
+        private static NetworkCredential s_networkCredential;
+
+        public static NetworkCredential NetworkCredential
+        {
+            get
+            {
+                if (s_networkCredential == null)
+                {
+                    lock (s_authLock)
+                    {
+                        if (s_networkCredential == null)
+                        {
+                            var response = BridgeClient.MakeResourceGetRequest(AuthenticationResourceName, null);
+                            var credential = new NetworkCredential();
+                            credential.UserName = response[UsernameKeyName];
+                            credential.Password = response[PasswordKeyName];
+                            if (string.IsNullOrEmpty(credential.UserName))
+                            {
+                                throw new NullReferenceException(UsernameKeyName);
+                            }
+
+                            if (string.IsNullOrEmpty(credential.Password))
+                            {
+                                throw new NullReferenceException(PasswordKeyName);
+                            }
+
+                            s_networkCredential = credential;
+                        }
+                    }
+                }
+
+                return s_networkCredential;
+            }
+        }
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Http/ClientCredentialTypeTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Http/ClientCredentialTypeTests.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// 
+
+using System.ServiceModel;
+
+using Infrastructure.Common;
+
+using Xunit;
+
+public static class Http_ClientCredentialTypeTests
+{
+    [Fact]
+    [OuterLoop]
+    public static void DigestAuthentication_Echo_RoundTrips_String_No_Domain()
+    {
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
+        string testString = "Hello";
+        BasicHttpBinding binding;
+
+        try
+        {
+            // *** SETUP *** \\
+            binding = new BasicHttpBinding(BasicHttpSecurityMode.TransportCredentialOnly);
+            binding.Security.Transport.ClientCredentialType = HttpClientCredentialType.Digest;
+            factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.Http_DigestAuth_NoDomain_Address));
+            factory.Credentials.HttpDigest.ClientCredential = BridgeClientAuthenticationManager.NetworkCredential;
+            serviceProxy = factory.CreateChannel();
+
+            // *** EXECUTE *** \\
+            string result = serviceProxy.Echo(testString);
+
+            // *** VALIDATE *** \\
+            Assert.True(result == testString, string.Format("Error: expected response from service: '{0}' Actual was: '{1}'", testString, result));
+
+            // *** CLEANUP *** \\
+            factory.Close();
+            ((ICommunicationObject)serviceProxy).Close();
+        }
+        finally
+        {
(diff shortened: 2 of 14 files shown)


mconnew:

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Http/ClientCredentialTypeTests.cs
```diff
@@ -0,0 +1,51 @@
...
+using Infrastructure.Common;
+
+using Xunit;
+
+public static class Http_ClientCredentialTypeTests
+{
+    [Fact]
+    [OuterLoop]
+    public static void DigestAuthentication_Echo_RoundTrips_String()
+    {
```
roncain: Nit: Recommend adding to method name "_No_Domain" or something to clue others into the fact no domain is required.  At some point we'll add [ConditionalFact] around any auth tests that do require it, and it will help keep them straight.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/DigestServiceAuthorizationManager.cs
```diff
@@ -0,0 +1,307 @@
...
+    {
+        private string _realm;
+        private readonly TimeSpan _nonceValidityTime = TimeSpan.FromMinutes(1);
+
+        public DigestServiceAuthorizationManager(string realm)
+        {
+            if (string.IsNullOrEmpty(realm))
+            {
+                throw new ArgumentException("realm");
+            }
```
roncain: ArgumentNullException?

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/DigestServiceAuthorizationManager.cs
```diff
@@ -0,0 +1,307 @@
...
+        {
+            object identitiesListObject;
+            if (!operationContext.ServiceSecurityContext.AuthorizationContext.Properties.TryGetValue("Identities",
+                out identitiesListObject))
+            {
+                identitiesListObject = new List<IIdentity>(1);
+                operationContext.ServiceSecurityContext.AuthorizationContext.Properties.Add("Identities", identitiesListObject);
+            }
+            var identities = identitiesListObject as IList<IIdentity>;
+            identities.Add(new GenericIdentity(digestState.Username, "GenericPrincipal"));
```
roncain: I think we would NullRef here if the TryGetValue above failed.  In that case, would we prefer to have a list containing only the GenericPrincipal, or an empty list.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/DigestServiceAuthorizationManager.cs
```diff
@@ -0,0 +1,307 @@
...
+
+            public DigestAuthenticationState(OperationContext operationContext, string realm)
+            {
+                _operationContext = operationContext;
+                _realm = realm;
+                _password = null;
+                _authorized = new bool?();
+                _authorizationHeader = GetAuthorizationHeader(operationContext, out _method);
+                if (_authorizationHeader.Length < 7)
+                {
```
roncain: Why 7?  Maybe add a const or local var with an explanatory name.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/DigestServiceAuthorizationManager.cs
```diff
@@ -0,0 +1,307 @@
...
+            public Dictionary<string, string> Parameters { get { return _authorizationParameters; } }
+
+            public string Password { set { _password = value; } }
+
+            public bool StaleNonce { get { return NonceExpiryTime < DateTime.Now; } }
+
+            public string Uri { get { return Parameters["uri"]; } }
+
+            public string Username { get { return Parameters["username"]; } }
+
```
roncain: Maybe make these string literal key names consts somewhere?

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/DigestServiceAuthorizationManager.cs
```diff
@@ -0,0 +1,307 @@
...
+                    _nonceString = nonceString.TrimEnd('=');
+                }
+            }
+
+            public Dictionary<string, string> Parameters { get { return _authorizationParameters; } }
+
+            public string Password { set { _password = value; } }
+
+            public bool StaleNonce { get { return NonceExpiryTime < DateTime.Now; } }
+
```
roncain: Nit: IsNonceStale? (StaleNonce makes make me think it returns a stale nonce).

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientAuthenticationManager.cs
```diff
@@ -0,0 +1,40 @@
...
+                if (s_networkCredential == null)
+                {
+                    lock (s_authLock)
+                    {
+                        if (s_networkCredential == null)
+                        {
+                            var response = BridgeClient.MakeResourceGetRequest(AuthenticationResourceName, null);
+                            var credential = new NetworkCredential();
+                            credential.UserName = response[UsernameKeyName];
+                            credential.Password = response[PasswordKeyName];
```
roncain: Nit: possibly worth an exception if the username and password are missing.  Your call.  I'm trying to think how best to diagnose auth failures that show up in the labs and looking for good ideas.

roncain:
LGTM, subject to the few observations I made.  It also passed OuterLoops on my box.  I didn't build for NET Native but don't see any reasons it would fail.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/DigestServiceAuthorizationManager.cs
mconnew: It's 7 because the word "Digest " with a trailing space is 7 characters long :). I'll add a constant.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/DigestServiceAuthorizationManager.cs
mconnew: If the TryGetValue fails, then it will run the code inside the if block which ensures the identitiesListObject is not null.

mconnew:
It might fail for .Net Native as this requires the content to be sent twice and .Net native has an issue with replay. I'll verify this.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/DigestServiceAuthorizationManager.cs
iamjasonp: Should we just check for the authz header StartsWith `Digest: `?

mconnew: Already do on line 51 calling IsRequestDigestAuth

mconnew:
I can confirm this fails on Net Native with this exception:
```csharp
  Caught Unexpected exception:System.ServiceModel.CommunicationException: net_http_client_execution_error ---> System.Net.Http.HttpRequestException: net_http_client_execution_error ---> System.NotSupportedException: NotSupported_CloningNotSupported, GetInputStreamAt
     at System.IO.NetFxToWinRtStreamAdapter.GetInputStreamAt(UInt64 position) in D:\dd\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\NetFxToWinRtStreamAdapter.cs:line 455
     at Windows.Storage.Streams.IRandomAccessStream__Impl.Vtbl.GetInputStreamAt__STUB(IntPtr pComThis, UInt64 unsafe_position, $Vtbl*** unsafe_stream__retval, Int32 CS$1$0000)
```
This is the same as we get with normal digest on Net Native

mconnew:

PR #565, opened 2015-12-02
Update project.json files to consistently use -rc2-23602 build

iamjasonp:
On advice of the corefx team, we need to move to specific build numbers rather
than using -release-* builds, so taking this chance to update them all.

Excluded are the ref projects, need to make sure that there are no repercussions
in doing so.

Note - only those packages that were referencing -beta-* or -rc2-* packages were updated to -rc2-23602.


iamjasonp:
@roncain, can you please take a look?

roncain:
LGTM.  Outerloops pass and initial TFS build looks okay.  Recommend proactive merge now to unblock TFS builds.

iamjasonp:
Thanks @roncain. I'll monitor the internal builds

Issue #47, opened 2015-05-27
Implement write buffering on send stream for Http transport


mconnew:
On desktop, we buffer the output stream so that lots of small writes that are done when serializing using streaming don't become lots of p/invokes. We don't currently buffer the output stream which will affect performance for some streaming scenarios.

mconnew:
This is now implemented

PR #573, opened 2015-12-03
Adding buffer to Http streamed mode transfer.

mconnew:


mconnew:

mconnew:
@KKhurin, can you test this code to see how much it improves the performance and memory usage of streamed mode HTTP

mconnew:
Fixes #47

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+
+        public BufferedStream(Stream stream, int bufferSize)
+        {
+            if (stream == null)
+                throw new ArgumentNullException("stream");
+
+            if (bufferSize <= 0)
+                throw new ArgumentOutOfRangeException("bufferSize");
+
+            Contract.EndContractBlock();
```
iamjasonp: Are we supporting all the Contract stuff? I seem to recall being asked not to use the `EndContractBlock` stuff before, but that might have been a long time ago

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+
+        private readonly SemaphoreSlim _sem = new SemaphoreSlim(1, 1);
+
+        public BufferedStream(Stream stream) : this(stream, DefaultBufferSize) {}
+
+
+        public BufferedStream(Stream stream, int bufferSize)
+        {
+            if (stream == null)
+                throw new ArgumentNullException("stream");
```
iamjasonp: Do you need to do validation on the type of incoming stream? E.g., `CanRead`?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+
+            if (bufferSize <= 0)
+                throw new ArgumentOutOfRangeException("bufferSize");
+
+            Contract.EndContractBlock();
+
+            Contract.Assert(!(stream is MemoryStream), "MemoryStream shouldn't be wrapped in a BufferedStream!");
+            Contract.Assert(!(stream is BufferedStream),
+                "BufferedStream shouldn't be wrapped in another BufferedStream!");
```
iamjasonp: NIT: I think we generally discourage using :exclamation: points in our error messages :smile:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+                return Task.FromCanceled<int>(cancellationToken);
+
+            EnsureNotClosed();
+
+            return FlushAsyncInternal(cancellationToken, this, _stream, _writePos, _readPos, _readLen);
+        }
+
+
+        private static async Task FlushAsyncInternal(CancellationToken cancellationToken,
+                                                     BufferedStream _this, Stream stream, int writePos, int readPos,
```
iamjasonp: NIT: why `_this` as a parameter? Is it just intended as a hint for the developer?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+            if (buffer.Length - offset < count)
+                throw new ArgumentException("offset/count");
+            Contract.EndContractBlock();
+
+            // Fast path check for cancellation already requested
+            if (cancellationToken.IsCancellationRequested)
+                return Task.FromCanceled<int>(cancellationToken);
+
+            EnsureNotClosed();
+            EnsureCanRead();
```
iamjasonp: A personal nit about semantics - I think that the EnsureNotClosed and EnsureCanRead should go above the fast path for cancellation requested. One shouldn't be able to try to do something invalid and have the method call succeed since it got cancelled. Not a blocker, just personal preference on my part.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+            if (buffer.Length - offset < count)
+                throw new ArgumentException("offset/count");
+            Contract.EndContractBlock();
+
+            // Fast path check for cancellation already requested
+            if (cancellationToken.IsCancellationRequested)
+                return Task.FromCanceled<int>(cancellationToken);
+
+            EnsureNotClosed();
+            EnsureCanWrite();
```
iamjasonp: Same comment about cancelling a potentially invalid operatoin

iamjasonp:
Nothing obvious is jumping out at me in terms of issues, but I'm going through the read and write with good ol' pencil and paper to check on the math and semaphores, so I can understand better. 

Are there any specific tests other than perf that could be written to check functionality?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+            Contract.Requires(_stream != null);
+
+            if (!_stream.CanWrite)
+                throw new NotSupportedException("write");
+        }
+
+        /// <summary><code>MaxShadowBufferSize</code> is chosen such that shadow buffers are not allocated on the Large Object Heap.
+        /// Currently, an object is allocated on the LOH if it is larger than 85000 bytes.
+        /// We will go with exactly 80 KBytes, although this is somewhat arbitrary.</summary>
+        private const int MaxShadowBufferSize = 81920; // Make sure not to get to the Large Object Heap.
```
iamjasonp: I presume there's practically zero risk that the LOH thresholds will be revised lower, right?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
```diff
@@ -167,7 +167,7 @@ internal static HttpContent Create(HttpChannelFactory<IRequestChannel> factory,
 
     public class StreamedMessageContent : MessageContent
     {
-        private Task _encodingTask;
+        private const int WriteBufferSize = 4096;
```
iamjasonp: Might be good to have a comment about we chose this number

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
mconnew: I was of two minds about this. This is an internal copy of the BufferedStream class from desktop and so the only inputs are from our own code. This makes me think that maybe for performance reasons I should strip the validation out and make if Asserts instead as we won't be passed arbitrary data that hasn't already been validated.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
mconnew: What are your thoughts about stripping all the validation out, at least in retail build to improve performance. Again, this code isn't gong to be used arbitrarily by external code.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
mconnew: This is a Contract.Assert so would only appear in a debug build of our code. Although with the nature of these particular checks, we can probably strip them out.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
mconnew: According to the original code comments, it was done this way because stream derives from MarshalByRefObject and that doesn't play well with async methods. That doesn't apply now. How much effort do you think I should put into cleaning up the code as opposed to leaving it the same as it is on desktop?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
mconnew: This is the semantics used by the base type Stream so we should be consistent.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
mconnew: There is so much code which has behavior based on the LOH threshold being 85K that I think changing the threshold would be considered a breaking change. It's now .Net dogma :)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: Good point. It was because HttpWebRequest used 4K internally for it's own buffer so I just copied that.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
mconnew: I think the issue is with anything that needs the contract rewriting compiler tools used to modify the assembly. I think I'm going to change the checks to Contract.Assert so that they are only there in debug builds as this is internal code only.

roncain: I tend to agree to stick with Assert for now.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+    /// The following should be true:
+    /// <![CDATA[
+    ///   * 0 <= _readPos <= _readLen < _bufferSize
+    ///   * 0 <= _writePos < _bufferSize
+    ///   * _readPos == _readLen && _readPos > 0 implies the read buffer is valid, but we're at the end of the buffer.
+    ///   * _readPos == _readLen == 0 means the read buffer contains garbage.
+    ///   * Either _writePos can be greater than 0, or _readLen & _readPos can be greater than zero,
+    ///     but neither can be greater than zero at the same time.
+    ///  ]]>
+    /// This class will never cache more bytes than the max specified buffer size.
```
roncain: Just a Nit question -- do we know how well [CDATA] is handled by Intellisense as well as external tools that parse these comments into API documentation?  My inclination is to take a conservative approach and avoid [CDATA] unless we're willing to verify it is handled correctly.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
roncain: BTW, one rule I've seen followed is to *not* throw ArgumentExceptions from internal code but only to Assert (unless the internal code is somehow callable through public code, like a virtual).  Otherwise you have dead code that will never execute.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+
+            _stream = stream;
+            _bufferSize = bufferSize;
+
+            // Allocate _buffer on its first use - it will not be used if all reads
+            // & writes are greater than or equal to buffer size.
+
+            if (!_stream.CanRead && !_stream.CanWrite)
+                throw new ObjectDisposedException("BufferedStream");
+        }
```
roncain: How do you know !CanRead & !CanWrite implies it was disposed?  This feels more like an Assert or ArgumentException with message.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
```diff
@@ -0,0 +1,1098 @@
...
+        {
+            if (_stream == null)
+                throw new ObjectDisposedException("BufferedStream");
+        }
+
+
+        private void EnsureCanSeek()
+        {
+            Contract.Requires(_stream != null);
+
```
roncain: I'll stop commenting on these, but think we should avoid Contract.Requires, etc.  Contract.Assert is probably our safest usage pattern for now.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/BufferedStream.cs
roncain: Oh, sorry I missed that.  If this is from the desktop, what is different?  It could save review and refactor time.

roncain:
LGTM.
@KKhurin and @mconnew -- it would be good to look at any perf and stress tests to make sure we have payloads that exercise both huge contiguous writes as well as a bunch of tiny writes let this buffered write stream show what it can do.

mconnew:

PR #575, opened 2015-12-03
Remove 2 minutes delay

### LINKED ISSUE #533: Find solutions to deal with the CRL caching to avoid this 2 minute wait
Currently we set CRL cache to refresh every 2 minutes. Thus the CRL test need to wait for 2 minutes to ensure it's updated.

This issue is to either find the solution, or work around it by changing the test to make a call to the service in a short period of time in a loop and times out in 2 minutes.

hongdai:
* It's not necessary now as Jason has made window one cache entry per certificate

Fixes #533


hongdai:

roncain:
LGTM

iamjasonp:
Ah cool - thanks for catching that.

PR #590, opened 2015-12-08
Updating existing tests with updated syntax and validation checking.

StephenBonikowsky:
* Using Asserts instead of StringBuilder for validation.
* Adding comments to compartmentalize the test into separate sections.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
@@ -19,25 +19,35 @@ public static class ServiceContractTests
     public static void DefaultSettings_Echo_RoundTrips_String_Buffered()
     {
         string testString = "Hello";
-
-        BasicHttpBinding binding = new BasicHttpBinding(BasicHttpSecurityMode.None);
-        binding.TransferMode = TransferMode.Buffered;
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
No functionality should have changed.

Anybody know if when an Assert fails will Xunit include the inner exception details when it throws?

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
```diff
@@ -19,25 +19,35 @@ public static class ServiceContractTests
     public static void DefaultSettings_Echo_RoundTrips_String_Buffered()
```
iamjasonp: Maybe these could be consolidated into a [Theory] test? It looks like much of this is repeated code.

iamjasonp:
The inner exception shows when a throw happens in a `[Fact]` test. 
Not sure about `[Theory]`, or within an `Assert.Throws<T>` block

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
StephenBonikowsky: Possibly, don't really want to get bogged down refactoring these tests, just some quick cleanup.

Issue #117, opened 2015-06-24
Clean up #pragmas and CLS compliance issues identififed in #116


iamjasonp:
Per #116: 

> Build warnings were produced due to fields not being CLS-compliant
> and Equals being overridden without overriding GetHashcode.
> These warnings caused the build to fail when building the
> TFS mirrored sources with warnings treated as errors.

We had to commit quickly in order to fix a build break - let's use this issue for further discussion on the issues that were identified in the code review for #116.

PR #602, opened 2015-12-10
Change MessageContent to internal class to fix CLS compliance

iamjasonp:
#116 disabled some compiler warnings after importing MessageContent.cs. We want to make sure that the class (and therefore the assembly) remains CLS compliant. 

@mconnew said in #116: 
> The compliance issue is that protected fields can't start with an underscore. What's the naming standard for protected fields? I would rather change the field names rather then disable the warning.

> If they are left as not CLS compliant, this simply means that those fields can't be accessed by code written in some other languages which only support CLS compliant assemblies. As this is only public to enable writing unit tests and isn't exposed on the contract, in practice this non-CLS compliance doesn't negatively affect any scenario.

The class was made public to enable unit tests... but our tests can only test against public contracts. Hence there's no point in making the class public. If we move this to be an internal class, the CLS compliance issues go away. 

Part of #117

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
@@ -13,13 +13,11 @@ using System.Linq;
 
 namespace System.ServiceModel.Channels
 {
-    public abstract class MessageContent : HttpContent
+    internal abstract class MessageContent : HttpContent
     {
-#pragma warning disable 3008	// not CLS-compliant
         protected Message _message;
         protected MessageEncoder _messageEncoder;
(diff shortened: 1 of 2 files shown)


iamjasonp:

iamjasonp:

iamjasonp:
Built in TFS on my local box, passes the build for System.Private.ServiceModel.*.csproj

roncain:
LGTM

iamjasonp:
Should be noted too that if anything inside the class is marked `[ClsCompliant(false)]`, then the entire class becomes noncompliant unless some special steps are done, e.g., providing an additional CLS-compliant field. 

System.Private.ServiceModel is CLS-compliant with this change - previously, technically, we were not.

Issue #576, opened 2015-12-04
Address certificate validation issues per #422


iamjasonp:
In #422, we moved certificate validation from relying on the certificate subject's CanonicalName (Desktop <= 4.5) to relying on Subject Alternative Names (behaviour in Desktop >= 4.6 by way of a quirking behaviour)

@bartonjs has identified a number of issues that need to be addressed going forward: 

- [x] https://github.com/dotnet/wcf/pull/422#discussion_r41986442 If you're trying for RFC 2818 matching logic, in the event there are no DNS values in the Subject Alternative Names list (or there's no SAN extension at all) you should fall back to the CN. So the previous code would go here.
- [x] https://github.com/dotnet/wcf/pull/422#discussion_r41986196  OID-info says that 2.5.29.7 was declared Obsolete in/prior-to the year 2000; you can probably safely ignore it. I don't think that Unix builds would even recognize it as anything to pretty print, instead just dumping the bytes in hex. And certainly SslStream won't use that one
- [x] https://github.com/dotnet/wcf/pull/422#discussion_r41989839 x-plat issues
- [x] https://github.com/dotnet/wcf/pull/422#issuecomment-148178353 increase code compat in x-plat
- [x] https://github.com/dotnet/wcf/pull/422#issuecomment-160690727 Additional testing (one test has been done to verify functionality merged in pull #422, but we could use a few more tests)

Note that some of these fixes will likely need a back-port into desktop WCF, and we'll have to take each of these issues back on a case-by-case basis. 

We will use this issue to track the issues that came from #422, and split them out as needed as time goes on and work is done on each item. The master list is above.

iamjasonp:
Done - as part of #603

PR #603, opened 2015-12-10
Allow certificate claim checking to fall back to using CN

iamjasonp:
We currently check only the Subject Alternative Name when looking at certificates. In order to match RFC 2818, in the event there are no DNS values in the SAN field (or no SAN extension at all), fall back to using CN (this is the behaviour that we use on Desktop pre-4.6) 

Relates to #576


iamjasonp:

roncain:
LGTM

iamjasonp:
@bartonjs could you also review please?

bartonjs:
Algorithmically, you've done the right thing here.  The remaining problem is that GetDnsFromExtensions returns things that aren't dNSName values (and doesn't work on Unix).  Lacking a SAN extension object to read from semantically (it's on my infinite TODO list), your best bet is to do something like:

* Pick a SubjectAltNames extension from somewhere, embed the bytes in your library.
  * Let's assume it contains two entries: dNSName:example.com, dNSName:*.example.com
  * Two is the minimum, but make sure that it doesn't contain things that'll trip you up.
* Make a Lazy<ParserSettings>
  * Load the bytes into an X509Extension object, get the string representation
  * Find `example.com` in the output
  * The character before it is the key-value delimiter (`=` on Windows, `:` on Unix)
  * The pair-separator sequence is now optional-comma and whatever whitespace (`,<space>` on Unix (IIRC), `\r\n` on Windows)
   * The dNSName string value is whatever is after the pair-separator, and before the key-value delimiter.
      * `DNS` on Unix (not localized)
      * `DNS Name` on Windows en-US (localized)

Then, in GetDnsFromExtensions, skip over anything that doesn't match the dNSName prefix (for example, the ipAddress entries).

iamjasonp:
@bartonjs yep - was planning to take the rest in steps (since it's logically easier to review, see #576). All of the issues you mentioned there will eventually be dealt with.

iamjasonp:
@bartonjs - wondering if I could cheat here a bit. Basically we're looking at parsing the following: 

<identifier><delimiter><value><separator(s)><identifier><delimiter><value><separator(s)>

is it sufficient to figure out that the identifier ("dnsName" or "DNS Name") is the string from the start of string to the first delimiter seen?

I agree it's probably safer to move forward from the end of <value> to the first non-whitespace, not-comma character though, just seeing if assuming we can assume identifier is `str[0] .. str[identifier-1]` makes the code less ugly :) 

Gist here with what I have for the moment: https://gist.github.com/iamjasonp/dbe3966aba99e8bc5d22; I ran this against Linux and it seems to work fine (In the gist, I stripped out a huge bunch of cruft around Console.WriteLine while I was playing around)

iamjasonp:
(gee, brings me back to when I was mucking around with the various XmlReaders/Writers :) )

bartonjs:
@iamjasonp Seems legit.  I couldn't remember if there was a header on it (or not), but when I just looked at an example there wasn't.

bartonjs:
@iamjasonp And, to be pedantic to avoid confusion: `dNSName` is the identifier in the spec (little d and all); `DNS Name` is the Windows en-US type identifier, and `DNS` is the Unix (OpenSSL) type identifier. :smile:

iamjasonp:
Pedantic is good :smile: 
I spent some time reading through the RFC. Whoooeeee... it's a doozy.

iamjasonp:
Update: things are looking good, seems like the OuterLoop test that broke in Linux are working again. I'll update this PR once it's in a state that's presentable, right now it's a big kludge.

iamjasonp:

iamjasonp:

iamjasonp:
I came up with a really clunky way to test this functionality: 

1. Have three certificates:

* CN = localhost
* CN = testmachine.example.com
* CN = testmachine
* Note that there are no Subject Alt Names for each of these 

2. Create endpoints for each of these three, and bind the cert to the endpoint 
3. Run a test through ALL THREE endpoints. 
4. Check that EXACTLY one of these tests will have passed - EXACTLY two will have failed with a MessageSecurityException due to certificate validation issues. 

The reason we need to do this is because I haven't found a good way to allow BridgeClient to report to the Bridge what endpoint it believes it's hitting. So the Bridge is listening and doesn't care where the requests are coming in, but there's a difference to the client between hitting "localhost" or "testmachine" or "testmachine.example.com". 

I guess in this way, we test the positive and negative cases in one go, so that's a win, right? :smile:

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
```diff
@@ -339,6 +375,59 @@ public IIdentity Identity
...
+        // We don't have a strongly typed extension to parse Subject Alt Names, so we have to do a workaround 
+        // to figure out what the identifier, delimiter, and separator is by using a well-known extension
+        private static class X509SubjectAlternativeNameConstants
+        {
+            public const string Oid = "2.5.29.17"; 
+            public static readonly string Identifier;
+            public static readonly char Delimiter;
+            public static readonly string Separator;
+
+            // static initializer runs only when one of the properties is accessed
```
iamjasonp: I ran this through the debugger and noted that accessing `X509SubjectAlternativeNameConstants.Oid` did not call the static initializer. It was called only when one of the `readonly` fields was accessed that it was created.

bartonjs: @iamjasonp Yep, `const` values are embedded as literals at the callsite, the type isn't involved at runtime.

iamjasonp:
@bartonjs - arggh, semantics: 

> If a subjectAltName extension of type dNSName is present, that MUST be used as the identity. Otherwise, the (most specific) Common Name field in the Subject field of the certificate MUST be used. Although the use of the Common Name is existing practice, it is deprecated and Certification Authorities are encouraged to use the dNSName instead.

So... the current behaviour is that GetDnsFromExtensions will return an empty `string` array. If there is a SubjectAlternativeName field that contains zero entries, it behaves the same as if there was no SubjectAlternativeName field at all. 

Thoughts on whether this is an important enough distinction?

bartonjs:
@iamjasonp I agree with the ninja edit, that you're doing the right thing in the case of SAN:Yes, DNS-SAN:No.

hongdai:
Ideally we want functional tests to be deterministic. But I agree we should not block the checkin as the test does provide desired coverage and it will take some time to investigate a deterministic way. Please open an issue for the test improvement.

iamjasonp:
@mconnew, could you take a review? Last time @roncain reviewed it was a much different change from today

iamjasonp:
> Ideally we want functional tests to be deterministic.

One of the unfortunate parts about this particular is that what is "correct" behaviour is only going to be defined at runtime, depending on what parameter you're passing into msbuild

I have, however, modified the test to be a [Theory] test that is a bit better at compartmentalizing what a correct an incorrect variation is.

iamjasonp:

iamjasonp:

iamjasonp:

iamjasonp:

iamjasonp:

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeCertificateCanonicalNameTests.cs
```diff
@@ -0,0 +1,119 @@
...
+            var fqdnEndpointUri = new Uri(Endpoints.Tcp_ClientCredentialType_Certificate_With_CanonicalName_Fqdn_Address);
+            if (string.Compare(domainNameHost, fqdnEndpointUri.Host, StringComparison.OrdinalIgnoreCase) != 0)
+            {
+                list.Add(new object[] { fqdnEndpointUri, fqdnEndpointUri.Host.IndexOf('.') > -1 });
+            }
+
+            return list;
+        }
+    }
+}
```
hongdai: Deterministic in the run time is sufficient. Thanks for being creative to solve the issue, Jason! LGTM from testing perspective.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
```diff
@@ -339,6 +375,59 @@ public IIdentity Identity
...
+            static X509SubjectAlternativeNameConstants()
+            {
+                // Extracted a well-known X509Extension
+                const string x509ExtensionBase64String = "MCSCFW5vdC1yZWFsLXN1YmplY3QtbmFtZYILZXhhbXBsZS5jb20=";
```
mconnew: Why not encode this directly as a byte array? Having it as a base64 encoded string doesn't make it any less of an opaque blob.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
```diff
@@ -339,6 +375,59 @@ public IIdentity Identity
...
+        private static class X509SubjectAlternativeNameConstants
+        {
+            public const string Oid = "2.5.29.17"; 
+            public static readonly string Identifier;
+            public static readonly char Delimiter;
+            public static readonly string Separator;
+
+            // static initializer runs only when one of the properties is accessed
+            static X509SubjectAlternativeNameConstants()
+            {
```
mconnew: Can this entire static constructor be placed inside a try/catch and then either set a flag which causes an exception to be thrown when any of the values are attempted to be retrieved such as the delimiter, or simply fail the process? Any exceptions in a static initializer are silently swallowed and can be the source of difficult to track down bugs.

mconnew:
LGTM, the only concern is the possibility of having an exception in the static block. I'm okay with a follow up PR to address that though.

iamjasonp:

iamjasonp:
Thanks @mconnew 

I'm trying to figure out why the outerloop build in CI will fail but the same failure doesn't repro locally for me, so there might be some light churn on the test.

iamjasonp:

iamjasonp:

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
iamjasonp: Good point. In order not to muck up the ordering of the commits here I'll merge this and immediately open up another PR to fix this issue.

iamjasonp:
FYI: the reason why is because when running a bridge as "localhost", the check for 

```csharp
            var domainNameEndpointUri = new Uri(Endpoints.Tcp_ClientCredentialType_Certificate_With_CanonicalName_DomainName_Address);
            list.Add(new object[] {
                domainNameEndpointUri,
                domainNameEndpointUri.Host.IndexOf('.') == -1});
```

meant that the Bridge was always presenting URIs as "localhost - and thus the "should test pass" would still resolve as true, as we were only looking for "." in the hostname. Fixed this, and now it passes. 

(I usually run the bridge separately, and use the same Bridge instance across multiple runs, so I never ran into this issue. This issue will surface when running `.\build.cmd /p:WithCategories=OuterLoop` if the Bridge has not been started before)

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
iamjasonp: I'll change this in the next PR, was just a quick way for me to test this which I never changed back

PR #622, opened 2015-12-10
Update instructions for running tests on Linux.

roncain:


roncain:
Merging before being out of office for 3 weeks.

PR #626, opened 2015-12-11
Disable WebSockets tests on Linux

roncain:
Adds [ActiveIssue(625, PlatformID.AnyUnix)] to WebSockets tests
because they cannot pass until support has been added.

Support of WebSockets on Linux in CoreFx is tracked by issue
https://github.com/dotnet/corefx/issues/2486


mconnew:
LGTM

Issue #604, opened 2015-12-10
Exclude System.Security.Principal.Windows for *NIX builds


roncain:
S.S.P.Windows is not supported on Linux and resolves to a desktop implementation assembly and fails at runtime.  The fix is:
1. Make a new project.json file just for Unix that excludes this dependency (as we did for the src\Windows\project.json
2. Make use of this type conditional based on OS
3. Make any scenario tests for this conditional on being Windows

Marked as a bug because our builds will start failing as soon as CoreFx treats the inclusion of this package in Linux a build error.

roncain:
Incidentally, to build for Linux, you use this command:
```
  build.cmd /p:OSGroup=Linux 
```

As long as we have project.lock.json files, we will need to remember to do this whenever we change anything in the project.json.

PR #627, opened 2015-12-11
Disable Linux tests using System.Security.Principal.Windows

roncain:
These tests invoke code that causes failures in OuterLoop due to
the fact System.Security.Principal.Windows is not supported on Linux
but we still depend on it.  This causes the Windows assembly to be
used instead.

When #604 is resolved to remove this dependency we will need to
fix these tests or make them conditional on OS.


mconnew:
LGTM

iamjasonp:
LGTM

PR #629, opened 2015-12-11
Update build tools, dir.props, and add facade projects

mellinoe:
This adds facade projects for all of the contract assemblies. Currently we are only building these  "facade" assemblies as part of a batch process in our internal build, so there is no way for external folks, or folks only working on GitHub, to build/deploy these themselves.

Right now, these are just included in the default build process in the default configuration. So if you run build.cmd from the root, they will only be built for the default configuration (CoreCLR). In corefx, we recently added the notion of a ".builds" file, which wraps a project such that running `msbuild` on that file will cause the wrapped project to be built for each applicable configuration. We may want to do the same thing here, but I didn't want to cause too much disruption to the regular workflow with this change.  Right now, if you want to build for other configurations, you need to pass that into the arguments, i.e. `msbuild System.ServiceModel.Http.csproj /p:TargetGroup=net46`.

As part of this, I needed to update the version of buildtools used, and grab a bunch of additional properties that we have set in our dir.props file in corefx. I just grabbed the latest version from corefx, and it didn't seem like anything important was lost in that process; it looks like the version replaced from this repo was an earlier version from corefx.

I ran the inner loop tests and, assuming I did it right, the outer loop tests as well. I ran `build.cmd /p:WithCategories=OuterLoop` from an elevated prompt and it seems that it ran all the outer loop tests with only 3 warnings.

cc: @roncain , @weshaggard


roncain:

roncain:
@iamjasonp and @mconnew -- would you mind reviewing this and buddy building on TFS if possible?  I'll try myself but am currently buddy building other shelvesets.

roncain:
FYI -- this passes OuterLoop for me too when I test locally. Haven't tested it building on TFS yet.

mellinoe:
Thanks for testing, @roncain . One thing to note is that this won't affect the things we ship out of TFS yet; we are still using the batch facade generation for now. Once all of the facades are building on GitHub we will start switching the build over to use them.

src/System.Private.ServiceModel/src/windows/project.json
```diff
@@ -57,7 +57,8 @@
     },
     "netcore50": {
       "dependencies": {
-        "System.Runtime.WindowsRuntime": "4.0.11-rc2-23608"
+        "System.Runtime.WindowsRuntime": "4.0.11-rc2-23608",
+        "Microsoft.TargetingPack.Private.NetNative": "1.0.0-rc2-23607"
```
iamjasonp: Just a question, what's the purpose of this particular targeting pack `Microsoft.TargetingPack.Private.NetNative`, and why isn't it part of the dnxcore set of dependencies as well?

iamjasonp:
Buddy building locally, and also on using the TFS gated build (not checkin) mechanism. Unfortunately it seems like the gated build is failing for other reasons.

iamjasonp:

iamjasonp:
TFS: gated build failed with errors
My own box: build failed with the following: 

```
"N:\dd\ProjectK\src\partitions.proj" (default target) (1) ->
"N:\dd\ProjectK\src\ndp\fxcore\dirs.proj" (BuildProjects target) (4:3) ->
"N:\dd\ProjectK\src\ndp\fxcore\src\dirs.proj" (Build target) (438) ->
"N:\dd\ProjectK\src\ndp\fxcore\src\Contracts\dirs.proj" (Build target) (440) ->
"N:\dd\ProjectK\src\ndp\fxcore\src\Contracts\Contracts.msbuild" (Build target) (441) ->
"N:\dd\ProjectK\src\ndp\fxcore\Open\src\Microsoft.CSharp\ref\Microsoft.CSharp.csproj" (Build target) (442) ->
  N:\dd\ProjectK\src\ndp\fxcore\Open\src\Microsoft.CSharp\ref\Microsoft.CSharp.csproj : error MSB4057: The target "Build" does not exist in the project.
```

mellinoe:
Thanks for doing a validation build, @iamjasonp . Unfortunately, it looks like the internal builds aren't stable right now :disappointed:. The Microsoft.CSharp failure shouldn't have been affected by this (well, it's not even this repo).

src/System.Private.ServiceModel/src/windows/project.json
mellinoe: The reason I added this target here is because this project needed to reference Windows.winmd, which is included in our "NetNative" targeting pack. This project uses contract references "System.Runtime.WindowsRuntime" and therefore doesn't need any of the targeting pack references for CoreCLR (i.e. mscorlib.dll)

weshaggard:
There are some build breaks internally that are being worked on right now. However @iamjasonp when you see a "missing Build target" it is generally a sign that you didn't follow the new workflow and do the Sync command before the Build, we are no longer doing restore as part of the build.

iamjasonp:
Thanks @weshaggard, forgot about the new dev workflow. I'll try again for a validation.

iamjasonp:
There's still some issues with me building. In principle, I see no issues with the code changes. 
LGTM subject to us finally getting a good buddy build...

mellinoe:
Let me know if there's anything I can do to help get a build running. There should be minimal effects on the TFS build, only the buildtools changes should affect there. The rest of these projects won't be built by the internal build yet until we modify it to use them instead of the batch build process.

iamjasonp:
@mellinoe - I'm calling this good, as I've built TFS clean and with the changes in this PR, came up with the same errors. Two machines were used. 

I see the System.Private.ServiceModel binary being binplaced 

LGTM

iamjasonp:
Merging as TFS builds still haven't turned stable yet, and we've waited days. No obvious break caused by this.

PR #630, opened 2015-12-11
Add sync variations and update existing tests to use correctly attrib…

### LINKED ISSUE #586: Add additional variations of scenario described in Issue467
Add synchronous variations.
Add Task variations.

### LINKED ISSUE #595: Use unique type not using DataContract attribute for Service Contract tests.
New tests validating Out and Ref args are using a complex type attributed with DataContract, need to update these tests to use a unique type without that attribute in order to verify that the Net Native toolchain can find a serializer.

StephenBonikowsky:
…uted type.

* The existing tests were using a type attributed with 'DataContract' this would allow the Net Native toolchain to know what serializer it needed. Updated the tests to use a non-attributed type so the Net Native toolchain is properly tested.
* Added synchronous variations of this scenario.
* Fixes #586
* Fixes #595


StephenBonikowsky:
Ran it against Net Native, still no failures.

PR #634, opened 2015-12-15
Add api surface area needed for Https

mconnew:
Also adding implementation for BasicHttpsBinding, NetHttpsBinding and removing some dead code.


mconnew:

src/System.Private.ServiceModel/src/System/ServiceModel/NetTcpBinding.cs
```diff
@@ -195,58 +195,6 @@ public override BindingElementCollection CreateBindingElements()
             return bindingElements.Clone();
         }
 
-        internal static bool TryCreate(BindingElementCollection elements, out Binding binding)
```
iamjasonp: Do you happen to know where TryCreate is usually used?

src/System.Private.ServiceModel/src/System/ServiceModel/HttpBindingBase.cs
```diff
@@ -293,21 +269,11 @@ internal virtual void CheckSettings()
...
 
-            // Message.ClientCredentialType = Certificate is not supported.
-            if (mode == BasicHttpSecurityMode.TransportWithMessageCredential)
-            {
-                BasicHttpMessageSecurity message = security.Message;
-                if ((message != null) && (message.ClientCredentialType == BasicHttpMessageCredentialType.Certificate))
-                {
-                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new NotSupportedException(SR.Format(SR.UnsupportedSecuritySetting, "Message.ClientCredentialType", message.ClientCredentialType)));
-                }
-            }
```
iamjasonp: I'd rather this be a throw PlatformNotSupported where `BasicHttpSecurityMode.TransportWithMessageCredential`, if we're not supporting it rather than removing the code. 

The Assert a couple of lines down will still pass.

iamjasonp: Oh, it's a couple of lines above...

src/System.Private.ServiceModel/src/System/ServiceModel/BasicHttpsBinding.cs
```diff
@@ -0,0 +1,98 @@
...
+        internal override EnvelopeVersion GetEnvelopeVersion()
+        {
+            return EnvelopeVersion.Soap11;
+        }
+
+        public override IChannelFactory<TChannel> BuildChannelFactory<TChannel>(BindingParameterCollection parameters)
+        {
+            if ((BasicHttpSecurity.Mode == BasicHttpSecurityMode.Transport ||
+                BasicHttpSecurity.Mode == BasicHttpSecurityMode.TransportCredentialOnly) &&
+                BasicHttpSecurity.Transport.ClientCredentialType == HttpClientCredentialType.InheritedFromHost)
```
iamjasonp: `InheritedFromHost` is a server-only concept, we probably shouldn't care about it here

src/System.Private.ServiceModel/src/System/ServiceModel/NetHttpsBinding.cs
```diff
@@ -0,0 +1,147 @@
...
+            if (securityMode == BasicHttpsSecurityMode.TransportWithMessageCredential)
+            {
+                throw ExceptionHelper.PlatformNotSupported(SR.Format(SR.UnsupportedSecuritySetting, "securityMode", securityMode));
+            }
+
+            Initialize();
+            _basicHttpsSecurity.Mode = securityMode;
+        }
+
+        [ComponentModel.DefaultValue(NetHttpMessageEncoding.Binary)]
```
iamjasonp: Do we still need these [ComponentModel] things around? I think it's only for config purposes

src/System.Private.ServiceModel/src/System/ServiceModel/NetHttpsBinding.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            set
+            {
+                if (value == null)
+                {
+                    throw FxTrace.Exception.ArgumentNull("value");
+                }
+
+                _basicHttpsSecurity = value;
```
iamjasonp: I would check for `securityMode == BasicHttpsSecurityMode.TransportWithMessageCredential` here

src/System.Private.ServiceModel/src/System/ServiceModel/NetHttpsBinding.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                return InternalWebSocketSettings;
+            }
+        }
+
+        public override IChannelFactory<TChannel> BuildChannelFactory<TChannel>(BindingParameterCollection parameters)
+        {
+            if ((BasicHttpSecurity.Mode == BasicHttpSecurityMode.Transport ||
+                BasicHttpSecurity.Mode == BasicHttpSecurityMode.TransportCredentialOnly) &&
+                BasicHttpSecurity.Transport.ClientCredentialType == HttpClientCredentialType.InheritedFromHost)
```
iamjasonp: InheritedFromHost probably can be removed

src/System.ServiceModel.Http/ref/System.ServiceModel.Http.cs
```diff
@@ -28,12 +28,31 @@ public sealed partial class BasicHttpSecurity
...
     public enum BasicHttpSecurityMode
     {
-        Message = 2,
         None = 0,
         Transport = 1,
-        TransportCredentialOnly = 4,
+        Message = 2,
         TransportWithMessageCredential = 3,
+        TransportCredentialOnly = 4,
     }
```
iamjasonp: NIT: tabs vs. spaces in this file

src/System.Private.ServiceModel/src/System/ServiceModel/NetHttpsBinding.cs
iamjasonp: Do we still need these [ComponentModel.DefaultValues] around in the implementation file? I thought it's only for config purposes when in the implementation file, but these should be specified in the ref assembly now, right?

iamjasonp:
Nothing glaringly obvious to me that's wrong. Outerloop failure is a flaky test, so I'm not worried about it.
I have some NIT- and question-only level comments as well as some very minor things - those should be fairly quick to address or ignore. 

Please make sure once the contract is revved, tests are written for the newly-introduced API

LGTM

src/System.Private.ServiceModel/src/System/ServiceModel/NetTcpBinding.cs
mconnew: It's used by the wsdl importer which we don't support in CoreCLR

src/System.Private.ServiceModel/src/System/ServiceModel/HttpBindingBase.cs
mconnew: You're right, I need to remove `mode == BasicHttpSecurityMode.TransportWithMessageCredential` from the Assert as this is not (current) supported.

src/System.Private.ServiceModel/src/System/ServiceModel/NetHttpsBinding.cs
mconnew: You're right. I need to remove these.

src/System.Private.ServiceModel/src/System/ServiceModel/NetHttpsBinding.cs
mconnew: This will be caught as soon as a channel factory is attempted to be made as CheckSettings in HttpBindingBase will get called as part of that. When I write the unit tests, I'll make sure to verify TransportWithMessageCredential fails.

mconnew:

PR #636, opened 2015-12-15
Adding test coverage for XmlDictionaryReaderQuotas.CopyTo method

### LINKED ISSUE #127: Update unit tests that validate System.Xml.XmlDictionaryReaderQuotas
4 unit tests are failing on Desktop runs due to the difference in behavior related to setting quotas. After these tests were implemented a constructor was added for XDRQ and also the .CopyTo and .ModifiedQuotas APIs were added to XmlDictionaryReaderQuotas. We should now be able to update the tests so they can pass in both K/N and Desktop and also include per @iamjasonp 's old recommendation...

"I think the test should eventually change to exercise the .CopyTo(XDRQ) method as well, when available."

StephenBonikowsky:
* Updated existing test to get this coverage.
* Fixes #127


hongdai:
LGTM.

PR #646, opened 2015-12-17
Remove lockfiles, update dependency versions, and enable dependency upgrade/validation tooling

dagood:
This applies changes like the ones here: https://github.com/dotnet/corefx/pull/4894 to WCF. This removes `project.lock.json` files and makes it reasonable to have repeatable builds without them.

The validation rules were already included in https://github.com/dotnet/wcf/pull/629, but validation wasn't enabled yet.

See https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/project-nuget-dependencies.md to get an overview of what these changes mean for dev workflow. (I added a link from the WCF docs to the CoreFX copy of this page--let me know if I should copy it over here instead.)

I'd suggest looking at each commit individually, since the lockfile removal changes are pretty big.

/cc @roncain @weshaggard 

### Running tests with floating dependencies

Since this pull request makes it impossible to use `--lock` to get the latest corefx dependencies, I added a new way to get floating dependencies in the test-runtime project.json. Use `/p:FloatingTestRuntimeDependencies=true`, for example:

    build.cmd /p:FloatingTestRuntimeDependencies=true

This copies `\src\Common\test-runtime\project.json` to an output directory and automatically changes all references to floating versions. For example, `4.1.0-rc2-12345` will turn into `4.1.0-*`. The `TestRuntimeProjectJson` and `TestRuntimeProjectLockJson` properties in `\dir.props` are changed to point at the generated project.json and the theoretical path where project.lock.json will appear once the project.json is restored.


iamjasonp:
@dagood, I think #647 needs to go in first, hang on while that gets verified...

iamjasonp:

iamjasonp:
In my opinion, the link to the [project dependency documentation in CoreFx](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/project-nuget-dependencies.md) is good as long as the WCF repo is kept reasonably up to date with any changes that might happen in CoreFx - we've been pretty good about keeping things synced as much as humanly possible. 

LGTM

dagood:
Thanks for the help, sounds good on the document link. I've pushed a rebase to merge in the buildtools update fix that was conflicting. It looks like the unix package dependency wasn't kept in sync so I updated that.

mconnew:
With this change, how do we build and run our tests using the latest available versions? With the previous system, we could easily regenerate all our project.lock.json files to whatever the latest version is and run our tests against them. We only needed to do anything manual when the prerelease version was bumped, eg beta7 -> beta8. I have a simple batch script to do this, and I know other WCF developers have their own variants too. We have caught multiple functional regressions in various libraries in corefx which weren't caught by the tests in the corefx repo itself. Being able to script running against the latest built and released versions is an important feature for the WCF team to catch functional issues in our dependencies in a timely manner and helps us maintain a stable product.

dagood:
@mconnew Hmm, good question, I didn't have that in mind when thinking about this. Currently the way to do that would be to find the latest prerelease specifier available on MyGet and:

1. Update the [`ValidationPattern`](https://github.com/dagood/wcf/commit/fc133e0eac928090226e0df0860436bbe3ebd35e#diff-0b192804a6349e8c26d2b027afbd89a2R76) `ExpectedPrerelease` metadata to the latest.
2. Run `msbuild /t:UpdateInvalidPackageVersions` to update all `project.json` files.

For 1, it might be easier to script a separate msbuild file included after `dir.props` that updates/overrides the `ExpectedPrerelease` rather than changing the `dir.props`, but I'm not sure what file that would be.

Unfortunately because of the way NuGet version parsing works (and because I use it as a dependency), currently I don't think you can just change `ExpectedPrerelease` to `*`. The `*` is ignored and becomes an empty string in the result, so the expected prerelease doesn't match.

---

Maybe I should make some changes to support this:

1. Make `ExpectedPrereleaseVersion` work with `*`. (Special-casing if necessary.)
2. For WCF, add a second set of `ValidationPattern`s that enforce `*` versions, switching with a property.

Then this workflow would just be:

1. `msbuild /p:FloatingCoreFXDependencies=true /t:UpdateInvalidPackageVersions ` to update, then
2. `msbuild /p:FloatingCoreFXDependencies=true [...]` to build.

weshaggard:
cc @ericstj

ericstj:
> With this change, how do we build and run our tests using the latest available versions?
That's controlled entirely by the test-runtime\project.json today.  Your projects themselves should be referencing the lowest stable version that supports the API they need.  The test runtime determines the value of the rest of the binaries that are brought in.  (this needs to be replaced with something less brute force in the future /cc @Chrisboh @jhendrixMSFT )

So really all you need to do to change the version of the baseline is to update the test-runtime, not the whole repo, and rebuild.

mconnew:
@ericstj, it's this bit which I don't see an easy way to automate:  
> So really all you need to do to change the version of the baseline is to update the test-runtime, not the whole repo, and rebuild.  
My understanding of the new mechanism is that I would need to manually go find out what the versions are and update the test-runtime to this version? Or have I missed something?

ericstj:
If you wanted to you could have a second test-runtime file that used the * versions and just pass that in to the build via the TestRuntimeProjectLockJson property.

The reason we stopped using *s in the build is that locking lock files is being deprecated and we needed the repo to be deterministic.  For your scenario you specifically don't want that determinism so you can have a separate file you use just for that purpose.

dagood:
Thanks Eric, that makes sense. This means I'll need to exclude that new test-runtime file from validation, so I think it makes sense to add it during this PR.

From what I see, using it would be adding `/p:TestRuntimeProjectJson=src\Common\floating-test-runtime\project.json` to the build command line.

@ericstj Would `TestRuntimeProjectLockJson` have to be changed too? For simplicity of the flow I'm thinking that I could add a property group for both that turns on with a `/p:FloatingTestRuntimeDependencies=true` or similar.

I also think restoring the floating project.json needs to be optional (excluded from the list of project.json files by default) so devs don't *always* have to download the latest packages as well as the specific prerelease level.

I'll get these changes in the pull request. Thanks!

dagood:
I added a commit with a `FloatingTestRuntimeDependencies` property/flag that you can set. A pretty major downside to this that I didn't realize earlier is that there's now a hefty new project.json file that needs to be kept in sync manually.

We could generate the floating test-runtime by copying the test-runtime project.json and running an auto-updating validation with different rules, but I'm not sure where that would fit in the targets and it might balloon the scope of these changes. Any other ideas?

mconnew:
Is there a way to specify the version number which only specifies the major and minor versions? That would allow fewer updates needing to be done. For example, specifying System.Collection to be 4.0.*.

dagood:
According to nuget docs you can (http://docs.nuget.org/consume/ProjectJson-Dependency#floating-versions) but when I try that with DNU restore it errors out, saying it's invalid:

    Restoring packages for E:\git\wcf\src\Common\test-runtime\floating\project.json
    ----------
    E:\git\wcf\src\Common\test-runtime\floating\project.json(0,0): Error: Microsoft.Dnx.Runtime.FileFormatException: '4.0.*' is not a valid version string. ---> Microsoft.Dnx.Runtime.FileFormatException: '4.0.*' is not a valid version string. ---> Microsoft.Dnx.Runtime.FileFormatException: '4.0.*' is not a valid version string. ---> System.ArgumentException: '4.0.*' is not a valid version string.
       at NuGet.VersionUtility.ParseVersionSpec(String value)
       at NuGet.VersionUtility.ParseVersionRange(String value)
       at Microsoft.Dnx.Runtime.ProjectReader.PopulateDependencies(String projectPath, IList`1 results, JsonObject settings, String propertyName, Boolean isGacOrFrameworkReference)
       --- End of inner exception stack trace ---
       at Microsoft.Dnx.Runtime.ProjectReader.PopulateDependencies(String projectPath, IList`1 results, JsonObject settings, String propertyName, Boolean isGacOrFrameworkReference)
       at Microsoft.Dnx.Runtime.ProjectReader.BuildTargetFrameworkNode(Project project, String frameworkKey, JsonObject frameworkValue)
       at Microsoft.Dnx.Runtime.ProjectReader.BuildTargetFrameworksAndConfigurations(Project project, JsonObject projectJsonObject, ICollection`1 diagnostics)
       --- End of inner exception stack trace ---
       at Microsoft.Dnx.Runtime.ProjectReader.BuildTargetFrameworksAndConfigurations(Project project, JsonObject projectJsonObject, ICollection`1 diagnostics)
       at Microsoft.Dnx.Runtime.ProjectReader.ReadProject(Stream stream, String projectName, String projectPath, ICollection`1 diagnostics)
       at Microsoft.Dnx.Runtime.Project.TryGetProject(String path, Project& project, ICollection`1 diagnostics)
       --- End of inner exception stack trace ---
       at Microsoft.Dnx.Runtime.Project.TryGetProject(String path, Project& project, ICollection`1 diagnostics)
       at Microsoft.Dnx.Tooling.RestoreCommand.<RestoreForProject>d__69.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at Microsoft.Dnx.Tooling.RestoreCommand.<>c__DisplayClass68_0.<<Execute>b__2>d.MoveNext()
    --- End of stack trace from previous location where exception was thrown ---
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
       at Microsoft.Dnx.Tooling.RestoreCommand.<Execute>d__68.MoveNext()
    ----------
    Restore failed
    '4.0.*' is not a valid version string.

@ericstj Do you know if this is just unsupported?

dagood:
I realized generating the file would be fairly simple, so I added a commit with the changes to do that. It copies the test-runtime project.json to somewhere like `bin\obj\Windows_NT.AnyCPU.Debug\Common\test-runtime\floating\project.json` and runs the validation task on it when the flag is set.

I've tried it with batch restore and an individual project restore and it worked in both cases.

roncain:
Thanks for doing this @dagood!  The changes LGTM but I'm late to the party and would appreciate @mconnew and/or @iamjasonp reviewing the latest changes.  I will also experimentally pull this PR into a TFS enlistment and ensure it builds clean there, probably early tomorrow morning. I'll also try running both flavors (floating and fixed).

dagood:
@roncain Thanks! I copied the repository into my enlistment a while back to test and it worked for me, so hopefully it still will.

mconnew:
@dagood, can you explain with your latest changes how we would run our tests against the latest available versions?

roncain:
I presume we will use this via something like build /p:FloatingTestRuntimeProperties=true.  It will be interesting to try this with the new build.sh on Linux. @dagood , is that correct?

dagood:
@roncain @mconnew Yep, it's `/p:FloatingTestRuntimeDependencies=true`. For example, I've been doing `build.cmd /p:FloatingTestRuntimeDependencies=true` to restore + build + run tests. Are there other flows for running tests I should make sure work?

dagood:
I found `/p:WithCategories=OuterLoop` in the developer guide and tried it out successfully. I can't tell for sure where it's getting the dlls from, but if it's `bin/tests/[platform]/[project]/dnxcore50/` then it looks good. (The dll versions copied there do depend on whether `FloatingTestRuntimeDependencies` is true.)

I also added a small section to that doc with usage info for the property.

roncain:
For the record, I ran this on Linux:
sudo bash ./build.sh /p:WithCategories=OuterLoop /p:BridgeHost=roncainl1 /p:FloatingTestRuntimeDependencies=true  > out.log 2>&1 

It looks like there were no new failures running this way, and I see the packages folder has pulled in the 23705 packages published a few hours ago. Not all packages were 23705, but it appears the ones that weren't were still uploading to myget when I ran the test.  I can repeat the test but would prefer to get this merged and finesse any dangling issues if they exist.

  @dagood assuming this worked for you in TFS too (TFS is currently broken for me), then I'm still okay to merge (LGTM).

@mconnew or @iamjasonp -- any concerns with merging?

dagood:
I haven't tried it very recently in TFS. (Only with the original lockfile removal.) I have an enlistment that should be clean, so I'll try it there.

mconnew:
LGTM now that we have `FloatingTestRuntimeDependencies` I'm happy with this change. Thanks @dagood for adding support for this, it will help with our stability a lot.

roncain:
LGTM.  @dagood would you mind being the one to push Merge pull request (our convention)?
I've done a full build of TFS with these changes in the enlistment, and there were no build errors.
I've also built this in Linux, both with and without the FloatingTestRuntimeDependencies property set, and the only test failures are ones we already know about.

Once this is merged, we should monitor the rolling OTG build, but it all looks good to me.

roncain:
Chatted with @dagood, and I will do the Merge.

Issue #270, opened 2015-08-13
Either enable Digest authentication test without a domain or add an instruction to run it in a domain


hongdai:
It's preferable to run it in the CI machine. If not possible, we should add an instruction to run it in a domain.

roncain:
We ultimately need all developers to be able to run this test, even if it requires special setup and xunit traits.  I understand we may be blocked on the xunit traits right now.  We should explore what new things we can do now that Bridge can run on a different machine.

hongdai:
We have enabled a digest authentication test without a domain. However, it's still valuable to keep the domain test as it's the major scenario. The remaining work item here is to add the instruction to the manual test document.

roncain:
I think the remaining work items are for me to use [ConditionalFact] on this test and add documentation for it.

roncain:
Possible approaches are:
1. Compare env var USERDOMAIN to machine name,  If not same, it is domain joined (quick and dirty)
2. Compare env var LOGONSERVER and COMPUTERNAME.  If not same, is domain joined (quick and dirty)
3. Create new WebAPI endpoint that requires windows auth, then have BridgeClient attempt to contact it via HttpClient using default credentials.  Harder but is more accurate test.  Trade-off -- BridgeClient doesn't support windows auth, so test needs to be conditional.

roncain:
Closing because we have added [ConditionalFact] to assert this test runs only on a domain-joined machine.  We still need #1007 and #958 to be resolved to detect when we are domain-joined before this test will run again.

PR #653, opened 2015-12-22
Remove Digest test that requires a domain

### LINKED ISSUE #69: Investigate if it's possible to enable basic authentication test without a domain
Basic authentication test is currently disabled as the test needs to be run on a domain with domain user. We need to investigate if it's possible to run it without a domain.

hongdai:
* this test is no longer needed as Matt has added a digest test without
the need of a domain.

Fixes #69


hongdai:
Not ready for review yet.

hongdai:
Ready for review now.

mconnew:
Is there any benefit in having a test which uses digest in a domain? If the test client/server are in a domain, it would be useful to be able to run this test.

hongdai:
@mconnew Agree it's useful to keep the test for running in a domain. I will update the PR.

iamjasonp:
LGTM

roncain:
LGTM.  I will work on issue #270 to make this test run conditionally.

PR #658, opened 2015-12-29
Enable and document linux (Ubuntu) build instructions

mconnew:


Documentation/building/unix-instructions.md
```diff
@@ -0,0 +1,75 @@
...
+```
+### Switch to the mono snapshot build
+```
+. mono-snapshot mono
+```
+
+# Known Issues
+If you see errors along the lines of `SendFailure (Error writing headers)` you may need to import trusted root certificates:
+```
+mozroots --import --sync
```
iamjasonp: Do you know what the conditions are which causes this to be necessary? 
I wonder if it's a good thing to just include them in the "standard" instructions above. Is it a destructive operation?

dir.props
```diff
@@ -10,7 +10,7 @@
 
   <!-- Build Tools Versions -->
   <PropertyGroup>
-    <BuildToolsVersion>1.0.25-prerelease-00134</BuildToolsVersion>
+    <BuildToolsVersion>1.0.25-prerelease-00137</BuildToolsVersion>
```
roncain: Should we go up to 141 to match CoreFx?  As long as 137 doesn't break TFS builds, we can do that at a later time.

Documentation/building/unix-instructions.md
mconnew: I have no idea. I copied this from CoreFx. I created a second VM to verify the instructions and didn't run that command and everything built fine. I left it in as the only root cert stuff done that I'm aware of is downloading from nuget/myget and we do that too. It could be an issue with a different unix variation.

dir.props
mconnew: I'll do a quick(-ish) verification of using 141 in TFS as my change doesn't modify anything else related to the TFS build.

roncain:
LGTM -- I'm inclined to get this merged asap so we can all try it out.  I can't build TFS successfully right now so I can't prove whether the buildtools version will break it.  So we just need to watch the rolling build or try it manually.

dir.props
mconnew: Verified 141 builds in TFS without issue. I'm also going to modify my PR to have an extra prior commit to the build script and build instructions with the version from CoreFx so that the diff can easily be seen.

roncain:
LGTM -- I can now run build.sh to run InnerLoop and OuterLoop tests with this PR.  We may want to consider calling the cert cleanup exe in a different PR.

PR #660, opened 2016-01-04
Remove OpenExistingOnly flag

hongdai:
* This flag is no longer needed according to Jason. 
Removing it because it requires we pre-create cert folders on Linux.


roncain:
@hongdai -- LGTM as long as you have tested this in Linux (after manually deleting the cert folders you added to make the tests pass).

hongdai:
I have tested it in Linux and verified the cert folders were recreated.

PR #662, opened 2016-01-04
Add a test to cover OperationContextScope major scenario

### LINKED ISSUE #655: Need to add a scenario test for OperationContextScope
WCF Clients can use OperationContextScope to establish current operation context. This is useful for things like modify outgoing message headers and other properties.

hongdai:
* we currently does not have OperationContextScope coverage. This is to
ensure we cover the major usage.

Fixes #655

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.ServiceModel;
 using System.ServiceModel.Channels;
 using System.ServiceModel.Description;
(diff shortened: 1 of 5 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +174,48 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
...
+    [OuterLoop]
+    public static void OperationContextScope_HttpRequestCustomMessageHeader_RoundTrip_Verify()
+    {
+        BasicHttpBinding binding = new BasicHttpBinding();
+        MyClientBase<IWcfService> client = new MyClientBase<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
+        IWcfService serviceProxy = client.ChannelFactory.CreateChannel();
+
+        string customHeaderName = "OperationContextScopeCustomHeader";
+        string customHeaderNS = "http://Microsoft.WCF.OperationContextScope_HttpRequestCustomMessageHeader_RoundTrip_Verify";
+        string customHeaderValue = "CustomHappyValue";
```
roncain: Does the namespace need to be exactly this string?  I think we should probably adopt a convention of using tempuri for this kind of usage to show it's not a real namespace or URI.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +174,48 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
...
+
+                string result = serviceProxy.GetRequestCustomHeader(customHeaderName, customHeaderNS);
+
+                Assert.Equal(customHeaderValue, result);
+            }
+
+            //Call outside of scope should not have the custom header
+            string outofScopeResult = serviceProxy.GetRequestCustomHeader(customHeaderName, customHeaderNS);
+            Assert.Equal(null, outofScopeResult);
+            ((ICommunicationObject)client).Close();
```
roncain: If would be more informative to use Assert.True(oosr == null, "some descriptive message") so we could diagnose a failure on a lab machine.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
```diff
@@ -174,6 +174,48 @@ public static void ClientBaseOfT_Sync_RoundTrip_Call_Using_NetTcpTransport()
...
+                MessageHeader header
+                  = MessageHeader.CreateHeader(
+                  customHeaderName,
+                  customHeaderNS,
+                  customHeaderValue
+                  );
+                OperationContext.Current.OutgoingMessageHeaders.Add(header);
+
+                string result = serviceProxy.GetRequestCustomHeader(customHeaderName, customHeaderNS);
+
```
roncain: Can we add a new server side operation that gets called from this method and returns the custom header information?  That way we could evaluate that the server received the custom header we set inside this method. The server should not do any validation, just return all the custom headers.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
hongdai: Make sense. I'll change it to tempuri.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
hongdai: Agreed. I'll make the change.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/ClientBaseTests.cs
hongdai: The server currently find the value of the specified custom header from it's incoming headers , it returns null if not found. The the test fails if it gets null result.

Return all custom headers could make the test validation more obvious. I'll make the change.

hongdai:

roncain:
LGTM

PR #663, opened 2016-01-04
Exclude System.Security.Principal.Windows for *NIX builds

### LINKED ISSUE #604: Exclude System.Security.Principal.Windows for *NIX builds

mconnew:
Fixes #604

### DIFF
--- a/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
+++ b/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
@@ -28,6 +28,9 @@
     <ProjectJson>$(MsBuildThisFileDirectory)windows\project.json</ProjectJson>
     <ProjectLockJson>$(MsBuildThisFileDirectory)windows\project.lock.json</ProjectLockJson>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(IsNETNativeLibrary)' != 'true' AND '$(TargetsWindows)' == 'true'">
+    <DefineConstants>>$(DefineConstants);SUPPORTS_WINDOWSIDENTITY</DefineConstants>
+  </PropertyGroup>
   <ItemGroup>
     <Compile Include="$(CommonPath)\System\NotImplemented.cs">
       <Link>Common\System\NotImplemented.cs</Link>
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimSet.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimSet.cs
@@ -36,9 +36,7 @@ namespace System.IdentityModel.Claims
         {
             get
             {
-#if FEATURE_NETNATIVE // NegotiateStream
-                throw ExceptionHelper.PlatformNotSupported("Windows Stream Security is not supported on UWP yet");
-#else 
+#if SUPPORTS_WINDOWSIDENTITY // NegotiateStream
                 if (s_windows == null)
                 {
                     List<Claim> claims = new List<Claim>(2);
@@ -48,7 +46,9 @@ namespace System.IdentityModel.Claims
                     s_windows = new DefaultClaimSet(claims);
                 }
                 return s_windows;
-#endif // FEATURE_NETNATIVE
+#else 
+                throw ExceptionHelper.PlatformNotSupported("Windows Stream Security is not supported on UWP yet");
+#endif // SUPPORTS_WINDOWSIDENTITY
             }
         }
 
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimSet.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimSet.cs
@@ -47,7 +47,7 @@ namespace System.IdentityModel.Claims
                 }
                 return s_windows;
 #else 
-                throw ExceptionHelper.PlatformNotSupported("Windows Stream Security is not supported on UWP yet");
+                throw ExceptionHelper.PlatformNotSupported(ExceptionHelper.WinsdowsStreamSecurityNotSupported);
 #endif // SUPPORTS_WINDOWSIDENTITY
             }
         }
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WindowsStreamSecurityUpgradeProvider.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WindowsStreamSecurityUpgradeProvider.cs
@@ -223,7 +223,7 @@ namespace System.ServiceModel.Channels
                 remoteSecurity = CreateClientSecurity(negotiateStream, _parent.ExtractGroupsForWindowsAccounts);
                 return negotiateStream;
 #else
-                throw ExceptionHelper.PlatformNotSupported("NegotiateStream is not supported on UWP yet"); 
+                throw ExceptionHelper.PlatformNotSupported(ExceptionHelper.WinsdowsStreamSecurityNotSupported); 
 #endif // SUPPORTS_WINDOWSIDENTITY
             }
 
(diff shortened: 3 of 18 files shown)


src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
```diff
@@ -28,6 +28,9 @@
     <ProjectJson>$(MsBuildThisFileDirectory)windows\project.json</ProjectJson>
     <ProjectLockJson>$(MsBuildThisFileDirectory)windows\project.lock.json</ProjectLockJson>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(IsNETNativeLibrary)' != 'true' AND '$(TargetsWindows)' == 'true'">
+    <DefineConstants>>$(DefineConstants);SUPPORTS_WINDOWSIDENTITY</DefineConstants>
+  </PropertyGroup>
```
roncain: Nit: we originally chose "FEATURE" as the prefix for our conditional compiles, so FEATURE_WINDOWSIDENTITY might be more consistent.  Not required, just observing it.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimSet.cs
```diff
@@ -48,7 +46,9 @@ public static ClaimSet Windows
                     s_windows = new DefaultClaimSet(claims);
                 }
                 return s_windows;
-#endif // FEATURE_NETNATIVE
+#else 
+                throw ExceptionHelper.PlatformNotSupported("Windows Stream Security is not supported on UWP yet");
+#endif // SUPPORTS_WINDOWSIDENTITY
```
roncain: Should we remove a reference to UWP from the exception message?  Might be misleading.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WindowsStreamSecurityUpgradeProvider.cs
```diff
@@ -224,7 +222,9 @@ protected override Stream OnAcceptUpgrade(Stream stream, out SecurityMessageProp
 
                 remoteSecurity = CreateClientSecurity(negotiateStream, _parent.ExtractGroupsForWindowsAccounts);
                 return negotiateStream;
-#endif // FEATURE_NETNATIVE
+#else
+                throw ExceptionHelper.PlatformNotSupported("NegotiateStream is not supported on UWP yet"); 
+#endif // SUPPORTS_WINDOWSIDENTITY
```
roncain: I'll stop mentioning that we might want to remove UWP from all these PNSE exception messages.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
```diff
@@ -791,15 +791,15 @@ static internal bool IsDefault(NetworkCredential credential)
...
-#else
+#if SUPPORTS_WINDOWSIDENTITY
                 using (WindowsIdentity self = WindowsIdentity.GetCurrent())
                 {
                     return self.User.Value;
                 }
+#else
+                // There's no way to retrieve the current logged in user Id in UWP apps
+                // so returning a username which is very unlikely to be a real username;
+                return "_______****currentUser****_______";
```
roncain: How about a name like "____CURRENTUSER_NOT_AVAILABLE____" or something more informative?  Also the comment uses UWP.

roncain:

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
mconnew: I see FEATURE as our feature, whereas SUPPORTS I see as meaning the platform we're building on supports this dependency.

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimSet.cs
mconnew: It's all through the code. I'll go through and update each exception message to stop at the word "supported". If only this was a string resource or a constant, it would be so much easier. What do you think about adding a constant to ExceptionHelper for this message?

roncain: Your call on whether to make a new constant for it.  Could also be a helper method that returns a PNSE with that message in it.

roncain:
LGTM, subject to your choice of how to fix the exception message.  CI keeps failing in package restore so we'll have to buddy build.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
```diff
@@ -137,7 +137,6 @@ public static void DigestAuthentication_RoundTrips_Echo()
 
     [Fact]
     [OuterLoop]
-    [ActiveIssue(604, PlatformID.AnyUnix)]
     public static void NtlmAuthentication_RoundTrips_Echo()
```
roncain: LGTM.
I ran this PR on Linux, and now that these tests are no longer excluded, they are run and they fail.
Should we open a new issue and keep an [ActiveIssue] on these or just leave them failing and investigate the failure?  Don't block the merge for this question.

    Https_ClientCredentialTypeTests.NtlmAuthentication_RoundTrips_Echo [FAIL]
        Test Case: NtlmAuthentication FAILED with the following errors:     Error: Unexpected exception was caught while doing the basic echo test for variation...
        'BasicHttpBinding with NTLM authentication'
        Exception: System.ServiceModel.CommunicationException: An error occurred while sending the request. ---> System.Net.Http.HttpRequestException: An error occurred while sending the request. ---> System.Net.Http.CurlException: Peer certificate cannot be authenticated with given CA certificates
           at System.Net.Http.CurlHandler.MultiAgent.FinishRequest(EasyRequest completedOperation, CURLcode messageResult)

mconnew: Those tests are using default credentials. We should have a test which uses explicit credentials with NTLM. I can't remember whether NTLM requires a domain or whether you can authenticate against a local machine account. If the later, then we could have NTLM tests running regularly.

roncain: I opened #670 to add a test with explicit credentials.  @mconnew what do you recommend we do with this test now?  Removing ActiveIssue makes it fail.  Are you saying it cannot succeed on Linux using default credentials ever, or is there a new bug here that needs to be opened? Ideally we want all test failures either deactivated with [ActiveIssue] or left failing for further investigation.

roncain:

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
mconnew: I'm not sure how the NTLM authentication code works on *NIX. If there's a concept of being logged in, there might be some default credentials which could work. @stephentoub , as you've done some work with the libcurl implementation of HttpClientHandler, do you know the answer to this?

roncain:
LGTM.  We can resolve how to handle the failing NTLM test post-merge.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
stephentoub: To my knowledge there's currently no support for NTLM and default credentials.  @vijaykota and @shrutigarg might be able to speak to whether that support is coming / will be possible in the future?

vijaykota: Apologies for the delayed response. libcurl and hence HttpClient on Unix support NTLM with explicit credentials. However default credentials are a no-op on Unix. @kapilash did propose a way to implement this using .netrc files but there is currently no planned work for this.

PR #669, opened 2016-01-06
Add scenario test for FaultException using primitive type.

roncain:
This PR adds a new scenario test that demonstrates primitive types
can be used in FaultContracts.  It is known to succeed in CoreCLR
but expected to fail in NET Native.

Making this test succeed in NET Native will occur in a different PR.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Contract/Fault/FaultExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Contract/Fault/FaultExceptionTests.cs
@@ -89,4 +89,42 @@ public static class FaultExceptionTests
 
         Assert.True(false, "Expected FaultException<FaultDetail> exception, but no exception thrown.");
     }
+
+    [Fact]
+    [OuterLoop]
+    public static void FaultException_Throws_With_Int()
+    {
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
+        BasicHttpBinding binding = null;
+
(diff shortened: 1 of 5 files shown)


roncain:
Testing appears to show the latest toolchain is correctly handling FaultException<int> now because this test passes in NET Native without any other changes.  I recommend we leave the rd.xml file untouched so that we don't break any apps using the released toolchain if they update to later WCF packages.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Fault/FaultExceptionTests.cs
```diff
@@ -89,4 +89,42 @@ public static void UnexpectedException_Throws_FaultException()
...
+            serviceProxy = factory.CreateChannel();
+
+            // *** EXECUTE *** \\
+            thrownException = Assert.Throws <FaultException<int>>(() =>
+            {
+                serviceProxy.TestFaultInt(expectedFaultCode);
+            });
+
+            // *** VALIDATE *** \\
+            Assert.Equal(expectedFaultCode, thrownException.Detail);
```
iamjasonp: CHECK: that this propagates in .NET Native as well. I think that we've had issues with exception messages not propagating correctly, and I don't recall if the `.Detail` does.

iamjasonp: Ah, I see the comment that says this passes in .NET Native already, missed that earlier.

iamjasonp:
LGTM

PR #673, opened 2016-01-07
Adding .builds file and applying new conventions to S.P.SM.csproj

### LINKED ISSUE #637: Update naming conventions for target frameworks in WCF repo
Names used for target frameworks currently used in the WCF repo are not the same as those used in the CoreFx repo. 

We should update our repo to use the same names as specified in https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/project-guidelines.md to make things consistent. 

When doing this work, we also need to do the same for the ToF wrappers. 

(related comment: https://github.com/mellinoe/wcf/commit/c27cd216b45f7b1c117fb4734be7e01dddd40721#commitcomment-14939710)

StephenBonikowsky:
* Once this looks good we can do the same work for the rest of the WCF projects.
* Fixes #637

### DIFF
--- a/src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
+++ b/src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
@@ -2,7 +2,6 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <ItemGroup>
-    <Project Include="System.Private.ServiceModel.csproj" />
     <Project Include="System.Private.ServiceModel.csproj">
       <OSGroup>Linux</OSGroup>
     </Project>
--- a/src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
+++ b/src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <ItemGroup>
+    <Project Include="System.Private.ServiceModel.csproj" />
+    <Project Include="System.Private.ServiceModel.csproj">
+      <OSGroup>Linux</OSGroup>
+    </Project>
+    <Project Include="System.Private.ServiceModel.csproj">
+      <OSGroup>Windows_NT</OSGroup>
+    </Project>
+    <Project Include="System.Private.ServiceModel.csproj">
+      <OSGroup>Windows_NT</OSGroup>
+      <TargetGroup>netcore50</TargetGroup>
+    </Project>
+    <Project Include="System.Private.ServiceModel.csproj">
+      <TargetGroup>net46</TargetGroup>
+    </Project>
+  </ItemGroup>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.traversal.targets))\dir.traversal.targets" />
+</Project>
--- a/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
+++ b/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
@@ -15,12 +15,24 @@
     <PlatformTarget>AnyCPU</PlatformTarget>
     <ProjectGuid>{9E50E7BF-CD6E-4269-A6DD-59FD0BD6C0FD}</ProjectGuid>
     <CommonPath Condition="'$(CommonPath)' == ''">..\..\Common\src</CommonPath>
-    <IsNETNativeLibrary Condition="'$(TargetGroup)' == 'netcore50'">true</IsNETNativeLibrary>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(IsNETNativeLibrary)' == 'true'">
+
+  <!-- Default configurations to help VS understand the options -->
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Linux_Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Linux_Release|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_Debug|AnyCPU'"/>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_Release|AnyCPU'"/>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_netcore50_Debug|AnyCPU'"/>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_netcore50_Release|AnyCPU'"/>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'net46_Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'net46_Release|AnyCPU'" />
+
+  <PropertyGroup Condition="'$(TargetGroup)' == 'netcore50'">
     <DefineConstants>$(DefineConstants);FEATURE_NETNATIVE</DefineConstants>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(IsNETNativeLibrary)' != 'true'">
+  <PropertyGroup Condition="'$(TargetGroup)' != 'netcore50'">
     <DefineConstants>$(DefineConstants);FEATURE_CORECLR</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetsWindows)' == 'true'">
@@ -41,7 +53,7 @@
     <None Include="project.json" />
     <None Include="windows\project.json" />
   </ItemGroup>
-  <ItemGroup Condition="'$(IsNETNativeLibrary)' == 'true'">
+  <ItemGroup Condition="'$(TargetGroup)' == 'netcore50'">
     <EmbeddedResource Include="$(MsBuildThisFileDirectory)Resources\$(AssemblyName).rd.xml" />
     <TargetingPackReference Include="Windows" />
   </ItemGroup>
--- a/src/System.ServiceModel.Http/src/System.ServiceModel.Http.builds
+++ b/src/System.ServiceModel.Http/src/System.ServiceModel.Http.builds
@@ -2,7 +2,6 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <ItemGroup>
-    <Project Include="System.ServiceModel.Http.csproj" />
     <Project Include="System.ServiceModel.Http.csproj" >
       <TargetGroup>net46</TargetGroup>
     </Project>
--- a/src/System.ServiceModel.Http/src/System.ServiceModel.Http.builds
+++ b/src/System.ServiceModel.Http/src/System.ServiceModel.Http.builds
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <ItemGroup>
+    <Project Include="System.ServiceModel.Http.csproj" />
+    <Project Include="System.ServiceModel.Http.csproj" >
+      <TargetGroup>net46</TargetGroup>
+    </Project>
+    <Project Include="System.ServiceModel.Http.csproj" >
+      <OSGroup>Windows_NT</OSGroup>
+    </Project>
+    <Project Include="System.ServiceModel.Http.csproj" >
+      <OSGroup>Windows_NT</OSGroup>
+      <TargetGroup>netcore50</TargetGroup>
+    </Project>
+  </ItemGroup>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.traversal.targets))\dir.traversal.targets" />
+</Project>
--- a/src/System.ServiceModel.NetTcp/src/System.ServiceModel.NetTcp.builds
+++ b/src/System.ServiceModel.NetTcp/src/System.ServiceModel.NetTcp.builds
@@ -2,7 +2,6 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <ItemGroup>
-    <Project Include="System.ServiceModel.NetTcp.csproj" />
     <Project Include="System.ServiceModel.NetTcp.csproj" >
       <TargetGroup>net46</TargetGroup>
     </Project>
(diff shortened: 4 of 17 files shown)


src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
```diff
@@ -15,12 +15,20 @@
...
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'netcore50_Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'netcore50_Release|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'dnxcore50_Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'dnxcore50_Release|AnyCPU'" />
+
+  <PropertyGroup Condition="'$(TargetGroup)' == 'netcore50'">
     <DefineConstants>$(DefineConstants);FEATURE_NETNATIVE</DefineConstants>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(IsNETNativeLibrary)' != 'true'">
+  <PropertyGroup Condition="'$(TargetGroup)' != 'netcore50'">
```
roncain: IsNetNativeLibrary is being set explicitly in our ToF wrapper projects' settings file when it detects it is being run against N -- so that it would trigger these conditional properties and constants.  So we may need to alter the TFS Tof WCF.settings.targets file to set $(TargetGroup) instead.  We could make the settings file set both properties first so that it works with both the unmodified OSS code as well as the modified OSS code.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <ItemGroup>
+    <Project Include="System.Private.ServiceModel.csproj" />
+	<Project Include="System.Private.ServiceModel.csproj">
+      <OSGroup>Linux</OSGroup>
+	  <TargetGroup>dnxcore50</TargetGroup>
```
roncain: Do you know if we need explicit entries for OSX and FreeBSD too?  We are currently being build and run for OSX in the Helix labs today.  I'm still trying to understand when explicit entries are needed in this ItemGroup.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
```diff
@@ -15,12 +15,20 @@
...
+  <!-- Default configurations to help VS understand the options -->
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'netcore50_Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'netcore50_Release|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'dnxcore50_Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'dnxcore50_Release|AnyCPU'" />
+
+  <PropertyGroup Condition="'$(TargetGroup)' == 'netcore50'">
     <DefineConstants>$(DefineConstants);FEATURE_NETNATIVE</DefineConstants>
```
roncain: In a future PR (not this one), I would like to change the constant FEATURE_NETNATIVE to be 'netcore50' and FEATURE_CORECLR to be dnxcore50.  Part of the rename goal was to try to use consistent naming across MSBuild properties, conditional compile if's, files, etc.  Just something to think about for the next sets of PR's.

roncain: I think we need a similar conditional property that adds FEATURE_CORECLR to $(DefineConstants) if the $(TargetGroup) is dnxcore50

roncain:
It is my understanding this file will not be built automatically until we duplicate changes CoreFx made in their build.proj (which does a traversal for *.builds).  Do you want to make those changes in this PR so we can test that this builds correctly?  Or do you prefer to get all the other pieces in place before enabling it?  These can be separate PR's or commits, however you prefer (though I notice other teams seem to be doing single PR's that make all the changes at once).

src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
StephenBonikowsky: I wondered about this too, I think we just add what we want to support? Are we ready for OSX and FreeBSD?

StephenBonikowsky:
It makes sense that we include everything needed in one PR that is needed to make any changes work.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
roncain: Yes, I think we are ready.  It looks like Helix is already running our legacy ToF tests on at least Ubuntu, Centos, and OSX (and Windows CoreCLR).  So I'd say, yes, let's add entries to say we build for OSX and FreeBSD and see what Helix reveals.

StephenBonikowsky: My understanding now is that you only explicitly list an <OSGroup> if you need to build it differently (as in forked code) so at this point we build Linux differently (because of dependency on System.Net.Http.WinHttpHandler) and Windows (because we build a netcore50 flavor in addition to dnxcore50).  
I will be updating this PR based on my new understanding.

StephenBonikowsky:
ToDo For This PR:
- Update Build.proj
- Update the façade .builds files that are affected by the "TARGETS_WINDOWS" #ifdef used to fork builds for Linux. @mconnew I couldn't find which public contract was affected by this, do you know?

StephenBonikowsky:
@weshaggard and @mellinoe would you be willing to take a look at the .builds files I created and the changes I made to the configuration lists of the .csproj files? One thing I wasn't sure about...

In .builds files do I need the section with OSGroup Windows_NT where I don't specify a TargetGroup, since I already specified one for TargetGroup netcore50 I wasn't sure if I needed to for our dnxcore50 flavor or if that was already done by default with the first <Project Include=...

src/System.ServiceModel.Http/src/System.ServiceModel.Http.builds
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <ItemGroup>
+    <Project Include="System.ServiceModel.Http.csproj" />
+    <Project Include="System.ServiceModel.Http.csproj" >
+      <TargetGroup>net46</TargetGroup>
+    </Project>
+    <Project Include="System.ServiceModel.Http.csproj" >
```
weshaggard: Do you want this project to build for every OSGroup or just Windows_NT? If you want just Windows_NT then keep this entry and delete the first Project in this item group if you want every then delete this one and keep the first one.

src/System.ServiceModel.NetTcp/src/System.ServiceModel.NetTcp.builds
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <ItemGroup>
+    <Project Include="System.ServiceModel.NetTcp.csproj" />
+    <Project Include="System.ServiceModel.NetTcp.csproj" >
+      <TargetGroup>net46</TargetGroup>
+    </Project>
+    <Project Include="System.Private.NetTcp.csproj">
```
weshaggard: Same question as with System.SErviceModel.Http

src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
weshaggard: You should only ever list entries for different OSGroup's if you have something unique for those OSGroups. Otherwise if you are pure IL and don't have anything unique don't list any OSGroup and your project will be "AnyOS" and thus will be available on any OS.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <ItemGroup>
+    <Project Include="System.Private.ServiceModel.csproj" />
+    <Project Include="System.Private.ServiceModel.csproj">
```
weshaggard: As I called out in my other comment you should only have the first entry in this item group which means it default to AnyOS and delete the other entires with specific OSGroup defined, unless you actually have conditions in your projects for different OSGroups, which I don't believe you guys do yet.

src/System.ServiceModel.Security/src/System.ServiceModel.Security.csproj
```diff
@@ -10,8 +10,10 @@
...
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|AnyCPU'" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|AnyCPU'" />
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'netcore50_Debug|AnyCPU'" />
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'netcore50_Release|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_Debug|AnyCPU'"/>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_Release|AnyCPU'"/>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_netcore50_Debug|AnyCPU'"/>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_netcore50_Release|AnyCPU'"/>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'net46_Debug|AnyCPU'" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'net46_Release|AnyCPU'" />
```
weshaggard: For the line below this '<TargetingPackReference Include="mscorlib" Condition="'$(TargetGroup)' != 'netcore50aot'" />` you should change that to `<TargetingPackReference Include="mscorlib" Condition="'$(TargetGroup)' == 'net46'" />` because you only need the mscorlib reference when you are targeting net46. The other targets will use System.Private.ServiceModel.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
```diff
@@ -38,7 +50,7 @@
     <None Include="project.json" />
     <None Include="windows\project.json" />
```
weshaggard: What is the windows\project.json file for?

src/System.Private.ServiceModel/src/System.Private.ServiceModel.builds
roncain: @weshaggard -- is it correct to assume we should add a separate item for OSX and FreeBSD too?  We conditionally use TargetsWindows conditionally and want to be able to build for FreeBSD and OSX as well as Linux.  We don't conditionally compile anything specific for OSX or Linux, just Windows vs non-Windows.

weshaggard: We still have to workout the full OS support graph but for now I would say only add the entries for the OS's you distinctly need to build for. If it is just Windows vs Unix, I would just have the Windows and add Linux as a representative for Unix for now. We will likely add a Unix flavor at some point.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
roncain: @dagood -- could you elaborate on the windows\project.json file addition?  Thanks.

src/System.ServiceModel.Http/src/System.ServiceModel.Http.builds
roncain: @weshaggard -- because all these libraries are facades over System.Private.ServiceModel, I assume we want the shape of this ItemGroup in each façade libary to exactly mirror S.P.ServiceModel.builds, differing only in the project file they specify.  Is this correct?

weshaggard: I don't think that is necessary. In general I think the .builds file for a given library should remain agnostic to its dependencies and only build for the configurations that it needs itself. Given these are a facade I would expect you only need the "Empty", "net46", and "netcore50" for this guy. It does mean however you should explicitly pick a version of S.P.SM to reference. See https://github.com/dotnet/corefx/blob/master/src/System.Collections/src/System.Collections.csproj#L20 where I explicitly always pick the Windows_NT version of its dependency. It really doesn't matter which you pick as long as the public surface is the same which it should be in this case.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
dagood: I see that the windows project.json adds `System.Net.Http.WinHttpHandler` to the `dnxcore50` framework dependencies and `Microsoft.TargetingPack.Private.NetNative` to `netcore50`. I don't know what makes this necessary, though. (I also don't remember being involved in this addition--is there a specific reason you pinged me?)

src/System.ServiceModel.Security/src/System.ServiceModel.Security.csproj
StephenBonikowsky: This has been there since @mellinoe  initially added these files, I'll test if its needed separately from this PR.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
StephenBonikowsky: @mconnew has the details but these are dependencies needed for a specific WCF scenario to work. These dependencies are not supported on Linux which is why they are pulled in on a Windows only condition.

StephenBonikowsky:
Regarding the 3d commit, I didn't update...


@roncain

roncain:
Trying to pull this PR gives me merge conflicts in dir.props and build.proj.  @stebon you probably need to revert these files and reapply only the edits we need on top of what's there.

src/System.ServiceModel.Http/src/System.ServiceModel.Http.builds
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <ItemGroup>
+    <Project Include="System.ServiceModel.Http.csproj" >
```
mconnew: I'm not sure the list of variations is correct here. I think we need just net46 and "everything else". On net46, we forward to desktop so don't need any implementation. On everything else we type forward to system.private.servicemodel.

mconnew:
One small issue about what filters to apply for the various implementations on the contract assemblies, but I don't think you should block on that. So :+1:

StephenBonikowsky:

src/System.ServiceModel.Http/src/System.ServiceModel.Http.builds
weshaggard: Yes that is likely correct. If you can verify that the resulting binary for each configuration is the same I agree you should collapse them.

src/System.ServiceModel.Primitives/src/System.ServiceModel.Primitives.builds
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <ItemGroup>
+    <Project Include="System.ServiceModel.Primitives.csproj" >
+      <TargetGroup>net46</TargetGroup>
+    </Project>
+    <Project Include="System.ServiceModel.Primitives.csproj" >
```
weshaggard: Similar these can probably be collapsed

PR #674, opened 2016-01-07
Modify CertificateGenerator to fix Linux testing issues

iamjasonp:
See individual commits for change description

### DIFF
--- a/src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
+++ b/src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
@@ -35,7 +35,8 @@ namespace WcfTestBridgeCommon
 
         // This can't be too short as there might be a time skew between machines,
         // but also can't be too long, as the CRL is cached by the machine
-        private TimeSpan _crlValidityPeriod = TimeSpan.FromMinutes(2);
+        private TimeSpan _crlValidityGracePeriodStart = TimeSpan.FromMinutes(5);
+        private TimeSpan _crlValidityGracePeriodEnd = TimeSpan.FromMinutes(5);
 
         // Give the cert a grace period in case there's a time skew between machines
         private readonly TimeSpan _gracePeriod = TimeSpan.FromHours(1);
@@ -281,7 +282,8 @@ namespace WcfTestBridgeCommon
             _certGenerator.SetSignatureAlgorithm(_signatureAlthorithm);
 
             X509Name authorityX509Name = CreateX509Name(_authorityCanonicalName);
-            
+            var serialNum = new BigInteger(64 /*sizeInBits*/, _random).Abs();
+
             var keyPair = isAuthority ? _authorityKeyPair : _keyPairGenerator.GenerateKeyPair();
             if (isAuthority)
             {
@@ -291,10 +293,10 @@ namespace WcfTestBridgeCommon
                 var authorityKeyIdentifier = new AuthorityKeyIdentifier(
                     SubjectPublicKeyInfoFactory.CreateSubjectPublicKeyInfo(_authorityKeyPair.Public),
                     new GeneralNames(new GeneralName(authorityX509Name)),
-                    new BigInteger(7, _random).Abs());
+                    serialNum);
 
-                _certGenerator.AddExtension(X509Extensions.AuthorityKeyIdentifier, true, authorityKeyIdentifier);
-                _certGenerator.AddExtension(X509Extensions.KeyUsage, true, new KeyUsage(X509KeyUsage.DigitalSignature | X509KeyUsage.KeyAgreement | X509KeyUsage.KeyCertSign | X509KeyUsage.KeyEncipherment | X509KeyUsage.CrlSign));
+                _certGenerator.AddExtension(X509Extensions.AuthorityKeyIdentifier, false, authorityKeyIdentifier);
+                _certGenerator.AddExtension(X509Extensions.KeyUsage, false, new KeyUsage(X509KeyUsage.DigitalSignature | X509KeyUsage.KeyAgreement | X509KeyUsage.KeyCertSign | X509KeyUsage.KeyEncipherment | X509KeyUsage.CrlSign));
 
             }
             else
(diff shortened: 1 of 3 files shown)


iamjasonp:

iamjasonp:
I have most of the Linux certificate security tests in the `Security.TransportSecurity.Tests` test class previously failing in Linux succeed by:

* pulling this PR 
* start the Bridge in Windows, allowing remote access <br/>
  `./startBridge -allowRemote`
* `cd ~/git/wcf`
* `pushd /usr/local/share/ca-certificates`
* `sudo wget -O wcfbridge-ca.crt http://<bridgeHost>:44283/resource/WcfService.CertificateResources.CertificateAuthorityResource?exportAsPem=true`
* `sudo update-ca-certificates`
* `popd`
* `./build.sh /p:WithCategories=OuterLoop /p:BridgeHost=<bridgeHost>`

OpenSSL and therefore cURL uses a "bundle" of SSL root CA certificates to determine whether or not the cert presented is trusted. We need to add the WCF Bridge root cert to the bundle for our tests to pass.

iamjasonp:
Tests still failing: 
* Tcp_ClientCredentialTypeTests.TcpClientCredentialType_Certificate_EchoString 
* Tcp_ClientCredentialTypeTests.TcpClientCredentialType_Certificate_CustomValidator_EchoString

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
```diff
@@ -291,10 +293,10 @@ private X509CertificateContainer CreateCertificate(bool isAuthority, bool isMach
                 var authorityKeyIdentifier = new AuthorityKeyIdentifier(
                     SubjectPublicKeyInfoFactory.CreateSubjectPublicKeyInfo(_authorityKeyPair.Public),
                     new GeneralNames(new GeneralName(authorityX509Name)),
-                    new BigInteger(7, _random).Abs());
+                    serialNum);
```
roncain: Just a question -- the refactored code now uses 64-bits when before it used 7.  Is that also part of this fix?

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
```diff
@@ -291,10 +293,10 @@ private X509CertificateContainer CreateCertificate(bool isAuthority, bool isMach
...
 
-                _certGenerator.AddExtension(X509Extensions.AuthorityKeyIdentifier, true, authorityKeyIdentifier);
-                _certGenerator.AddExtension(X509Extensions.KeyUsage, true, new KeyUsage(X509KeyUsage.DigitalSignature | X509KeyUsage.KeyAgreement | X509KeyUsage.KeyCertSign | X509KeyUsage.KeyEncipherment | X509KeyUsage.CrlSign));
+                _certGenerator.AddExtension(X509Extensions.AuthorityKeyIdentifier, false, authorityKeyIdentifier);
```
roncain: Nit: not required for merge, but it would be better for readability if the parameters all used their names.  To find out what the change from 'true' to 'false' meant, I had to go look at code not in our repo (BouncyCastle).

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
```diff
@@ -364,9 +365,14 @@ private X509CertificateContainer CreateCertificate(bool isAuthority, bool isMach
...
             // Our CRL Distribution Point has the serial number in the query string to fool Windows into doing a fresh query 
             // rather than using a cached copy of the CRL in the case where the CRL has been previously accessed before
-            var crlDistributionPoints = new DistributionPoint[1] {
+            var crlDistributionPoints = new DistributionPoint[2] {
                 new DistributionPoint(new DistributionPointName(
                     new GeneralNames(new GeneralName(
+                        GeneralName.UniformResourceIdentifier, string.Format("{0}?serialNum={1}", _crlUri, serialNum.ToString(radix: 16))))),
+                        null,
+                        null),
+                    new DistributionPoint(new DistributionPointName(
```
roncain: Sub-atomic Nit: outdent the 2nd 'new DistributionPoint. Not required for merge.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
```diff
@@ -35,7 +35,8 @@ public class CertificateGenerator
 
         // This can't be too short as there might be a time skew between machines,
         // but also can't be too long, as the CRL is cached by the machine
-        private TimeSpan _crlValidityPeriod = TimeSpan.FromMinutes(2);
+        private TimeSpan _crlValidityGracePeriodStart = TimeSpan.FromMinutes(5);
+        private TimeSpan _crlValidityGracePeriodEnd = TimeSpan.FromMinutes(5);
 
         // Give the cert a grace period in case there's a time skew between machines
         private readonly TimeSpan _gracePeriod = TimeSpan.FromHours(1);
```
roncain: If there is a time skew between machines, should we be using the "now" of the client (i.e. pass it in as part of the request)?  It might be worth considering (not in this PR) how to compute start and end times that are guaranteed to be consistent with whichever machine is going to validate them (or that has a start time of the earliest of the 2 times and an end time of the latest) to guarantee validity on both machines.

roncain:
LGTM, subject to at least considering some of the Nit's mentioned.  I'm fine if we merge this and than make a separate PR to automate it.  We are still missing the cert cleanup step from the new build.sh, so it might be good for a single PR to automate the manual steps you listed plus cert cleanup.

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
iamjasonp: Yes, this is to fix bug in certificate issuer serial number. Copy+paste issue that dates way back

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateGenerator.cs
iamjasonp: For this particular case, it might not be possible as the request for the CRL is a call made by Windows or OpenSSL based on the CRL URI baked into the certificate itself (i.e., I don't see how it's modifiable per-request)

If we are keen to solve the skew problem for this purpose, we might consider, upon PUT for an Authority certificate or another well-known CertificateResource, to send the "now" of the client over, and then assuming the machine being tested won't correct its time in the meantime. 

In a real-world environment, it's likely that a skew of five minutes will cause problems anyway with various forms of security anyway (e.g., Kerberos). I arrived at five minutes because in Windows: 

> If the KDC verifies that the time it decrypts from the authenticator is within a specified amount of the local time on the KDC (the default is five minutes), the system can assume that the credentials presented are genuine.

https://technet.microsoft.com/en-us/library/cc780011%28v=ws.10%29.aspx?f=255&MSPPError=-2147217396

(there are a couple of other articles stating the same, so it should be a reasonable assumption that clocks should be within five minutes within a given environment)

iamjasonp:
Agreed. Though I'm thinking about how to automate this "insert CA certificate" and "remove CA certificate" steps, which is why it's not already in this PR. 

That might be a little hairy, so I think this should go in first to unblock (most) testing w.r.t. certificates on Linux, even if there's an annoying manual step in between.

roncain:
LGTM to merge as it is and automate in a follow-up.

PR #675, opened 2016-01-07
Call Dispose() on the XmlWriter after use.

### LINKED ISSUE #51: Confirm correct behaviour in usages of System.Xml.XmlWriter.Dispose()
In all cases within WCF where we use an XmlWriter rather than XmlDictionaryWriter, check that:
* `Dispose()` is called on the XmlWriter after use
* All Xml tags/elements/contents are manually closed and doesn't rely on `Dispose()` to do this work

**Background:**
In desktop .NET, [System.Xml.XmlWriter](http://referencesource.microsoft.com/#System.Xml/System/Xml/Core/XmlWriter.cs) has a virtual `Close()` method that is overridden by, say, [System.Xml.XmlTextWriter](http://referencesource.microsoft.com/#System.Xml/System/Xml/Core/XmlTextWriter.cs) that performs actions like closing all outstanding opened elements. 

In .NET Core, [System.Xml.XmlWriter](https://github.com/dotnet/corefx/blob/master/src/System.Xml.ReaderWriter/src/System/Xml/Core/XmlWriter.cs) does not expose a `Close()` method anymore, and neither does [System.Xml.XmlTextWriter](https://github.com/dotnet/corefx/blob/master/src/Common/src/System/Xml/XmlTextWriter.cs). You are expected to call the `Dispose()` method instead. 

We *don't* need to check XmlDictionaryWriter for this issue. `Dispose() == Close()` since [XmlDictionaryWriter](https://github.com/dotnet/corefx/blob/master/src/System.Private.DataContractSerialization/src/System/Xml/XmlDictionaryWriter.cs) (and therefore, derived classes) has a virtual `Close()` method. `XmlDictionaryWriter.Dispose()` will call `XmlDictionaryWriter.Close()`

shmao:
Fixes #51

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/Message.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/Message.cs
@@ -451,20 +451,28 @@ namespace System.ServiceModel.Channels
                 return base.ToString();
             }
 
-            StringWriter stringWriter = new StringWriter(CultureInfo.InvariantCulture);
+            
             XmlWriterSettings settings = new XmlWriterSettings();
             settings.Indent = true;
-            XmlWriter textWriter = XmlWriter.Create(stringWriter, settings);
-            XmlDictionaryWriter writer = XmlDictionaryWriter.CreateDictionaryWriter(textWriter);
-            try
+
+            using (StringWriter stringWriter = new StringWriter(CultureInfo.InvariantCulture))
             {
-                ToString(writer);
-                writer.Flush();
-                return stringWriter.ToString();
-            }
-            catch (XmlException e)
-            {
-                return SR.Format(SR.MessageBodyToStringError, e.GetType().ToString(), e.Message);
+                using (XmlWriter textWriter = XmlWriter.Create(stringWriter, settings))
+                {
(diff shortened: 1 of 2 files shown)


shmao:

shmao:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/Message.cs
```diff
@@ -451,20 +451,24 @@ public override string ToString()
...
-            try
-            {
-                ToString(writer);
-                writer.Flush();
-                return stringWriter.ToString();
-            }
-            catch (XmlException e)
+
+            using (StringWriter stringWriter = new StringWriter(CultureInfo.InvariantCulture))
+            using (XmlWriter textWriter = XmlWriter.Create(stringWriter, settings))
```
roncain: I think it would be better to use curly braces for each nested statement for clarity -- even though it will indent quite a bit.

shmao:

iamjasonp:
LGTM

roncain:
LGTM. OuterLoops pass for me (note that CI is temporarily down).

PR #678, opened 2016-01-11
Enabled more ETW events, including adding activity id tracing

mconnew:
Still to be done:
1. Remove unused events
2. Add entry to rd.xml to allow EventAttribute reflection data to be made available at runtime


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CoreClrSocketConnection.cs
```diff
@@ -550,6 +550,20 @@ protected override void WriteCore(byte[] buffer, int offset, int size, bool imme
             }
         }
 
+        protected override void TraceWriteStart(int size, bool async)
+        {
+            var remoteEndpoint = (IPEndPoint)_socket.RemoteEndPoint;
+            string remoteEndpointAddressString = remoteEndpoint.Address + ":" + remoteEndpoint.Port;
```
roncain: Is _socket guaranteed to be not null? I see TraceSocketReadStop below is checking for null.  Maybe a Contract.Assert here?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CoreClrSocketConnection.cs
```diff
@@ -580,6 +594,20 @@ protected override int ReadCore(byte[] buffer, int offset, int size, TimeSpan ti
...
         }
 
+        protected override void TraceSocketReadStop(int bytesRead, bool async)
+        {
+            var remoteEndpoint = (IPEndPoint)_socket.RemoteEndPoint;
+            string remoteEndpointAddressString = remoteEndpoint.Address + ":" + remoteEndpoint.Port;
+            if (!async)
+            {
+                WcfEventSource.Instance.SocketReadStop((_socket != null) ? _socket.GetHashCode() : -1, bytesRead, remoteEndpointAddressString);
+            }
```
roncain: Hmmm, it checks _socket for null here but not when it reads _socket.RemoteEndpoint above?  When can it be null?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
```diff
@@ -393,7 +393,20 @@ protected override void WriteCore(byte[] buffer, int offset, int size, bool imme
                 }
                 throw;
             }
+        }
 
+        protected override void TraceWriteStart(int size, bool async)
+        {
```
roncain: Is this an exact copy of the code for CoreCLRSocketConnection? Not for this PR, but we will eventually need to rename this file to follow the new naming guidelines and probably make it a partial type to avoid duplicating code.

src/System.Private.ServiceModel/src/Internals/System/Runtime/Diagnostics/EtwDiagnosticTrace.cs
```diff
@@ -173,7 +173,47 @@ public TracePayload GetSerializedPayload(object source, TraceRecord traceRecord,
...
+            }
+
+            if (exception != null)
+            {
+                // We want to keep the ETW trace message to under 32k. So we keep the serialized exception to under 28k bytes.
+                serializedException = ExceptionToTraceString(exception, MaxExceptionStringLength);
+            }
+
+            return new TracePayload(serializedException, eventSource, DiagnosticTraceBase.AppDomainFriendlyName, extendedData, string.Empty);
         }
```
roncain: AppDomainFriendlyName is currently always String.Empty now, right?  Any concerns with that?

roncain:
Will this break in NET Native before the rd.xml change?  If not, I'm fine checking in what's here before moving onto the TODO's. Otherwise, I'd like to ensure it does not break in NET Native first.

mconnew:
Doesn't break .Net Native. When WcfEventSource is initializing, an exception is thrown which EventSource swallows and it then just doesn't do anything. This hasn't changed from the original code.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
mconnew: Not a copy. Getting the remote ip address and port is done differently.

src/System.Private.ServiceModel/src/Internals/System/Runtime/Diagnostics/EtwDiagnosticTrace.cs
```diff
@@ -173,7 +173,47 @@ public TracePayload GetSerializedPayload(object source, TraceRecord traceRecord,
...
+                            writer.Flush();
+                            stringWriter.Flush();
+
+                            extendedData = sb.ToString();
+                        }
+                    }
+                }
+                finally
+                {
+                    StringBuilderPool.Return(sb);
```
iamjasonp: Do we need to StringBuilder.Clear() the StringBuilder before returning to the pool?

roncain: It looks like Return does a Clear before enqueing.

roncain:
LGTM subject to verifying the _socket==null questions raised and ensuring it really does not break in NET Native (I know confidence is high, but...).  Passed OuterLoops for me.

src/System.Private.ServiceModel/src/Internals/System/Runtime/Diagnostics/EtwDiagnosticTrace.cs
mconnew: No concern. It's only useful for being able to tell events apart from multiple domains in a single process. My thinking was any non-empty string will make the event bigger.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CoreClrSocketConnection.cs
mconnew: This is same behavior as on Desktop. We only call Write after a socket is connected so _socket will be non-null. I'll add a Contract.Assert

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CoreClrSocketConnection.cs
mconnew: I don't know why it does that as on desktop it's technically possible that the call made there to fetch the remote endpoint could hit a null socket. To be on the safe side, I'll protect it here.

mconnew:
I haven't made any changes to rd.xml as I have been unable to get event logs in .Net Native to work. I keep getting a MissingMetadataException happening (which gets swallowed by EventSource so it does no harm) with the error `Reflection_InsufficientMetadata_EdbNeeded: System.Diagnostics.Tracing.EventAttribute.`. I will work on that as a follow-up task to prevent delaying these changes.

mconnew:

mconnew:
Single failure in outerloop tests is caused by changes made in build workflow and is unrelated. I'm going to merge.

huangjinshe:
@mconnew Is there any solution for this exception?  I'm using RC1 in my project.  it will always throw `Reflection_InsufficientMetadata_EdbNeeded: System.Diagnostics.Tracing.EventAttribute.` when every time send the first request to the WCF.(only in Release mode)

PR #679, opened 2016-01-11
Create supported-feature matrix

hongdai:
* This is just to clone RC1 release note. Once RC2 test completes, it will
be updated. All related release notes will be updated once we complete RC2 testing.


roncain:
LGTM.  We will obviously be editing this file over the coming days but should get it checked in to allow that.

Issue #587, opened 2015-12-07
TextEncodingTests.SameBinding_SecurityModeNone_Text_EchoString_Roundtrip Failed Sporadically


shmao:
Running Test: TextEncodingTests.SameBinding_SecurityModeNone_Text_EchoString_Roundtrip
  Assert.True:Test case FAILED with errors:     Error: Unexpected exception was caught while doing the basic echo test for variation...
  'Client:: CustomBinding/TcpTransport/NoWindowsStreamSecurity/TextMessageEncoding = None
  Server:: CustomBinding/TcpTransport/NoWindowsStreamSecurity/TextMessageEncoding = None'
  Exception: System.IO.IOException: An existing connection was forcibly closed by the remote host.
  
  An existing connection was forcibly closed by the remote host.
   ---> System.Exception: An existing connection was forcibly closed by the remote host.
  
  An existing connection was forcibly closed by the remote host.
  
     --- End of inner exception stack trace ---
     at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() in f:\dd\ndp\fxcore\CoreRT\src\System.Private.CoreLib\src\System\Runtime\ExceptionServices\ExceptionDispatchInfo.cs:line 60
     at System.IO.StreamOperationAsyncResult.ThrowWithIOExceptionDispatchInfo(Exception e) in D:\dd\ProjectK\src\ndp\fxcore\src\System.Runtime.WindowsRuntime\System\IO\StreamOperationAsyncResult.NETNative.cs:line 13
     at System.IO.StreamOperationAsyncResult.ProcessCompletedOperation() in D:\dd\ProjectK\src\ndp\fxcore\src\System.Runtime.WindowsRuntime\System\IO\StreamOperationAsyncResult.cs:line 231
     at System.IO.WinRtToNetFxStreamAdapter.EndRead(IAsyncResult asyncResult) in D:\dd\ProjectK\src\ndp\fxcore\src\System.Runtime.WindowsRuntime\System\IO\WinRtToNetFxStreamAdapter.cs:line 509
     at System.IO.WinRtToNetFxStreamAdapter.Read(Byte[] buffer, Int32 offset, Int32 count) in D:\dd\ProjectK\src\ndp\fxcore\src\System.Runtime.WindowsRuntime\System\IO\WinRtToNetFxStreamAdapter.cs:line 564
[...]

Issue #521, opened 2015-11-13
Tcp_ClientCredentialTypeTests.SameBinding_SecurityModeTransport_ClientCredentialTypeCertificate_EchoString test fails in NET NAtive


roncain:
Running Test: Tcp_ClientCredentialTypeTests.SameBinding_SecurityModeTransport_ClientCredentialTypeCertificate_EchoString
  Caught Unexpected exception:System.IO.IOException: An existing connection was forcibly closed by the remote host.
  
  An existing connection was forcibly closed by the remote host.
   ---> System.Exception: An existing connection was forcibly closed by the remote host.
  
  An existing connection was forcibly closed by the remote host.
  
     Exception_EndOfInnerExceptionStack
     at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() in f:\dd\ndp\fxcore\CoreRT\src\System.Private.CoreLib\src\System\Runtime\ExceptionServices\ExceptionDispatchInfo.cs:line 60
     at System.IO.StreamOperationAsyncResult.ThrowWithIOExceptionDispatchInfo(Exception e) in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\StreamOperationAsyncResult.NETNative.cs:line 13
     at System.IO.StreamOperationAsyncResult.ProcessCompletedOperation() in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\StreamOperationAsyncResult.cs:line 231
     at System.IO.WinRtToNetFxStreamAdapter.EndRead(IAsyncResult asyncResult) in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\WinRtToNetFxStreamAdapter.cs:line 509
     at System.IO.WinRtToNetFxStreamAdapter.Read(Byte[] buffer, Int32 offset, Int32 count) in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\WinRtToNetFxStreamAdapter.cs:line 564
     at System.IO.BufferedStream.Read(Byte[] array, Int32 offset, Int32 count) in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\BufferedStream.cs:line 560
[...]

Issue #520, opened 2015-11-13
TCP_ServiceCertExpired_Throw_MessageSecurityException test fails in NET Native


roncain:
Running Test: ExpectedExceptionTests.TCP_ServiceCertExpired_Throw_MessageSecurityException
  Caught Unexpected exception:System.IO.IOException: An existing connection was forcibly closed by the remote host. (Excep_FromHResult 0x80072746) ---> System.Exception: An existing connection was forcibly closed by the remote host. (Excep_FromHResult 0x80072746)
     Exception_EndOfInnerExceptionStack
     at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() in f:\dd\ndp\fxcore\CoreRT\src\System.Private.CoreLib\src\System\Runtime\ExceptionServices\ExceptionDispatchInfo.cs:line 60
     at System.IO.StreamOperationAsyncResult.ThrowWithIOExceptionDispatchInfo(Exception e) in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\StreamOperationAsyncResult.NETNative.cs:line 13
     at System.IO.StreamOperationAsyncResult.ProcessCompletedOperation() in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\StreamOperationAsyncResult.cs:line 231
     at System.IO.WinRtToNetFxStreamAdapter.EndRead(IAsyncResult asyncResult) in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\WinRtToNetFxStreamAdapter.cs:line 509
     at System.IO.WinRtToNetFxStreamAdapter.Read(Byte[] buffer, Int32 offset, Int32 count) in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\WinRtToNetFxStreamAdapter.cs:line 564
     at System.IO.BufferedStream.Read(Byte[] array, Int32 offset, Int32 count) in E:\ProjectK\src\NDP\FxCore\src\System.Runtime.WindowsRuntime\System\IO\BufferedStream.cs:line 560
     at System.ServiceModel.Channels.RTSocketConnection.ReadCore$catch$2() in E:\ProjectK\src\NDP\FxCore\WcfOpen\src\System.Private.ServiceModel\src\System\ServiceModel\Channels\RTSocketConnection.cs:line 430
[...]

Issue #624, opened 2015-12-10
The ServiceCertRevoked test fails in ToF for CoreCLR


roncain:
It may be sporadic, but I just saw it fail running ToF against CoreCLR 
 --------------------------------
  Running Test: ExpectedExceptionTests.TCP_ServiceCertRevoked_Throw_MessageSecurityException
  Assert.True:Expected: SecurityNegotiationException, Actual: no exception
  Xunit.AssertTestException: Assert.True: Expected: SecurityNegotiationException, Actual: no exception
     at Xunit.Assert.HandleFail(String assertionName, String message) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 592
     at Xunit.Assert.True(Boolean condition, String message) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 147
     at ExpectedExceptionTests.TCP_ServiceCertRevoked_Throw_MessageSecurityException() in e:\ProjectK\src\NDP\FxCore\WcfOpen\src\System.Private.ServiceModel\tests\Scenarios\Client\ExpectedExceptions\ExpectedExceptionTests.cs:line 532
     at EntryPointMain.b__d() in e:\ProjectK\src\QA\ToF\PK\x86\rel\IL\FX\Conformance\System.Private.ServiceModel\4.0.0.0\WcfOpen\Scenarios\Client\ExpectedExceptions\Client.ExpectedExceptions.Tests\Client.ExpectedExceptions.Tests.main.cs:line 26
     at CoreFXTestLibrary.Internal.Runner.RunTestMethod(TestInfo t) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Internal\Runner.cs:line 170
  ---- Test FAILED ---------------

hongdai:
It passed on my enlistment machine with the latest bit. @roncain , do you have more data such as how frequently you see it fails? So far it only repros on your machine.

roncain:
It fails 100% of the time so far.  Verified again today 1/4/2016.  Could this be a timing issue with the CRL test logic?

roncain:
It also fails with a different exception running in ToF against NET Native:
  Running Test: ExpectedExceptionTests.TCP_ServiceCertRevoked_Throw_MessageSecurityException
[...]

PR #687, opened 2016-01-12
Fix bug where async write was reporting completion before the Flush completed

mconnew:
Here is this sequence of events when the data is supposed to be sent immediately.  
* _socket.WriteAsync is called and returns a task (t1)
* t1.ContinueWith is called on this task where the delegate calls FlushAsync. ContinueWith returns a task t2.
* If t2 is Completed immediately, return AsyncCompletionResult.Completed

The problem is the FlushWriteImmediate method was an async void method so we weren't waiting on the flush completing and t1.ContinueWith would complete immediately which caused us to report the Write had completed. Another Write would then be called which would cause the new data to be appended to the write buffer which hadn't been cleared yet from the previous Flush. FlushAsync would then be called again and the entire buffer including the data from the first Write call would get sent. This had the effect of two consecutive immediate write calls in quick succession would cause the data from the first write to be sent on the socket twice.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
@@ -22,7 +22,7 @@ namespace System.ServiceModel.Channels
     internal class RTSocketConnection : SocketConnection
     {
         // callback static delegates
-        private static Action<Task, object> s_flushWriteImmedaite = new Action<Task, object>(FlushWriteImmediate);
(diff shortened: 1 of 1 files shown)


roncain:
Are there any specific issue numbers we can include and close with this checkin?

mconnew:
I think I have at least one bug assigned to me. I'll find the correct bug number and reference it.

roncain:
LGTM.
If there is a test currently disabled due to this issue, it would be great to reactivate it with this PR.

mconnew:
Fixes #524 
Fixes #521 
Fixes #587 
Fixes #520 
This also will cause the correct exception to be thrown on Net Native as mentioned in issue #624

PR #695, opened 2016-01-13
Enable NegotiateStream tests for non-Unix platforms

iamjasonp:
Includes two changes: 

* Modifies System.Private.ServiceModel.csproj to allow TargetsWindows condition to be set - this is consistent with https://github.com/dotnet/corefx/commit/9e1da13cf6d5ff4b37b1867beb3411a2fe61da6a
* Enables the NegotiateStream tests

### DIFF
--- a/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
+++ b/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
@@ -1,9 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
+  <PropertyGroup>
+    <Configuration Condition="'$(Configuration)'==''">Windows_Debug</Configuration>
+  </PropertyGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


iamjasonp:

roncain:
LGTM

iamjasonp:
I do find it weird that we are specifying a default configuration directly in the csproj like so: 

```
  <PropertyGroup>  
      <Configuration Condition="'$(Configuration)'==''">Windows_Debug</Configuration>  
  </PropertyGroup>  
```

However, it seems like what's happening in dotnet/corefx@9e1da13, so I've adopted the approach here.

@weshaggard, is this intended to be a long-term solution? If not, is there any change in the pipeline in dotnet/corefx that we might have to be aware of so that we update when dotnet/corefx does? 

We needed this configuration to allow the $(TargetsWindows) condition to be set as that's how we condition for the #ifs

weshaggard:
The default configuration is mainly about ensuring that the project can build by itself or for other projects that reference it without specifying it Configuration they get a reasonable default. It is only needed for projects where "Debug" isn't a valid configuration.

In this csproj it actually lists Debug/Release as validate configurations but in the .builds file they are not so those configuration should also be removed from the project otherwise you will have issues if anyone builds those configs in VS.

iamjasonp:
Thanks for the explanation and feedback @weshaggard

iamjasonp:

mconnew:
The fix to remove the > has just been checked in so you don't need it.

mconnew:
LGTM

iamjasonp:
Thanks for the heads up, @mconnew. Rebased to master again and now doing a final build/test before merging.

iamjasonp:

PR #694, opened 2016-01-13
Supporting package building in WCF.

StephenBonikowsky:
* Add .pkgproj files and all needed supporting files to enable building our own packages.

PR #701, opened 2016-01-14
Add comment to the dir.props file diffs from corefx

iamjasonp:
* Add comment to the dir.props file where it differs from corefx
* Remove the workaround we had used in System.Private.ServiceModel.csproj when the fix to $(OSGroup) wasn't in place. 

See https://github.com/dotnet/wcf/pull/694#commitcomment-15440494 for discussion:

> @mconnew and I had a chat about this.

> I'm reasonably convinced now that we can leave this here. and revert the default configuration change in System.Private.ServiceModel.csproj. This way all of our projects will automatically get the benefit of the $(TargetsWindows) being set correctly, rather than solving this on a per-project basis.

> We should, however, comment clearly that we've diverged from CoreFx here so that we're not surprised down the line.

cc @mconnew

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -222,6 +222,11 @@
     <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('Linux'))">Linux</OSGroup>
     <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('OSX'))">OSX</OSGroup>
     <OSGroup Condition="'$(OSGroup)'=='' and $(Configuration.StartsWith('FreeBSD'))">FreeBSD</OSGroup>
+
+    <!-- 
+    This logic differs from corefx's dir.props. Most things we build are not "AnyOS", so unless explicilty specified, we want to 
+    inherit the OS environment that we are building under 
+    --> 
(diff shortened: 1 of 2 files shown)


roncain:
LGTM but would you mind waiting for the dev workflow PR to be merged? dir.props is under heavy 3-way merge change right now to combine new dev workflow and package projects.

iamjasonp:
Just to double check - 

@weshaggard, are there any concerns with us having modified dir.props in the wcf repo such that

`<OSGroup Condition="'$(OSGroup)'==''">AnyOS</OSGroup>`
becomes
`<OSGroup Condition="'$(OSGroup)'==''">$(OSEnvironment)</OSGroup>`

?

Thanks

weshaggard:
OSGroup should default to AnyOS otherwise you will get inconsistent results and duplicated builds when someone builds all OSGroups on one machine. It isn't a good idea to default it to OSEnviroment. Why do you think that is necessary?

iamjasonp:
After some further research and team discussions, we have concluded that this particular PR is not necessary. #707 should take care of the problems we were trying to fix and ties up some loose ends left from #694, #695.

PR #703, opened 2016-01-14
Fix type-o in the NetTcp.builds file

roncain:


roncain:
@StephenBonikowsky -- I haven't done an exhaustive search but found this trying to experiment adding a src.builds file.

roncain:
#704 will fix this from the TFS side.  Will cancel this PR when that happens.

roncain:
#704 merged so this became redundant.  Closing it.

StephenBonikowsky:
@weshaggard caught and fixed this in TFS.
This went in with the PR that we merged when Jenkins was down otherwise it would have been caught :cry:

PR #707, opened 2016-01-14
Enable the packaging targets to build

roncain:
This PR copies files and.or edits from CoreFx to enable the packaging targets to build.  It does not yet build successfully, pending the merge of some open PR's to update some pkgproj files. I will rebase and update when those PR's are merged.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -12,7 +12,7 @@
       <!-- For the root traversal default filter the OSGroup to the OSEnvironment which is the OS we are running on -->
       <FilterToOSGroup Condition="'$(BuildAllOSGroups)' != 'true'">$(OSEnvironment)</FilterToOSGroup>
     </Project>
-    <Project Include="src\packages.builds" Condition="'$(BuildPackages)' == 'true'">
+    <Project Include="src\packages.builds">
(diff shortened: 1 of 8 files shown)


roncain:
@ericstj and @weshaggard -- your feedback is welcome on this.  I'm trying to bring WCF up to parity with CoreFX w.r.t. these files.  Except for errors related to the contents of the pkgproj, it appears to build okay.  I will not merge till the pkgproj files are updated or will add them here if preferred.

StephenBonikowsky:
This will also fix the other test issue we have been discussing.
Once this is rebased with Eric's fixes and it builds it should be good to go.

LGTM

weshaggard:
LGTM. 

As an FYI not what we have changed the dev workflow the plan is to move a lot of these common props and targets files into BuildTools itself so it will help eliminate all this duplication and make things more consistent. Given that we were restoring BuildTools as part of the build it wasn't possible for a lot of these but now that we are restoring it outside of msbuild we can fix a lot of this duplication.

roncain:
Builds clean in GitHub and builds nupkgs for the WCF libraries.  I will verify it builds inside TFS before merging.

Issue #462, opened 2015-10-26
Update System.Runtime.Serialization.Xml dependency to 4.1.0


roncain:
This depends on issue https://github.com/dotnet/corefx/issues/4132 being fixed first.
Currently, WCF depends on 4.0.10.0 because the dependency on 4.1.0 broke the TFS lab builds (due to API's not existing in desktop and TestNet).

zhenlan:
We will wait serialization to finalized its support story for desktop and testnet.

zhenlan:
@roncain Please go ahead and make the change. TEST.NET should not be a problem for this anymore.

roncain:
Fixed by #715 which included this version increase.

PR #715, opened 2016-01-15
Use minimum dependencies

mconnew:


mconnew:
I'm still testing this, but the new build workflow actually compiles against netcore50 dependencies and caught a missing using statement which I then put back in. I'll be attempting to build and run tests in TFS with this code to verify nothing seems broken there. This also reduces the number of packages we need to download overall (although we now download non-windows dependencies so the reduction isn't as great as it would have been a week ago).

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/DnsCache.cs
```diff
@@ -1,17 +1,13 @@
...
-using System.Net.NetworkInformation;
 using System.Net.Sockets;
 using System.Runtime;
-using System.ServiceModel;
-using System.ServiceModel.Dispatcher;
-using System.Text;
 using System.Threading.Tasks;
 #if FEATURE_NETNATIVE
+using System.Collections.Generic;
+
```
roncain: I can't tell from the PR -- is this in alphabetical order?  Or are we keeping NET Native usings all together and apart?

src/System.Private.ServiceModel/src/windows/project.json
```diff
@@ -1,49 +1,49 @@
...
+    "System.Reflection.Extensions": "4.0.0",
+    "System.Reflection.Primitives": "4.0.0",
+    "System.Reflection.TypeExtensions": "4.0.0",
+    "System.Resources.ResourceManager": "4.0.0",
+    "System.Runtime": "4.0.20",
+    "System.Runtime.Extensions": "4.0.10",
+    "System.Runtime.Handles": "4.0.0",
+    "System.Runtime.InteropServices": "4.0.0",
     "System.Runtime.Serialization.Xml": "4.0.10",
-    "System.Security.Claims": "4.0.1-rc2-23712",
```
roncain: I commented in the wrong place -- would you mind making S.R.S.Xml be 4.1.0 to fix #462

roncain:
Generally looks good.  Before the L.G.T.M. questions are 1) would you mind updating S.R.S.Xml dependency? (2) Do we know why the CI tests failed in Release? (3) Is there a big perf hit without the buffer pool?

roncain:
Also, I'd like to buddy build this in TFS when it's ready.  Considering a pending 3-day weekend and that the official builds have been broken for 2 days, a new breakage for this one checkin should be avoided at all costs.

mconnew:
1) I'm doing that in my local TFS test run now. Built fine, running ToF tests at the moment.
2) The release build failed with this error:
`Could not load file or assembly 'System.Private.DataContractSerialization, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference.`
Hopefully this will be fixed by the upgrade to S.R.S.Xml.
3) The buffer pool code was never used in this code base. On Desktop, WCF needs to provide the buffer to ClientWebSocket, but doesn't in CoreFx. The buffer pool code used Marshel.SizeOf which kept one of our dependencies at a higher level.

Additionally, I've got another small change. I need to raise System.Text.Encoding up to 4.0.10 as some of our dependencies do that and that was a version conflict warning.

mconnew:

roncain:
LGTM subject to no significant OuterLoop regressions introduced by the changes. We chatted offline and agreed to merge to make the release branch if possible.

mconnew:
Outerloop tests are failing as they look to be using the Linux binaries for running on windows. This is unrelated to this change and is something to be fixed as a result of build changes made in the last few days.

mconnew:
All TFS ToF tests which were expected to pass have completed.

PR #725, opened 2016-01-20
Message encoder content type fix

mconnew:
Because we no longer have the ContentType class, we had previously added a PNSE exception in MessageEncoder.IsContentTypeSupported. This change restores the desktop framework implementation replacing ContentType with MediaTypeHeaderValue for parsing.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Encoding/Encoders/TextEncodingTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Encoding/Encoders/TextEncodingTests.cs
@@ -48,4 +48,36 @@ public static class TextEncodingTests
 
         Assert.True(errorBuilder.Length == 0, "Test case FAILED with errors: " + errorBuilder.ToString());
     }
+
+    [Fact]
+    [OuterLoop]
+    public static void TextMessageEncoder_WrongContentTypeResponse_Throws_ProtocolException()
+    {
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
+        string testContentType = "text/blah";
+        Binding binding = null;
+
+        try
+        {
+            // *** SETUP *** \\
+            binding = new BasicHttpBinding(BasicHttpSecurityMode.None);
+            factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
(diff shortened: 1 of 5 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Encoding/Encoders/TextEncodingTests.cs
```diff
@@ -48,4 +48,36 @@ public static void SameBinding_SecurityModeNone_Text_EchoString_Roundtrip()
 
         Assert.True(errorBuilder.Length == 0, "Test case FAILED with errors: " + errorBuilder.ToString());
     }
+
+    [Fact]
+    [OuterLoop]
+    public static void TextMessageEncoder_WrongContentTypeResponse_Throws_ProtocolException()
+    {
```
roncain: It would be good to add a positive test to demonstrate what happens with a known good content type. And possibly another test where the content type is parsable but not expected.   In fact, what does happen if the WCF service sets a content type that does not agree with what the client expected (e.g. saying it returned json when it returned xml)?

roncain:
Question only -- is there an associated issue for this that gets fixed by this PR?

src/System.Private.ServiceModel/tests/Scenarios/Encoding/Encoders/TextEncodingTests.cs
mconnew: All our other tests are known good content types. Returning the wrong content type proves we are examining the content type and throw an exception if it's wrong. Not throwing an exception on all other tests shows we accept a good value.

mconnew:
No associated issue. This was reported via gitter and I threw the fix together quickly.

src/System.Private.ServiceModel/tests/Scenarios/Encoding/Encoders/TextEncodingTests.cs
roncain: It would still be good to test what happens when you set the content type to some legal parsable value *other* than the normal one for the response.  Does it flow back to the client as is? Cause a different kind of exception on the client?  Cause an exception on the server?

roncain:
LGTM. The additional test I suggested is not required for this PR.

I recommend we open a issue from the gitter discussion to be able to defend this mini-feature addition in our ship room.

mconnew:
Fixes #733

PR #735, opened 2016-01-22
Refactored stress tests and adopted them to run perf

KKhurin:
Parity between sync and async coverage
Merged with Hong's duplex changes for NetHttp binding
Added Duplex NetHttp streaming scenario
Added the server source code (will provide the instructions on how to setup the IIS server)


src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/project.json
```diff
@@ -1,37 +1,30 @@
...
-    "System.ServiceModel.Security": "4.0.1-rc2-23712",
-    "System.Console": "4.0.0-rc2-23712",
-    "System.Collections": "4.0.11-rc2-23712",
-    "System.Linq": "4.0.1-rc2-23712",
-    "System.Threading": "4.0.11-rc2-23712",
-    "System.Diagnostics.Debug": "4.0.11-rc2-23712",
-    "System.Collections.Specialized": "4.0.1-rc2-23712",
-    "System.Private.ServiceModel": "4.0.1-rc2-23712",
-    "System.Diagnostics.Tools": "4.0.1-rc2-23712",
-    "System.ServiceModel.Duplex": "4.0.1-rc2-23712"
```
iamjasonp: This will cause a build break as build.proj will verify dependency versions for 23712. 

Your two options are: 

1. Modify the `build.proj` file to verify for rc2-23719, and at the same time, touch all `project.json`s 
2. Revert the versions to rc2-23712 in this PR, then touch `build.proj` and all `project.json` files in a subsequent PR. 

I recommend option 2.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjectsServer/DuplexService.svc.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Runtime.Serialization;
+using System.ServiceModel;
+using System.ServiceModel.Web;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace WcfService1
+{
+    // NOTE: You can use the "Rename" command on the "Refactor" menu to change the class name "Service1" in code, svc and config file together.
+    // NOTE: In order to launch WCF Test Client for testing this service, please select Service1.svc or Service1.svc.cs at the Solution Explorer and start debugging.
```
iamjasonp: NIT: Any need for this? Seems like project boilerplate code

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
```diff
@@ -68,11 +64,15 @@ public T CallFuncAndRecordStats<T>(Func<T> func)
             catch (Exception e)
```
iamjasonp: NIT: I don't see `e` used 
(Question: are you building with Warnings as Errors?)

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/project.json
roncain: Recommend option 2 also.  It is vital each PR be atomic, so any product wide touch to our build.proj and project.json's must contain only those changes and nothing else.  For example, we need the option to update the release branch with only the project.json touches.  I also think the version will be 23721 or more by the time the next PR happens.

StephenBonikowsky:
CI tests are failing with a build error...

d:\j\workspace\dotnet_wcf\windows_debug_prtest\build.proj(58,5): error : Dependency validation error: for System.ServiceModel.NetTcp 4.1.0-rc2-23719 in d:\j\workspace\dotnet_wcf\windows_debug_prtest\src\System.Private.ServiceModel\tests\Stress\SharedPoolsOfWCFObjects\project.json package prerelease is 'rc2-23719', but expected 'rc2-23712' for packages matching '^((System\..*)|(Microsoft\.CSharp)|(Microsoft\.NETCore.*)|(Microsoft\.Win32\..*)|(Microsoft\.VisualBasic))(?<!TestData)$'

KKhurin:
Fixed the build errors by syncing the build number and getting rid of the csproj files from the server side.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
```diff
@@ -10,10 +13,119 @@ namespace SharedPoolsOfWCFObjects
...
+            var test = new Program();
+            test.ProcessRunOptions(args);
+            if (test._paramProgram2Run == Program2Run.Stress)
+            {
+                test.RunStress();
+            }
+            else
+            {
+                test.RunPerf();
+            }
```
roncain: Just a question -- it is now possible to run these either as Stress or as Perf tests, but the files are all located under a folder called tests\Stress.  I wonder if we can come up with a better name for the folder.  Just

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
```diff
@@ -10,10 +13,119 @@ namespace SharedPoolsOfWCFObjects
...
+        }
+
+        private bool ProcessRunOptions(string[] args)
+        {
+            //var paramHostName = Environment.GetEnvironmentVariable("HostName");
```
roncain: Nit: it would be nice to eliminate these commented out sections.  Or if you occasionally use them for private testing, put them under a conditional compile check.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
```diff
@@ -21,57 +133,101 @@ public void Main(string[] args)
...
-                            System.Diagnostics.Debugger.Break();
+                            TestUtils.ReportFailure(e.ToString());
+                            throw;
                         }
 
-                        if (i % 1000 == 0)
+                        if (i % 100 == 0)
                         {
-                            Console.WriteLine(ttt + " " + i);
+                            Console.WriteLine(DateTime.Now.ToString() + " " + ttt + " " + i);
```
roncain: Just a question: if these tests are run as Perf tests, does this use of Console interfere with the metrics?

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
```diff
@@ -79,136 +235,381 @@ private static void DoTheRun(int iterations)
...
+                        }
+                        catch (Exception e)
+                        {
+                            TestUtils.ReportFailure(e.ToString());
+                            throw;
+                        }
+#if DEBUG
+                        if (i % 100 == 0)
+                        {
+                            Console.WriteLine(DateTime.Now.ToString() + " " + ttt + " " + i);
```
roncain: Just a question: I notice Console is used other places not inside an 'if DEBUG'.  Should they all be inside 'if DEBUG' checks?

roncain:
@KKhurin It looks like you will need to rebase again before we can do buddy-builds or outerloop tests, but otherwise it looks good to me.  I'll mark it with the real LGTM when we can verify all Windows CI tests pass.  We don't expect failure, but it makes sure there is not some special code somewhere that tries to build and run these tests accidentally.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
KKhurin: Good question. These tests were originally designed for stress (concurrently use shared state, trigger failures, catch unexpected errors, etc.) and only later adopted to collect performance metrics. The dedicated performance tests should (in theory) look totally different (much simpler and targeted to get certain number, like CPU cycles per request in a tightly controlled environment). These current tests are more like "throughput bottleneck detection" and "overall system elasticity" type of tests. They are not very well suited for precise profiling as there is still too much randomness in the way we make requests. So a descriptive name would be "Stress, Throughput, Startup And Bottlenecks". Maybe add a readme.md to describe the exact purpose of these tests?

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
KKhurin: Agreed.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
KKhurin: This method isn't used when running perf (and no threads that actually make requests in perf write to Console). Perhaps I should rename the method to avoid confusion.

However your suggestion to use a different logging mechanism still makes perfect sense.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/Program.cs
KKhurin: #756

iamjasonp:
@KKhurin when are these going to make it in? This PR has been sitting around for a while, and the longer it sits the less likely the rebase will be clean :)

jasonwilliams200OK:
@KKhurin, could you please rebase it?

```sh
git clone https://github.com/kkhurin/wcf
cd wcf
git checkout PerfAndStressTests
git remote add dotnet https://github.com/dotnet/wcf
git pull --rebase dotnet master
# fix the conflicts (if any),
# if there were conflicts which you fixed manually then:
# git add .;git rebase --continue

git push -f
# and done! :)
```

Issue #731, opened 2016-01-21
Some OuterLoop tests fail when built with explicit configuration


roncain:
If you build and run OuterLoop tests by explicitly setting $(Configuration), 3 tests fail.  More importantly, it means the product builds for the wrong target framework if the configuration is specified on the command line.

Command to run the tests is:
```
build.cmd /p:Configuration=Debug /p:WithCategories=OuterLoop
```
The 3 tests that fail (apparently thinking they are for the wrong target framework) are:

Tcp_ClientCredentialTypeTests.SameBinding_DefaultSettings_EchoString [FAIL] (TaskId:6796)
Tcp_ClientCredentialTypeTests.SameBinding_SecurityModeTransport_EchoString [FAIL] (TaskId:6796)
HttpsTests.ServerCertificateValidation_EchoString [FAIL] (TaskId:6796)

StephenBonikowsky:
Fixed by PR #744

StephenBonikowsky:
Clarification
These tests will still fail if "Debug" or "Retail" are passed-in to the build script as the Configuration. This is due to the convoluted logic in the build process.
While we can find the exact place this happens it is debatable whether this is worth the time to do as WCF Core is not OS neutral, it must be built either against WIndows or Linus (Unix).

So this bug is not fixed, just avoided by changing Jenkins to not pass-in just "Debug" or "Retail".

PR #739, opened 2016-01-22
Fixing command line used by Jenkins to build and run CI tests.

StephenBonikowsky:
* The value used for the parameter "/p:Configuration" will now be the OS name plus either Debug or Release.
* Like this: Windows_NT_Debug
* When the configuration is just "Debug" or "Release" WCF OuterLoop test projects are running against the wrong build of the product, there is broken logic in the build process around this which has not yet been figured out, this is a short term solution.
* Fixes Issue #731

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -130,7 +130,7 @@ def project = 'dotnet/wcf'
             label('windows')
             steps {
                 // Use inline replacement
-                batchFile("build.cmd /p:Configuration=${configuration} /p:OSGroup=${os}")
+                batchFile("build.cmd /p:Configuration=${os}_${configuration} /p:OSGroup=${os}")
                 // Pack up the results for max efficiency
                 batchFile("C:\\Packer\\Packer.exe .\\bin\\build.pack .\\bin")
             }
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
Since this fix is for the Jenkins CI process there is no way to test this until after it is merged.
So merging, and if it doesn't work I will revert it.

netci.groovy
```diff
@@ -274,7 +274,7 @@ Utilities.addGithubPRTrigger(prCCJob, 'Code Coverage Windows Debug', '@dotnet-bo
     def newRollingJob = job(Utilities.getFullJobName(project, jobName, false)) {
         label('windows-elevated')
         steps {
-            batchFile("build.cmd /p:Configuration=${configuration} /p:WithCategories=OuterLoop")
+            batchFile("build.cmd /p:Configuration=${os}_${configuration} /p:WithCategories=OuterLoop")
```
roncain: Do we need OSGroup in these like we do in the ones above?

StephenBonikowsky: The issue we are fixing seems to be isolated to the Configuration parameter. OSGroup as it is being used in any of the calls to Build.cmd is working as expected.

PR #741, opened 2016-01-22
Fixing mistake in PR #739

StephenBonikowsky:
* Line 277 the variable 'os' is not known, hardcoding it since calls to build.cmd and running OuterLoops can only be on Windows.


StephenBonikowsky:
@dotnet-bot test ci please

PR #742, opened 2016-01-22
Document how to collect and analyze ETW traces

### LINKED ISSUE #562: Document how to collect and analyze ETW traces
Task is to create a new .md file under Documentation that describes how to collect and analyze ETW traces.  If there are differences between doing this with CoreCLR or NET Native, those should be called out.

This task should also consist of an investigation how EventSource behaves on Linux with the intent to describe how to configure it there too.

hongdai:
* This is to document how customers should collect and analyze ETW traces

Fixes #562

### DIFF
--- a/Documentation/HowToUseETW.md
+++ b/Documentation/HowToUseETW.md
@@ -1,0 +1,66 @@
+# Collecting and Analyzing ETW Traces 
+
+## Supported
+ETW logs are available on Windows platform only currently.
+
+## Collect ETW logs for WCF Server
+* If you use your own WCF Server, it is recommended to follow https://msdn.microsoft.com/en-us/library/ms733025(v=vs.110).aspx.
+* If you run the scenario tests, a locally-hosted WCF service will be started automatically, and it is recommended
+ to follow "How to Collect ETW logs uses logman" below.
+
+## How to Collect ETW logs use logman
+
+For logman parameters usage, please see https://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/nt_command_logman.mspx?mfr=true 
+
+Logman collects ETW logs from all WCF applications on the machine. 
+* If you run both WCF Client and Server on the same machine, you will get ETW logs for both processes in the same file. It 
+  can be differentiated from process IDs.
+
+* If you run WCF Client and Server on different machines. You need to collect the logs using the same command on both machines.
+
+### Start the logging
+
+ * Option 1: Generate a single file that does circular logging with Verbose level, For example
+ 
+ ```
+logman start WCFETWTracing -p "Microsoft-Windows-Application Server-Applications" 0xFFFFFFFF  0x5 -bs 64 -nb 120 320 -ets -ct perf -f bincirc -max 500 -o 
+
+c:\temp\WCFEtwTrace.etl
+```
+ * Option 2: Generate a single file that does circular logging with Informational level, For example
+ 
+ ```
+logman start WCFETWTracing -p "Microsoft-Windows-Application Server-Applications" 0xFFFFFFFF  0x4 -bs 64 -nb 120 320 -ets -ct perf -f bincirc -max 500 -o 
+
+c:\temp\WCFEtwTrace.etl
+```
+ * Option 3: Generate multiple files that rolls logging every 10 minutes with Informational level, For example
+ 
+```
+logman create trace WCFETWTracing -p  "Microsoft-Windows-Application Server-Applications" 0xFFFFFFFF 0x4 -o C:\temp\WCFETWTrace -f bin -cnf 00:10:00 -v 
+
+mmddhhmm
+
+logman start WCFETWTracing 
+```
+ * 	Option 4: Generate a single file that does circular logging with Warning level, For example
+ 
+ ```
+ logman start WCFETWTracing -p "Microsoft-Windows-Application Server-Applications" 0xFFFFFFFF 0x3 -bs 64 -nb 120 320 -ets -ct perf -f bincirc -max 500 -o 
+
+c:\temp\WCFEtwTrace.etl
+```
+
+### Stopping the logging when done
+	logman stop WCFETWTracing -ets
+
+## Tools for viewing/analyzing ETW events
+ * eventvwr
+   1. Launch eventvwr.
+   2. Window Logs -> Saved Logs, then open the event log
+ * perfview
+   1. Download perfview from http://www.microsoft.com/en-us/download/details.aspx?id=28567 
+   2. Open perfview.exe
+       * Open the event log file by typing the path into the file path window.
+       * Click on Events from the list.
+ * svcperf, see instructions on http://svcperf.codeplex.com/


StephenBonikowsky:

roncain:
LGTM subject to @mconnew comments

Documentation/HowToUseETW.md
```diff
@@ -0,0 +1,46 @@
+# Supported
+ETW logs are available on CoreCLR only currently.
```
mconnew: These instructions are windows only. I don't the mechanism to collect logs on Unix.

mconnew:
LGTM subject to making it clear this is currently windows only.

Documentation/HowToUseETW.md
```diff
@@ -0,0 +1,46 @@
+# Supported
```
iamjasonp: That's kind of a weird title. Consider making the first line title something like "Collecting and Analyzing ETW Traces" with "Support" being a subheading

Documentation/HowToUseETW.md
```diff
@@ -0,0 +1,46 @@
...
+
+# How to Collect ETW logs use logman
+
+
+## Start the logging
+
+ * Option 1: Generate a single file that does circular logging with Verbose level, For example
+ 
+ ```
+logman start WCFETWTracing -p "Microsoft-Windows-Application Server-Applications" 0xFFFFFFFF  0x5 -bs 64 -nb 120 320 -ets -ct perf -f bincirc -max 500 -o c:\temp\WCFEtwTrace.etl
```
iamjasonp: Do you think it's necessary here to talk about what each parameter here does, or perhaps add a link to MSDN that states what they do?

iamjasonp:
LGTM, subject to @mconnew's comments (consider my comments as nits)

Documentation/HowToUseETW.md
roncain: I recommend you also add a statement saying on which machine you execute this command.  For example, when we run the Bridge on a separate machine, is this the command you use on the client machine or the server machine?  When the client and server are on different machines, would you use this same command on both machines to collect ETW event for both client and server independently? When client and server are on the same machine, would this make the ETW events be a combination of client and server events?

Also, please say when this command can be entered.  For example, must you do it before you start the Bridge?  Can you do it after the Bridge is already running?  If you stop and restart the Bridge, do you need to enter it again?

Documentation/HowToUseETW.md
```diff
@@ -0,0 +1,46 @@
...
+ * eventvwr
+   1. Launch eventvwr.
+   2. Window Logs -> Saved Logs, then open the event log
+ * perfview
+   1. Download perfview from http://www.microsoft.com/en-us/download/details.aspx?id=28567 
+   2. Open the perfview.exe
+       * Open the event log file by typing the path into the file path window.
+       * Click on Events from the list.
+ * svcperf, see instruction on http://svcperf.codeplex.com/
+
```
roncain: Nit :"see instructions..." (plural)

Documentation/HowToUseETW.md
hongdai: This documentation is specific for end users, not only for OSS developers. Hence I left bridge and server out of the picture.

That reminds me to open a issue tracking ETW documentation for developers.

Documentation/HowToUseETW.md
hongdai: Will add windows specific.

Documentation/HowToUseETW.md
hongdai: Make sense. Will do.

Documentation/HowToUseETW.md
hongdai: As for the logman parameters, I will add a MSDN documentation link.

Documentation/HowToUseETW.md
hongdai: Will fix it.

Documentation/HowToUseETW.md
roncain: I recommend the document be structured so it is clear how to use ETW as an end-user to debug an app, and how to use ETW as an OSS developer to debug product issues.  In both cases, we need to be clear which commands are executed on the client machine and which are executed on a server machine.  I think a typical debug scenario will want traces from both client and server, for both end users and OSS users.

StephenBonikowsky:

Documentation/HowToUseETW.md
hongdai: For end users, collect/analyze the server ETW traces is the same as any Desktop WCF services.

Only OSS users use bridge, which needs documentation as it does NOT enable typical desktop tracing experience.

The debugging experience for both users are different because of the bridge. I think it's easier for end users to read without mention bridge in the same document.

roncain: This document appears only in OSS, therefore it is only OSS people who will read it.  The right document to reach true end users would be a blog (which is something our team should ultimately consider seriously, along with samples).  It raises a good point -- who is the audience these OSS documentation files speak to?  All the other MD's talk directly to the OSS user (how to build, how to test, what the Bridge is), so I think the purpose of the Documentation folder is to speak to OSS folks.

Therefore I think it is worth at least pointing out any special considerations needed if they are using ETW to look at the behavior of the OuterLoop tests vs. debugging their own application.  A URL to using ETW for WCF on the full framework would also be beneficial.

This doesn't mean a big restructure -- just a few extra paragraphs to help them understand how they can use ETW for the full E2E debugging experience.

hongdai: Thanks for clarifying the OSS documentation's audience. I will make this doc for only OSS users then.

Documentation/HowToUseETW.md
```diff
@@ -0,0 +1,46 @@
...
+## Stopping the logging when done
+	logman stop WCFETWTracing -ets
+
+# Tools for viewing/analyzing ETW events
+ * eventvwr
+   1. Launch eventvwr.
+   2. Window Logs -> Saved Logs, then open the event log
+ * perfview
+   1. Download perfview from http://www.microsoft.com/en-us/download/details.aspx?id=28567 
+   2. Open the perfview.exe
```
iamjasonp: Nit: superfluous "the"

Documentation/HowToUseETW.md
```diff
@@ -0,0 +1,65 @@
+# Collecting and Analyzing ETW Traces 
+
+## Supported
+ETW logs are available on Windows platform only currently.
+
+## Collect ETW logs for WCF Server
+* If you use your own WCF Server, it's recommended to follow https://msdn.microsoft.com/en-us/library/ms733025(v=vs.110).aspx.
+* If you run outerloop tests that use bridge, it's recommended to follow "How to Collect ETW logs uses logman" below.
+
```
roncain: Nit: no need to mention the bridge or outerloop here (we don't yet document those well).  Though you might say something like "If you run the scenario tests, a locally-hosted WCF service will be started automatically, and it is recommended..."

roncain:
LGTM.  There is a nit level comment about mentioning the bridge, but it is optional.

Documentation/HowToUseETW.md
hongdai: Agreed.

hongdai:
I will merge as the fix is nit level.

PR #743, opened 2016-01-22
Fixing rule to ignore /Tools folder

mconnew:
The rule was also matching src/System.Private.ServiceModel/tools/* which meant new files in that folder were ignored

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ syntax: glob
 ### VisualStudio ###
 
 # Tool Runtime Dir
-[Tt]ools/
+/[Tt]ools/
 
 # User-specific files
 *.suo


mconnew:
This fix is going to need to be manually verified. The command `git check-ignore -v filename` is useful for verifying this.

mconnew:
This rule was originally copied from a change made in this [PR](https://github.com/dotnet/corefx/pull/5310) from CoreFx. @joperezr you might want to consider applying this fix to CoreFx.

joperezr:
@mconnew Good catch! I'll add that to corefx as well. Thanks!

roncain:
LGTM

PR #745, opened 2016-01-25
Make Bridge honor max idle timeout set from command line

roncain:
Prior to this change, the Bridge initialized its max idle
timeout to a fixed value and did not adjust for a value being
passed on the commandline at startup.

It was necessary to fix this issue so that a Bridge launched
on a lab machine can be given a maximum idle timeout.  This
permits the Bridge to close gracefully regardless how a test
run completes.

### DIFF
--- a/src/System.Private.ServiceModel/tools/setupfiles/EnsureBridgeRunning.cmd
+++ b/src/System.Private.ServiceModel/tools/setupfiles/EnsureBridgeRunning.cmd
@@ -1,0 +1,28 @@
+echo off
+setlocal
+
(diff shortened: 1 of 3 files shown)


mconnew:
Don't block on the answer to this question, but could we have a situation where a test contacts the bridge to set everything up, and then runs for longer than the timeout without talking to the non-WCF part of the bridge? I'm thinking perf/stress test scenarios.

mconnew:
LGTM

roncain:
Thanks @mconnew -- this change will not reduce the timeout, only honor it.  I expect us to set it to a number smaller than the default (1 day) in typical outerloops but longer than the default in stress/perf scenarios.

PR #746, opened 2016-01-25
Escape *'s in validation regex

dagood:
Same changes as https://github.com/dotnet/corefx/pull/5665.

MSBuild uses `*` as a special character in `Include`, so I [escaped](https://msdn.microsoft.com/en-us/library/bb383819.aspx) the regexes. I also removed the TestData rule because it doesn't seem to be validateable: some packages are 1.0.0-prerelease and others are 1.0.1-prerelease, but TestData packages aren't produced in lockstep versions so this is normal.

cc @roncain @weshaggard

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -80,11 +80,8 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


weshaggard:
LGTM

roncain:
LGTM

PR #751, opened 2016-01-25
Fixed a bug with MessageHeader.CreateHeader().

### LINKED ISSUE #740: Add Tests Covering Using MessageHeader.CreateHeader with XmlSerializerFormatAttribute

shmao:
When XmlSerializerFormatAttribute is marked on a OperationContract, .Net Native tool assums that the operation wants to serialize objects using XmlSerializer and so pre-generates XmlSerializer for all relevant types. However, MessageHeader.CreateHeader() internally used DataContractSerializer to serialize the header object, even if the operation is marked with XmlSerializerFormatAttribute.

The fix is adding directives in the rd.xml to tell .Net Native tool to generate DataContractSerializer for the 'value' parameter of the MessageHeader.CreateHeader() method.

I also added tests covering the case of using MessageHeader.CreateHeader together with XmlSerializerFormatAttribute.

Fix #740

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
@@ -14,6 +14,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Xml;
+using System.Xml.Serialization;
 
 namespace TestTypes
 {
@@ -1229,4 +1230,24 @@ public class NonInstantiatedType
 public class UniqueType
 {
     public string stringValue;
+}
+
+// This type should be exclusively used by test OperationContextScope_HttpRequestCustomMessageHeader_RoundTrip_Verify.
+[XmlType(Namespace = "urn:TestWebServices/MyWebService/")]
+public class MesssageHeaderCreateHeaderWithXmlSerializerTestType
+{
+    private string message;
+
+    [XmlElement(Order = 0)]
+    public string Message
+    {
+        get
+        {
+            return message;
+        }
+        set
+        {
+            message = value;
+        }
+    }
 }
--- a/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WcfService.cs
+++ b/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WcfService.cs
@@ -251,6 +251,18 @@ namespace WcfService
             return infos;
         }
 
+        public string GetIncomingMessageHeadersMessage(string customHeaderName, string customHeaderNS)
+        {
+            MessageHeaders headers = OperationContext.Current.IncomingMessageHeaders;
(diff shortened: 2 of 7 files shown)


shmao:

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/WcfService.cs
```diff
@@ -251,6 +251,24 @@ public string GetRequestCustomHeader(string customHeaderName, string customHeade
...
+                MessageHeaders headers = OperationContext.Current.IncomingMessageHeaders;
+                TestMessageHeader header = headers.GetHeader<TestMessageHeader>(customHeaderName, customHeaderNS);
+                if (header != null)
+                {
+                    return header.Message;
+                }
+            }
+            catch
+            {
+
```
roncain: I recommend we don't catch and ignore all exceptions in this operation.  Better to let any exception propagate back to the client.

shmao: I see. I will make a change. Thanks.

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
```diff
@@ -1230,3 +1231,18 @@ public class UniqueType
 {
     public string stringValue;
 }
+    
+[XmlType(Namespace="urn:TestWebServices/MyWebService/")]
+public class TestMessageHeader {        
+    private string message;
```
roncain: Nit: we should probably choose a name and include a comment that makes it very clear that this type must not be used by any other service or operation or [ServiceKnownType].  Otherwise someone might reuse this type in another test and accidentally tell SG to make a DataContract serializer for it, which would defeat what this test is currently testing.

@StephenBonikowsky might have recommendations because he has other types like this that say "no one else can use me".

roncain:
LGTM, subject to renaming the test type and not ignoring exceptions.

src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestTypes.cs
```diff
@@ -1230,3 +1231,18 @@ public class UniqueType
...
+    [XmlElement(Order=0)]
+    public string Message {
+        get {
+            return message;
+        }
+        set {
+            message = value;
+        }
+    }
+}
```
iamjasonp: NIT: Formatting for new code

iamjasonp:
LGTM, subject to Ron's comments. 

Small nit: Would prefer if the commit title incorporated "Add directive to rd.xml" as that's what this commit really ends up doing - even though it manifests itself as a bug in MessageHeader.CreateHeader().

shmao:

roncain:
LGTM.  Thanks @shmao

shmao:
Thanks for reviewing the PR, @roncain and @iamjasonp . I'm merging it.

PR #753, opened 2016-01-26
Remove extra line ending in BuilToolsVersion.txt

iamjasonp:
This was causing a build break in Linux, and a directory was being created as "1.0.25-prerelease-00154^M" causing all sorts of weirdness


roncain:
LGTM

iamjasonp:
The best part about this issue was that when I tried to cd into ~/git/wcf/Tools/<toolsDir>, bash would read the ^M (CR) and adjust my cursor accordingly...

Thanks @roncain, will merge after CI finishes

PR #755, opened 2016-01-27
Change X509 to not use handles in *nix

iamjasonp:
Please wait to review; there are still some issues that still aren't resolved yet

--- 

In Linux, we use OpenSSL as a backing for our certificate-based operations.
However, when retrieving a handle to the certificate via X509Certificate2.Handle,
OpenSSL does not pass along the PrivateKey to us

Pass along the certificate without trying to re-create the certificate in Linux
so we can obtain the private key

### DIFF
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
@@ -3,6 +3,8 @@
 
 using System.Collections.Generic;
 using System.IdentityModel.Policy;
+using System.Runtime;
+using System.Runtime.InteropServices;
 using System.Security.Claims;
 using System.Security.Cryptography.X509Certificates;
 using System.Security.Principal;
@@ -28,6 +30,11 @@ namespace System.IdentityModel.Claims
 
         internal X509CertificateClaimSet(X509Certificate2 certificate, bool clone)
         {
+            Fx.Assert(
+                !clone || RuntimeInformation.IsOSPlatform(OSPlatform.Windows),
+                "Certificates MUST NOT be cloned on non-Windows platforms"
+            );
+
             if (certificate == null)
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("certificate");
             _certificate = clone ? new X509Certificate2(certificate.Handle) : certificate;
@@ -509,6 +516,11 @@ namespace System.IdentityModel.Claims
         internal X509Identity(X509Certificate2 certificate, bool clone, bool disposable)
             : base(X509, X509)
         {
+            Fx.Assert(
+                !clone || RuntimeInformation.IsOSPlatform(OSPlatform.Windows),
+                "Certificates MUST NOT be cloned on non-Windows platforms"
+            );
+
             _certificate = clone ? new X509Certificate2(certificate.Handle) : certificate;
             _disposable = clone || disposable;
         }
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509SecurityTokenProvider.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509SecurityTokenProvider.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.IdentityModel.Tokens;
+using System.Runtime;
+using System.Runtime.InteropServices;
 using System.Security.Cryptography.X509Certificates;
 using System.ServiceModel;
 using System.Threading;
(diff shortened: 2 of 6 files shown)


src/System.Private.ServiceModel/src/System/IdentityModel/Claims/X509CertificateClaimSet.cs
```diff
@@ -509,8 +514,13 @@ public X509Identity(X500DistinguishedName x500DistinguishedName)
         internal X509Identity(X509Certificate2 certificate, bool clone, bool disposable)
             : base(X509, X509)
         {
+#if TARGETS_WINDOWS
             _certificate = clone ? new X509Certificate2(certificate.Handle) : certificate;
             _disposable = clone || disposable;
+#else
```
mconnew: This overload is designed to allow a caller to specify whether to clone the certificate or not. I would prefer to modify the caller to use this overload and pass in false/false for clone/disposable.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509SecurityTokenProvider.cs
```diff
@@ -20,7 +20,11 @@ public X509SecurityTokenProvider(X509Certificate2 certificate)
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("certificate");
             }
 
+#if TARGETS_WINDOWS
```
mconnew: I would prefer to have an overload on X509SercurityTokenProvider constructor to specify whether to clone or not. This can then be passed along in the GetTokenCoreAsync call where a new X509SecurityToken is constructed as that already has an overload for whether to clone or not. That way we can avoid as many `#if` blocks as possible. We already have most of the functionality needed throughout this code to achieve this with minimal changes and minimal platform specific code.

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/X509SecurityToken.cs
```diff
@@ -52,9 +52,15 @@ internal X509SecurityToken(X509Certificate2 certificate, string id, bool clone,
 
             _id = id;
 
+#if TARGETS_WINDOWS 
             _certificate = clone ? new X509Certificate2(certificate.Handle) : certificate;
```
mconnew: We already have an overload on the constructor to specify whether to clone. We don't need to make any changes in this class, just in the consumer.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -262,14 +262,21 @@ private void SetupServerCertificate(SecurityToken token)
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(
                     SR.InvalidTokenProvided, _serverTokenProvider.GetType(), typeof(X509SecurityToken))));
             }
+
+#if TARGETS_WINDOWS
             _serverCertificate = new X509Certificate2(x509Token.Certificate.Handle);
```
mconnew: As far as I can tell, the _serverCertificate is never used for clients so this is effectively dead code. We only call SetupServerCertificate when _serverTokenProvider is not null. The only place where _serverTokenProvider is set to anything except null is in the second constructor method on line 46 which is never used. I would rather remove dead code than add more conditional statements to dead code :smile:

mconnew:
I don't think this is the right approach to the issue. I need the same/similar fix for similar reasons in netcore (I need to add a custom X509Extension which basically attaches a UWP Certificate object to the X509Certificate2 so that I don't end up modifying a ridiculous number of classes) so I need the certificate to not be copied. We should specify whether the certificate should be cloned at the top level and have this propagate through using all the existing code to suppress cloning.

mconnew:
I've put together a fix in my own local repo (as I need it for come UWP certificate work) which uses `System.Runtime.InteropServices.RuntimeInformation` to switch at runtime the cloning behavior. The only time we need to use an #if is when the difference in platform requires dependency changes and referencing classes which don't exist on a particular platform.

mconnew:
PR #761 should address this issue in a cleaner way. After testing that change, I suggest closing this PR.

mconnew:
I've checked in the change which achieves the same goal. Closing PR

iamjasonp:
I don't think the PR #761 achieves the goal of this PR completely. I will be reopening this PR shortly.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
iamjasonp: Good catch! :)

roncain:
@iamjasonp -- let me know when you want this PR reviewed or buddy-built.  I'm holding off until I hear.

iamjasonp:
Note: there's investigation going on from the System.Net guys on this, so this one is on hold while we are investigating.

iamjasonp:
@dotnet-bot test Linux Debug Build and Test, test Linux Release Build and Test

iamjasonp:
@dotnet-bot  test Linux Release Build and Test

iamjasonp:
Closing this for now - after updating to the newest packages this looks to have been fully solved without this change.

Issue #702, opened 2016-01-14
SOAP Headers not working in NetCore


shmao:
Jeff reported the issue at a [MSDN blog](http://blogs.msdn.com/b/webdev/archive/2015/12/15/wcf-connected-service-visual-studio-extension-preview-for-asp-net-5-projects.aspx),

> As far as I can tell this does not handle SOAP headers. Can you confirm that they are not yet supported under .NET Core? If they are not, are there any potential workarounds that do not involve changing the server? Thanks!
> 
> After doing a bit of digging, it looks like the part that is missing (or at least the first part) is the implementation of System.ServiceModel.Dispatcher.OperationFormatter.XmlElementMessageHeader.OnWriteHeaderAttributes. Presumably because it used XmlNodeReader, which has not been implemented either. I can't find much of anything on why that isn't implemented, but my first guess would be that it's because not much of anything seems to use it.

jeffhube:
Here is a simple project to reproduce the issue: [SoapHeaderTest.zip](https://github.com/dotnet/wcf/files/91355/SoapHeaderTest.zip)

```
Unhandled Exception: System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.ServiceModel.Dispatcher.OperationFormatter.XmlElementMessageHeader.OnWriteHeaderAttributes(XmlDictionaryWriter writer, MessageVersion messageVersion)
   at System.ServiceModel.Channels.Message.WriteMessagePreamble(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.Message.OnWriteMessage(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.BufferedMessageWriter.WriteMessage(Message message, BufferManager bufferManager, Int32 initialOffset, Int32 maxSizeQuota)
   at System.ServiceModel.Channels.TextMessageEncoderFactory.TextMessageEncoder.WriteMessageAsync(Message message, Int32 maxMessageSize, BufferManager bufferManager, Int32 messageOffset)
   at System.ServiceModel.Channels.TextMessageEncoderFactory.TextMessageEncoder.WriteMessage(Message message, Int32 maxMessageSize, BufferManager bufferManager, Int32 messageOffset)
   at System.ServiceModel.Channels.BufferedMessageContent.EnsureMessageEncoded()
   at System.ServiceModel.Channels.BufferedMessageContent.TryComputeLength(Int64& length)
   at System.Net.Http.HttpContent.GetComputedOrBufferLength()
[...]

PR #758, opened 2016-01-27
Tests covering using MessageHeaderAttribute with XmlSerializerFormatAttribute

### LINKED ISSUE #676: Add Tests covering using MessageHeaderAttribute with XmlSerializerFormatAttribute
We should also verify whether this test would pass in Net Native.

shmao:
Added two tests covering using MessageHeaderAttribute with XmlSerializerFormatAttribute. One test is for the scenario where the request type uses [MessageHeader]; the other one is for the scenario where the response type uses [MessageHeader].

The first test is currently failing due to #702. I've marked the test with [ActiveIssue].

The second test currently passed in .Net Core, but failed in .Net Native. The fix for the second case would be mainly in toolchain, but it also includes some library changes (see XmlSerializerOperationBehavior.cs). These two parts of the fix have no dependency on each other.

Fix #676


src/System.Private.ServiceModel/src/System/ServiceModel/Description/XmlSerializerOperationBehavior.cs
```diff
@@ -412,6 +412,12 @@ private void CreateHeaderDescriptionTable(MessageDescription message, MessageInf
                         else if (headersMapping != null)
                         {
                             XmlMemberMapping memberMapping = headersMapping[headerNameIndex++];
+#if FEATURE_NETNATIVE
+                            // We cannot do the checks in Net Native because the mapping
+                            // object is just a fake mapping instance which has a valid key only.
+                            // The checks should have already been done in SG.exe
+                            headerDescriptionTable.Add(header.Name, header.Namespace, header);
```
roncain: Nit: possibly rephrase comment to not mention SG.exe, which has meaning only to our team.   Maybe just say it was already done during NET Native precompilation.

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
```diff
@@ -222,4 +222,59 @@ MessageHeader header
...
+        var input = new XmlMessageContractTestRequest("1");
+
+        try
+        {
+            // *** EXECUTE *** \\
+            XmlMessageContractTestResponse response = serviceProxy.EchoMessageResponseWithMessageHeader(input);
+
+            // *** VALIDATE *** \\
+            Assert.True(response != null);
+            Assert.Equal(input.Message, response.Message);
```
roncain: Assert.True or Assert.False should provide a meaningful error message so we can debug failures in the lab.

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
```diff
@@ -222,4 +222,59 @@ MessageHeader header
...
+        IXmlMessageContarctTestService serviceProxy = factory.CreateChannel();
+        var input = new XmlMessageContractTestRequestWithMessageHeader("1");
+
+        try
+        {
+            // *** EXECUTE *** \\
+            XmlMessageContractTestResponse response = serviceProxy.EchoMessageResquestWithMessageHeader(input);
+
+            // *** VALIDATE *** \\
+            Assert.True(response != null);
```
roncain: Same comment -- would be better to have a message describing why Assert.True fails.

roncain:
Just so I understand, you are saying the 2nd test will fail when run in NET Native, so it should not surprise us when we see this new failure in NET Native.  Can you refer to an issue describing what the toolchain fix is?

shmao:
[TFS Bug 184966] is used for tracking the fix in the toolchain. Since the change would be checked-in in TFS, I didn’t open any GitHub issue for this.

src/System.Private.ServiceModel/src/System/ServiceModel/Description/XmlSerializerOperationBehavior.cs
```diff
@@ -412,6 +412,12 @@ private void CreateHeaderDescriptionTable(MessageDescription message, MessageInf
                         else if (headersMapping != null)
                         {
                             XmlMemberMapping memberMapping = headersMapping[headerNameIndex++];
+#if FEATURE_NETNATIVE
```
mconnew: We have a lot of #if blocks that we need to eventually remove. The only technical reason for an #if block is if the code won't compile if left in. Could we do a runtime check instead to detect we've been compiled to native to skip the checks?

roncain: Interesting question -- I think we found RuntimeInformation was not supported in NET Native yet.  If it is, it would be very useful for places like this.

shmao: Good suggestion. I will make a change on this.

src/System.Private.ServiceModel/src/System/ServiceModel/Description/XmlSerializerOperationBehavior.cs
shmao: I see. I will change the comments.

src/System.Private.ServiceModel/src/System/ServiceModel/Description/XmlSerializerOperationBehavior.cs
mconnew: Unfortunately it's kind of clunky how to do this. If you look at the [source](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.InteropServices.RuntimeInformation/src/System/Runtime/InteropServices/RuntimeInformation/RuntimeInformation.cs) for RuntimeInformation, you'll see that the string FrameworkDescription is constructed like this:
```csharp
return string.Format("{0} {1}", FrameworkName, typeof(object).GetTypeInfo().Assembly.GetName().Version);
```
Where FrameworkName is conditionally compiled to be the constant value `".NET Native"`. This means we need the following code:
```csharp
    bool isNativeCompiled = RuntimeInformation.FrameworkName.StartsWith(".NET Native");
```
This is a relatively expensive string operation to do each time. Can you add a static helper method to a (hopefully pre-existing) class in our code which caches the result of this so it's a simple bool check. I'm adding a dependency on RuntimeInformation in another PR that I'm just about to merge in the next hour or so.

shmao:

roncain:
LGTM

mconnew:
LGTM

PR #761, opened 2016-01-28
Enable client certificates on HTTPS

### LINKED ISSUE #597: Turning on client cert for HTTPS in CoreCLR and UWP
This will need add new api to CoreCLR.
UWP may not be able to do this because it will require changes in HttpClientHanderler owned by Networking.

mconnew:
Small fixes to enable client certs on coreclr (windows only).  
Bring in a copy of the UWP HttpClient and modify for to enable certificates.

Fixes #597


src/Common/test-runtime/project.json
```diff
@@ -40,11 +40,13 @@
     "System.Runtime.Extensions": "4.0.10",
     "System.Runtime.Handles": "4.0.0",
     "System.Runtime.InteropServices": "4.0.0",
+	"System.Runtime.InteropServices.RuntimeInformation": "4.0.0-rc3-23725",
```
roncain: I thought this was not available for NET Native.  Also nit: indent is different from others.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509SecurityTokenProvider.cs
```diff
@@ -12,61 +12,39 @@ namespace System.IdentityModel.Selectors
...
-
-        public X509Certificate2 Certificate
-        {
-            get { return _certificate; }
         }
 
         protected override async Task<SecurityToken> GetTokenCoreAsync(CancellationToken cancellationToken)
         {
-            return await Task.FromResult<SecurityToken>(new X509SecurityToken(_certificate));
+            return await Task.FromResult<SecurityToken>(new X509SecurityToken(_certificate, _clone, _clone));
```
roncain: Nit: possibly use the named parameter style so we know what (...,_clone, _clone) means without looking it up.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509SecurityTokenProvider.cs
```diff
@@ -12,61 +12,39 @@ namespace System.IdentityModel.Selectors
...
         }
 
         public void Dispose()
         {
-            System.ServiceModel.Security.SecurityUtils.ResetCertificate(_certificate);
+            if (_clone)
+            {
+                System.ServiceModel.Security.SecurityUtils.ResetCertificate(_certificate);
+            }
```
roncain: Just a question: should we null out _certificate here and check for it if Dispose is called again?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
```diff
@@ -0,0 +1,475 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
roncain: I believe the new naming convention is <file><group>.cs, so I believe this should probably be HttpClientHandler.netcore50.cs.  Eventually we need to rename our other RTxxx.  I see not all CoreFx projects are consistent -- some use file.NETNative.cs.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
```diff
@@ -0,0 +1,475 @@
...
+using RTApiInformation = Windows.Foundation.Metadata.ApiInformation;
+using RTHttpBaseProtocolFilter = Windows.Web.Http.Filters.HttpBaseProtocolFilter;
+using RTHttpCacheReadBehavior = Windows.Web.Http.Filters.HttpCacheReadBehavior;
+using RTHttpCacheWriteBehavior = Windows.Web.Http.Filters.HttpCacheWriteBehavior;
+using RTHttpCookieUsageBehavior = Windows.Web.Http.Filters.HttpCookieUsageBehavior;
+using RTPasswordCredential = Windows.Security.Credentials.PasswordCredential;
+
+namespace System.ServiceModel.Channels
+{
+    internal class RTHttpClientHandler : DelegatingHandler
```
roncain: Is there a plan to tag the code we will eventually need to remove when certs are officially handled in NET Native?  It's not entirely clear to me what duplicated code will need to be removed.  It doesn't need to block this PR but it might be nice to open a new issue identifying what code to remove/replace at that future date, and possibly add some new FEATURE_XXX we can use to identify this kind of code.  Thoughts?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
```diff
@@ -0,0 +1,475 @@
...
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException("value");
+                }
+                if (!UseCookies)
+                {
+                    throw new InvalidOperationException(string.Format(CultureInfo.InvariantCulture,
+                        "net_http_invalid_enable_first", "UseCookies", "true"));
```
roncain: Questions: why doesn't this use SR.Format, and will it produce a meaningful error?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
```diff
@@ -0,0 +1,475 @@
...
+
+        public bool PreAuthenticate
+        {
+            get { return true; }
+            set
+            {
+                if (value != PreAuthenticate)
+                {
+                    throw new PlatformNotSupportedException(String.Format(CultureInfo.InvariantCulture,
+                        "net_http_value_not_supported", value, "PreAuthenticate"));
```
roncain: Similar kind of question about whether we need SR.Format and what kind of message this will produce.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
```diff
@@ -0,0 +1,475 @@
...
+                                    // layer. So, we need to add all applicable cookies that are received to the
+                                    // CookieContainer.
+                                    if (RTCookieUsageBehaviorSupported || !cookie.HttpOnly)
+                                    {
+                                        CookieContainer.Add(response.RequestMessage.RequestUri, cookie);
+                                    }
+                                }
+                            }
+                            catch (Exception)
+                            {
```
roncain: It's hard to know what was copied from elsewhere, but this silent catch/ignore for all exception types seems dangerous.  If it is a pure duplicate of code and will be removed at some point, I guess it's okay.  But if this is our code to maintain, I'd like to know whether this catch can be constrained to only specific exception types.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpHandlerToFilter.cs
```diff
@@ -0,0 +1,262 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
roncain: Whatever naming convention we adopt for the other files of this nature, it needs to apply to all of them.  HttpHandlerToFilter.netcore50.cs?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceModelHttpMessageHandler.NetNative.cs
```diff
@@ -53,8 +41,8 @@ public bool UseDefaultCredentials
 
         public bool UseProxy
         {
-            get { return _innerHandler.UseProxy; }
-            set { _innerHandler.UseProxy = value; }
+            get { return true; }
+            set { /* RTHttpClient can only use the system defined proxy so this is a no-op */ }
```
roncain: Should this throw if you try to set?  Assert?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/X509CertificateInitiatorClientCredential.CoreCLR.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
roncain: Aside -- I have not heard a naming convention what we call "CoreCLR", but it looks like CoreFx is using this same naming pattern for similar purposes.  So I think this is the right file name for us.

src/System.Private.ServiceModel/src/windows/project.json
```diff
@@ -32,11 +32,13 @@
     "System.Runtime.Extensions": "4.0.10",
     "System.Runtime.Handles": "4.0.0",
     "System.Runtime.InteropServices": "4.0.0",
+	"System.Runtime.InteropServices.RuntimeInformation": "4.0.0-rc3-23725",
```
roncain: Same observation -- is this package available for NET Native?  API Catalog says it isn't.  Also the indent on this line is different than the others

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/RequestLoggingServiceBehavior.cs
```diff
@@ -0,0 +1,138 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
roncain: Just checking -- this looks like a test enhancement unrelated to the cert feature.  If so, okay for this PR but would prefer we make separate PR's for separate features.  Don't block on this.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/RequestLoggingServiceBehavior.cs
```diff
@@ -0,0 +1,138 @@
...
+                if (requestMessageProperty != null)
+                {
+                    log.Append(requestMessageProperty.Method).Append(' ');
+                }
+            }
+
+            log.AppendLine(message.Properties.Via.ToString());
+            log.Append("To Header:").AppendLine(message.Headers.Action);
+            log.Append("Authentication ").AppendLine(authenticated ? "succeeded" : "failed");
+            Console.WriteLine(log.ToString());
```
roncain: Not for this PR, but at some point we need have a central mechanism for logging or tracing from the Bridge.  As written, we will see this spew but not know where it is coming from or why.

src/Common/test-runtime/project.json
mconnew: This was a tab vs space issue.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509SecurityTokenProvider.cs
mconnew: Great idea, the second parameter is `disposable` and that's totally non-obvious.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509SecurityTokenProvider.cs
mconnew: We don't on desktop. I'm hesitant to add extra checks when it hasn't been an issue. Also, at least at the moment, this is an internal only class so we are in full control of any usage.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
mconnew: This isn't a partial class of HttpClientHandler but is a full implementation in its own right so those naming conventions don't really apply. Having HttpClientHandler.netcore50.cs implies there's a base file without the suffix.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
mconnew: This entire class along with several others will outright disappear. Currently we have ServiceModelHttpMessageHandler which has partial implementations for CoreClr, NetNative and Windows. HttpsChannelFactory in multiple places checks various properties to see if certain capabilities are there before trying to use them. When client cert usage becomes available cross plat, the change will start at HttpsChannelFactory and it will be obvious which code to remove as the classes just won't be referenced any more. I'll open a bug and mention the files which will need to be touched/removed though as we don't know when this will become available.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
mconnew: The SR values only exist in NCL code in CoreFx. On NetNative, the string resources are stripped anyway. I'll modify this to look like what would normally be returned. For example this one will become `"net_http_invalid_enable_first, UseCookies, true"`.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
mconnew: This is straight copied code. This entire class will disappear eventually.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpHandlerToFilter.cs
mconnew: Same as above, this isn't a partial implementation with netcore only code in this file. This is the entire implementation.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
roncain: Actually the naming convention doesn't imply it is a partial class (that I know).  It just means "this file contains code only for XXX".  I think we need to choose some convention for files that are strictly for one OS or target group, regardless how they are structured inside.  I thought the new naming convention was a reasonable proposal for that.  The fact we include "**\*.cs" makes it especially non-obvious which of our files are OS/target specific.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceModelHttpMessageHandler.NetNative.cs
mconnew: This is the same behavior as we had before as this is copied from HttpClientHandler in netcore50. _innerHandler.UseProxy used to call into the netcore50 implementation of HttpClientHandler.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/X509CertificateInitiatorClientCredential.CoreCLR.cs
mconnew: Additionally, this file will be going away once HttpClientHandler supports certificates.

src/System.Private.ServiceModel/src/windows/project.json
mconnew: It's available for Net Native. Indent is tab vs space and I'll fix it.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/RequestLoggingServiceBehavior.cs
mconnew: I separated the PR into two commits. One has product code changes, the other has test code changes. That way any cherry picking won't pick up any test related code.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/RequestLoggingServiceBehavior.cs
mconnew: That's exactly what this class facilitates. You can provide a delegate with any functionality you want for the logging so this is completely flexible, but at the same time has a reasonable "out of the box" implementation for test debugging purposes. I would also want to add a message inspector to be able to log the reply message too, but I didn't need that for my purposes so didn't add it.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTHttpClientHandler.cs
roncain: You know, given that this decision involves some discussion, I'm fine with the file names you have now, and we can visit for (possible) file renames later, after this is merged.

roncain:
LGTM -- subject to the changes you mentioned about string.format.  I think the file rename issue should not block this PR.

PR #762, opened 2016-01-28
Fix WebSockets duplex streaming when it completes async

### LINKED ISSUE #760: System.ArgumentException in async duplex channel.
We get a System.ArgumentException in async duplex streaming with the message: "The asynchronous result object used to end this operation was not the object that was returned when the operation was initiated"

The meaningful part of the call stack is:
~~~
System_Private_ServiceModel!System.ServiceModel.Channels.TransportDuplexSessionChannel.OnWriteComplete(System.Object)
System_Private_ServiceModel!System.ServiceModel.Channels.WebSocketTransportDuplexSessionChannel.StreamWriteCallback(System.IAsyncResult)
~~~

mconnew:
Fixes #760

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportDuplexSessionChannel.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportDuplexSessionChannel.cs
@@ -533,7 +533,7 @@ namespace System.ServiceModel.Channels
                     AsyncCompletionResult completionResult;
                     if (this.IsStreamedOutput)
                     {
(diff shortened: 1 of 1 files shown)


roncain:
LGTM.
Question: why did we discover this only in stress tests?  Are we missing a functional test that does async duplex streaming?

mconnew:
This would only have failed when the streamed write went async. Normally it would immediately get put into some socket buffer somewhere and report that it completed synchronously. In a stress test, it could be contention in the kernel or the kernel not allocating as much buffer to each socket as it now has more of them or one of several other possibilities which now causes the write to complete asynchronously. That means we rely on the callback to complete the task object. When it completes synchronously, we set the task result immediately ourselves so it didn't matter that the state object was incorrect.

roncain:
LGTM. Thanks @mconnew.

PR #767, opened 2016-01-29
Deletes Bridge.Build.Tasks

roncain:
This project was originally intended to hold MSBuild tasks to
deal with the Bridge, but is now obsolete and interferes with
building within the ToF environment.  Its only existing task was
to ask the Bridge to release its resources, but that functionality
has since been moved into Bridge.exe itself and is invoked after
tests have been run.


iamjasonp:
Buddy built + LGTM

PR #770, opened 2016-02-01
Allow Bridge to know whether it was manually started

roncain:
To enable automation of the Bridge in OuterLoop test
runs in the lab, it is necessary for the Bridge itself
to know whether it was manually started by the developer
or automatically started by a test.

The StartBridge.cmd script now propagates that information
to the Bridge.  And a new -stopIfAutoStart switch to Bridge.exe
will stop the Bridge only if it was not manually started.

The old KeepBridgeRunning environment variable is no longer used.


roncain:
Reviewers: I need this feature to allow me to start and conditionally stop the Bridge in ToF runs. Without this, the ToF runs will start and stop the Bridge every test.

zhenlan:
This brings to the question what is considered manual start vs. auto start. Sounds like all we need is a -autoStop switch and we will use proper values in different configurations (either lab run or dev run).

roncain:
Yes. We can work on phrasing, but StartBridge.cmd is "manual" and EnsureBridgeRunning is "auto". I am using these in exactly the ways you describe. Do you have a concrete proposal to change spelling or use in this PR? BTW, "auto" is the default for all Bridge starts, except for the manual cmd to "StartBridge". Outerloops, including Tof use "auto".

zhenlan:
It's more than phrasing but more on design perspective. "stop" is a feature of bridge but whether bridge will stop when it runs locally (-stopIfLocal) or when it is started manually (-stopIfAutoStart) should be up to the caller.

> but StartBridge.cmd is "manual" and EnsureBridgeRunning is "auto"

If I have a choice, I would prefer not having to remember this :smiley:. Whoever new to this will also scratch his/her head and wonder why.

> "auto" is the default for all Bridge starts, except for the manual cmd to "StartBridge". Outerloops, including Tof use "auto".

Sounds like all we need is one flag with two states that does not have to be tied with how the bridge is started (who knows one may use StartBridge.cmd in some automation too one day). Can we just reuse the "-stop" parameter and pass it to bridge only in configurations you consider as "auto" today?

src/System.Private.ServiceModel/tools/setupfiles/CleanupWCFTestService.cmd
```diff
@@ -1,20 +1,11 @@
...
-    pushd %~dp0..\..\..\..\bin\wcf\tools\Bridge
-    echo Invoking Bridge.exe -reset %* ...
-    call Bridge.exe -reset %*
-    popd
-    echo The Bridge was left running because BridgeKeepRunning is true
-)
+echo Releasing Bridge resources and stopping it if running locally.
+pushd %~dp0..\..\..\..\bin\wcf\tools\Bridge
+echo Invoking Bridge.exe -stopIfAutoStart -reset %* ...
+call Bridge.exe -stopIfAutoStart -reset %*
```
iamjasonp: Just a thought: I think our use of binary "true/false" operation modes might be hindering our thinking. 

For example, `stopIfAutoStart` is kind of weirdly worded to me, and I wonder if we should do something like "script mode" vs. "manual mode" vs. "long-running mode"

i.e., 
if you're running under "script" mode, then you expect a script to kill the Bridge (or it to kill itself after a while)
if you're running "manually" you are expected to start/stop yourself
if you're running under "longrunning timeout" mode - it can time out by itself after inactivity

And you'd specify this by using an option like `-mode=manual` or `mode=longrunning`

Not blocking for checkin

iamjasonp:
LGTM, buddy built and no problems found. Consider this comment: https://github.com/dotnet/wcf/pull/770#discussion_r51560198; it might simplify our future logic a bit

src/System.Private.ServiceModel/tools/setupfiles/CleanupWCFTestService.cmd
roncain: The unspoken design goal here is that I can start the Bridge manually for debugging, and no test script is permitted to shut it down.  Or another process can be managing the lifetime of the Bridge and does not want test scripts to stop it. It is literally a question of who owns the lifetime of the Bridge. To respond to @zhenlen, the Bridge should remember how it was started, because it will condition under what circumstances it should be stopped.  There is no persistence mechanism between test runs or machines other than the Bridge itself.

I think I will refactor to a single boolean flag "BridgeAutoStart" that is set from any test script that causes the Bridge to start.  It will make the new command "-stopIfAutoStart" make more sense.

roncain: Regarding remembering "StartBridge" means "manually started and won't be stopped by tests" and "EnsureBridge" means "automatically started and can be stopped by tests", that sounds like a matter of choosing the right name.  Suggestions to improve? Just "StartBridge" and "AutoStartBridge?"  "StartLongRunningBridge?"  Note that these CMD's are only wrappers around how Bridge.exe gets invoked.  The ToF test scripts will not use them because they directly invoke Bridge.exe (the ToF tests run in an environment where there is no source or script code, so the Bridge already needs to have been built).

roncain: @Regarding @iamjasonp suggestion about start modes -- interesting and worth considering.  My main concern is that the max-idle-timeout is slightly orthogonal and can be applied to a Bridge started either manually or by a script.  Which brings us back to a 2-state value -- it was started by a script or it was started manually.  I rely on that max idle timeout to rescue CI machines that may have inadvertently not completed a test run cleanly.

iamjasonp: So I see two pivots based on this discussion: 

* The Bridge should (not) be allowed to be terminated by test script
* The Bridge should (not) shut down after a timeout occurs (use case: when running locally, i *never* want the bridge to shut down by itself, but on CI boxes, shut them down after x hours to rescue the machines. This can be achieved by setting the timeout to be infinite).

Perhaps "AutoShutdown" or "ShutdownAfterTest" might be a more useful moniker for StartBridge/EnsureBridge, given the different modes seem to be talking about how we intend to *stop* the bridge, rather than how we intend to *start* it.

Aside: FWIW, I personally always found I needed to do a double-take when thinking about "StartBridge" and "EnsureBridge" and how it interacts with timeouts.

roncain:
@Zhenlan, @iamjasonp and @hongdai -- this latest iteration comes back to what Zhenlan was proposing, namely let the script decide how to stop the bridge by letting it know how it was started. I will be sending out a TFS codeflow with ToF targets that use these same conventions.

roncain:

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -74,9 +90,11 @@ private static void Main(string[] args)
         // Issues a GET request to the Bridge to determine whether it is alive.
         // A return of 'true' means the Bridge is healthy.  A return of 'false'
         // indicates the Bridge is not healthy, and 'errorMessage' describes the problem.
-        private static bool PingBridge(string host, int port, out string errorMessage)
+        // A healthy response will also populate properties with the current Bridge configuration.
+        private static bool PingBridge(string host, int port, out string errorMessage, out Dictionary<string, string> properties)
```
zhenlan: I don't see we actually need and use the new "properties" anywhere. Without adding the properties parameter here, most change in this PR can go away.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -16,6 +16,15 @@ namespace Bridge
...
     internal class Program
     {
+        // Exit codes returned by this executable
+        private const int Bridge_ExitCode_Success = 0;
+        private const int Bridge_ExitCode_Failure = -1;
+        private const int Bridge_ExitCode_AlreadyRunning = 1;
+        private const int Bridge_ExitCode_Started = 2;
+        private const int Bridge_ExitCode_Stopped = 3;
+        private const int Bridge_ExitCode_NotRunning = 4;
+        private const int Bridge_ExitCode_NotStopped = 5;
```
zhenlan: We don't really need the exit code 2, 3, 4 and 5. For the purpose of this PR, all we want to know is if the bridge is already running (exit code 1). For all other commands (eg. start, stop etc.), an exit code of either success (0) or failure (-1) will be sufficient.

src/System.Private.ServiceModel/tools/setupfiles/EnsureBridgeRunning.cmd
```diff
@@ -30,9 +20,29 @@ if '%BridgeResourceFolder%' == '' (
...
+  echo EnsureBridgeRunning: The Bridge was already running
+  set BridgeStartedByScript=false
+  goto done
+  
+:done
 popd
 
-
+endlocal & set BridgeStartedByScript=%BridgeStartedByScript%
 exit /b
```
zhenlan: Keeping an env variable (BridgeStartedByScript) floating around between 4 scripts (EnsureBridgeRunning.cmd, ReleaseBridge.cmd, build.cmd and StartBridge.cmd) could be a headache for long term maintenance.

Why don't we just do "exit /b %BridgeReturnCode%" here (may need move endlocal below)? In Build.cmd, it will check/record the errorlevel of this script. At the time of release, it will pass either -reset or -stop to the bridge.exe depending on the earlier errorlevel code. We can get rid of ReleaseBridge.cmd completely. Finally, in the StartBridge.cmd, it will simply ignore the errorlevel code from this scrip. In this way, each script has full control of the lifetime of bridge, all logic is self-contained and there is no magic variables.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -253,56 +284,90 @@ private static void ResetBridge(CommandLineArguments commandLineArgs)
...
             {
-                Console.WriteLine("The Bridge is already running.");
-                Environment.Exit(0);
+                // Ping first in case the Bridge is running already or on another machine
+                if (PingBridge(commandLineArgs.BridgeConfiguration.BridgeHost,
+                                               commandLineArgs.BridgeConfiguration.BridgePort,
+                                               out errorMessage,
+                                               out properties))
+                {
+                    BridgeConfiguration bridgeConfiguation = new BridgeConfiguration(properties);
```
zhenlan: We don't need this.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -16,6 +16,15 @@ namespace Bridge
 {
     internal class Program
     {
+        // Exit codes returned by this executable
+        private const int Bridge_ExitCode_Success = 0;
+        private const int Bridge_ExitCode_Failure = -1;
+        private const int Bridge_ExitCode_AlreadyRunning = 1;
```
zhenlan: Maybe we should add a comment here... AlreadyRunning is reserved only for -require command.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -181,7 +207,7 @@ private static void StopBridge(CommandLineArguments commandLineArgs)
                 Console.WriteLine("A problem was encountered stopping the Bridge:{0}{1}",
                                     Environment.NewLine, problem);
                 Console.WriteLine("Forcing local resource cleanup...");
-                BridgeController.StopBridgeProcess(1);
+                BridgeController.StopBridgeProcess(Bridge_ExitCode_Stopped);
```
zhenlan: As soon as we added more exit code than we actually need, we will be confused what exit code to use at each case. So as I recommended earlier, we should just stick with success or failure for most commands (except -require for now).

For the case of this line, we should exit with failure and the line below should be success. They both exit with the same code now.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -253,56 +284,90 @@ private static void ResetBridge(CommandLineArguments commandLineArgs)
...
+            // Otherwise we would recursively start this same logic in a new process.
+            List<string> originalArgList = new List<string>(commandLineArgs.OriginalArgs);
+            for (int i = originalArgList.Count-1; i >= 0; --i)
+            {
+                string[] argParts = (originalArgList[i].Substring(1)).Split(':');
+                if (argParts.Length > 0 && String.Equals(argParts[0], "require", StringComparison.OrdinalIgnoreCase))
+                {
+                    originalArgList.RemoveAt(i);
+                }
+            }
```
zhenlan: Are we already using something like "/require:123" somewhere?

startBridge.cmd
```diff
@@ -1,20 +1,18 @@
...
-goto done
+REM Undo any decision EnsureBridgeRunning made about starting the
+REM Bridge so the next test run will leave the Bridge running.
+set BridgeStartedByScript=false
 
-:error
-echo An error occurred starting the Bridge
-popd
-exit /b l
+echo Because you started the Bridge manually, it will remain running until you stop it manually.
```
zhenlan: Not stopping bridge is just the behavior we decided for this script regardless if someone typed startBridgel.cmd in the console window or it is called by another script. Maybe a simple reminder like "The Bridge will remain running until you stop it manually." will be sufficient.

zhenlan:
Thank you @roncain for coming up this solution :cool:. The behavior is way more intuitive and easy to explain now.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
roncain: /require:nnn is supported for flexibility, but the hashset approach didn't handle it properly.  I found some instances where the default 1 minute was too short (high load) and did use /require:180 to allow more time.  I'll make the default reasonable, but still want to support this syntax as an emergency out that can be provided by the script.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
roncain: Sorry could you clarify?  What don't we need and why don't we need it?  The initial ping allows someone to start the Bridge on another machine and let ToF naturally detect and use it.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
roncain: I think I'd prefer to keep it.  It was an omission in the original design to have no way for a Bridge to obtain the BridgeConfiguration of a Bridge that was already running.  The content already exists in the response, so rather than throw it away, we at least capture it.   It was cheap to capture and will be useful in other scenarios.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
roncain: Okay -- simplifying to the first 3 only.

src/System.Private.ServiceModel/tools/setupfiles/EnsureBridgeRunning.cmd
roncain: Agreed -- losing any global environment variable and relying on exit /b to propagate return code through the scripts.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
roncain: Reverted to only 0 (success), -1 (failure) and 1 (-require found Bridge already running).

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
iamjasonp: +1, please keep this change. I've run into this issue before where I didn't have a copy of the BridgeConfiguration and wanted it.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -253,56 +284,90 @@ private static void ResetBridge(CommandLineArguments commandLineArgs)
...
+                // Ping failed.  If the Bridge process is running, try again until it responds.
+                // If the Bridge process is not even running, no need to keep retrying.
+                // We have to consider the Bridge.exe running that is executing this code
+                // as well as the Bridge.exe running as a service hoster.
+                bool bridgeRunning = Process.GetProcessesByName("Bridge").Count() >= 2;
+                if (!bridgeRunning)
+                {
+                    Console.WriteLine("The Bridge.exe process is not running so it will be started now.");
+                    break;
+                }
```
iamjasonp: Might be good to Thread.Sleep for a little bit instead of tight looping?

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
iamjasonp: Reminder: please clean up these exit codes if no longer used :)

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -318,10 +383,12 @@ private static Process StartBridgeInNewProcess(CommandLineArguments commandLineA
...
+            Dictionary<string, string> properties;
 
             if (PingBridge(commandLineArgs.BridgeConfiguration.BridgeHost,
                                            commandLineArgs.BridgeConfiguration.BridgePort,
-                                           out errorMessage))
+                                           out errorMessage,
+                                           out properties))
             {
                 Console.WriteLine("The Bridge is already running.");
                 Environment.Exit(0);
```
iamjasonp: NIT: now that the variables are defined, use ExitCode_Success instead of 0

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
zhenlan: I don't see bridgeConfiguation is used anywhere after it is created here. Did I miss anything?

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
```diff
@@ -253,56 +284,90 @@ private static void ResetBridge(CommandLineArguments commandLineArgs)
...
+                    return Bridge_ExitCode_AlreadyRunning;
+                }
+
+                Console.WriteLine("  -- the ping failure response was: {0}", errorMessage);
+
+                // Ping failed.  If the Bridge process is running, try again until it responds.
+                // If the Bridge process is not even running, no need to keep retrying.
+                // We have to consider the Bridge.exe running that is executing this code
+                // as well as the Bridge.exe running as a service hoster.
+                bool bridgeRunning = Process.GetProcessesByName("Bridge").Count() >= 2;
```
iamjasonp: Just a thought - are there any other known programs that might be running on a dev machine also called "Bridge"? It's quite a generic name.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
zhenlan: In general I recommend us not to create something until we actually use it. We have seen so many times in the past things were created because they were thought to be useful but never used and caused confusions.

In this case, if you guys feel strongly we will use it, you can keep it. Maybe we can create another overload, so we don't have to change all existing code that calls PingBridge.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
zhenlan: Sure, as long as you do use it. Yes, change default is a good idea too.

src/System.Private.ServiceModel/tools/test/Bridge/Bridge/Program.cs
roncain: Note: good comments but the level of review on this PR exceeds my ability to respond and still make progress on ToF.  I am going to debug what I just pushed up and come back to consider these later.  Whew, we need this level of feedback on product code PR's :smile:

zhenlan: Sure. If it is not already the case, we need better code review for product code PR's :smiley: .

roncain:

roncain:
Merging now because OuterLoops pass on 2 machines (CI seems to not run them).  I will address remaining feedback in follow-up PR.  I need this in TFS so I can light it up in ToF.

PR #774, opened 2016-02-02
Enable running against latest dependencies

hongdai:
* Use FloatingTestRuntimeDependencies=true to run against latest dependencies
* Fix the build script to allow config TestFixedRuntimeProjectJson
* Note that versions "4.0.11-rc3-23725" will be changed to "4.0.11-*" in
the build process and copy to a location used by the tests.
* Add default configration explcitly to work around the issue it will fail
if no build configration is specified

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -19,6 +19,16 @@ if %errorlevel% equ 0 (
   set outloop=true
 )
 
+echo %* | findstr /i /C:"FloatingTestRuntimeDependencies=true"  1>nul
+set _FloatingDependencies=!ERRORLEVEL!
+if '!_FloatingDependencies!'=='0' (
+  set TestFixedRuntimeProjectJson=%~dp0src\Common\test-runtime\LatestDependencies\project.json
+  echo %* | findstr /i /C:"/p:Configuration="  1>nul
+  if !ERRORLEVEL! neq 0 (
+  set _defaultBuildConfig=/p:Configuration=Windows_NT_Debug
+  )
+)
+
 if not defined VisualStudioVersion (
     if defined VS140COMNTOOLS (
         call "%VS140COMNTOOLS%\VsDevCmd.bat"
@@ -70,7 +80,7 @@ call :build %*
 goto :AfterBuild
 
 :build
-%_buildprefix% msbuild "%_buildproj%" /nologo /maxcpucount /verbosity:minimal /nodeReuse:false /fileloggerparameters:Verbosity=diag;LogFile="%_buildlog%";Append %* %_buildpostfix%
+%_buildprefix% msbuild "%_buildproj%" %_defaultBuildConfig% /nologo /maxcpucount /verbosity:minimal /nodeReuse:false /fileloggerparameters:Verbosity=diag;LogFile="%_buildlog%";Append %* %_buildpostfix%
 set BUILDERRORLEVEL=!ERRORLEVEL!
 goto :eof
 
--- a/src/Common/test-runtime/LatestDependencies/project.json
+++ b/src/Common/test-runtime/LatestDependencies/project.json
@@ -1,0 +1,75 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore.Platforms": "1.0.1-rc3-23729",
+    "Microsoft.NETCore.TestHost": "1.0.0-rc3-23729",
+    "Microsoft.NETCore.Console": "1.0.0-rc3-23729",
+    "coveralls.io": "1.4",
+    "OpenCover": "4.6.166",
+    "ReportGenerator": "2.3.4",
+    "Microsoft.DotNet.xunit.performance.analysis": "1.0.0-alpha-build0025",
+    "Microsoft.DotNet.xunit.performance.runner.Windows": "1.0.0-alpha-build0025",
+    "System.Collections": "4.0.11-rc3-23729",
+    "System.Collections.Concurrent": "4.0.11-rc3-23729",
+    "System.Collections.NonGeneric": "4.0.1-rc3-23729",
+    "System.Collections.Specialized": "4.0.1-rc3-23729",
+    "System.ComponentModel": "4.0.1-rc3-23729",
+    "System.ComponentModel.EventBasedAsync": "4.0.11-rc3-23729",
+    "System.Diagnostics.Contracts": "4.0.1-rc3-23729",
+    "System.Diagnostics.Debug": "4.0.11-rc3-23729",
+    "System.Diagnostics.Tools": "4.0.1-rc3-23729",
+    "System.Diagnostics.Tracing": "4.0.21-rc3-23729",
+    "System.Globalization": "4.0.11-rc3-23729",
+    "System.IO": "4.0.11-rc3-23729",
+    "System.IO.Compression": "4.1.0-rc3-23729",
+    "System.Linq": "4.0.1-rc3-23729",
+    "System.Linq.Expressions": "4.0.11-rc3-23729",
+    "System.Linq.Queryable": "4.0.1-rc3-23729",
+    "System.Net.Http": "4.0.1-rc3-23729",
+    "System.Net.Primitives": "4.0.11-rc3-23729",
+    "System.Net.WebHeaderCollection": "4.0.1-rc3-23729",
+    "System.Net.WebSockets": "4.0.0-rc3-23729",
+    "System.Net.WebSockets.Client": "4.0.0-rc3-23729",
+    "System.ObjectModel": "4.0.11-rc3-23729",
+    "System.Reflection": "4.0.11",
+    "System.Reflection.DispatchProxy": "4.0.1-rc3-23729",
+    "System.Reflection.Extensions": "4.0.1-rc3-23729",
+    "System.Reflection.Primitives": "4.0.1-rc3-23729",
+    "System.Reflection.TypeExtensions": "4.0.1-rc3-23729",
+    "System.Resources.ResourceManager": "4.0.1-rc3-23729",
+    "System.Runtime": "4.0.21-rc3-23729",
+    "System.Runtime.Extensions": "4.0.11-rc3-23729",
+    "System.Runtime.Handles": "4.0.1-rc3-23729",
+    "System.Runtime.InteropServices": "4.0.21-rc3-23729",
+    "System.Runtime.Serialization.Xml": "4.1.0-rc3-23729",
+    "System.Security.Claims": "4.0.1-rc3-23729",
+    "System.Security.Cryptography.X509Certificates": "4.0.0-rc3-23729",
+    "System.Security.Principal": "4.0.1-rc3-23729",
+    "System.Text.Encoding": "4.0.11-rc3-23729",
+    "System.Threading": "4.0.11-rc3-23729",
+    "System.Threading.Tasks": "4.0.11-rc3-23729",
+    "System.Threading.Timer": "4.0.1-rc3-23729",
+    "System.Xml.ReaderWriter": "4.0.11-rc3-23729",
+    "System.Xml.XDocument": "4.0.11-rc3-23729",
+    "System.Xml.XmlDocument": "4.0.1-rc3-23729",
+    "System.Xml.XmlSerializer": "4.0.11-rc3-23729",
+    "xunit": "2.1.0",
+    "xunit.console.netcore": "1.0.2-prerelease-00120",
+    "xunit.runner.utility": "2.1.0"
+  },
+  "frameworks": {
+    "dnxcore50": {
+      "imports": "portable-net45+win8", 
+      "dependencies": {
+        "System.Net.Http.WinHttpHandler": "4.0.0-rc3-23729",
+        "System.Net.Security": "4.0.0-rc3-23729"
+      }
+    }
+  },
+  "runtimes": {
+    "win7-x64": {},
+    "win7-x86": {},
+    "ubuntu.14.04-x64": {},
+    "osx.10.10-x64": {}
+  }
+}
+
--- a/dir.props
+++ b/dir.props
@@ -425,8 +425,8 @@
   <PropertyGroup>
     <TestRuntimePath>$(SourceDir)Common/test-runtime/</TestRuntimePath>
 
-    <TestFixedRuntimeProjectJson>$(TestRuntimePath)project.json</TestFixedRuntimeProjectJson>
-    <TestFixedRuntimeProjectLockJson>$(TestRuntimePath)project.lock.json</TestFixedRuntimeProjectLockJson>
+    <TestFixedRuntimeProjectJson Condition="'$(TestFixedRuntimeProjectJson)' == ''">$(TestRuntimePath)project.json</TestFixedRuntimeProjectJson>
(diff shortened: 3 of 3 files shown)


build.cmd
```diff
@@ -67,7 +73,7 @@ call :build %*
 goto :AfterBuild
 
 :build
-%_buildprefix% msbuild "%_buildproj%" /nologo /maxcpucount /verbosity:minimal /nodeReuse:false /fileloggerparameters:Verbosity=diag;LogFile="%_buildlog%";Append %* %_buildpostfix%
+%_buildprefix% msbuild "%_buildproj%" /p:Configuration=%defaultBuildConfigure% /nologo /maxcpucount /verbosity:minimal /nodeReuse:false /fileloggerparameters:Verbosity=diag;LogFile="%_buildlog%";Append %* %_buildpostfix%
```
zhenlan: When defaultBuildConfigure is not defined, will this cause anything unexpected?

hongdai: when defaultBuildConfigure is not defined, that means FloatingTestRuntimeDependencies is not defined. It will use the minimum dependencies as usual, nothing unexpected.

hongdai: I'll need to rebase the commitment to get the latest update version change from Jason.

hongdai:
I also need to update the file versions in the floating test runtime project.jason file.

build.cmd
zhenlan: Understood. Then we will just pass "/p:Configuration= /nologo ..." to msbuild. Is this intentional (when defaultBuildConfigure is undefined)?

hongdai: Good point! It's not intentional although looks like msbuild just ignore it in this case. I'll make a change to get rid of it.

build.cmd
```diff
@@ -18,6 +18,12 @@ if %errorlevel% equ 0 (
   set outloop=true
 )
 
+echo %* | findstr /i /C:"FloatingTestRuntimeDependencies=true"  1>nul
```
iamjasonp: This is a little bit of a misnomer: it's not really "floating test runtime dependencies", just the "latest" ones in test-runtime. 
Perhaps there's a way for us to creatively retrieve the latest version number of (say) Microsoft.NetCore from myget.org, then do a replace on the project.json file for the actual latest at the time.

iamjasonp: As an example, this PR references `rc3-23729` in test-runtime, but the packages just revved to `rc3-23802`

build.cmd
```diff
@@ -67,7 +73,7 @@ call :build %*
 goto :AfterBuild
 
 :build
-%_buildprefix% msbuild "%_buildproj%" /nologo /maxcpucount /verbosity:minimal /nodeReuse:false /fileloggerparameters:Verbosity=diag;LogFile="%_buildlog%";Append %* %_buildpostfix%
+%_buildprefix% msbuild "%_buildproj%" %defaultBuildConfig% /nologo /maxcpucount /verbosity:minimal /nodeReuse:false /fileloggerparameters:Verbosity=diag;LogFile="%_buildlog%";Append %* %_buildpostfix%
```
iamjasonp: Should adding %defaultBuildConfig% be conditioned on if /p:Configuration= isn't already otherwise specified? 

Or... if we do `msbuild /p:Configuration=Configuration1 /p:Configuration=Configuration2`, which one wins?

build.cmd
```diff
@@ -18,6 +18,12 @@ if %errorlevel% equ 0 (
   set outloop=true
 )
 
+echo %* | findstr /i /C:"FloatingTestRuntimeDependencies=true"  1>nul
+if %errorlevel% equ 0 (
+  set TestFixedRuntimeProjectJson=%~dp0src\Common\test-runtime\LatestDependencies\project.json
+  set defaultBuildConfig=/p:Configuration=Windows_NT_Debug
```
iamjasonp: variable names should follow convention and start with _ for variables defined in the script

roncain: I am also editing build.cmd so we are going to conflict -- just head's up for one of us.  I also just discovered %ERRORLEVEL% and ERRORLEVEL don't expand correctly when used multiple times in the same CMD file unless ENABLEDELAYEDEXPANSION is set.  You may want to look at #770.

roncain: Question: why was it necessary to set Configuration?  And is the intent only to work on Windows?  You should probably check with @StephenBonikowsky to understand the consequences of building Windows and Linux packages when we decide we want to build them with the latest dependencies.

build.cmd
hongdai: The property FloatingTestRuntimeDependencies was originally designed for testing the latest dependencies by Corefx team.  The design has been reviewed by both Corefx and WCF team, and implemented by Corefx. The implementation did not work as expected and I figured out it's because TestFixedRuntimeProjectJson cannot be configured. This PR fixes it.

hongdai: As for rc3-23729, it's required by build validation rules defined in our dirs.prop. The build will fail if it does not match.
 <ValidationPattern Include="^((System\..%2A)|(Microsoft\.CSharp)|(Microsoft\.NETCore.%2A)|(Microsoft\.TargetingPack\.Private.%2A)|(Microsoft\.Win32\..%2A)|(Microsoft\.VisualBasic))(?&lt;!TestData)$">
      <ExpectedPrerelease>rc3-23729</ExpectedPrerelease>
    </ValidationPattern>

build.cmd
iamjasonp: I think ""Build and test on Linux" be part of build.sh and not build.cmd - maybe that's acceptable here to default to Windows_Debug.

hongdai: Will fix it.

build.cmd
hongdai: Matt told me the last one wins, and I have verified it's the case :-)

build.cmd
hongdai: @roncain This PR is only for Windows support. Linux is the next step. Build.cmd only used by Windows build process. Linux does not uses it.

roncain: Would you please open issues to cover scenarios 2 and 3 above?  All 3 are important scenarios.  Thanks.

src/Common/test-runtime/LatestDependencies/project.json
```diff
@@ -0,0 +1,75 @@
+{
```
roncain: I don't think we want to hard-code the "latest" dependencies in a file like this.  It would be better to have a special project.json like this that contains a placeholder value for each, like:
```
  "Microsoft.NETCore.TestHost": "1.0.0-rc3-$USEVERSION$",
```
Then if we ask to build with latest versions, an MSBuild task would copy that to some known location (in bin) and replace $VERSION with a value we specify on the command line (e.g. /p:UseVersion="*" or /p:UseVersion="23729").  This will make it much more flexible because we can ask to test with any specific version we want, not just "the latest on myget".  I know my comments make the scope of this change bigger than you'd planned, but I really think what I'm proposing is the right design for this feature.

I thought @dagood initially had a task that did something similar.  It's possible we broke it along the way.

build.cmd
roncain: I think we need to invent a new MSBuild property, something like /p:TestVersion=23729.  If it is not defined in dir.props, it would be set to some value we choose.  And everywhere we see 23729 (or its equivalent) should become something like:
```
<ExpectedPrerelease>rc3-$(TestVersion)</ExpectedPrerelease>
```
We need to get hardcoded versions out of the files that change frequently and not have the same literal in several places.

build.cmd
roncain: It's always good practice in targets files to set properties *only if they have not already been set*.  This would allow one of us to pass it in as a parameter to build.cmd.  CoreFx frequently breaks this rule, but we need to get more methodical about it for our own use.

build.cmd
hongdai: For easy maintenance, I'd suggest we keep our build scripts consistent with corefx build scripts if possible.    Corefx build scripts uses <ExpectedPrerelease>rc3-23729</ExpectedPrerelease>.

I agree corefx build scripts are not always best practice. But easy maintenance weighs more here, I think.

src/Common/test-runtime/LatestDependencies/project.json
dagood: Yes, https://github.com/dotnet/wcf/commit/22998db02e379092318754d728b3b17445d7aa1c added it during https://github.com/dotnet/wcf/pull/646. If you change https://github.com/dotnet/wcf/blob/master/dir.targets#L29 to some property, you could pass in desired prerelease identifiers perhaps with `*` as the default.

build.cmd
hongdai: Make sense. I'll add a check.

build.cmd
hongdai: I have opened issue #777 for scenario 2. Scenario 3(E2E) is a manual verification, just need to manual change the version to the latest in the E2E app. No automation work needed.

roncain: I think scenario 3 still requires us to be able to build WCF packages that depend on the latest versions.  And then we can do the manual E2E verification using those.  It's not automated, but it would still require making sure build.cmd built the right packages with the new dependencies.

build.cmd
roncain: Okay not to change in our repo.   I'll open a PR in CoreFx next time I need to hand-edit the value.  CoreFx needs to make this a configurable property.

build.cmd
hongdai: E2E App are built from VS. build.cmd is NOT used. E2E App tests the dependencies external customers are going to get. Typically the E2E App will get the dependencies from a pre-release testing feed, such as https://www.myget.org/F/aspnetrc1/api/v2

src/Common/test-runtime/LatestDependencies/project.json
hongdai: I'm currently limited by the only version is ALLOWED(**otherwise I get a build failure**) in the project.json file is whatever specified in:
<ExpectedPrerelease>rc3-23729</ExpectedPrerelease>.
 
@dagood  unfortunately it's no longer work as for the reason above.

@roncain the design you proposed will need corefx level build scripts fixes. Also, allow testing any version is an edge scenario. I think we should unblock ourselves for the time being.

dagood: I'm not sure why that would prevent it from working--the copied and modified project.json should be in `bin` where the validator doesn't look. I'm missing the context for the pull request though, so maybe I don't understand the scenario.

build.cmd
zhenlan: > I think scenario 3 still requires us to be able to build WCF packages that depend on the latest versions.

Just to clarify... it's not a goal to build WCF packages using latest version dependencies.

roncain: @zhenlan I can understand it not being a goal of this PR, but I think being able to build WCF packages with latest (or just newer) package dependencies is a critical test scenario for us. How else can we know the packages we deliver will work for customers (prior to being deep in an RC release cycle)?  We can take this conversation off this PR and open a separate issue to discuss it if you think it is appropriate.  @hongdai -- don't block this PR on this discussion.

src/Common/test-runtime/LatestDependencies/project.json
roncain: @hongdai I would like to understand a little more about the build failure.  If I understand the intent of this PR, I believe it is very similar to what @dagood had working with #646. If this PR is attempting to solve the same issue then it makes sense to get dagood's solution working again, or build on it.  If they are solving different issues then ignore me.
Also, we regularly take corefx build script fixes, so if that solves any problems in this PR, we should probably do that first in a separate PR.  If you can tell me the issues, maybe I can take on that PR. Thoughts?

hongdai: This PR is building on top of @dagood  work with #646. $646 enables the scenario where you want to get the latest versions for what is specified in the DEFAULT test runtime project.json. It does not allow us to change to a different location. We need both default and a different project.json for both minimum dependencies and the latest dependencies. My change is to enable it points to a different location.

As for the build failures, it's because the validation rule check all project.jason file versions according to  what is defined in the dirs.prop.

@dagood we are discussing a non default jason file under src directory, we use it to test latest dependencies. All version in this file will be changed into * in the build time and copy to the bin directory.

build.cmd
zhenlan: For WCF package building, we always build against the dependencies we specified in the product project.json file, which is the minimum dependency. To know if there are issues with latest dependency, we run our tests against the WCF packages plus latest version dependent packages. Note, we don't rebuild WCF packages against latest here. If you worry about build break by latest dependencies (eg. API changes), our test against latest will find out. If you worry about functional break (eg. API behavior change), building against latest won't tell us that anyway and only test can.

src/Common/test-runtime/LatestDependencies/project.json
hongdai: @roncain Basically we will need to change the validation rule to allow different versions and * in corefx build script dirs.prop.

roncain:
You will need to rebase and merge (probably build.cmd), but otherwise LGTM.
Be aware you may need to change to !errorlevel! rather than %errorlevel%

src/Common/test-runtime/LatestDependencies/project.json
iamjasonp: Wait... the validation rule should _not_ try to accept * or different versions when we are doing a normal build... we really do want that validation to work, and we want package consistency. 

Maybe the correct thing to do is to put the project.json in some other directory as suggested, and condition the floating dependency project.json to include only on `$(FloatingTestRuntimeDependencies=true`, and also in dir.props, condition such that validation does _not_ happen when `$(FloatingTestRuntimeDependencies=true`

When `$(FloatingTestRuntimeDependencies=false`, then the floating dependency project.json should not factor into the build at all.

roncain:
LGTM, subject to the rebase and use of !errorlevel!.  Incidentally, @iamjasonp has #781 open to move to later dependencies.  Just a head's up if his PR is merged first because it uses more recent versions.

src/Common/test-runtime/LatestDependencies/project.json
hongdai: @iamjasonp We have two solutions here:
1) the current solution: put the project.json under the src directory, and maintain the file versions to match the normal build. It will be copied to the obj directory and changed to * in the build process, the obj directory is not checked for matching version.
2) put in a different directory that not scanned by the validation rule. It cannot be src, cannot be obj and bin. What directory you suggest?

iamjasonp: Is something like "project.json.floating", which gets renamed when we need it possible? 

In my mind, it's just crappy to see another project.json file being maintained for no great reason, but I gues we just find+replace across all files anyway, and it's not too bad. 

I'm just being a bit of a devil's advocate here to see if there's something cleaner 
Your call on what you like best - I'm done being pedantic and will not pursue this any further :smile:

hongdai: I build it on top of  @dagood work with #646, it does not allow names other than project.json. I agree it's crappy to see another project.json file being maintained without great reason. But I wanted to minimize the build script changes,  hence minimize the risk of breaking other existing scenarios.

iamjasonp: Agreed.

iamjasonp:
I think you had a couple of changes you were going to make - but otherwise LGTM

roncain:
LGTM -- the merge conflicts are fixed now.

PR #776, opened 2016-02-03
Create Certificate Authority "bootstrapper" for Linux

iamjasonp:
On Windows, the BridgeClient is capable of installing CA certificates on the system. However, in Linux, that is not possible as we do not have access to the machine certificate stores in write mode. 

The "bootstrapper" and associated scripts need to do the following: 

* Activate and access the Bridge through the BridgeClient (basically, acting like it's a test so we pick up whatever configuration we intend to use during the test) 
* Download the CA certificate from the Bridge and save it to a temporary location
* Copy the certificate to the machine root store (this may vary by distro, but for now we've assumed a location that OpenSSL+Ubuntu uses) 
* Run a script to update the certs

---

Note: This isn't hooked into the build yet - some thinking needs to be done on that front due to the ordering of the build

Note 2: The Windows cmd file is just a helper used to test the program in Windows (a bit friendlier of an environment). It's not necessary to keep in the final checkin, but it's kind of nice to have a helper to get the cert as a file when debugging anyway


src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateBootstrapper.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
+
+show_usage() 
+{
+    echo ""
+    echo "WCF Core Bridge CA Certificate Bootstrapper" 
+    echo "    Usage: $0 [bridge-host] [certificate-file]"
+    echo ""
+    echo "    bridge-host - hostname where Bridge is running" 
+    echo "    certificate-file - Full file path where the retrieved certificate should be saved"
```
iamjasonp: NOTE: certificate-file probably will get hard-coded to ~/tmp/wcfbridge-ca.crt or some similar name and location. However, for debugging purposes, I allow myself the flexibility to stash the cert wherever I like so I can look at it.

src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateBootstrapper.sh
```diff
@@ -0,0 +1,107 @@
...
+    __cafile=$1
+
+    cp -f "$__cafile" /usr/local/share/ca-certificates
+
+    if [ $? -ne 0 ]; then 
+        return $? 
+    fi 
+
+    echo "Updating root certificates with cert from '$__cafile'"
+    update-ca-certificates
```
iamjasonp: Probably need to do a check to see if this command is present as a prerequisite to executing the entire script

src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateBootstrapper.sh
```diff
@@ -0,0 +1,107 @@
...
+    echo "    certificate-file - Full file path where the retrieved certificate should be saved"
+    echo ""
+}
+
+run_bootstrapper() 
+{
+    __bridge_host=$1
+    __outfile=$2
+
+     # Locate any CoreRun.exe to be able to execute BridgeCertificateBootstrapper.exe
```
roncain: Nit: possibly echo __bridgeHost and __outFile.  Not required but I've found it useful for debugging.

src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateBootstrapper.sh
```diff
@@ -0,0 +1,107 @@
...
+   
+    if [ "$__bootstrapper_path" == "" ]; then 
+        echo "Could not find 'BridgeCertificateBootstrapper.exe' under path '$__binpath'" 
+        exit 1
+    fi 
+
+    echo "Found bootstrapper at '$__bootstrapper_path'" 
+    
+    export BridgeHost=$__bridge_host
+
```
roncain: Is it possible for __bridgeHost to not be set?  If so, this would overwrite any BridgeHost environment variable set elsewhere.  I wonder if we should test before exporting.  Also, nit level, but the echo's below may want to say you can pass it in as a parameter or set BridgeHost outside.

src/System.Private.ServiceModel/tools/test/BridgeCertificateBootstrapper/Bootstrapper.cs
```diff
@@ -0,0 +1,97 @@
...
+                Console.WriteLine(certAsPemUri);
+                Console.WriteLine();
+
+                HttpClient client = new HttpClient();
+                var response = client.GetAsync(certAsPemUri).GetAwaiter().GetResult();
+
+                certificateBytes = response.Content.ReadAsByteArrayAsync().GetAwaiter().GetResult();
+
+                Console.WriteLine("   ... success");
+                return true;
```
roncain: Nit: possibly say "successfully read {0} bytes from Bridge".  Then any error happening later won't have "success..." written to the console.  Optional.

src/System.Private.ServiceModel/tools/test/BridgeCertificateBootstrapper/Bootstrapper.cs
```diff
@@ -0,0 +1,97 @@
...
+                Console.WriteLine("  Bridge hostname is: '{0}'", uri.Host);
+                Console.WriteLine("  Retrieving certificate from:");
+                
+                Console.WriteLine(certAsPemUri);
+                Console.WriteLine();
+
+                HttpClient client = new HttpClient();
+                var response = client.GetAsync(certAsPemUri).GetAwaiter().GetResult();
+
+                certificateBytes = response.Content.ReadAsByteArrayAsync().GetAwaiter().GetResult();
```
roncain: Recommend you use the pattern the BridgeClient uses in other places to check the status code and reason, and report what they are if they are not OK.  There might not be content in the case where status != OK.

src/System.Private.ServiceModel/tools/test/BridgeCertificateBootstrapper/Bootstrapper.cs
```diff
@@ -0,0 +1,97 @@
...
+        DisplayBanner(); 
+
+        if (args.Length < 1)
+        {
+            DisplayUsage();
+            return -1;
+        }
+
+        var filename = args[0];
+
```
roncain: Is it possible we might have to do Directory.Create of the directory where this file will be written?  I'm thinking usage might name an outputfile in a non-existent folder.  Alternatively consider testing for existence of the folder before attempting the write.

src/System.Private.ServiceModel/tools/test/BridgeCertificateBootstrapper/Bootstrapper.cs
```diff
@@ -0,0 +1,97 @@
...
+        byte[] certificateBytes; 
+        if (GetCertificate(out certificateBytes))
+        {
+            File.WriteAllBytes(filename, certificateBytes);
```
roncain: Should probably try/catch around the WriteAllBytes and report failures.  One can foresee IO exceptions.

src/System.Private.ServiceModel/tools/test/BridgeCertificateBootstrapper/BridgeCertificateBootstrapper.csproj
```diff
@@ -0,0 +1,36 @@
...
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>BridgeCertificateBootstrapper</RootNamespace>
+    <AssemblyName>BridgeCertificateBootstrapper</AssemblyName>
+    <SignAssembly>false</SignAssembly>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <ProjectGuid>{5AC60BBF-8007-46D8-9BDC-CD137409E45B}</ProjectGuid>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="Bootstrapper.cs" />
+  </ItemGroup>
```
roncain: Nit, possibly a more specific name than "BootStrapper.cs"?  Maybe just CertificateInstaller.cs?

src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateBootstrapper.cmd
```diff
@@ -0,0 +1,25 @@
+echo off
+setlocal
+
```
roncain: Nit: possibly some REM's up front that describe intended usage of this CMD. I usually forget them...

src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateBootstrapper.sh
iamjasonp: Changed - $__outfile == $__cafile, so I just named it the same thing

src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateBootstrapper.sh
iamjasonp: Right now, I mandate it to be set at the script (so you'd run the script with something like `sudo ./BridgeCertificateBootstrapper.sh $BridgeHost ~/tmp/wcf-ca.crt`). The script will not clobber BridgeHost as a new shell is spun up, kind of like `setlocal` in DOS.

I _think_ it will simplify things when we bake this into the build scripts, but I'll make a call on making it optional or not as I bake things in.

src/System.Private.ServiceModel/tools/test/BridgeCertificateBootstrapper/BridgeCertificateBootstrapper.csproj
iamjasonp: Done.

src/System.Private.ServiceModel/tools/test/BridgeCertificateBootstrapper/Bootstrapper.cs
iamjasonp: Good catch - fixed.

src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateBootstrapper.sh
iamjasonp: Ah, another wrinkle - it seems like sudo doesn't preserve environment variables (you have to run `sudo -E command` to preserve the current environment to sudo) 

So basically even if in the user environment, $BridgeHost is set, $BridgeHost is likely not set when running undo sudo

iamjasonp:
Builds will fail until #781 is merged.

iamjasonp:
We can probably stage this - basically commit the tool and start using it, and then integrate this into the build process once we figure out what the integration should look like.

roncain:
LGTM, subject to #781 going in first.  I agree we should stage this.  Get the tool in and available for manual use and then later merge it into the build script.

iamjasonp:
@roncain - I'll commit now without updating the package versions to rc3-23803, and then update all package versions in a subsequent PR; #781 need not go in first now.

PR #782, opened 2016-02-05
Add unpack step to netci.groovy for corefx binaries

iamjasonp:
Our Linux tests were not being run as corefx binaries were not being unpacked from the corefx build.pack. Unpack the artifacts from build.pack so that ./run-tests.sh can run in CI and produce us with test results

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -76,6 +76,7 @@ def project = 'dotnet/wcf'
                 }
             
                 // Unpack
+                shell("unpacker ./corefx/bin/build.pack ./corefx/bin")
                 shell("unpacker ./bin/build.pack ./bin")
                 shell("""
 ./run-test.sh --coreclr-bins \${WORKSPACE}/coreclr/bin/Product/Linux.x64.Release \\


iamjasonp:
@dotnet-bot test ci please

iamjasonp:
@StephenBonikowsky, can you please review this change?

iamjasonp:
... with full knowledge that this will (should) get supplanted by build.sh at some point soon. However at the moment the practice is to build on Windows boxes and then copy to another box to test (corefx does this right now).

roncain:
LGTM, but I don't know how to evaluate it.  @iamjasonp and @StephenBonikowsky -- do you know what if anything is preventing us from moving to build.sh now?

iamjasonp:
It seems like the non-Windows test machines are there for one reason only - to test - this seems by convention in Jenkins. 

One major reason I see offhand why we'd refrain from using the Ubuntu machines to build is that there's 9 in the pool vs 72 for Windows build machines. So our builds would likely take longer (fewer machines), and we'd be hampering others who use the pool significantly by increasing the amount of time we take on those boxes.

roncain:
@iamjasonp -- so you are saying we will always keep run-test.sh around because it is faster?  I see CoreFx's netci.groovy seems to offer both flavors, though I don't know how they are used.  A build of CoreFx is definitely time consuming.

iamjasonp:
@roncain - 

>  I see CoreFx's netci.groovy seems to offer both flavors,

Looks like they made a change more recently than I synced my repo :D Based on my observation though, it seems ilke they only do this for outerloops and native builds - the innerloops still aren't using build.sh and are relying on a build-on-windows-then-copy mechanism.

If we use build.sh, we won't be building corefx - we'll merely be restoring packages from myget for corefx, so I'm not worried about that. 

That said - we aren't doing any Linux outerloops yet - we can cross that Bridge when we get there and there's a bit of work left :) The only reasons why I'm using build.sh really stem from me wanting to be a good citizen and not use a linux machine until necessary (i.e., only the short time needed for an innerloop test).

StephenBonikowsky:
LGTM

iamjasonp:
I'll merge this now. I think the decision to solely use ./build.sh warrants further discussion. Let me open up an issue for that.

PR #787, opened 2016-02-07
Fix PortManager addresses for resource domain

mconnew:
The remote addresses property wasn't being set in the test resource app domain. This fix sets the PortManager.RemoteAddresses value in the new app domain to match what's in the default app domain.


roncain:
LGTM.  Question only: how did this issue manifest itself?  Did some test fail?  Or did you manually observe firewall rules were wrongly formed?

mconnew:
I was trying to test between two hosts in different subnets. The TestResource app domain was using the RemoteAddresses value of LocalSubnet and the firewall rules were only being applied to the Bridge resource manager.

PR #793, opened 2016-02-09
Package S.P.SM as dotnet5.4 and consolidate facade builds.

### LINKED ISSUE #697: Consolidate the facades into a single build as dotnet*.
Investigate just having one build of our façade projects.

### LINKED ISSUE #698: Package System.Private.ServiceModel as dotnet* with a netcore50 specific build for UWP.

StephenBonikowsky:
* Package System.Private.ServiceModel as target framework .Net5.4
* The 5 ServiceModel facades only need to be built once for WcfCore.
* Fixes #698
* Fixes #697


wcf.targets
```diff
@@ -47,4 +47,9 @@
      <WcfToolsOutputPath>$(WcfRootFolder)bin\Wcf\tools\</WcfToolsOutputPath>
   </PropertyGroup>
 
+  <!-- Override ValidateIgnoreReference from Target 'ValidatePackageTargetFramework' in PackageLibs.targets -->
+  <ItemGroup>
+    <ValidateIgnoreReference Include="System.Private.ServiceModel" />
+  </ItemGroup>
+
```
StephenBonikowsky: @ericstj it took me a little while to discover this in PackageLibs.targets to get passed the indirect reference failures. I noticed that System.Private.Uri was added to this collection in the coreclr repo, I believe for Wcf we can just override it safely in our own repo.

It all builds and the packages look correct, could you confirm?

iamjasonp:
Buddy built - LGTM

roncain:
LGTM.  Would love to hear comments from @ericstj but don't block on this, as they can be addressed post-merge.

wcf.targets
ericstj: What error where you getting before you did this?  I'd expect ValidatePackageTargetFramework to walk the graph to S.P.SM and evaluate all of its references to determine if you are setting the right generation.

src/System.ServiceModel.Duplex/src/System.ServiceModel.Duplex.csproj
```diff
@@ -19,10 +19,7 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'net46_Release|AnyCPU'" />
 
   <ItemGroup>
-    <TargetingPackReference Include="mscorlib" Condition="'$(TargetGroup)' != 'netcore50aot'" />
-    <TargetingPackReference Include="System.ServiceModel" Condition="'$(TargetGroup)' == 'net46'" />
-    <ProjectReference Include="$(SourceDir)System.Private.ServiceModel\src\System.Private.ServiceModel.csproj" Condition="'$(TargetGroup)' != 'net46'" >
-      <AdditionalProperties>OSGroup=Windows_NT</AdditionalProperties>
+    <ProjectReference Include="$(SourceDir)System.Private.ServiceModel\src\System.Private.ServiceModel.csproj" >
```
ericstj: Usually project references should match the set of properties defined in the builds file for the project they are referencing.  Since S.P.SM doesn't have an `AnyOS` build, but this project is building as `AnyOS`, this project reference would result in an extra build of S.P.SM that we'd never ship.  Instead you could choose one existing build (EG: the windows_nt one) to treat as the reference assembly for S.P.SM.  And example of this is the change @weshaggard made to System.IO: https://github.com/dotnet/corefx/commit/05a42a3841b706e1bba6cbaeedd630baf3a1a394

wcf.targets
ericstj: Also, ValidateIgnoreReference is only something we want to use for classic assemblies.  It's essentially saying that these assemblies don't follow the rules of generations so don't expect to be able to calculate their generation.  For S.P.SM it should be a normal assembly built on contracts, so you should be able to validate its generation.

src/System.ServiceModel.Duplex/src/System.ServiceModel.Duplex.csproj
```diff
@@ -5,7 +5,7 @@
     <AssemblyName>System.ServiceModel.Duplex</AssemblyName>
     <AssemblyVersion>4.0.1.0</AssemblyVersion>
     <IsPartialFacadeAssembly>true</IsPartialFacadeAssembly>
```
zhenlan: I know it's not the change of this PR, but S.SM.Duplex is _not_ a partial façade assembly.

src/System.ServiceModel.Http/src/System.ServiceModel.Http.csproj
```diff
@@ -5,7 +5,7 @@
     <AssemblyName>System.ServiceModel.Http</AssemblyName>
     <AssemblyVersion>4.1.0.0</AssemblyVersion>
     <IsPartialFacadeAssembly>true</IsPartialFacadeAssembly>
```
zhenlan: S.SM.Http is not a partial façade either. Actually none of our 5 contracts is partial façade. Not sure what impact this property has to the build, so just to point out.

wcf.targets
StephenBonikowsky: It cannot resolve a number of indirect dependencies (listed below) because it could not determine the generation.
Currently I'm setting the PackageTargetFramework to dotnet5.4, these dependencies are all new and not included in the list of assemblies for "5.4" in Generations.json
They should be listed in the section for "5.5" but they are not yet.
The tools package we pull down is already out of date from the file in the [buildtools](https://github.com/dotnet/buildtools/blob/master/src/Microsoft.DotNet.Build.Tasks.Packaging/src/PackageFiles/Generations.json) repo, it looks like they are moving to a new naming convention, and "5.5" will now be "1.4".
Should I open an Issue in the buildtools repo to add these new contracts to "1.4"?

So in the meantime I think we should ignore validation on S.P.SM. 

System.Security.Cryptography.Primitives
System.Security.Principal.Windows
System.Security.Cryptography.X509Certificates
System.Security.Claims
System.Collections.Specialized
System.Collections.NonGeneric
System.Reflection.TypeExtensions
System.Net.Security
System.Xml.XmlDocument
System.Net.WebSockets
System.Net.WebSockets.Client
System.Net.NameResolution
System.Security.Cryptography.Algorithms
System.Reflection.DispatchProxy
System.Net.Http.WinHttpHandler
System.Security.Cryptography.Encoding
System.Runtime.InteropServices.RuntimeInformatio

ericstj: That's happening because the facade projects aren't referencing these libraries, they aren't coming as part of the project reference, and NuGet isn't flowing them because the x-plat NuGet doesn't yet call into msbuild to determn

You could either directly reference all these packages or disable S.P.SM *only* in the facade projects.  Plese don't disable globally or you'll lose all of your validation.

src/System.ServiceModel.Duplex/src/System.ServiceModel.Duplex.csproj
StephenBonikowsky: Hmm, does it work differently on my own box than wherever we build the official packages? I don't see S.P.SM getting built twice on my machine, just once for Windows.

ericstj: I see, you are [defaulting the OSGroup in the S.P.SM project](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj#L4).  In that case it won't build twice but it will be a race condition between the project reference (missing OSGroup global property) and the builds-file reference (with the OSGroup global property).
I've been working on cleaning these up as well as adding a logger to detect them.

src/System.ServiceModel.Http/src/System.ServiceModel.Http.csproj
ericstj: That's OK.  Full facade is just a special case of partial facade that is all type forwards.

src/System.ServiceModel.Duplex/src/System.ServiceModel.Duplex.csproj
StephenBonikowsky: So is the best practice in this case to remove it from S.P.SM and add it to the façade projects when referencing S.P.SM?

ericstj: Yes, as a temporary solution.  You'll be able to remove it once nuget starts flowing the dependencies between project-references.

src/System.ServiceModel.Http/src/System.ServiceModel.Http.csproj
zhenlan: Thanks for the clarification @ericstj.

ericstj: NP.  The name of that property has bugged me too /cc @mellinoe.

src/System.ServiceModel.Duplex/src/System.ServiceModel.Duplex.builds
```diff
@@ -7,13 +7,8 @@
...
     -->
-    <Project Include="System.ServiceModel.Duplex.csproj" >
-      <OSGroup>Windows_NT</OSGroup>
-    </Project>
-    <Project Include="System.ServiceModel.Duplex.csproj" >
-      <OSGroup>Windows_NT</OSGroup>
-      <TargetGroup>netcore50</TargetGroup>
-    </Project>
+    <Project Include="System.ServiceModel.Duplex.csproj" />
+
```
weshaggard: nit: empty line

weshaggard:
LGTM

wcf.targets
StephenBonikowsky: I will add a reference to the S.P.SM package.

StephenBonikowsky:

PR #781, opened 2016-02-05
Update packages to rc3-23809

iamjasonp:

PR #795, opened 2016-02-10
Update build sh and runt-test.sh to use enterprise myget feed

iamjasonp:


iamjasonp:
Note that this is just an additional commit over #781. This is to test commit 3c0d6ea

roncain:
@iamjasonp -- I don't understand the relation between #781 and this PR.  Are you saying this would be used instead of #781?  Merged after it?  It looks like this is logically just #781 plus edits for the .sh scripts, so I'm curious why it didn't just add to #781.  These are only questions -- don't block on it.
LGTM unless you have some other comments you want to include.

iamjasonp:
@roncain - What you said is exactly correct. 

I had to go AFK after committing the sh scripts, and I wanted to be able to merge #781 to master independently in case the scripts for any reason didn't work right. Looks like it has worked fine.

PR #799, opened 2016-02-11
Add +x permissions for init-tools.sh

iamjasonp:
No substantive changes here: 

`chmod a+x init-tools.sh`

So that init-tools.sh doesn't always show up as a change in git after running build.sh


roncain:
LGTM -- thanks for doing this.

Issue #769, opened 2016-02-01
.Net native: ExpectedExceptionTests.DuplexCallback_Throws_FaultException_DirectThrow failed


hongdai:
Running Test: ExpectedExceptionTests.DuplexCallback_Throws_FaultException_DirectThrow
  Assert.True:Expected 'System.ServiceModel.FaultException`1[FaultDetail]' to be thrown, however 'System.ServiceModel.CommunicationException' was thrown.
  Xunit.AssertTestException: Assert.True: Expected 'System.ServiceModel.FaultException`1[FaultDetail]' to be thrown, however 'System.ServiceModel.CommunicationException' was thrown.
     at Xunit.Assert.HandleFail(String assertionName, String message) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 605
     at Xunit.Assert.Throws(Action action, Object[] CS$0$0000) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 76
     at ExpectedExceptionTests.DuplexCallback_Throws_FaultException_DirectThrow(DuplexTaskReturnServiceCallback callbackService) in e:\ProjectK\src\NDP\FxCore\WcfOpen\src\System.Private.ServiceModel\tests\Scenarios\Client\ExpectedExceptions\ExpectedExceptionTests.cs:line 372
     at EntryPointMain.b__9() in e:\ProjectK\src\QA\ToF\PN\x86\rel\IL\FX\Conformance\System.Private.ServiceModel\4.0.0.0\WcfOpen\Scenarios\Client\ExpectedExceptions\Client.ExpectedExceptions.Tests.main.cs:line 22
     at System.Func$1.InvokeOpenStaticThunk()
     at System.ServiceModel.Dispatcher.MessageRpcProcessor.Invoke($MessageRpc rpc)
     at CoreFXTestLibrary.Internal.Runner.RunTestMethod$catch$0() in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Internal\Runner.cs:line 170
  ---- Test FAILED ---------------

hongdai:
Note that running test with only one worker failed as well.

hongdai:
This is caused by a Missing metadata exception:
[...]

PR #811, opened 2016-02-12
Add OperationFault to rd.xml

hongdai:
* This is needed to fix issue #769 from our side.
* Toolchain does not know we need this type and remove it during reduce
dependencies. Adding to rd.xml to explictly tell the toolchain we need it.
* I have verified the rd.xml fix worked by working around the toolchain bug 
189994: Reflection blocking applying to namespaces in unblocked assemblies.

### DIFF
--- a/src/System.Private.ServiceModel/src/Resources/System.Private.ServiceModel.rd.xml
+++ b/src/System.Private.ServiceModel/src/Resources/System.Private.ServiceModel.rd.xml
@@ -59,6 +59,11 @@
           </Method>
         </Type>
       </Namespace>
+      <Namespace Name="System.ServiceModel.Dispatcher">
+          <Type Name="FaultFormatter">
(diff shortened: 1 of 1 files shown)


roncain:
LGTM

src/System.Private.ServiceModel/src/Resources/System.Private.ServiceModel.rd.xml
```diff
@@ -59,6 +59,11 @@
           </Method>
         </Type>
       </Namespace>
+      <Namespace Name="System.ServiceModel.Dispatcher">
+          <Type Name="FaultFormatter">
+            <Type Name="OperationFault{T}" Dynamic="Required All" />
```
roncain: Ah, so this is the syntax for nested classes.  Good to know.  This comment originated from an experimental version of CodeFlow

iamjasonp:
LGTM

PR #827, opened 2016-02-18
Update json files for xplat nuget

StephenBonikowsky:
Includes a fix to a path issue related to newlines in dir.props.

### DIFF
--- a/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
+++ b/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
@@ -1,8 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
-  <PropertyGroup>
-    <Configuration Condition="'$(Configuration)'==''">Windows_Debug</Configuration>
-  </PropertyGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
     <AssemblyName>System.Private.ServiceModel</AssemblyName>
@@ -15,7 +12,7 @@
     <DebugSymbols>true</DebugSymbols>
     <PlatformTarget>AnyCPU</PlatformTarget>
     <ProjectGuid>{9E50E7BF-CD6E-4269-A6DD-59FD0BD6C0FD}</ProjectGuid>
-    <PackageTargetFramework Condition="'$(TargetGroup)' == ''">DNXCore50</PackageTargetFramework>
+    <PackageTargetFramework Condition="'$(PackageTargetFramework)'==''">dotnet5.4</PackageTargetFramework>
     <CommonPath Condition="'$(CommonPath)' == ''">..\..\Common\src</CommonPath>
   </PropertyGroup>
 
--- a/src/System.ServiceModel.Duplex/src/project.json
+++ b/src/System.ServiceModel.Duplex/src/project.json
@@ -1,21 +1,39 @@
 {
+  "dependencies": {
+	"System.Collections.NonGeneric": "4.0.0",
+	"System.Collections.Specialized": "4.0.0",
+	"System.Net.WebSockets": "4.0.0-rc3-23809",
+	"System.Net.WebSockets.Client": "4.0.0-rc3-23809",
+	"System.Reflection.DispatchProxy": "4.0.0",
+	"System.Reflection.TypeExtensions": "4.0.0",
+	"System.Runtime.InteropServices.RuntimeInformation": "4.0.0-rc3-23809",
+	"System.Security.Claims": "4.0.0",
+	"System.Security.Cryptography.Algorithms": "4.0.0-rc3-23809",
+	"System.Security.Cryptography.Encoding": "4.0.0-rc3-23809",
+	"System.Security.Cryptography.Primitives": "4.0.0-rc3-23809",
+	"System.Security.Cryptography.X509Certificates": "4.0.0-rc3-23809",
+	"System.Xml.XmlDocument": "4.0.0"
+	},
   "frameworks": {
-    "dnxcore50": {
-      "dependencies": {
-        "Microsoft.NETCore.Platforms": "1.0.1-rc3-23809",
-        "Microsoft.TargetingPack.Private.CoreCLR": "1.0.0-rc3-23809"
-      }
-    },
-    "netcore50": {
-      "dependencies": {
-        "Microsoft.NETCore.Platforms": "1.0.1-rc3-23809",
-        "Microsoft.TargetingPack.Private.CoreCLR": "1.0.0-rc3-23809"
-      }
-    },
-    "net46": {
-      "dependencies": {
-        "Microsoft.TargetingPack.NETFramework.v4.6": "1.0.1"
-      }
-    }
+	"dnxcore50": {
+		"dependencies": {
+		  "Microsoft.NETCore.Platforms": "1.0.1-rc3-23809",
+		  "System.Net.Http.WinHttpHandler": "4.0.0-rc3-23809",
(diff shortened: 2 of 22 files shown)


StephenBonikowsky:
@dagood could you review the .json forks?
WCF is comprised of 5 full facades, all 5 of which reference the same private implementation.
With this in mind do I also need to fork the .json file for System.Private.ServiceModel?
Also, we already have a fork with an additional .json file under S.P.SM\src\windows for which we have logic in the .csproj file. Will the new logic you added affect this .json file?

To be certain this works in TFS with xplat nugget I was going to create a shelveset there and build but it currently seems to be broken.

src/System.ServiceModel.Duplex/src/project.json
```diff
@@ -1,21 +1,33 @@
...
+	"System.Reflection.DispatchProxy": "4.0.0",
+	"System.Reflection.TypeExtensions": "4.0.0",
+	"System.Runtime.InteropServices.RuntimeInformation": "4.0.0-rc3-23809",
+	"System.Security.Claims": "4.0.0",
+	"System.Security.Cryptography.Algorithms": "4.0.0-rc3-23809",
+	"System.Security.Cryptography.Encoding": "4.0.0-rc3-23809",
+	"System.Security.Cryptography.Primitives": "4.0.0-rc3-23809",
+	"System.Security.Cryptography.X509Certificates": "4.0.0-rc3-23809",
+	"System.Xml.XmlDocument": "4.0.0"
+	},
```
dagood: Why are these not in the `dnxcore50` dependencies section?

dagood:
The forks look reasonable to me, just had one question that I put in as a comment on the diff. I'll have to try restoring/building them tomorrow morning.

I don't think any fork is needed for System.Private.ServiceModel. The detailed log for the restore errors this is fixing lists only these project.jsons having errors:

	Errors in ...src\System.ServiceModel.Duplex\src\project.json
	Errors in ...src\System.ServiceModel.Http\src\project.json
	Errors in ...src\System.ServiceModel.NetTcp\src\project.json
	Errors in ...src\System.ServiceModel.Primitives\src\project.json
	Errors in ...src\System.ServiceModel.Security\src\project.json

I'm not 100% sure about the actual build after restoring, though.

The logic in `dir.props` only happens when a project.json is found in a folder matching the framework name, and even if it did find a project.json in S.P.SM/src/netcore50, the csproj overrides the property anyway. The logic in `dir.props` shouldn't affect S.P.SM.

src/System.ServiceModel.Duplex/src/project.json
StephenBonikowsky: I'll fix and amend the commit.

StephenBonikowsky:
Thanks for the review Davis, let me know how the restore/build goes tomorrow, I will try as well if I can get a successful build.

iamjasonp:
The changes look LGTM
I buddy built on TFS, and it failed for some reasons that don't look related to this; as such, I can't confirm this builds entirely correctly.

roncain:
LGTM but we need to get a clean TFS build with it to ensure no build break.  I'm still trying to get a clean TFS build that passes ToF tests even without this PR.  Then I'll rebuild from scratch with this PR.

roncain:
@dotnet-bot test Innerloop Linux Release Build and Test

roncain:
LGTM.  I was able to buddy-build this in TFS from a clean state (clean /all && sync && build)  and saw no new failures in ToF tests for CoreCLR.  I think we have validated as much as we can and recommend we merge and watch OTG builds.

dagood:
Thanks @roncain. I hadn't been able to run tests yet, but I did get these changes to sync packages and build successfully.

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
```diff
@@ -1,8 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
-  <PropertyGroup>
```
weshaggard: By removing this you are blocking anyone from building this project by itself as without it the configuration will default to "empty" and will either not build successfully or build something different then the default.

roncain: This could be the recent regression issue #838

weshaggard: @roncain I wouldn't expect this to break that scenario as the default configuration should never be in use when building from the root project which uses the builds files.

StephenBonikowsky: @weshaggard I removed this based on @ericstj 's suggestion. See https://github.com/dotnet/wcf/pull/793/files#r52515944

Is this a case of going with the lesser of two evils? What do you advise?

weshaggard: Yeah I couldn't see the comment from @ericstj so I'm not sure what he was suggesting. However having the default is only a convenience thing for folks building in VS or the csproj directly and don't want to always be forced to pass the default configuration. It isn't actually required for the general build system.

roncain: @weshaggard and @ericstj -- curiously, with that line omitted, I had OuterLoops fail consistently on 2 different machines due to having built the product with $(TargetsWindows) false.  This was the command where they failed:
```
  build /p:WithCategories=OuterLoop
```
Yet on @StephenBonikowsky 's machine they passed.  Both of us are using the same freshly synced and cleaned master branch. By adding just this back to S.P.SM.csproj, it passed for me:
```
<PropertyGroup>  
   <Configuration Condition="'$(Configuration)'==''">Windows_Debug</Configuration>  
</PropertyGroup>  
```
Without that line, the only way I could get OuterLoops to succeed was with this command:
```
build /p:Configuration=Windows_NT_Debug /p:OSGroup=Windows_NT /p:WithCategories=OuterLoop
```
We will investigate but it feels like it could be a race condition that gives different results on different machines. It appeared it had built both AnyOS as well as Windows_NT binaries. Neither of my machines where this failed had any unusual Environment variables set.

StephenBonikowsky: WCF test projects have project references to S.P.SM.csproj, the default configuration for these projects was just “Debug” and in their reference to S.P.SM.csproj they did not specify “OSGroup”, therefore; an AnyOS build of S.P.SM.csproj was being triggered.

The race condition comes into play when the individual test projects are being created and the test assembly together with any dependent assemblies are being copied/bin-placed specifically in the wcf\bin\AnyOS.AnyCPU.Debug folder. It looks like for those test folders created while the AnyOS version of S.P.SM has not yet completed building it will find the Windows version already built and copy that to the test folder then once the AnyOS version of S.P.SM has completed building the remaining test folders get that version of S.P.SM.

In Ron’s case a test project with tests of a feature that only exists in the Windows build had the AnyOS version of S.P.SM, in my case (probably because my machine is slower and the AnyOS version of S.P.SM had not finished building) the Windows version of S.P.SM was in that particular test folder.
By the end of the process however both of us had the AnyOS version of S.P.SM in …\wcf\bin\ AnyOS.AnyCPU.Debug\ System.Private.ServiceModel.

I don’t have a solution for the race condition but we can avoid this entirely by adding the default “Windows_Debug” back to the S.P.SM.proj file for which I am creating a PR. In addition, I am updating all of our test projects and other supporting projects with this same default, this way we won’t have an “AnyOS” bin directory at all and it will save a little time when we build locally.

Regarding Eric’s comments in the other PR and whether or not there is another race condition related to having that default setting in S.P.SM we can discuss in the new PR.

weshaggard: You should consider adding the binclash logger (https://github.com/dotnet/corefx/commit/c604732f4a781bc931444641268aa9edd81310fd) to wcf repo to help find any conflicts.

StephenBonikowsky: I will try using it.

PR #829, opened 2016-02-18
Fix ClientCredential.Clone method which missed Windows credentials

iamjasonp:
The initial porting work meant that some methods had functionality stripped to allow for compile. In this case, we stripped out the clone capability for the WindowsClientCredential and neglected to add it back when we were turning this capability back on

This is required for WindowsStreamSecurityUpgradePRovider to to successfully pass our creds along to the underlying NegotiateStream

(a big omission... a simple change... but boy this took forever to hunt down...)


iamjasonp:

roncain:
LGTM

PR #839, opened 2016-02-19
Allow MsBuild properties at build time to flow to the Bridge.

roncain:
This PR makes it possible for Bridge configuration properties
expressed at build time to be built into the test binaries.
And it also allows Build.exe to accept the MSBuild style of
property arguments.

This enables 2 important scenarios for our multi-machine effort.

1. Doing "build /p:BridgeHost=xyz" will generate the appropriate
TestProperties at compile time so that the generated test binaries
carry this information. Then if they are uploaded to Helix to
execute, they already carry the information necessary to communicate
with a Bridge on a different machine.  There is no need to set
environment variables on the execution machine for this.

2. By allowing Bridge.exe to parse the MSBuild style property
arguments, it allows our build scripts to flow Bridge configuration
information to the Bridge startup/ping scripts so they agree with
what the tests expect.

Prior to this PR, it would have been necessary to set environment
variables on both the client and execution machines to have a
shared Bridge configuration.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -63,7 +63,7 @@ set _buildpostfix=^> "%_buildlog%"
 
 if "%outloop%" equ "true"  (
     pushd %setupFilesFolder%
-    call EnsureBridgeRunning.cmd
+    call EnsureBridgeRunning.cmd %*
     set _bridgeReturnCode=!ERRORLEVEL!
     echo EnsureBridgeRunning returned !_bridgeReturnCode!
     popd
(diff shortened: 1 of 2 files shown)


roncain:
Reviewers, I would like to merge this soon because it enables the scenario of submitting a job to Helix pre-compiled to communicate with the Bridge on another machine.  We need this to begin building our automated nightly run scripts.
/cc @iamjasonp @StephenBonikowsky @hongdai @mconnew

hongdai:
LGTM.

PR #845, opened 2016-02-23
Fix test projects to not build AnyOS flavor of S.P.SM

### LINKED ISSUE #750: Remove 'AnyOS' builds from all WCF projects.
The core issue is that if a test project builds as ‘AnyOS’ it will run against an ‘AnyOS’ flavor of WCF Core causing the failures we have seen. This is because of the build & run logic we have inherited from corefx.

Since we don’t support our product building against ‘AnyOS’ we should fix the issue on our side that triggers ‘AnyOS’ flavors to get build. This is most likely caused by test or related projects such as the Common or Infrastructure projects.

One point of clarification. WCF Core built against 'AnyOS' results in a build that works for Linus (Unix), the product forks based on whether it targets Windows.

### LINKED ISSUE #838: Some OuterLoop tests fail using default build.cmd
Using the default 'build.cmd /p:WithCategories=OuterLoop' in Windows, these tests now fail:
HttpsTests.ServerCertificateValidation_EchoString
HttpsTests.ClientCertificate_EchoString
Tcp_ClientCredentialTypeTests.SameBinding_DefaultSettings_EchoString
Tcp_ClientCredentialTypeTests.SameBinding_SecurityModeTransport_EchoString

They fail with:
```
System.PlatformNotSupportedException : Windows Stream Security is not supported (TaskId:6061)
Stack Trace: (TaskId:6061)                              c:\OSS\wcf\src\System.Private.ServiceModel\src\System\ServiceModel\Channels\WindowsStreamSecurityUpgradeProvider.cs(392,0): at System.ServiceModel.Channels.WindowsStreamSecurityUpgradeProvider.WindowsStreamSecurityUpgradeInitiator.OnInitiateUpgradeAsync(Stream stream, OutWrapper`1 remoteSecurity) (TaskId:6061)
                              c:\OSS\wcf\src\System.Private.ServiceModel\src\System\ServiceModel\Channels\StreamSecurityUpgradeInitiatorBase.cs(91,0): at System.ServiceModel.Channels.StreamSecurityUpgradeInitiatorBase.<InitiateUpgradeAsync>d__14.MoveNext() (TaskId:6061)
 ```

They fail because SUPPORTS_WINDOWSIDENTITY is incorrectly set, probably because $(OSGROUP) is AnyOS and $(TargetsWindows) != true.  

The only workaround is to explicitly  do:
```
build.cmd  /p:Configuration=Windows_NT_Debug /p:OSGroup=Windows_NT
```
 which explains why our CI jobs succeeds.

This is a recent regression, and I believe we should ensure build.cmd correctly sets SUPPORTS_WINDOWSIDENTITY so that these tests pass without requiring us to set the Configuration manually.  Otherwise our own developer guide build instructions are incorrect.

StephenBonikowsky:
* Test projects were causing S.P.SM to be built as AnyOS.
* S.P.SM forks based on whether it is a Windows build or not, so this was producing a non-windows build of S.P.SM being used with tests that expect a Windows build of S.P.SM.
* Fixes #750
* Fixes #838

### DIFF
--- a/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
+++ b/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
+  <PropertyGroup>
+    <Configuration Condition="'$(Configuration)'==''">$(OS)_Debug</Configuration>
+  </PropertyGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
     <AssemblyName>System.Private.ServiceModel</AssemblyName>
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/Infrastructure.Common.csproj
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/Infrastructure.Common.csproj
@@ -1,9 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <Configuration Condition="'$(Configuration)'==''">$(OS)_Debug</Configuration>
+  </PropertyGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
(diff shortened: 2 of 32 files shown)


StephenBonikowsky:
Adding default Configuration back into S.P.SM.csproj per feedback from @weshaggard that...

> By removing this you are blocking anyone from building this project by itself as without it the configuration will default to "empty" and will either not build successfully or build something different then the default.

@ericstj you had mentioned previously that keeping it would cause...

> a race condition between the project reference (missing OSGroup global property) and the builds-file reference (with the OSGroup global property).

... you were referring to the project reference to S.P.SM from the 5 façade projects. 

So I'm not sure if I should keep it or remove it?

weshaggard:
I would suggest add the binclash logger to build.cmd as well.

iamjasonp:
@dotnet-bot test Innerloop Windows_NT Debug Build and Test

iamjasonp:
@dotnet-bot test Innerloop Windows_NT Release Build and Test

StephenBonikowsky:
I tried using binclash logger and it did help find a few issues.
I will be adding it to our repo in an upcoming PR.

For this PR, I built and ran on the K enlistment and found no issues after rebasing.
I am going to merge.

PR #850, opened 2016-02-25
Change BridgeCertificateInstaller.sh to sudo as current user

iamjasonp:
We need to make a call to the Bridge as the original user we sudo-ed from, as we need to write to the cert store for the current user, not as root. 

@dotnet-bot skip ci please - we are not equipped to do outerloop Linux tests yet in CI.


roncain:
LGTM.

Issue #592, opened 2015-12-08
Support NegotiateStream after it lights up in dotnet/corefx


iamjasonp:
Currently, System.Net.Security.NegotiateStream is stubbed out but not yet implemented. 

* [x] Windows implementation - dotnet/corefx#1844 
* [x] Linux implementation - dotnet/corefx#2483

iamjasonp:
dotnet/corefx#5156 is the implementation port

iamjasonp:
This was merged, and packages should be out. I'm updating the packages to test this.

iamjasonp:
To do: 

* [x] Update project.json files to package version rc2-23712
* [ ] Update project.lock.json (they're to be deprecated, but while they exist we still have to update them
* [ ] Update the build scripts to have SUPPORTS_WINDOWSIDENTITY be defined
* [ ] Allow TransportSecurity tests with NegotiateStream to run

(Currently, build.cmd targets OSGroup = AnyOS, which prevents `SUPPORTS_WINDOWSIDENTITY` from being defined)

iamjasonp:
We're lit up in Windows! We are waiting for this to be available on Linux.

roncain:
Added label to indicate we are blocked in the Linux implementation by https://github.com/dotnet/corefx/issues/2483

iamjasonp:
I'm privately building dotnet/corefx#5773 and seeing how far we can go.

iamjasonp:
UPDATE: we are now at a point where we seem to be acquiring the Kerberos ticket, and we can see that WCF is doing the right things when calling NegotiateStream.AuthenticateAsClientAsync. However, the connection is still not yet established, and we're working through this one.

iamjasonp:
This seems lit up, as tested with bits that I've manually built from the corresponding corefx NegotiateStream dev branch (https://github.com/iamjasonp/corefx/tree/devNegoUnix). 

Here are the details that need to be ironed out: 

- Testing will work from Linux->Windows only if: 
  - the Bridge is running as LOCAL SERVICE
  - client credentials are specified ahead of test time using `kinit` in Linux
  - or by specifying it explicitly on the factory 

```csharp 
factory = new ChannelFactory<IWcfService>(
    binding,
    new EndpointAddress(new Uri(Endpoints.Tcp_DefaultBinding_Address))
);


factory.Credentials.Windows.ClientCredential.Domain = "redmond";
factory.Credentials.Windows.ClientCredential.UserName = "someUser";
[...]

PR #852, opened 2016-02-25
Turn on build include NegoStream

iamjasonp:
Using this to test build only, please do not merge. 
Testing out #592

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/SpnEndpointIdentity.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/SpnEndpointIdentity.cs
@@ -1,37 +1,31 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.IdentityModel.Claims;
+using System.Runtime;
+using System.Security.Principal; 
+
 namespace System.ServiceModel
 {
     public class SpnEndpointIdentity : EndpointIdentity
-    {
-        private static TimeSpan s_spnLookupTime = TimeSpan.FromMinutes(1);
-        private Object _thisLock = new Object();
-        private static Object s_typeLock = new Object();
-
-        public static TimeSpan SpnLookupTime
-        {
-            get
-            {
-                return s_spnLookupTime;
-            }
-            set
-            {
-                if (value.Ticks < 0)
-                {
-                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentOutOfRangeException("value", value.Ticks,
-                                                    SR.ValueMustBeNonNegative));
-                }
-                s_spnLookupTime = value;
-            }
-        }
-
+    { 
         public SpnEndpointIdentity(string spnName)
         {
             if (spnName == null)
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("spnName");
 
-            throw ExceptionHelper.PlatformNotSupported("SpnEndpointIdentity is not supported");
+            base.Initialize(Claim.CreateSpnClaim(spnName));
+        }
+
+        public SpnEndpointIdentity(Claim identity)
+        {
+            if (identity == null)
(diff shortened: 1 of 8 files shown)


roncain:
@iamjasonp -- do you want review comments?

iamjasonp:
No reviews needed yet. If you see anything flagrant, do feel free to comment, but I'll need to clean up some files, write some tests, etc. first.

iamjasonp:
These bits are not dependent on NegotiateStream being fully working - let's merge these first and not wait. 

@roncain, could you review?

iamjasonp:
Note that there are no tests yet for NegotiateStream itself - this is still dependent on some work in System.Net

iamjasonp:
@dotnet-bot test code coverage please

src/System.Private.ServiceModel/src/System/ServiceModel/SpnEndpointIdentity.cs
```diff
@@ -1,37 +1,31 @@
...
 
+using System.IdentityModel.Claims;
+using System.Runtime;
+using System.Security.Principal; 
+
 namespace System.ServiceModel
 {
     public class SpnEndpointIdentity : EndpointIdentity
-    {
-        private static TimeSpan s_spnLookupTime = TimeSpan.FromMinutes(1);
```
roncain: Question only: is this code being removed because it is dead?  I was comparing to full framework and was curious.

src/System.Private.ServiceModel/src/project.json
```diff
@@ -53,7 +53,8 @@
       "dependencies": {
         "System.Net.NameResolution": "4.0.0-rc3-23809",
         "System.Net.Security": "4.0.0-rc3-23809",
-        "System.Net.Sockets": "4.1.0-rc3-23809"
+        "System.Net.Sockets": "4.1.0-rc3-23809",
+        "System.Security.Principal.Windows": "4.0.0-rc3-23809"
```
roncain: Dumb question, but is S.S.Principal.Windows now supported on *nix?  A few weeks ago this dependency broke WCF and a few other CoreFx libraries.

roncain:
LGTM subject only to the dumb question about the S.Security.Princiipal.Windows dependency.

roncain:
One other question (doesn't block a merge) -- @iamjasonp you said there are no tests for NegotiateStream itself.  Were you saying there are no CoreFx tests for it or no tests for it at all?  I don't understand how this is being tested or what work in System.Net is necessary before these tests can exist.  Could you please elaborate a little?

iamjasonp:
In corefx, everything negotiatestream is currently checked into dev/negotiatestream (i.e. not in master yet)

S.Security.Principal.Windows - this was why I was holding off on committing this, but it seems like it builds in our Linux and Windows environments (see CI). What else should we do to increase our confidence that this won't break everything? :)

src/System.Private.ServiceModel/src/System/ServiceModel/SpnEndpointIdentity.cs
iamjasonp: Yep. In Desktop, we try to map SPNs with Windows Security Identifiers (SIDs) if possible. 
We don't do this in CoreFx, hence dead code here. Thought it'd look nicer :)

roncain:
The dependency on S.Security.Principal.Windows allowed things to build okay, but tests failed because they tried to load a desktop version of that DLL.  It caused the entire test project to fail.  So a test for us would be an OuterLoop pass against only packages.

iamjasonp:
@dotnet-bot test outerloop please, test latest dependencies please

roncain:
LGTM.
I'm having some issues unrelated to these changes trying to buddy build on Linux to test OuterLoops, but since CI passed, I think we should merge.

iamjasonp:
Note that I have not turned on the OuterLoop tests, as they will not pass in a Linux environment.
We need to do those tests by manual testing.

roncain:
LGTM.

PR #858, opened 2016-02-26
Fix FQDNs if the server hosted on a Cloud machine

### LINKED ISSUE #855: fqdn logic in the bridge does not work when the server is an Azure machine
In Azure machine:
   var fqdn = Dns.GetHostEntry("127.0.0.1").HostName;
   var hostname = fqdn.Split('.')[0];

fqdn above is the same as hostname. In order to get fully qualified server name, we need to append domain name, which is cloudapp.net

hongdai:
* On cloud machines, the current logic to get FQDN does not work.
We need to append domain name cloudapp.net.

Fix #855

### DIFF
--- a/src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateManager.cs
+++ b/src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateManager.cs
@@ -36,6 +36,8 @@ namespace WcfTestBridgeCommon
 
         public static EventHandler BridgeConfigurationChanged;
 
+        private const string CloudServerDomain = "cloudapp.net";
+
         // Any change to the Bridge resource folder uninstalls all the certificates
         // installed by those resources.
         public static void OnResourceFolderChanged(string oldFolder, string newFolder)
@@ -202,6 +204,10 @@ namespace WcfTestBridgeCommon
 
                 var fqdn = Dns.GetHostEntry("127.0.0.1").HostName;
                 var hostname = fqdn.Split('.')[0];
+                if (fqdn == hostname)
+                {
+                    fqdn = string.Format("{0}.{1}", fqdn, CloudServerDomain);
+                }
 
                 // always create a certificate locally for the current machine's fully qualified domain name, 
                 // hostname, and "localhost". 
--- a/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/EndpointResource.cs
+++ b/src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/EndpointResource.cs
@@ -15,11 +15,12 @@ namespace WcfService.TestResources
     {
         private const string ResourceResponseUriKeyName = "uri";
         private const string ResourceResponseFQHNKeyName = "hostname";
+        private const string CloudServerDomain = "cloudapp.net";
 
         protected static int SixtyFourMB = 64 * 1024 * 1024;
         private static Dictionary<string, ServiceHost> s_currentHosts = new Dictionary<string, ServiceHost>();
         private static object s_currentHostLock = new object();
-        protected static string s_fqdn = Dns.GetHostEntry("127.0.0.1").HostName;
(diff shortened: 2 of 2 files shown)


hongdai:
test outerloop

src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/CertificateManager.cs
```diff
@@ -36,6 +36,8 @@ public static class CertificateManager
 
         public static EventHandler BridgeConfigurationChanged;
 
+        private const string CloudServerDomain = "cloudapp.net";
```
roncain: We should not hard-code this. If there is no way to determine this dynamically, we should make it a TestProperty and BridgeConfiguration property.  Like BridgeDomain. Then we can pass it to build.cmd with a /p:BridgeDomain=cloudapp.net the same way we pass BridgeHost.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/EndpointResource.cs
```diff
@@ -110,5 +111,17 @@ private Uri BuildUri(ResourceRequestContext context)
             builder.Scheme = Protocol;
             return builder.Uri;
         }
+
+        private static string Getfqdn()
+        {
```
roncain: We'll need to pass BridgeConfiguration in if this becomes a property. So we may not want to cache it as a static.

hongdai: The domain for cloud VMs is extremely unlikely to change as it will break a lot of existing customers. In my opinion, it only make senses to pass around if it's likely to change.

roncain: I don't believe it is correct to assume the Bridge is on an Azure VM in this case. I think this value should be configurable, though we can certainly default to cloudapp.net in TestProperties. We allow configuration of host name, ports, etc. and this feels like the same. If there is a way to determine this value dynamically it would be best, but configurable is a better option if we can't. TestProperties was invented to allow smart defaults with the ability to override via MsBuild or environment variables. I think we should make this a new TestProperties value. I can make a quick PR to make this easier if it helps the multi-machine testing effort. In fact, we could do that PR *after* you merge and don't need to block on the discussion.

hongdai: Dynamically determine the value is the best, unfortunately I haven't found a good way.

Agree merge it for now to unblock the multi-machine lab runs. We can do the necessary improvement later on.

roncain:
LGTM if it works for x-machine. Would you mind opening an issue to make it configurable? You can assign it to me.

src/System.Private.ServiceModel/tools/test/SelfHostWcfService/TestResources/EndpointResource.cs
hongdai: https://github.com/dotnet/wcf/issues/864 is opened to trace the improvement.

zhenlan:
Wouldn't this break current dev experience whoever starts bridge on a workgroup machine?

hongdai:
The current dev experience such as CI is client and server are in the same machine, it uses localhost to validate server certificate. That's why CI outerloop passed.

This is just a temporary fix. I have an issue opened to track necessary improvement.

zhenlan:
No, I am talking about running bridge on one machine (windows, workgroup) and running tests from another (windows or Linux), which is supposed to be working today (before the change). 

Azure is a special environment, we should not make it a default fall back case in our code. I am wondering if we can cheat the resolution of 127.0.0.1 by modifying the hosts file.

hongdai:
I don't think anybody has tried this configure - it's unknown if it's working before.

If cheat the resolution of 127.0.0.1 works, that probably the best. I'll give it a try.

hongdai:
It worked :-) I'm going to revert the changes.

PR #868, opened 2016-03-01
Fix default test timeout

hongdai:
* Currenty test timeout default value is 20 minutes. A test
could hang for 20 minutes before it times out. A typical test
should not run more than 1 minute.


roncain:
LGTM.
Be aware this affects only the tests that consume TestProperties for their timeout.

PR #870, opened 2016-03-02
Automatically build BridgeCertificateInstaller

iamjasonp:


iamjasonp:
@roncain - this should fix the problem you're seeing about running OuterLoops in Linux

roncain:
LGTM.  Thanks.

PR #872, opened 2016-03-03
Syncing build scripts to corefx.

StephenBonikowsky:
* Changes to dir.props and dir.targets made in corefx cause failures in WCF repo.
* Use of Roslyn compiler keeps around the task "VBCSCompiler.exe" which must be manually killed after each local build before running "git clean -xdf"
* There are a few files that either exist in corefx but not in Wcf or vice versa, am not using this PR to investigate those discrepancies.
* For sections where we have WCF specific changes I have added the comments "*** start WCF Content ***" and "*** end WCF Content ***", this should make it easier in the future to identify sections that can safely be ignored when diffing the files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,10 @@ syntax: glob
 ### VisualStudio ###
 
 # Tool Runtime Dir
+# *** start WCF Content ***
+# The syntax used in the corefx version would also ignore WCF changes under "wcf\src\System.Private.ServiceModel\tools"
 /[Tt]ools/
+# *** end WCF Content ***
 
 # User-specific files
 *.suo
@@ -23,6 +26,9 @@ bld/
 [Bb]in/
 [Oo]bj/
 msbuild.log
+
+# Cross building rootfs
+cross/rootfs/
 
 # Visual Studio 2015
 .vs/
--- a/BuildToolsVersion.txt
+++ b/BuildToolsVersion.txt
@@ -1,1 +1,1 @@
-1.0.25-prerelease-00183
+1.0.25-prerelease-00184
--- a/build.cmd
+++ b/build.cmd
@@ -1,12 +1,16 @@
 @echo off
+
+:: *** start WCF Content ***
 setlocal ENABLEDELAYEDEXPANSION
 REM DelayedExpansion needed to allow %ERRORLEVEL% to be set/used multiple times
+:: *** end WCF Content ***
 
 :: Note: We've disabled node reuse because it causes file locking issues.
 ::       The issue is that we extend the build with our own targets which
 ::       means that that rebuilding cannot successfully delete the task
 ::       assembly. 
 
+:: *** start WCF Content ***
 set outloop=false
 set setupFilesFolder=%~dp0src\System.Private.ServiceModel\tools\setupfiles
 
@@ -28,27 +32,50 @@ if '!_FloatingDependencies!'=='0' (
   set _defaultBuildConfig=/p:Configuration=Windows_NT_Debug
   )
 )
+:: *** end WCF Content ***
 
+:ReadArguments
+:: Read in the args to determine whether to run the native build, managed build, or both (default)
+set "__args=%*"
+if /i [%1] == [native] (set __buildSpec=native&&set "__args=%__args:~6%"&&shift&&goto Tools)
+if /i [%1] == [managed] (set __buildSpec=managed&&set "__args=%__args:~7%"&&shift&&goto Tools)
+
+:Tools
+:: Setup VS
 if not defined VisualStudioVersion (
     if defined VS140COMNTOOLS (
         call "%VS140COMNTOOLS%\VsDevCmd.bat"
-        goto :EnvSet
+        goto :Build
     )
 
-    if defined VS120COMNTOOLS (
-        call "%VS120COMNTOOLS%\VsDevCmd.bat"
-        goto :EnvSet
-    )
-
-    echo Error: build.cmd requires Visual Studio 2013 or 2015. 
+    echo Error: build.cmd requires Visual Studio 2015.
+	:: *** start WCF Content ***
     echo        Please see https://github.com/dotnet/wcf/blob/master/Documentation/developer-guide.md for build instructions.
+	:: *** end WCF Content ***
     exit /b 1
 )
 
-:EnvSet
-
+:Build
+:: Restore the Tools directory
 call %~dp0init-tools.cmd
 
+:: Call the builds
+if "%__buildSpec%"=="managed"  goto :BuildManaged
+
+:BuildNative
+:: Run the Native Windows build
+echo [%time%] Building Native Libraries...
+call %~dp0src\native\Windows\build-native.cmd %__args% >nativebuild.log
+IF ERRORLEVEL 1 (
+    echo Native component build failed see nativebuild.log for more details.
+) else (
+    echo [%time%] Successfully built Native Libraries.
+)
+
+:: If we only wanted to build the native components, exit
+if "%__buildSpec%"=="native" goto :eof 
+
+:BuildManaged
 :: Clear the 'Platform' env variable for this session,
 :: as it's a per-project setting within the build, and
 :: misleading value (such as 'MCD' in HP PCs) may lead
@@ -63,6 +90,7 @@ set _binclashlog=%~dp0binclash.log
 set _buildprefix=echo
 set _buildpostfix=^> "%_buildlog%"
 
+:: *** start WCF Content ***
 if "%outloop%" equ "true"  (
     pushd %setupFilesFolder%
     call EnsureBridgeRunning.cmd %*
@@ -70,20 +98,23 @@ if "%outloop%" equ "true"  (
     echo EnsureBridgeRunning returned !_bridgeReturnCode!
     popd
 )
+:: *** end WCF Content ***
 
-call :build %*
+call :build %__args%
 
 :: Build
 set _buildprefix=
 set _buildpostfix=
 echo [%time%] Building Managed Libraries...
-call :build %*
+call :build %__args%
 
 goto :AfterBuild
 
 :build
+:: *** start WCF Content <WCF uses additional msbuild args> ***
 %_buildprefix% msbuild "%_buildproj%" %_defaultBuildConfig% /nologo /maxcpucount /v:minimal /clp:Summary /nodeReuse:false /flp:v=diag;LogFile="%_buildlog%";Append %* "/l:BinClashLogger,%_binclashLoggerDll%;LogFile=%_binclashlog%" %__args% %_buildpostfix%
 set BUILDERRORLEVEL=!ERRORLEVEL!
+:: *** end WCF Content ***
 goto :eof
 
 :AfterBuild
@@ -92,6 +123,7 @@ echo.
 findstr /ir /c:".*Warning(s)" /c:".*Error(s)" /c:"Time Elapsed.*" "%_buildlog%"
 echo [%time%] Build Exit Code = %BUILDERRORLEVEL%
 
+:: *** start WCF Content ***
 :doneBridge
 if "%outloop%" equ "true"  (
     pushd %~dp0bin\wcf\tools\Bridge
@@ -110,4 +142,6 @@ if "%outloop%" equ "true"  (
 )
 
 endlocal
+:: *** end WCF Content ***
+
 exit /b %BUILDERRORLEVEL%
--- a/dir.props
+++ b/dir.props
@@ -1,11 +1,13 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Condition="Exists('..\dir.props')" Project="..\dir.props" />
   <Import Project="src\BuildValues.props" />
-  
+
   <!--
     $(OS) is set to Unix/Windows_NT. This comes from an environment variable on Windows and MSBuild on Unix.
   -->
   <PropertyGroup>
+    <!-- Temp change to make OS X build behave as a Linux build -->
+    <OsEnvironment Condition="'$(OsEnvironment)'=='' AND '$(OS)'=='OSX'">Unix</OsEnvironment>
     <OsEnvironment Condition="'$(OsEnvironment)'==''">$(OS)</OsEnvironment>
   </PropertyGroup>
 
@@ -23,7 +25,7 @@
       <Value>True</Value>
     </AssemblyMetadata>
   </ItemGroup>
-  
+
   <!-- 
     Switching to the .NET Core version of the BuildTools tasks seems to break numerous scenarios, such as VS intellisense and resource designer 
     as well as runnning the build on mono. Until we can get these sorted out we will continue using the .NET 4.5 version of the tasks. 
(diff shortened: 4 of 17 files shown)


StephenBonikowsky:
@mellinoe Would you be aware of any reasons why your changes to dir.props and dir.targets would cause failures for WCF?

The Roslyn change was made by @joperezr (Jose Perez Rodriguez) but while I see his profile I can't seem to link him here. Any ideas?

LICENSE
```diff
@@ -1,21 +1,21 @@
...
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE 
-SOFTWARE.
\ No newline at end of file
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
zhenlan: We got this file from our legal team. It's not a straight copy from corefx. Although the content looks pretty much the same, I will not update it without having it reviewed with our legal team.

StephenBonikowsky: Ok, I'll remove it from the PR, wouldn't it be a bit more efficient if we had one such doc for all our .Net related products on GitHub though...?

.gitignore
```diff
@@ -3,7 +3,7 @@ syntax: glob
 ### VisualStudio ###
 
 # Tool Runtime Dir
-/[Tt]ools/
+[Tt]ools/
```
iamjasonp: Can you double check that this change will match only the root /Tools directory and not any of the underlying tools/ directories? 

For example, we have `~/git/wcf/src/System.Private.ServiceModel/tools` - I think this change will ignore that directory.

BuildToolsVersion.txt
```diff
@@ -1 +1 @@
-1.0.25-prerelease-00182
+1.0.25-prerelease-00182
```
iamjasonp: NIT: whitespace change only, revert (this is probably because I updated the buildtools earlier, #867)

LICENSE
iamjasonp: The explanation I had was that all repos are to be reasonably self-contained, so a LICENSE needs to exist per repo. That is, a LICENSE file should come when you clone a repo and not need to be externally referenced.

build.cmd
```diff
@@ -28,27 +32,48 @@ if '!_FloatingDependencies!'=='0' (
...
 
-    if defined VS120COMNTOOLS (
-        call "%VS120COMNTOOLS%\VsDevCmd.bat"
-        goto :EnvSet
-    )
-
-    echo Error: build.cmd requires Visual Studio 2013 or 2015. 
-    echo        Please see https://github.com/dotnet/wcf/blob/master/Documentation/developer-guide.md for build instructions.
+    echo Error: build.cmd requires Visual Studio 2015.
+    echo        Please see https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/developer-guide.md for build instructions.
```
iamjasonp: New URL points to `corefx`, not `wcf`

build.cmd
```diff
@@ -28,27 +32,48 @@ if '!_FloatingDependencies!'=='0' (
   set _defaultBuildConfig=/p:Configuration=Windows_NT_Debug
   )
 )
+:: *** end WCF Content ***
+
+:ReadArguments
+:: Read in the args to determine whether to run the native build, managed build, or both (default)
+set "__args=%*"
+if /i [%1] == [native] (set __buildSpec=native&&set "__args=%__args:~6%"&&shift&&goto Tools)
```
iamjasonp: We don't have any native libraries to build, right? Won't this cause the script to bail?

build.cmd
```diff
@@ -28,27 +32,48 @@ if '!_FloatingDependencies!'=='0' (
...
-:EnvSet
-
+:Build
+:: Restore the Tools directory
 call %~dp0init-tools.cmd
 
+:: Call the builds
+if "%__buildSpec%"=="managed"  goto :BuildManaged
+
+:BuildNative
```
iamjasonp: Ditto - we have no native bits to build.

build.sh
```diff
@@ -62,7 +56,7 @@ prepare_managed_build()
     if [ ! -e "$__nugetpath" ]; then
         which curl wget > /dev/null 2> /dev/null
         if [ $? -ne 0 -a $? -ne 1 ]; then
-            echo "cURL or wget is required to build WCF. Please see https://github.com/dotnet/WCF/blob/master/Documentation/building/unix-instructions.md for more details."
+            echo "cURL or wget is required to build corefx. Please see https://github.com/dotnet/corefx/blob/master/Documentation/building/unix-instructions.md for more details."
```
iamjasonp: `:s/corefx/WCF/g`

build.sh
```diff
@@ -82,27 +76,35 @@ prepare_managed_build()
...
+    fi
+
+    # Configure environment if we are doing a verbose build
+    if [ $__VerboseBuild == 1 ]; then
+        export VERBOSE=1
+    fi
 }
 
-build_managed_wcf()
+build_managed_corefx()
```
iamjasonp: `build_managed_corefx` or `build_managed_wcf` ? 

(yes, I know it means some differences when diffing between wcf and corefx repos, but this is a little confusing when you look at this repo in isolation)

build.sh
```diff
@@ -112,6 +114,49 @@ build_managed_wcf()
     echo Build Exit Code = $BUILDERRORLEVEL
 }
 
+build_native_corefx()
```
iamjasonp: Should we be keeping this?

build.sh
```diff
@@ -203,25 +301,64 @@ for i in "$@"
...
+    __buildnative=true
+fi
+
+# Disable the native build when targeting Windows.
+
+if [ "$__BuildOS" != "$__HostOS" ]; then
+    echo "Warning: cross compiling native components is not yet supported"
+    __buildnative=false
+fi
+
```
iamjasonp: Check logic of above and tune as needed for WCF

iamjasonp:
@dotnet-bot 
Test Innerloop Windows_NT Debug Build and Test
Test Innerloop Windows_NT Release Build and Test

iamjasonp:
@dotnet-bot Test Innerloop Windows_NT Debug Build and Test

iamjasonp:
@dotnet-bot Test Innerloop Windows_NT Release Build and Test

BuildToolsVersion.txt
roncain: If I recall, any extra whitespace in this file can cause build issues.  Whichever one we pick, we want an exact match with CoreFx.

---
In reply to: [54840341](https://github.com/dotnet/wcf/pull/872#discussion_r54840341) [](ancestors = 54840341)

build.sh
```diff
@@ -2,14 +2,18 @@
 
 usage()
 {
-    echo "Usage: $0 [BuildArch] [BuildType] [clean] [verbose] [platform]"
+    echo "Usage: $0 [managed] [native] [BuildArch] [BuildType] [clean] [verbose] [clangx.y] [platform] [cross]"
+    echo "managed - optional argument to build the managed code"
+    echo "native - optional argument to build the native code"
     echo "The following arguments affect native builds only:"
```
roncain: Echoing comments from @iamjasonp -- we have no native parts to build, so this usage statement is wrong.  I know we're trying to keep these in sync, but we need to keep the -native switch and labels out of our scripts.

@StephenBonikowsky -- have you tried builds that specify these optional arguments like -cross?  We need to verify the new ones being added don't cause build failures.

build.sh
roncain: I'm afraid I agree we should either consciously diverge or make a PR to CoreFx to generalize their names.  I don't think it is widely appreciated how difficult it is becoming to stay in sync with their changes.

build.sh
roncain: No.  We need to expunge the native command line switches and script logic. I'm afraid this has to remain one of the diffs we need to ignore each time we sync.  Or find some creative way to indicate "this is CoreFx logic only" that makes it a NOP in our repo.

build.sh
```diff
@@ -230,23 +367,47 @@ __BinDir="$__rootbinpath/$__BuildOS.$__BuildArch.$__BuildType/Native"
...
 if [ $BUILDERRORLEVEL != 0 ]; then
     exit $BUILDERRORLEVEL
 fi
 
+# Configure environment if we are doing a cross compile.
+if [ "$__CrossBuild" == 1 ]; then
+    export CROSSCOMPILE=1
+    if ! [[ -n "$ROOTFS_DIR" ]]; then
+        export ROOTFS_DIR="$__scriptpath/cross/rootfs/$__BuildArch"
+    fi
```
roncain: We'll need to test this if we keep it.  That means building with -cross and verifying both with and without ROOTFS_DIR set.  This testing could be done after the merge, providing a normal TFS build succeeds.

dir.props
```diff
@@ -1,11 +1,13 @@
...
   <Import Condition="Exists('..\dir.props')" Project="..\dir.props" />
   <Import Project="src\BuildValues.props" />
-  
+
   <!--
     $(OS) is set to Unix/Windows_NT. This comes from an environment variable on Windows and MSBuild on Unix.
   -->
   <PropertyGroup>
+    <!-- Temp change to make OS X build behave as a Linux build -->
+    <OsEnvironment Condition="'$(OsEnvironment)'=='' AND '$(OS)'=='OSX'">Unix</OsEnvironment>
```
roncain: Wow -- I don't know the reason for this temp change in CoreFx, but it looks like it will block attempts to have OSX-specific code, as I believe is needed for crypto.  Don't need to change this PR.  The CoreFx repo needed an issue number here.

run-test.sh
```diff
@@ -230,9 +247,9 @@ runtest()
 
   echo
   echo "Running tests in $dirName"
-  echo "./corerun xunit.console.netcore.exe $testDllName -xml testResults.xml -notrait category=failing -notrait category=OuterLoop -notrait category=$xunitOSCategory" -notrait Benchmark=true
+  echo "./corerun xunit.console.netcore.exe $testDllName -xml testResults.xml -notrait category=failing $OuterLoop -notrait category=$xunitOSCategory" -notrait Benchmark=true
   echo
-  ./corerun xunit.console.netcore.exe $testDllName -xml testResults.xml -notrait category=failing -notrait category=OuterLoop -notrait category=$xunitOSCategory -notrait Benchmark=true
+    ./corerun xunit.console.netcore.exe $testDllName -xml testResults.xml -notrait category=failing $OuterLoop -notrait category=$xunitOSCategory -notrait Benchmark=true
   exitCode=$?
```
roncain: We'll need to buddy build this for OuterLoops on Linux to test this logic

roncain:
Thanks for taking this on @StephenBonikowsky .  I know it is a thankless task but appreciate it being done.  I'd like to settle the issues around getting rid of the -native switch and build logic.  Before we merge, we're going to need to do a full TFS build as well as builds and OuterLoop runs under Linux.

BuildToolsVersion.txt
StephenBonikowsky: I'll make sure that after I rebase it is identical to corefx.

build.cmd
StephenBonikowsky: Thanks.

build.cmd
StephenBonikowsky: True we don't have any native libraries but this is a conditional check right? We don't need a native build therefore we don't specify that argument therefore we don't do a native build?

How do you mean the script will bail?

build.cmd
StephenBonikowsky: One of the purposes for this PR is to reduce the diffs between Wcf and corefx. So even though we don't need any of the native code paths as long as we don't go down those paths our diffs are cleaner.

build.sh
StephenBonikowsky: ty

build.sh
StephenBonikowsky: I'm not sure I agree with this, however I may be missing something. In this case it looks to me like this function is only called if "if $__buildnative;" evaluates to true (see line 396). Our WCF builds should set $__buildnative to false so we shouldn't have to worry about these native code paths and our diffs become much cleaner.

build.sh
StephenBonikowsky: We just need to make sure WCF builds are setting $__buildmanaged to true.

build.sh
StephenBonikowsky: I haven't tried using the script yet.
I think we need to discuss a few things.

StephenBonikowsky:
Agreed, lets discuss.

.gitignore
StephenBonikowsky: You are correct, for now I am reverting this change and adding comments. In later discussions we might consider renaming our tools directories.

StephenBonikowsky:
Regarding the Roslyn compiler issue I opened [https://github.com/dotnet/corefx/issues/6615](url) in the corefx repo.

roncain:
Is the Roslyn compiler used by default?
We should make sure we don't check in a change that exposes us to this issue if we can avoid it.  It sounds like it could put the CI machines into a bad state.

.gitignore
roncain: In the future I think we should correct this in CoreFx because their version is overly restrictive.  The pattern they have there would mean nobody could have a Tools or tools folder in the entire repo -- an unreasonable (and probably unintentional) restriction.

StephenBonikowsky:
It does use it by default, however; if there are any negative impacts on CI machines wouldn't it already be happening on corefx runs?

roncain:
We should not check-in anything that makes a git clean break.  Better to make our repo work properly while they fix the Roslyn bug, even if we diverge.  I recommend we reverse the default until they fix it, unless that breaks for a different reason :smile:

iamjasonp:
NOTE: BuildTools got bumped to -00184 as of a few hours ago

StephenBonikowsky:
I synced to corefx one more time, the only new diff was the build tools version as @iamjasonp indicated.

StephenBonikowsky:
CI tests seem to be gone....?

iamjasonp:
@dotnet-bot test this please

iamjasonp:
womp womp... I think there was a CI outage. It looks like the triggers are working once again :smile:

build.cmd
roncain: We chatted and agreed to stay as close to CoreFx as possible with optionally some comments explaining this.

iamjasonp:
Linux builds will fail spectacularly until you remove the leading `sh` command from https://github.com/dotnet/wcf/blob/master/build.sh#L97

build.cmd
iamjasonp: ACK

StephenBonikowsky:
The PR for the sh files is #878 and there are many updates to our repo's version after syncing to corefx including the item you mentioned.

I'm thinking that we should just merge both PRs.

PR #878, opened 2016-03-04
Updating WCF version of .sh files

StephenBonikowsky:
* Wrapping all sections where we differ with the corefx version with a WCF start/end comment so it is easier to see expected differences.
* Will open either Issue or PR in corefx to make some content be more generic and not specifically say "corefx" unless necessary.
* Will open another PR in corefx to update their myget feed links to use the enterprise feed.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,15 +1,23 @@
 #!/usr/bin/env bash
 
+# This file is leveraged from the corefx report
+# Some content such as anything related to building "native"
+# Does not apply to WCF and is therefore a no op
+
 usage()
 {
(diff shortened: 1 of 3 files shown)


StephenBonikowsky:
Not sure how much testing we want to do beyond the Linux CI tests to validate these scripts before merging?

Keep in mind that these scripts may have a dependency on the new tools version that is updated as part of PR #872. I am hoping to merge that soon, but if you test these changes you will likely need that PR as well.

roncain:
I think you might need to include BuildToolsVersion.txt to build successfully in Linux.  I see CI failed for a package restore that seems like what I am seeing in the current WCF master build on Linux.

PR #881, opened 2016-03-07
Change BuildCertificateInstaller.sh to chmod downloaded cert

iamjasonp:
The downloaded cert was in some circumstances not chmodded to world writable, which meant that cURL would have trouble accessing the certificates when trying to establish an HTTPS connection

Not sure why this wasn't happening on my main dev box, but this problem exists on another box I spun up. 

@roncain, this should fix the problem you've been seeing when running outerloops on Linux and seeing the HTTP Security.TransportSecurity.Tests tests fail. 

You may need to remove all certs inside `/usr/local/share/ca-certificates/*` and re-run the BridgeCertificateInstaller in order to have this take. Also, please check the directory permissions on `/usr/local/share/ca-certificates` to ensure it's world-readable

skip ci please

### DIFF
--- a/src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateInstaller.sh
+++ b/src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateInstaller.sh
@@ -32,6 +32,7 @@ install_root_cert()
 {
     __cafile=$1
 
+    chmod a+r "$__cafile"
     cp -f "$__cafile" /usr/local/share/ca-certificates
 
     if [ $? -ne 0 ]; then 


roncain:
Thanks @iamjasonp. My /usr/local/share/ca-certificates is not currently world readable and is currently owned by user:root, group:staff.  What code is responsible for creating this, and why do I need to manually chmod it?  I don't understand if it a point-in-time issue where it was created with older code or whether we should chmod it in the script.

roncain:
LGTM.
The question I asked about world-readable was to understand why this manual step is required.

iamjasonp:
Hmm - this is how mine looks today: 

ls -l /usr/local/share
```
drwxr-sr-x 2 root staff 4096 Mar  7 17:41 ca-certificates
```

It may not be necessary to manually chmod the ca-certificates directories (if you are able to `sudo update-ca-certificates`, and it succeeds, it should be good enough). However, the cert itself must be world-readable, otherwise cURL is not able to read the cert itself (update-ca-certificates symlinks into these cert files in /usr/local/share/ca-certificates). 

If you manually run cURL and you get: 
`curl: (35) error:0200100D:system library:fopen:Permission denied`

or running tests, you get something like 
`SSL peer certificate or SSH remote key was not OK`

You are probably hitting the problem seen here. 

Additional reading: http://serverfault.com/questions/411009/ubuntu-10-04-4-lts-curl-ssl-certificate-permission-denied

iamjasonp:
> My /usr/local/share/ca-certificates is not currently world readable and is currently owned by user:root, group:staff. What code is responsible for creating this, and why do I need to manually chmod it? I don't understand if it a point-in-time issue where it was created with older code or whether we should chmod it in the script.

No, I don't think we should chmod in the script... this should have been set up by default in a Linux distro setup to be in a state that works. I'm in general leery of doing a chmod on anything I don't own.

PR #884, opened 2016-03-07
Remove one of the duplicated commandline args

hongdai:
* commit e284227f820e2d0eb34d2b640e13acf71a7997e3 add
set "__args=%*", make __args the same as %*. thus we
pass commandline args twice to msbuild. This breaks
Helix, as Helix target does not allow two responses files
specified, even thougth they are the same
* The fix is to remove the redundancy.


hongdai:
test outerloop.

StephenBonikowsky:
LGTM

Issue #129, opened 2015-06-30
Reenable Fiddler-based URL's for Bridge using TestProperties


roncain:
The Bridge feature #115 will undo the UseFiddler conditional code. After that feature is merged, we need to update the point at which URL's are returned (Endpoints.cs?) to check TestProperties (#120) and conditionally substitute Fiddler URL's

roncain:
Not strictly required for beta7, but when we move to the Bridge we will lose the ability to use Fiddler to debug unless we fix this.  Fix needs to work with Bridge's returned URL's.

zhenlan:
@roncain is this already done?

iamjasonp:
Hmm, is this item strictly necessary? I guess i typically go and attempt to test cross-machine when I need Fiddler, and so if I allow Fiddler to take over my default HTTP(S) proxy, this problem goes away. 

A while back (I forget how long) I had luck passing build.cmd /p:BridgeHost=<my local machne's fqdn name> and was able to have Fiddler intercept traffic.

roncain:
@mconnew I think you were the main user of this before we moved to the Bridge and killed the old code.  Are you able to do what you want without this feature just by setting BridgeHost to another machine and run the tests cross-machine?  I wonder if we should close this issue and remove the Bridge and TestProperties that refer to this setting (no code does anything with it).

roncain:
@mconnew -- do you still need this feature?  If not, we should remove UseFiddlerUrl from TestProperties and the BridgeConfiguration.  They currently are inoperative anyway.

mconnew:
No we have cross machine working, it's less useful. I'm okay to remove them. If I find a need again in the future, I'll put them back.

roncain:
Closing issue because the old Bridge infrastructure has been removed, and we can easily use cross-machine testing when we want to use Fiddler, without the need for special URL handling.

PR #888, opened 2016-03-08
Code generation of TestProperties no longer uses CodeTaskFactory

roncain:
The CodeTaskFactory feature is not supported in the CoreCLR
version of MsBuild.  This resulted in build failure on Linux
when we attempted to code-gen TestProperties building
tests/Common/Infrastructure.

This PR eliminates the use of CodeTaskFactory and massively
simplifies code generation to a single CDATA section.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
@@ -1,74 +1,67 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <!-- 
-      TestProperty:
-        This item collection gives the name and value for each property to generate
-        into TestProperties.Properties. The included ItemSpec for each item will
-        become the property's key.  The value for the property will be taken from
-        the "Value" metadata, or if it is blank, from the "DefaultValue" metadata.
-        This pattern permits properties to specify their defaults here but to be
-        overridden from the command line.
-        Example: 'msbuild /p:UseFiddlerUrl=true' will override the default false.
-    -->
-    <ItemGroup>
-      <TestProperty Include="BridgeResourceFolder">
-        <Value>$(BridgeResourceFolder)</Value>
-        <DefaultValue>$(OutputPath)</DefaultValue>
-      </TestProperty>
-      <TestProperty Include="BridgeHost">
-        <Value>$(BridgeHost)</Value>
-        <DefaultValue>localhost</DefaultValue>
-      </TestProperty>
-      <TestProperty Include="BridgePort">
-        <Value>$(BridgePort)</Value>
-        <DefaultValue>44283</DefaultValue>
-      </TestProperty>
-      <TestProperty Include="BridgeHttpPort">
-        <Value>$(BridgeHttpPort)</Value>
-        <DefaultValue>8081</DefaultValue>
-      </TestProperty>
-      <TestProperty Include="BridgeHttpsPort">
-        <Value>$(BridgeHttpsPort)</Value>
-        <DefaultValue>44285</DefaultValue>
-      </TestProperty>
-      <TestProperty Include="BridgeTcpPort">
-        <Value>$(BridgeTcpPort)</Value>
-        <DefaultValue>809</DefaultValue>
-      </TestProperty>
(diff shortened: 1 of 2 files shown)


src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
```diff
@@ -1,74 +1,67 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <!--
```
roncain: Reviewers: to help understand this change... The old way was a CodeTaskFactory that scanned an item collection and pulled out metadata and wrote it to a file.  It was expected TestProperties would grow and that this would be easier to maintain.  But TestProperties is relatively stable, and it became far easier to just replace all that complexity with a CDATA section that would expand MsBuild properties and write to disk.  Future additions to TestProperties will require updating this CDATA section.

roncain:

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.targets
```diff
@@ -1,115 +1,85 @@
...
+    SanitizeTestProperties:
+      This target ensures the test property values are sanitized for proper
+      evaluation at runtime.
+  -->
+  <Target Name="SanitizeTestProperties">
+    <!-- BridgeResourceFolder needs to convert forward slashes to backward
+         slashes to be usable on Windows, where the Bridge runs.
+    -->
+    <PropertyGroup Condition="$(BridgeResourceFolder.Contains('/'))">
+      <BridgeResourceFolder>$(BridgeResourceFolder.Replace('/', '\'))</BridgeResourceFolder>
```
iamjasonp: Is this strictly necessary? I've been able to run things on the command line using forward slashes, but I agree it's better to sanitize rather than not.

iamjasonp: (I'm just a little leery of any escaping that needs to happen when you use backslashes, though it should be avoided since you already prefixed the above string literal with an @ )

roncain: Truthfully I don't know all the places that fail without this.  I know I can't copy and paste these mixed-slash paths from msbuild.log and use them reliably in CMD windows.  The old code used to do this programmatically, so I brought it forward for convenience.  The old code also did a lot more Unicode sanitization, but I think we can get away without it.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
```diff
@@ -1,74 +1,67 @@
...
+      <BridgeCertificatePassword>test</BridgeCertificatePassword>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(BridgeCertificateValidityPeriod)' == ''">
+      <BridgeCertificateValidityPeriod>7.00:00:00</BridgeCertificateValidityPeriod>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(UseFiddlerUrl)' == ''">
+      <UseFiddlerUrl>false</UseFiddlerUrl>
+    </PropertyGroup>
```
iamjasonp: Do we still use this? Consider using this opportunity to remove this if it's not in active use (see #129)

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.targets
roncain: Right -- I added the '@' to avoid the need to escape.  So far that is working for my testing of this.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
roncain: Correct, but I'd like to avoid mixing up PR's with orthogonal purposes, especially because they propagate to changes in the BridgeProperties.  I can make a separate PR to remove this without much trouble.

iamjasonp: Makes sense

iamjasonp:
LGTM

roncain:
Testing done -- Windows inner and outer, Linux inner and outer.
In progress -- ToF CoreCLR, but blocked on ToF Net Native.

PR #892, opened 2016-03-08
Checkin Helix lab run scripts

hongdai:
* Init check in the scripts used in the Helix lab run

### DIFF
--- a/src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
+++ b/src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
@@ -1,0 +1,88 @@
+REM
+REM This script sync to the latest WCF source, build and submit Helix lab runs.
+REM Arguments:
+REM %1: Lab scripts location
+REM %2: WCF local git source location
+REM %3: Response files location
+
+@echo off
+setlocal
+
+set BridgeHost=
+pushd %2
+pushd bin\wcf\tools\Bridge
+if [%ERRORLEVEL%]==[0] (
+    call Bridge.exe -stop
+    popd
+)
+
+#REM remove bin directory to ensure a rebuild
+call git clean -xdf
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while cleanning Git.
+    exit /b
+  )
+
+REM sync
+git checkout master
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while checking out master.
+    exit /b
+  )
+git pull
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while syncing master.
+    exit /b
+  )
+call startbridge.cmd  /allowRemote /remoteAddresses:*
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while starting the bridge.
+    exit /b
+  )
+REM replaceBuildNumber.ps1 will update build number for all response files.
+copy /y %1\* .
+copy /y %3\* .
+powershell.exe  %1\replaceBuildNumber.ps1
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while updating build number.
+    exit /b
+  )
+set BridgeHost=Wcfcoresrv4.cloudapp.net
+@echo kick off mac debug
+call build.cmd   @osx_debug.rsp
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while kicking off mac debug.
+    echo Continue submitting the rest of the jobs
+  )
+@echo kick off mac release
+call build.cmd   @osx_release.rsp
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while kicking off mac release.
+    echo Continue submitting the rest of the jobs
+  )
+@echo kick off windows debug
+call build.cmd  @windows_debug.rsp
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while kicking off windows debug.
+    echo Continue submitting the rest of the jobs
+  )
+@echo kick off window release
+call build.cmd  @windows_release.rsp
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while kicking off windows release.
+    echo Continue submitting the rest of the jobs
+  )
+@echo kick off Linux debug
+call build.cmd   @%ubuntu_debug.rsp
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while kicking off linux debug.
+    echo Continue submitting the rest of the jobs
+  )
+@echo kick off Linux release
+call build.cmd   @ubuntu_release.rsp
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while kicking off Linux release.
+  )
+
+popd
+endlocal
--- a/src/System.Private.ServiceModel/tools/test/HelixLabRun/replaceBuildNumber.ps1
+++ b/src/System.Private.ServiceModel/tools/test/HelixLabRun/replaceBuildNumber.ps1
@@ -1,0 +1,8 @@
+
+$buildStr = Get-Date -UFormat "%Y%m%d%H%M"
+$buildStr = $buildStr.Substring(3)
+$Files = Get-ChildItem *.rsp
+foreach ($templateFile in $Files)
+{
+    (get-content  $templateFile) | foreach-object {$_ -replace "tempBuildNumber", $buildStr} | set-content $templateFile
+}


src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,41 @@
+REM
```
iamjasonp: Please 
```
@echo off 
setlocal 
```

and 

```
endlocal
``` 

at the end 

If you're going to show status messages, please explicitly `echo` them rather than relying on REM to show the comments.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,41 @@
+REM 
+REM This script sync to the latest WCF source, build and submit Helix lab runs.
+REM Arguments:
+REM %1: Lab scripts location
```
iamjasonp: If the location of the lab scripts location is the same as this script, you can use %~dp0 instead of specifying it on the command line

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,41 @@
...
+REM %2: WCF local git source location
```
iamjasonp: Shouldn't this just be %~dp0\..\..\..\..\ ? Again, better to not specify it if it's possible, we've made assumptions on the location of these scripts already

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,41 @@
...
+
+set BridgeHost=
+cd %2
```
iamjasonp: Consider `pushd` and `popd` - someone running the script should not have their current working directory change after script runs.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,41 @@
...
+call build.cmd   @osx_debug.rsp
+REM kick off mac release
+powershell.exe  %1\replaceBuildNumber.ps1 -templateFile osx_release.rsp
+call build.cmd   @osx_release.rsp
+REM kick off window debug
+powershell.exe  %1\replaceBuildNumber.ps1 -templateFile window_debug.rsp
+call build.cmd  @window_debug.rsp
+REM kick off window release
+powershell.exe  %1\replaceBuildNumber.ps1 -templateFile window_release.rsp
+call build.cmd  @window_release.rsp
```
iamjasonp: NIT: Should be `windows` not `window` :smile:

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,41 @@
...
+del *.rsp
+
+REM sync to the latest
+git checkout master
+git pull
+
+REM start the bridge
+call startbridge.cmd  /allowRemote /remoteAddresses:*
+
+copy /y %1\*.rsp .
```
iamjasonp: Are there some files you haven't committed yet?

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,41 @@
...
+REM ensure bridge is stopped
+call stopbridge.cmd
+REM remove bin directory to ensure a rebuild
+del /s /q bin
+del *.rsp
+
+REM sync to the latest
+git checkout master
+git pull
+
```
roncain: For a script like this, I recommend checking ERRORLEVEL after every single operation.  If any step fails (like stopping the Bridge or the git pull, the script should report that and stop.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,41 @@
...
+REM %1: Lab scripts location
+REM %2: WCF local git source location
+
+set BridgeHost=
+cd %2
+REM ensure bridge is stopped
+call stopbridge.cmd
+REM remove bin directory to ensure a rebuild
+del /s /q bin
+del *.rsp
```
roncain: Why not 'git clean -xdf' instead of deleting bin and *.rsp?  If this script does a git pull, there is a good chance it may find newer package dependencies.  So I think it is cleaner to delete all the temporary build artifacts and ensure a clean state of the enlistment.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
roncain: I think the lab scripts may come from a different repo or be in a private location on the machine, so I don't think %~dp0 will work.  However, this script should definitely check that %1 has been set and points to a folder that contains what it expects (with an appropriate error if it does not).

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
roncain: I'm a little confused too, so maybe some REM's in the script will help explain.  Am I correct in assuming you expect this CMD to live in the git repo but be executed from some location other than the git repo folder?  I can understand how that might be required for a 'git pull' to succeed when this script changes.  It also means (I think) that any changes to this script need to be manually copied to a different location on the machine.

Here's another approach to consider.  Create a script that gets copied to %temp% and is started there.  That script could then do the 'git clean' and 'git pull' of the repo (without needing a %2 to know where it lives) and then calls back into the "real" launch script living in the repo.  That would allow the 'git pull' to retrieve a new copy of the "real" launch script and would not require us to manually maintain a separate folder.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: Will do.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: To be flexible, we should not add the assumption the lab script location is the same as this script.

@roncain, My goal is to keep the script simple and small so it's easier to read and modify. This script is only used for Helix lab automation by a schedule task, not used by Dev or CI experience. That's why I intentionally leave checking parameters out. I'm open to add it though if you feel strongly we should do it.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: Again it's for flexibility. I don't want to assume particular locations, so scripts can live anywhere.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: Agreed.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: Agreed.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: Agreed. Will do.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: rsp files cannot be committed as it contains azure account information.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: Will change.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: If adding more parameters can simplify script logic, I'd prefer adding more parameters. It's mainly used by a scheduled task, not human.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
roncain: This is probably worth discussing.  I think the goal should be that we are not required to manually update the lab machine as this script changes.  If we have a simple "bootstrap" script that only knows how to 'git clean' and 'git pull' and then executes a script in our repo to launch the builds, we will have a self-maintaining machine.  Think through the scenario of changing the contents of this CMD file -- what are the steps you would need today to update the lab machine to use it? With my proposal, the 'git pull' will get the new copy by default.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
roncain: Agreed -- for now.  But I think we will very soon want to allow humans to run a subset of this script to do trial runs of PR's.  We don't need to design for that now, but it is worth keeping in mind.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
roncain: Let's take the discussion to a different thread regarding the "git sync bootstrap" issue.  It does not need to block merging this PR.  I'm just waiting to see the changes for the other feedback before signing off.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
iamjasonp: Question only: where do they come from, and why should they sit in %1?

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
iamjasonp: Agreed. Consider (later) optional parameters, for example

roncain:
@hongdai -- what is the status of this PR?  I think @iamjasonp and I were waiting for the next iteration before signing off.

hongdai:
I wanted to get corefx input on what needs to be done for creating a WCF tab. Now I have all the information for next iteration.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: The template comes from VSO private Repro. I keep our own ones in both Server machine and my own machine, not shared. The plan is to copy them to the same location of %1 on the server machine.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+REM %2: WCF local git source location
+REM %3: Response files location
+
+@echo off
+setlocal
+
+set BridgeHost=
+pushd %2
+pushd bin\wcf\tools\Bridge
+if [%ERRORLEVEL%]==[0] (
```
roncain: The 2nd pushd will put current directory in the wrong place for the git clean.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+    call Bridge.exe -stop
+    popd
```
roncain: We have a StopBridge.cmd script.  I recommend you use that for consistency.  It avoids this logic to pushd to the Bridge bin folder too.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+)
+
+#REM remove bin directory to ensure a rebuild
+call git clean -xdf
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while cleanning Git.
+    exit /b
```
roncain: Nit: 'cleaning' instead of 'cleanning'

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+pushd %2
+pushd bin\wcf\tools\Bridge
+if [%ERRORLEVEL%]==[0] (
+    call Bridge.exe -stop
+    popd
+)
+
+#REM remove bin directory to ensure a rebuild
+call git clean -xdf
+if NOT [%ERRORLEVEL%]==[0] (
```
roncain: Recommend use clean.cmd

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+    exit /b
+  )
+call startbridge.cmd  /allowRemote /remoteAddresses:*
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while starting the bridge.
+    exit /b
+  )
+REM replaceBuildNumber.ps1 will update build number for all response files.
+copy /y %1\* .
+copy /y %3\* .
```
roncain: Copying * into our repo root is a little dangerous.  If there is a file not in our .gitignore, there could be difficulties with them appearing as new files that need to be added.  And if they have a file of the same name, it will overwrite what is in our repo (imagine if they had a build.cmd). I recommend we instead copy them into a separate folder (like bin\WCFTemp) and access them from there.  Then we know a clean.cmd will remove them, and there is no danger of an overwrite.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
hongdai: This is the same logic as stopbridge.cmd except I only pop if push has been successful(push could be fail if the tools\Bridge has not been built yet.)

We will need to change to use IIS/Self Hosted service once it's ready.  I'd suggest we don't spend more time on this PR in the meat time. I will cancel the PR if no objection.

roncain: Thanks @hongdai .Yes, please, *cancel* this PR if you don't want us reviewing it.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
+REM
+REM This script sync to the latest WCF source, build and submit Helix lab runs.
+REM Arguments:
+REM %1: Lab scripts location
+REM %2: WCF local git source location
+REM %3: Response files location
+
+@echo off
+setlocal
```
iamjasonp: NIT `@echo off` should be at the beginning of the script, and you should be showing usage info only when you need to (for example, not enough params passed) to prevent cluttering up any console output

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+
```
iamjasonp: Check that (at least) the correct number of args have been passed in

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+    exit /b
+  )
+
+REM sync
+git checkout master
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while checking out master.
+    exit /b
+  )
+git pull
```
iamjasonp: You should never be doing git pull operations as this does a git fetch and then a git merge. You will put the repo into a weird state (i.e., a state where they will no longer track origin/master and upstream/master) by doing this. 

It is _reasonably_ safe if you choose to do this on a lab box, but still not recommended. 
You're better off doing a git fetch then git merge --ff-only to ensure that there's no merge commits in your repo.

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while syncing master.
+    exit /b
```
iamjasonp: Recommend you exit /b with an errorlevel as well.

iamjasonp: Actually, no, you shouldn't be doing this; you never end up popd-ing. You should have a label at the end of the script (e.g., :done) where you popd then exit /b %captured_errorlevel%

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+  )
+REM replaceBuildNumber.ps1 will update build number for all response files.
+copy /y %1\* .
+copy /y %3\* .
+powershell.exe  %1\replaceBuildNumber.ps1
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while updating build number.
+    exit /b
+  )
+set BridgeHost=Wcfcoresrv4.cloudapp.net
```
iamjasonp: Should this be a param passed by command line?

src/System.Private.ServiceModel/tools/test/HelixLabRun/launchWCFOpenBuilds.cmd
```diff
@@ -0,0 +1,88 @@
...
+call build.cmd   @%ubuntu_debug.rsp
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while kicking off linux debug.
+    echo Continue submitting the rest of the jobs
+  )
+@echo kick off Linux release
+call build.cmd   @ubuntu_release.rsp
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while kicking off Linux release.
+  )
```
iamjasonp: NIT for all the above if close parens: align with `if`

You also don't need @echo, you've @echo off already - just echo.

PR #893, opened 2016-03-09
Implement supported parts of Spn/UpnEndpointIdentity

iamjasonp:
Reopening this since #852 caused a build break in .NET Native

Issue #851, opened 2016-02-25
Determine how to test NegotiateStream with explicit credentials specified


iamjasonp:
I would like a test case that allows us to test NegotiateStream + Kerberos with explicit credentials, for example, specified like this: 

``` csharp
factory = new ChannelFactory<IWcfService>(
    binding,
    new EndpointAddress(new Uri(Endpoints.Tcp_DefaultBinding_Address))
);


factory.Credentials.Windows.ClientCredential.Domain = "someDomain";
factory.Credentials.Windows.ClientCredential.UserName = "someUser";
factory.Credentials.Windows.ClientCredential.Password = "somePassword";
```

We need to 
1. figure out if there's any way to do testing without a domain joined server machine
2. figure out how to share secrets among machines so that they can be tested
3. set it all up automagically for automated testing

zhenlan:
Automation of this might not be simple. For now, we will add this to be part of our manual test guidance https://github.com/dotnet/wcf/blob/master/Documentation/manualtest-guide.md.

mconnew:
There's a new Azure service in preview which might help. [Azure Active Directory Domain Services](https://azure.microsoft.com/en-us/services/active-directory-ds/)

iamjasonp:
Tried this. This works! But it's not nicely automatable since it's still in preview. I suppose that when this is released for real, we might be able to provision this via API or something. 

In the meantime, I think we will have to make do with manually testing this.

StephenBonikowsky:
I am removing this Issue as the ActiveIssue used to skip these tests and instead adding the ConditionaFact 'Domain_Joined', if/when this is solved the tests will start running.

hongdai:
We now have a once per week manual run to cover it.

PR #898, opened 2016-03-10
NegotiateStream tests

iamjasonp:
Looking for some early feedback and discussion. Only review the NegotiateStream_*_Tests.cs files, please. 

I'm waiting for #893 to go in before rebasing this one. For some strange reason the rebasing is being weird, so you'll see a bunch of additional files. 

**Discussion:** 

* We need to figure out our story for how to specify usernames/passwords reasonably dynamically - this will probably involve using TestProperties so that they can be specified on the command line. 
* Furthermore, I would like to remove the [ActiveIssues] - rather, I would like to use some mechanism (like TestProperties or a ConditionalFact) to specify whether or not these tests should be run and should pass
* For example, one set of tests requires that the Bridge be run as LOCAL SYSTEM. This may require the Bridge to report on the context it's running as. 

This is the set of tests I am trying to do: 

|Host running as|targetidentity passed|Network credential|Expected Result|Actual|
|---|---|---|---|---|
|testuser|UPN -> realuser@DC.DOMAIN.COM|DC\realuser|Expected fail|Blocked (UPN missing)|
|testuser|UPN -> testuser@DC.DOMAIN.COM|DC\testuser|Expected pass | Blocked (UPN missing)|
|testuser|SPN -> mymachine|DC\testuser|Expected fail | Blocked (Assert) |
|LOCAL SYSTEM|UPN -> testuser@DC.DOMAIN.COM|DC\testuser|Expected fail | Fail|
|LOCAL SYSTEM|SPN -> mymachine|DC\testuser|Expected pass | Pass|
|LOCAL SYSTEM|SPN -> mymachine|DC\realuser|Expected pass | Pass |
|LOCAL SYSTEM|SPN -> mymachine|(none) |Expected pass | Pass|
|User (Server needs SPN reg'n)|SPN -> bridge/mymachine |(none / any valid) |Expected Pass | (Not passing - under investigation) |
|LOCAL SYSTEM (Server needs SPN reg'n)|SPN -> bridge/mymachine|(none / any valid) |Expected Pass | (Under investigation -> next ) |

Note too that there will need to be additional pivots on the server - that is: 

* Server supports Kerberos auth, falling back to NTLM auth if necessary
* Server supports only Kerberos auth
* Server supports only NTLM auth
* Server supports neither

However, I posit that the server pivots are going to be pri 2 while client pivots are going to be pri 1

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.cs
@@ -1,0 +1,301 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Infrastructure.Common;
+using System;
+using System.Security.Cryptography.X509Certificates;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using System.ServiceModel.Security;
+using Xunit;
+
+public static class NegotiateStream_Http_Tests
+{
+    // The tests are as follows:
+    //
+    // NegotiateStream_*_AmbientCredentials
+    //     Windows: This should pass by default without any code changes
+    //       Linux: This should not pass by default 
+    //              Run 'kinit user@DC.DOMAIN.COM' before running this test to use ambient credentials
+    //              ('DC.DOMAIN.COM' must be in capital letters) 
+    //              If previous tests were run, it may be necessary to run 'kdestroy -A' to remove all
+    //              prior Kerberos tickets
+    // 
+    // NegotiateStream_*_With_ExplicitUserNameAndPassword
+    //     Windows: Edit the s_UserName and s_Password variables to a user valid on your Kerberos realm
+    //       Linux: Edit the s_UserName and s_Password variables to a user valid on your Kerberos realm
+    //              If previous tests were run, it may be necessary to run 'kdestroy -A' to remove all
+    //              prior Kerberos tickets
+    // 
+    // NegotiateStream_*_With_ExplicitSpn
+    //     Windows: Edit the s_Spn variable to match a valid SPN for the server 
+    //       Linux: Edit the s_Spn variable to match a valid SPN for the server 
+    //   
+    //     By default, the SPN is the same as the host's fully qualified domain name, for example, 
+    //     'host.domain.com'
+    //     On a Windows host, one has to register the SPN using 'setspn', or run the process as LOCAL SYSTEM
+    //     by using a tool like psexec and running 'psexec -s -h <WcfBridge.exe>' 
+    // 
+    // NegotiateStream_*_With_Upn
+    //     Windows: Edit the s_Upn field to match a valid UPN for the server in the form of 
+    //              'user@DOMAIN.COM'
+    //       Linux: This scenario is not yet supported - dotnet/corefx#6606
+    //
+    // NegotiateStream_*_With_ExplicitUserNameAndPassword_With_Spn
+    //     Windows: Edit the s_Spn variable to match a valid SPN for the server
+    //              Edit the s_UserName and s_Password variables to a user valid on your Kerberos realm
+    //       Linux: Edit the s_Spn variable to match a valid SPN for the server
+    //              Edit the s_UserName and s_Password variables to a user valid on your Kerberos realm
+    // 
+    // NegotiateStream_*_With_ExplicitUserNameAndPassword_With_Upn
+    //     Windows: Edit the s_Upn variable to match a valid UPN for the server
+    //              Edit the s_UserName and s_Password variables to a user valid on your Kerberos realm
+    //       Linux: This scenario is not yet supported - dotnet/corefx#6606
+
+    private const string s_Domain = "domain";
+    private const string s_Username = "azureuser";
+    private const string s_Password = "hunter2";
+    private const string s_Spn = "host";
+    private const string s_Upn = "azureuser@DOMAIN.COM";
+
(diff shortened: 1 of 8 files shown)


roncain:
Agree about TestProperties and things like passwords and user names.  This will allow them to flow into the binaries and run from other machines as well, without being passed in the clear.  Technically we could also leverage this in ToF using environment variables to set the TestProperties.  I have mixed feelings about using [ConditionalFact] vs xunit -notrait to condition which tests run, but lean slightly toward [ConditionalFact] for this usage because the tests will appear as "skipped" in the logs (as opposed to being silently skipped with -notrait).  We can also make [ConditionalFact] smart so that it autodetects when it can but also allows a TestPropertyt to drive it when it cannot. Worth a discussion because whichever we choose will set a precedent.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.cs
```diff
@@ -0,0 +1,301 @@
...
+
+public static class NegotiateStream_Http_Tests
+{
+    // The tests are as follows:
+    //
+    // NegotiateStream_*_AmbientCredentials
+    //     Windows: This should pass by default without any code changes
+    //       Linux: This should not pass by default 
+    //              Run 'kinit user@DC.DOMAIN.COM' before running this test to use ambient credentials
+    //              ('DC.DOMAIN.COM' must be in capital letters)
```
roncain: Should we automate this kinit and kdestroy on Linux?

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.cs
```diff
@@ -0,0 +1,301 @@
...
+    //     Windows: This should pass by default without any code changes
+    //       Linux: This should not pass by default 
+    //              Run 'kinit user@DC.DOMAIN.COM' before running this test to use ambient credentials
+    //              ('DC.DOMAIN.COM' must be in capital letters) 
+    //              If previous tests were run, it may be necessary to run 'kdestroy -A' to remove all
+    //              prior Kerberos tickets
+    // 
+    // NegotiateStream_*_With_ExplicitUserNameAndPassword
+    //     Windows: Edit the s_UserName and s_Password variables to a user valid on your Kerberos realm
+    //       Linux: Edit the s_UserName and s_Password variables to a user valid on your Kerberos realm
```
roncain: I like your proposal to use TestProperties, and think all the instructions that say "edit xyz..." should just call out the TestProperty name.  We could also use [ConditionalFact] to skip these tests automatically if the TestProperties required are not specified.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Tcp_Tests.cs
```diff
@@ -0,0 +1,297 @@
...
+using Infrastructure.Common;
+using System;
+using System.Security.Cryptography.X509Certificates;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using System.ServiceModel.Security;
+using Xunit;
+
+public static class NegotiateStream_Tcp_Tests
+{
```
roncain: Question only: I recently added under the TransportSecurity tests a number of tests that use TransferMode.Streamed.  I did this because when I enabled NetTcp streaming, I noticed code coverage gaps in stream-specific areas in StreamedFramingRequestChannel. Could I ask you to look at code coverage for that class and decide whether there are still any gaps that might be filled if you added NegotiateStream tests that used TransferMode.Streamed? It can be a different PR if we see a need for it.  It looks like Buffered vs Streamed take different code paths that are related to transport security.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.cs
iamjasonp: I'm generally leery of doing this as it has effect across the entire user session (we can't isolate this to happen only during a run)

If we know the password, we should be able to automate kinit, and if we want tests to run one after another, then this may force our hand to kdestroy after each test happens. We may have to force such tests to run in sequence.

iamjasonp:
@roncain - could you please take a look at the TestProperties changes? For some reason, the TestProperties are not being read in if I build. There's probably something really obvious that I missed. 

I've turned on the sanity test for Windows, but keeping all tests off otherwise until we figure out automation. Given that we're in discussion to deprecate the Bridge for another solution, I will not spend additional time attempting to automate this for now. #851 tracks the investigation to automate.

iamjasonp:

iamjasonp:
Btw, to clarify - the scenarios above have been verified manually on Ubuntu.

roncain:
LGTM.
I ran both inner and outer loops on Windows but have not verified on Ubuntu.
Regarding the TestProperties not being read in -- I verified that msbuild.log has a TestProperties property that contained all your changes @iamjasonp .  And I verified that the generated file (see $(GeneratedTestPropertiesFileName) in msbuild.log) looked correct.  One thing to consider is adding to the common\Infrastructure\tests some verification that there are at least non-empty entries in TestProperties for the new names.  These tests verify code-gen was successful without trying to verify the actual values.

iamjasonp:
@dotnet-bot test Innerloop Windows_NT Debug Build and Test

iamjasonp:
@dotnet-bot test Innerloop Windows_NT Release Build and Test, test Innerloop Windows_NT Debug Build and Test

PR #883, opened 2016-03-07
Enable AllowCookies=true.

shmao:
Invoking operations on a channel that has AllowCookies = true currently fails with:

> Unhandled Exception: System.InvalidOperationException: When using CookieUsePolicy.UseSpecifiedCookieContainer, the CookieContainer property must not be null.

The root cause is that we didn't assign a CookieContainer to the winhttp handler.

Fix #853.

PR #899, opened 2016-03-10
Fix an Issue caused by PR #883.

shmao:
PR #883 breaks one Tof test project because the PR used C#6 syntax.


shmao:
I changed the test to use [Fact] and verified that the test than could run in Net Core/Native. But I changed it back to use [Theory] because,
1. [Theory] seems to be the right attribute to use for the case in OSS;
2. changing it to [Fact] requires either duplicating the test or adding extra control logic in the test, which I think would complicate the test itself;
3. currently there are 60 tests being marked with [Theory], I don't think we should treat this test differently to make this particular test work in Net Native. 

I'm merging the PR to unblock the TOF test.

roncain:
Unfortunately this means we have 60 tests not being run against NET Native.   I agree it is not the job of this PR to fix that, but everyone needs to be aware we have a large test gap in NET Native now.  And it is growing.

PR #897, opened 2016-03-09
Restoring build scripts back to using Roslyn.

StephenBonikowsky:
* After removing the WCF dependency on the old version of MSBuild via Mono which was not supported in Roslyn we can now remove the temporary hack we had in place that allowed us to continue using the old version of MSBuild.

PR #903, opened 2016-03-11
Fix .NET Native build of S.P.SM due to mscorlib reference.

StephenBonikowsky:
* The NET Native assembly for System.private.ServiceModel can't resolve the references in Windows.winmd, because they refer to mscorlib.
* Adding a reference to Microsoft.NETCore.Portable.Compatibility, which has an mscorlib façade in it fixes the problem.


StephenBonikowsky:
This will fix an issue introduced by #897

roncain:
I can buddy build TFS tomorrow but can't currently test it. It would be great to know TFS builds clean but if the build is already broken, I think you need to either revert or merge.

StephenBonikowsky:
Discussed with @zhenlan and believe we should merge, I just checked the Jenkins' logs and can confirm that the issue is gone from the Windows checks, just waiting for the Linux checks.

PR #905, opened 2016-03-11
Fix AllowCookies logic to not set cookie container to null

roncain:
The recent PR #883 for AllowCookies added code to assign the
cookie container to null when AllowCookies was false.  But on
NET Native, this causes an ArgumentNullException.  This caused
all Http OuterLoop tests to fail.

The fix is to set the cookie container only when AllowCookies is true.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
@@ -273,7 +273,11 @@ namespace System.ServiceModel.Channels
                     var clientHandler = GetHttpMessageHandler(to, clientCertificateToken);
                     clientHandler.AutomaticDecompression = DecompressionMethods.Deflate | DecompressionMethods.GZip;
                     clientHandler.UseCookies = _allowCookies;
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
LGTM

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -273,7 +273,11 @@ private HttpCookieContainerManager GetHttpCookieContainerManager()
                     var clientHandler = GetHttpMessageHandler(to, clientCertificateToken);
                     clientHandler.AutomaticDecompression = DecompressionMethods.Deflate | DecompressionMethods.GZip;
                     clientHandler.UseCookies = _allowCookies;
-                    clientHandler.CookieContainer = _allowCookies ? _httpCookieContainerManager.CookieContainer : null;
+                    if (_allowCookies)
+                    {
+                        clientHandler.CookieContainer = _httpCookieContainerManager.CookieContainer;
```
mconnew: This will still break actual cookie usage in UWP apps. We need to fix this more generally rather than avoid triggering the issue in the general case. I'm okay with this as a short term fix.

mconnew:
LGTM as a short term fix. Still need to revisit this.

shmao:
LGTM.

PR #906, opened 2016-03-11
Fat package conversion of System.Private.ServiceModel

chcosta:
This change enables the Unix OSGroup, disables build filtering, and creates a fat package instead of split packages for the System.Private.ServiceModel library.

This change is in conjunction with the other fat package conversions in CoreFx - https://github.com/dotnet/corefx/commit/be08afd9701053f5f9369762db58b87c8ca9559f

/cc @weshaggard, @ericstj


roncain:
I've buddy built on Windows and Linux, and all looks good.  I haven't done a TFS full build but we can watch rolling builds.

StephenBonikowsky:
Spoke with Chris, just to be clear, after this change we will no longer have two separate runtime packages as we have had for win7 and unix, now we will just have on S.P.SM package that contains all the needed assets for the supported OSes.

StephenBonikowsky:
Also, I am hoping that this change will also resolve the issue with System.IO.Compression that has prevented us from advancing to a newer package reference version.
[(https://github.com/dotnet/corefx/issues/6603#issuecomment-195145593)]

chcosta:
@StephenBonikowsky , correct.  You'll still have the same S.P.SM identity package, but that package will contain the win7 and unix runtime assets instead of a runtime.json file referencing those split runtime packages.

StephenBonikowsky:
Great, thanks for the confirmation Chris.

Issue #890, opened 2016-03-08
Disable tests that are known to fail in NET Native


roncain:
We currently have a dozen or so tests that fail in NET Native for reasons we understand.  We need to silence these failures but continue to track them as active issues.  Most are currently marked with the Test Care Failure label.

Recommend conditional #if FEATURE_NETNATIVE be placed around an [ActiveIssue] in front of all these tests, specifying the appropriate active issue.  The condition should be true only when we build in the ToF world today and does not imply we will maintain 2 separate sets of tests for netcore50 and dnxcore50.  If and when these run in Helix against NET Native, we can revisit how to set FEATURE_NETNATIVE when building the test binaries.

StephenBonikowsky:
Fixed by #917

PR #907, opened 2016-03-11
Disable tcp secure streaming tests on Linux

roncain:
The recent enabling of NegotiateStream in WCF Linux code
caused all the existing tcp secure streaming tests to fail
on Linux because CoreFx has not yet checked in the
implementation of NegotiateStream.

After discussion with @iamjasonp we decided to disable these
tests for the same issue being addressed by PR #898.  Once
CoreFx publishes a working NegotiateStream, we will either
reactivate these tests or fold them into the new NegotiateStream
tests in #898.


zhenlan:
LGTM as we only disable these tests for *nix.

Will UWP pick up these tests too (expect failure)?

StephenBonikowsky:
:+1:

roncain:
These do fail for UWP and are new candidates for the #if FEATURE_NETNATIVE [ActiveIssue] change recommended in #890.

iamjasonp:
(after the fact LGTM)

PR #910, opened 2016-03-11
Updating some packages to use stable versions.

StephenBonikowsky:
* We want to use as many stable versions of packages as possible in our default building environment.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/project.json
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/project.json
@@ -1,9 +1,9 @@
 {
   "dependencies": {
     "Microsoft.NETCore.Platforms": "1.0.1-rc3-23915",
-    "System.Collections.Concurrent": "4.0.10.0",
+    "System.Collections.Concurrent": "4.0.10",
     "System.ComponentModel.EventBasedAsync": "4.0.10",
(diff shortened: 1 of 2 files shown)


roncain:

src/Common/test-runtime/project.json
```diff
@@ -30,7 +30,7 @@
     "System.Net.WebSockets": "4.0.0-rc3-23809",
     "System.Net.WebSockets.Client": "4.0.0-rc3-23809",
     "System.ObjectModel": "4.0.0",
-    "System.Reflection": "4.0.10",
+    "System.Reflection": "4.0.0",
```
zhenlan: Since 4.0.10 system.reflection is also stable version, I think it's ok to stay with it. Similarly for 4.0.20 system.runtime. Do we have a build issue if we stick with these versions?

iamjasonp:
LGTM subject to an answer to @zhenlan's question above.

src/Common/test-runtime/project.json
StephenBonikowsky: Thinking through this I agree.
We are currently working in a Gen 1.3 + world so we should use the stable released version for Gen 1.3 unless required to move to a pre-release version due to development changes.

In this case v4.0.10 is the stable version for Gen 1.3 while v4.0.0 is the stable version for Gens 1.0/1.1/1.2

These were updated correctly in PR #914 so after re-basing this PR these changes are gone.

zhenlan:
LGTM

Issue #902, opened 2016-03-10
Decide how SpnEndpointIdentity.SpnLookupTime property setter should behave


roncain:
PR #901 added back the SpnLookupTime to SpnEndpointIdentity because it was part of the public contract.  But it was added only as an auto-property to un-break a TFS build.  We need to decide whether it is preferable for the setter to (a) be a NOP, (b) behave like the desktop, or (c) throw PNSE

iamjasonp:
After looking I'm thinking we might as well implement this in CoreFx like we have on Desktop. It's a very lightweight implementation, so I see no harm in implementing it. 

The only internal usage I can find is http://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/SpnEndpointIdentity.cs,110 

... it's in a method which we are unable to support in CoreFx, as it attempts to map a given SPN to a Windows SID.

roncain:
@iamjasonp -- I agree that the replicating the behavior of the property setter from the full framework seems best, even though we have no code to consume that property now.

PR #913, opened 2016-03-15
Implement SpnEndpointIdentity.SpnLookupTime

iamjasonp:
SpnEndpointIdentity.SpnLookupTime was not used previously as it wasn't used; it ended up causing a TFS build break, so an AutoProperty was implemented to unbreak the build. 

In #902, we have a discussion about what the real CoreFx implementation should look like. This PR shows what it looks like if we just took the desktop implementation (my preferred option) 

Note that the SpnLookupTime API is still unused, but this is essentially a harmless addition.


iamjasonp:
(Please continue discussion in #902. Do not merge until discussion is concluded)

roncain:
LGTM.  I agree with the opinion @iamjasonp gives in #902.

Issue #832, opened 2016-02-19
PlatformNotSupportedException: Windows Stream Security is not supported


StephenBonikowsky:
|Failing Test Cases |
|--------------------------|
|**Tcp_ClientCredentialTypeTests.SameBinding_DefaultSettings_EchoString**|
|**Tcp_ClientCredentialTypeTests.SameBinding_SecurityModeTransport_EchoString**|

StephenBonikowsky:
Call Stack

  Running Test: Tcp_ClientCredentialTypeTests.SameBinding_DefaultSettings_EchoString
  Caught Unexpected exception:System.PlatformNotSupportedException: Windows Stream Security is not supported
     at System.ServiceModel.Channels.WindowsStreamSecurityUpgradeProvider.WindowsStreamSecurityUpgradeInitiator.OnInitiateUpgradeAsync(Stream stream, OutWrapper`1 remoteSecurity)
     at System.ServiceModel.Channels.StreamSecurityUpgradeInitiatorBase.d__14.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     at System.ServiceModel.Channels.ConnectionUpgradeHelper.d__3.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     at System.ServiceModel.Channels.ClientFramingDuplexSessionChannel.d__10.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     at System.ServiceModel.Channels.ConnectionPoolHelper.d__15.MoveNext()
  --- End of stack trace from previous location where exception was thrown ---
     at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #917, opened 2016-03-16
Adding ActiveIssues for tests failing only on NET Native

StephenBonikowsky:
* By wrapping the ActiveIssue in an #if #endif based on FEATURE_NETNATIVE we should be able to only skip these test on NET Native and not on NET Core.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
@@ -75,7 +75,11 @@ public static class WebSocketTests
 
     [Fact]
     [OuterLoop]
+#if !FEATURE_NETNATIVE
     [ActiveIssue(420, PlatformID.AnyUnix)]
+#else
+    [ActiveIssue(526)]
+#endif
     public static void WebSocket_Http_Duplex_BinaryStreamed()
     {
         NetHttpBinding binding = null;
@@ -167,7 +171,11 @@ public static class WebSocketTests
 
     [Fact]
     [OuterLoop]
+#if !FEATURE_NETNATIVE
     [ActiveIssue(420, PlatformID.AnyUnix)]
+#else
+    [ActiveIssue(526)]
+#endif
     public static void WebSocket_Https_Duplex_BinaryStreamed()
     {
         BinaryMessageEncodingBindingElement binaryMessageEncodingBindingElement = null;
@@ -262,9 +270,13 @@ public static class WebSocketTests
 
     [Fact]
     [OuterLoop]
+#if !FEATURE_NETNATIVE
     [ActiveIssue(4429)] // In #CoreFX
     [ActiveIssue(470)]
     [ActiveIssue(420, PlatformID.AnyUnix)]
+#else
+    [ActiveIssue(526)]
+#endif
     public static void WebSocket_Https_Duplex_TextStreamed()
     {
         TextMessageEncodingBindingElement textMessageEncodingBindingElement = null;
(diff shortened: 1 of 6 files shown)


StephenBonikowsky:
This PR needs to be validated on an N from K run.
I'm building my K enlistment and will update here once I have been able to validate the changes.
Anyone who has a recent and valid K build can feel free to try as well.

roncain:
I'm trying this out with ToF NET Native now

roncain:
Would you mind adding these new tests to that list? ActiveIssue is #832.  They are under Security\TransportSecurity\Tcp\StreamingTests.cs


    StreamingTests.NetTcp_TransportSecurity_StreamedRequest_RoundTrips_String
    StreamingTests.NetTcp_TransportSecurity_StreamedResponse_RoundTrips_String
    StreamingTests.NetTcp_TransportSecurity_Streamed_RoundTrips_String
    StreamingTests.NetTcp_TransportSecurity_Streamed_TimeOut_Long_Running_Operation
    StreamingTests.NetTcp_TransportSecurity_Streamed_Async_RoundTrips_String
    StreamingTests.NetTcp_TransportSecurity_StreamedRequest_Async_RoundTrips_String
    StreamingTests.NetTcp_TransportSecurity_StreamedResponse_Async_RoundTrips_String
    StreamingTests.NetTcp_TransportSecurity_Streamed_RoundTrips_String_WithSingleThreadedSyncContext
    StreamingTests.NetTcp_TransportSecurity_Streamed_Async_RoundTrips_String_WithSingleThreadedSyncContext

StephenBonikowsky:
I'll add those as well Ron, I can confirm that it works as expected.
I am currently checking on some additional failures I am getting to see if they are legitimate or not.

StephenBonikowsky:
Verified all of this works on NET Native.
Had to add the #if !FEATURE_NETNATIVE because otherwise the tests would still run and then fail, test logic only seems to pay attention to the first ActiveIssue.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
```diff
@@ -161,7 +161,9 @@ public static void SameBinding_Soap12_EchoString()
     }
 
     [Fact]
-    [ActiveIssue(544, PlatformID.AnyUnix)]
+#if FEATURE_NETNATIVE
+    [ActiveIssue(544)] // Server certificate validation not supported in NET Native
+#endif
```
zhenlan: Do we expect this test will pass on *nix now?

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -75,7 +75,12 @@ public static void WebSocket_Http_RequestReply_BinaryStreamed()
 
     [Fact]
     [OuterLoop]
+#if !FEATURE_NETNATIVE
     [ActiveIssue(420, PlatformID.AnyUnix)]
+#endif
+#if FEATURE_NETNATIVE
+    [ActiveIssue(526)]
+#endif
```
zhenlan: You can use #if...__#else__...#endif instead.

roncain:
LGTM.
I also verified on NET Native, and it looks good.  I can't verify on Linux because we are currently broken running on Linux due to System.Private.Uri.  This should be fixed in CoreFx via https://github.com/dotnet/corefx/blob/be08afd9701053f5f9369762db58b87c8ca9559f/src/System.Private.Uri/pkg/unix/System.Private.Uri.pkgproj#L11 but we need to wait for new packages to be build.  Don't block the merge based on this.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
StephenBonikowsky: Not sure, @mconnew would know. The issue implies that it works on Windows but not Unix, there is a later comment that says it is now supported on UWP.
If Matt fixed it and it now works on Unix he should verify it and remove the ActiveIssue?

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: Done, amending the PR with this update.

PR #919, opened 2016-03-17
Syncing to the latest version of the Clean script.

StephenBonikowsky:
* The script was updated to check for and kill the long running VBCSCompiler.exe task before continuing cleaning tasks.

### DIFF
--- a/clean.cmd
+++ b/clean.cmd
@@ -12,6 +12,7 @@ if [%1] == [] (
 set clean_targets=
 set clean_src=
 set clean_tools=
+set clean_environment=
 set clean_all=
 set clean_successful=true
 
@@ -45,9 +46,15 @@ if /I [%1] == [/t] (
   goto Next
 )
 
+if /I [%1] == [/e] (
+  set clean_environment=true
+  goto Next
+)
+
 if /I [%1] == [/all] (
   set clean_src=
   set clean_tools=
+  set clean_environment=
   set clean_targets=Clean;CleanPackages;CleanPackagesCache;
   set clean_all=true
   goto Begin
@@ -61,9 +68,10 @@ shift /1
 goto Loop
 
 :Begin
+if /I [%clean_environment%] == [true] (
+  call :CleanEnvironment
+)
 
-echo Running init-tools.cmd
-call %~dp0init-tools.cmd
 
 if /I [%clean_src%] == [true] (
   echo Cleaning src directory ...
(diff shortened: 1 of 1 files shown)


clean.cmd
```diff
@@ -125,4 +139,11 @@ if NOT [%ERRORLEVEL%]==[0] (
...
   set clean_successful=false
 )
+exit /b
+
+:CleanEnvironment
+REM If VBCSCompiler.exe is running we need to kill it
+echo Stop VBCSCompiler.exe execution.
+echo. >> %cleanlog% && echo Stop VBCSCompiler.exe execution. >> %cleanlog% 
+for /f "tokens=2 delims=," %%F in ('tasklist /nh /fi "imagename eq VBCSCompiler.exe" /fo csv') do taskkill /f /PID %%~F >> %cleanlog%
 exit /b
```
StephenBonikowsky: Yes I know, but that's how they checked it in to corefx, sigh. Trying to keep the diffs clean.

StephenBonikowsky:
So for those of us that used "git clean -xdf", it looks to me like the closest thing is just "clean /all".

roncain:
LGTM. Thank you @StephenBonikowsky .
Observation for future: we may want to use clean.cmd in our CI and nightly build scripts

zhenlan:
LGTM

@roncain I don't know if we run clean on CI today. Did you see issues with CI due to not clean? Also apparently, clean.cmd is windows only so it cannot run on Linux CI.

roncain:
@Zhenlan -- I think we may want to consider running clean.cmd on Windows and clean.sh on Linux CI's, and possibly nightly builds.  It looks like our repo is already up to date for clean.sh.  I haven't seen CI breakage personally, but VBCSCompile breaks my day-to-day workflow every few minutes, so it's easy to imagine it will get us sooner or later.  Plus, a centralized clean script will always be where new logic goes, so it would be better to rely on the "blessed" clean script and not duplicate code. Example: the clean on TFS also cleans out some private NuGet caches.  I think it's time to embrace the workflow scripts.

iamjasonp:
I've seen things break before due to this issue when running back-to-back CI runs. 

We don't _typically_ see CI breakage typically because we both: 
1. don't run CI often enough
2. run in a pool of machines rather than just one

Script LGTM. We can take on incorporating clean.cmd to CI at a later time. We may also need to consider incorporating this into the under-construction git sync web service

PR #924, opened 2016-03-21
Add PR sync scripts to repo for CI

iamjasonp:
These scripts are for allowing the CI system to call out to a WCF sync service for outerloop tests. This is not incorporated into the CI system yet, as the relevant pieces on the test server/client are not yet done. 

skip ci please

### DIFF
--- a/src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
+++ b/src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
@@ -2,23 +2,24 @@
 setlocal
 
 REM 
-REM This calls a web service, specified on the command line, which syncs a 
-REM remote server's repo to the correct PR. 
+REM This notifies a the test web service, specified on the command line,
+REM that a test is about to start
 REM This script is meant to be called from a GitHub Pull Request. 
 REM 
 
 set __args=%*
 
 echo.
-echo  WCF Remote Github Repo sync script
+echo  WCF Test Service Pull Request sync script
 echo. 
 
-if '%__args%'=='' (
+if "__args%"=="" (
     echo   [%~n0] A URL must be specified for the pull request synchronization URL
-    echo     Usage:  %~n0 [sync url] [pr-number (optional)]
+    echo     Usage:  %~n0 [sync url] [optional pr-number]
     echo   sync-url  - URL on remote server for PR synchronization
     echo   pr-number - PR number to sync to
     echo.
+    echo   Example:  %~n0 http://wcfci-sync-server/repo/sync.ashx 
     echo   Example:  %~n0 http://wcfci-sync-server/repo/sync.ashx 404
     set __EXITCODE=1
     goto done
@@ -26,7 +27,7 @@ if '%__args%'=='' (
 
 set __SYNC_HOST_URL=%1
 
-if '%2'=='' (
+if "%2"=="" (
     if '%ghprbPullId%'=='' (
         echo   [%~n0] This script should only be called only from the context of a GitHub Pull Request
         echo   [%~n0] 'ghprbPullId' environment variable not set
@@ -38,7 +39,7 @@ if '%2'=='' (
 ) else (
     set __PR_ID=%2
     echo   [%~n0] WARNING: This script should usually only be called only from the context of a GitHub Pull Request
-    echo   [%~n0] PR ID overridden: '%__PR_ID%'
+    echo   [%~n0] PR ID overridden: '%2'
 ) 
 
 set __REQUEST_URL=%__SYNC_HOST_URL%?pr=%__PR_ID%
--- a/src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
+++ b/src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
@@ -1,0 +1,53 @@
+@echo off 
+setlocal
+
+REM 
+REM This calls a web service, specified on the command line, which syncs a 
+REM remote server's repo to the correct PR. 
+REM This script is meant to be called from a GitHub Pull Request. 
+REM 
+
+set __args=%*
+
+echo.
+echo  WCF Remote Github Repo sync script
+echo. 
+
+if '%__args%'=='' (
+    echo   [%~n0] A URL must be specified for the pull request synchronization URL
+    echo     Usage:  %~n0 [sync url] [pr-number (optional)]
+    echo   sync-url  - URL on remote server for PR synchronization
+    echo   pr-number - PR number to sync to
+    echo.
+    echo   Example:  %~n0 http://wcfci-sync-server/repo/sync.ashx 404
+    set __EXITCODE=1
+    goto done
+)
+
+set __SYNC_HOST_URL=%1
+
+if '%2'=='' (
+    if '%ghprbPullId%'=='' (
+        echo   [%~n0] This script should only be called only from the context of a GitHub Pull Request
+        echo   [%~n0] 'ghprbPullId' environment variable not set
+        set __EXITCODE=1
+        goto done
+    ) else ( 
+        set __PR_ID=%ghprbPullId%
+    )
+) else (
+    set __PR_ID=%2
+    echo   [%~n0] WARNING: This script should usually only be called only from the context of a GitHub Pull Request
+    echo   [%~n0] PR ID overridden: '%__PR_ID%'
+) 
+
+set __REQUEST_URL=%__SYNC_HOST_URL%?pr=%__PR_ID%
+echo   [%~n0] Making call to '%__REQUEST_URL%'
+powershell -NoProfile -ExecutionPolicy unrestricted -Command "(New-Object Net.WebClient).DownloadString('%__REQUEST_URL%'); 
+
+set __EXITCODE=%ERRORLEVEL% 
+
+:done
+endlocal
+exit /b %__EXITCODE% 
+
--- a/src/System.Private.ServiceModel/tools/setupfiles/sync-pr.sh
+++ b/src/System.Private.ServiceModel/tools/setupfiles/sync-pr.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+
+show_banner()
+{
+    echo ""
+    echo "WCF Remote Github Repo sync script"
+    echo ""
+}
+
+show_usage() 
+{
+    echo "    A URL must be specified for the pull request synchronization URL"
+    echo "    Usage: $0 [sync-url] [pr-number (optional)]"
+    echo "    sync-url  - URL on remote server for PR synchronization"
+    echo "    pr-number - PR to sync to"
+    echo ""
+    echo "    Example:  $0 http://wcfci-sync-server/repo/sync.ashx 404"
+    echo ""
+}
+
+run_request() 
+{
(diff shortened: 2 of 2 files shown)


src/System.Private.ServiceModel/tools/setupfiles/sync-pr.sh
```diff
@@ -0,0 +1,76 @@
...
+        fi
+
+        # curl has HTTPS CA trust-issues less often than wget, so lets try that first.
+        which curl > /dev/null 2> /dev/null
+        if [ $? -ne 0 ]; then
+           wget -q -O- ${__sync_url}?pr=${__ghprbPullId}
+        else
+           output=`curl -sw "\n%{http_code}" ${__sync_url}?pr=${__ghprbPullId}`
+           echo $output
+           retCode=`echo ${output} | grep "200\$"`
```
iamjasonp: I have to do this kind of calisthenics here since curl always returns a 0 error code as long as a response was received by the server, regardless of the server's response code

The way this CURL outputs is that the HTTP return code always sits at the end of the returned string, for example: 

"Syncing to PR = 901... OK 200"
"Bad Request 400"

The grep string will look for a 200 at the end. 

Note that wget will fail with exit code 1 when the server returns an error code

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
```diff
@@ -0,0 +1,43 @@
...
+    echo.
+    echo   Example:  %~n0 http://wcfci-sync-server/repo/sync.ashx
+    set __EXITCODE=1
+    goto done
+)
+
+set __SYNC_HOST_URL=%1
+
+if '%ghprbPullId%'=='' (
+    echo   [%~n0] This script should only be called only from the context of a GitHub Pull Request
```
roncain: Can we pass this in as a parameter (with optional substitution of environment variable)?  Environment variables as required input parameters are often problematic.  If there is some reason this is not possible, then show_usage should explain what environment variables are required.

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
```diff
@@ -0,0 +1,43 @@
...
+
+set __SYNC_URL=%__SYNC_HOST_URL%?pr=%ghprbPullId%
+echo   [%~n0] Making call to '%__SYNC_URL%'
+powershell -NoProfile -ExecutionPolicy unrestricted -Command "(New-Object Net.WebClient).DownloadString('%__SYNC_URL%'); 
+
+set __EXITCODE=%ERRORLEVEL% 
+
+:done
+endlocal
+exit /b %__EXITCODE%
```
roncain: Nit: newline.  Not mandatory.

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.sh
```diff
@@ -0,0 +1,76 @@
...
+    echo "    sync-url - URL on remote server for PR synchronization"
+    echo ""
+    echo "    Example:  $0 http://wcfci-sync-server/repo/sync.ashx"
+}
+
+run_request() 
+{
+    __sync_url=$1
+    __ghprbPullId=$2
+
```
roncain: See comments about the CMD version.  Here the PR is an input parameter, which is what I would prefer for both.  It also belongs in the show_usage() description.

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.sh
```diff
@@ -0,0 +1,76 @@
...
+    __request_uri=${__sync_url}?pr=${__ghprbPullId}
+
+    echo "Making a call to '${__request_uri}'" 
+
+    if [ ! -e "$__nugetpath" ]; then
+        which curl wget > /dev/null 2> /dev/null
```
roncain: Why the test against $__nugetpath?  And if so, what should happen in the "else"?

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.sh
```diff
@@ -0,0 +1,76 @@
...
+        if [ $? -ne 0 -a $? -ne 1 ]; then
+            echo "cURL or wget is required to make this request. Please see https://github.com/dotnet/wcf/blob/master/Documentation/building/unix-instructions.md for more details."
+            exit 1
+        fi
+
+        # curl has HTTPS CA trust-issues less often than wget, so lets try that first.
+        which curl > /dev/null 2> /dev/null
+        if [ $? -ne 0 ]; then
+           wget -q -O- ${__sync_url}?pr=${__ghprbPullId}
+        else
```
roncain: Nit: could use __request_uri instead of recomposing the URL + PR

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.sh
```diff
@@ -0,0 +1,76 @@
...
+# Check parameters
+
+if [ -z "$1" ]; then 
+    show_usage
+    exit 1
+fi 
+
+if [ -z "$ghprbPullId" ]; then 
+    show_usage
+    echo "This script should only be called only from the context of a GitHub Pull Request"
```
roncain: Hmmm -- I thought this script took it from $2.  Am I misreading it?  My preferred option is to take everything as an input parameter but to optionally use environment variables if the input parameters are not specified.  Show_usage() should probably describe any required environment variables and/or call out what the defaults are for missing parameters.  I know no one is likely to call this script but CI, but it's show_usage is the best documentation.

roncain:
LGTM, subject to some questions/suggestions which can be dealt with post-merge.

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.sh
iamjasonp: The $2 you see above is a variable passed into the function - $1, $2... $n only mean the command line param when you're outside a function. As soon as you enter a function, then the $1, $2... $n now mean the variables that got passed into it (see line 68 on how this got called)

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.sh
iamjasonp: Oops, copy+paste error on this one. Line 27 should go away. 
This is to figure out whether or not we have curl or wget installed.

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
iamjasonp: Makes sense; I'm just trying hard to ensure that this script runs in the context of a GitHub PR only and I don't necessarily want this to be commonly used in the dev workflow. $ghprPullId is set by Jenkins, so it's a reasonably good indicator that we're running within CI. 

Thoughts? Is your preference still to use a named param?

roncain: Not a deal breaker, but I still prefer parameters that optionally default to environment variables.  I know it is meant to be called by CI only, but the environment variable is a hidden contract, like using a global variable to pass method parameters.  I presume whatever logic calls this can simply pass the environment variable as the 2nd arg. Don't block progress on this -- it's just an attempt to reduce the "magic" that tends to surround CI scripts in general.

iamjasonp: Fair enough - I'll see what I can do with reducing magic here :smile: env var magic makes life so easy though!

zhenlan:
Thank you @iamjasonp. I think this will do the job! :smiley: 

I have one design feedback. The idea of these scripts is to notify test server - I am about to start testing PR "x". It's up to the server to decide what to do to get things prepared - whether it's syncing the enlistment or something else, these scripts don't have to know or assume.

What does this mean? If you can clean up these scripts wherever it made assumption that the server is going to do sync (mostly naming and comments), that will great! Decoupling this logic will give us more flexibility in case we need to tweak anything on the server and avoid any confusions.

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
zhenlan: It's more than preference. I think the question here is - who should own the knowledge of environment variable "ghprbPullId", the caller or the callee? From your current design, these two scripts are more like utilities. The caller knows about the Jenkins/CI Job, knows about which service to call and (should also) knows about which PR to notify the server. These two scripts are just to get the job done.

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
```diff
@@ -0,0 +1,43 @@
+@echo off 
+setlocal
+
+REM 
+REM This calls a web service, specified on the command line, which syncs a 
+REM remote server's repo to the correct PR. 
+REM This script is meant to be called from a GitHub Pull Request. 
+REM If you must, you can fake it by setting the 'ghprbPullId' environment variable
```
zhenlan: I would remove this. I think your previous comment (disclaimer) is good - this is only meant to run on CI (so devs, don't try to run it on your box :))

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
iamjasonp: I'm more of a fan of the "self contained" approach personally. I've had way too many times where I was trying to locally repro something and found the scripts unnavigable due to magic. 

In general, I think an optional parameter helps us in the case where we need to specify a clear set of repro steps running against a server. 

For day to day use, I will allow the ghprb param to be set automatically.

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
iamjasonp: Done

src/System.Private.ServiceModel/tools/setupfiles/sync-pr.cmd
zhenlan: Sorry, I may not make my point clear. There should be only one person (caller vs. callee) who owns the knowledge of ghprbPullId env var. If you make it as a parameter, then caller owns it and the callee should fail if it is not passed in (so not optional). It's the magic that the callee still works even without parameter passed in makes the maintenance hard and error prone.

I am okay with the other way too (does not accept it as a parameter) - as you said, intentionally make it harder for anyone to use it on non-CI environment :smile:.

iamjasonp:
@zhenlan - great suggestion :) updated the sync scripts' usage to talk about the "WCF Test Service Pull Request sync script" so it's clear what the service is used for, with less emphasis on what the nitty gritty is (we can update the name further as the purpose of the scripts evolve)

The default use case is still the same (i.e., we will still call `sync.sh http://server/pr/sync.ashx` with a ghprbPullId specified by the environment, but allows us the option of overriding and syncing to a PR if absolutely necessary. 

This is still a hurdle for someone to do, and we make it clear we don't _want_ you doing it, but it'll be reasonably easy for someone to script if absolutely necessary.

iamjasonp:
Merging this now so I can test out the CI boxes. All comments should have been addressed

PR #925, opened 2016-03-21
Update23918

ericstj:
Update packages to rc3-23918.

I had to fix a few places where projects were downgrading themselves to packages which didn't support the platform they were targeting.  The only shipping project that was changed was System.Private.ServiceModel.  All the versions upgraded too here are supported in all NETStandard1.3 platforms, including NETCore50, so this is a safe change.

/cc @iamjasonp


roncain:
Thanks @ericstj .  Buddy testing this now...

roncain:

ericstj:
Tests are failing to restore overlapped on Linux.  I've fixed that in build 921, will just update to latest.

iamjasonp:
Thanks @ericstj. I gave this a crack earlier but ran into some downgrading problems; I see what you've done now - makes sense that these packages needed to be changed too

roncain:
@ericstj  -- I resynced to the 23921 changes but still see this error in OuterLoops on Windows
```
                        DuplexClientBaseTests.DuplexClientBaseOfT_OverHttp_Call_Throws_InvalidOperation [FAIL] (TaskId:2118)
                           System.TypeInitializationException : The type initializer for 'System.Net.SystemNetworkCredential' threw an exception. (TaskId:2118)
                           ---- System.IO.FileNotFoundException : Could not load file or assembly 'System.Security.SecureString, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified. (TaskId:2118)
                           Stack Trace: (TaskId:2118)
                                 at System.Net.Http.WinHttpHandler..ctor() (TaskId:2118)
                                 at System.Net.Http.HttpClientHandler..ctor() (TaskId:2118)
                                 at System.Net.Http.HttpClient..ctor() (TaskId:2118)
```

iamjasonp:
+1 to Ron's comments - I still see problems with Outerloops (pulled about 8am PDT)

roncain:

src/System.Private.ServiceModel/src/project.json
```diff
@@ -50,12 +50,14 @@
   },
```
zhenlan: Thanks @ericstj. Should we update versions of packages below in src/project.json here too to be consistent with the change you made in src/windows/project.json?

"System.Collections.Concurrent": "4.0.10",
"System.ObjectModel": "4.0.10",
"System.Xml.XmlSerializer": "4.0.10"

ericstj: So that outerloop failure is because you're still using an old version of System.Net.WinHttpHandler.  Let me have a look who's bringing that in.

ericstj: You can if you like but it isn't strictly required.  I was primarily trying to do the smallest change to unblock.

What I'd like to see us do is use the latest packages for all "netstandard" sections and let the version of netstandard determine the right references rather than all of this cherry picking.

ericstj:
So that outerloop failure is because you're still using an old version of System.Net.WinHttpHandler.  We need to track that down, I would have expected test-runtime to bring in latest but its possible its getting downgraded somewhere.

ericstj:
I think I understand what's happening.  There are a few invalid major.minor.build entries in projects causing NuGet to find the closest one, which is sometimes very old.

StephenBonikowsky:
I found that System.ServiceModel.Http was using the incorrect version which caused it to go find an rc2 package.
Also, System.Linq.Expressions was using an old minor version in one project, I think several references to the 4.0.0 stable version in other projects may need to be moved forward as well.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/project.json
```diff
@@ -6,8 +6,8 @@
     "System.Net.Http": "4.0.0",
     "System.Net.Primitives": "4.0.10",
     "System.Security.Cryptography.X509Certificates": "4.1.0-rc3-23921",
-    "System.ServiceModel.Duplex": "4.0.1-rc3-23921",
-    "System.ServiceModel.Http": "4.0.11-rc3-23921",
+    "System.ServiceModel.Duplex": "4.1.0-rc3-23921",
```
StephenBonikowsky: I don't think  Duplex needed to be updated, we don't have a 4.1 version for it.
After fixing the Duplex versions I get an issue with System.Linq.Expressions, looks like something is downgrading it to 4.0.0, working on it.

StephenBonikowsky: Eric, the error I'm getting is...
1>EXEC : error : System.Linq.Expressions 4.0.0 provides a compile-time reference assembly for System.Linq.Expressions on DNXCore,Version=v5.0, but there is no run-time assembly compatible with win7-x64. [C:\OSS\wcf\build.proj]

I updated every project that had an explicit dependency to it to use version "4.0.11-rc3-23921", I still get the failure, and then I saw that xunit 2.1.0 depends on the 4.0.0 version transitively, so I'm wondering if I need to add an explicit reference to S.Linq.Expressions in all our test projects to get around this issue?

StephenBonikowsky: After updating all the test projects to System.Linq.Expressions 4.0.11-rc3-23921 I no longer have any package restoration failures but now I am back to the System.IO.FileNotFoundException for System.Security.SecureString version 4.0.0

So looks like somehow we are still getting an old version of System.Net.Http.WinHttpHandler

Continuing to investigate.

StephenBonikowsky: Searching for System.Net.Http.WinHttpHandler I can only find an old version in lock.json files under C:\OSS\wcf\Tools test-runtime and tool-runtime, not sure if that is the source?

The exact same version number found there "4.0.0-rc2-23816" is also in the packages that were pulled down.

I have to leave for the day, I pushed my changes made on top of Eric's to my repo.
https://github.com/StephenBonikowsky/wcf/tree/ericstj-update23918

ericstj:
I went ahead and give this the same treatment I gave CoreFx.  Updating all the test assemblies to latest.  This was effectively what was happening before with the lineup.

StephenBonikowsky:

StephenBonikowsky:
Thanks for doing this Eric, we discussed this and agree that because of the new "netstandard" model we have moved to where there is just one package that contains all supported assets it now makes sense to just use the latest package versions for all our dependencies.

Because this PR is going to have to go to ship room for RC2 could you squash this down to one PR? They seem to frown on multiple commit PRs :)

* Ron is finishing up the last Outer Loop run on Linux, so we just need to wait for his all clear.

ericstj:
I can squash to two commits.  I want to avoid having the tool-generated changes mixed with human-changes as it creates a lot of noise.

roncain:
LGTM subject to the 2-commit squash.  I've verified this builds and runs our OuterLoops in Linux, which is a big breakthrough.  We see failures that indicate there are NegotiateStream issues for us to investigate.  I think it's critical to merge this asap.  @ericstj once you merge, I'll go ahead and prepare the PR to move it to RC2 unless you prefer to do it.

ericstj:
@roncain you can take it from here.

Issue #885, opened 2016-03-07
AV in ServiceModel.Channels.RequestChannel+<WaitForPendingRequestsAsync>


KKhurin:
We got an AV (null ref) in WaitForPendingRequestsAsync because _closedTcs == null.
Here is the call stack:
~~~
0:016> !clrstack
OS Thread Id: 0xfd4 (16)
 Call Site
 System.ServiceModel.Channels.RequestChannel+d__19.MoveNext()
 System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[[System.ServiceModel.Channels.RequestChannel+d__19, System.Private.ServiceModel]](d__19 ByRef)
 System.ServiceModel.Channels.RequestChannel.WaitForPendingRequestsAsync(System.TimeSpan)
 System.ServiceModel.Channels.HttpChannelFactory`1+HttpClientRequestChannel+d__22[[System.__Canon, mscorlib]].MoveNext()
 System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[[System.ServiceModel.Channels.HttpChannelFactory`1+HttpClientRequestChannel+d__22[[System.__Canon, eModel]](d__22 ByRef)
 System.ServiceModel.Channels.HttpChannelFactory`1+HttpClientRequestChannel[[System.__Canon, mscorlib]].OnCloseAsync(System.TimeSpan)
 System.ServiceModel.Channels.CommunicationObject+d__57.MoveNext()
 System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[[System.ServiceModel.Channels.CommunicationObject+d__57, System.Private.ServiceModel]](d__57 ByRef)
 System.ServiceModel.Channels.CommunicationObject.System.ServiceModel.IAsyncCommunicationObject.CloseAsync(System.TimeSpan)
 System.ServiceModel.Channels.ServiceChannel+d__145.MoveNext()
 System.Runtime.CompilerServices.AsyncTaskMethodBuilder.Start[[System.ServiceModel.Channels.ServiceChannel+d__145, System.Private.ServiceModel]](d__145 ByRef)
 System.ServiceModel.Channels.ServiceChannel.OnCloseAsync(System.TimeSpan)
 System.ServiceModel.Channels.CommunicationObject+d__57.MoveNext()
[...]

PR #920, opened 2016-03-18
Fix async close race condition in RequestChannel

roncain:
Stress tests revealed a race condition during an async close
of RequestChannel.  A TCS used to signal completion of the close
could become null if the final pending request completed
asynchronously during the close.

Fixes #885

PR #926, opened 2016-03-21
Port to RC2 the fix to the async close issue #885

roncain:
PR #920 fixed issue #885 in master.  This PR cherry-picks
that fix into the RC2 branch.

Original description was:
Stress tests revealed a race condition during an async close
 of RequestChannel. A TCS used to signal completion of the close
 could become null if the final pending request completed
 asynchronously during the close.


iamjasonp:
LGTM

PR #935, opened 2016-03-23
Fix NullReferenceException in SendRequestAsync during async streaming

### LINKED ISSUE #931: AV in System.ServiceModel.Channels.HttpChannelFactory.HttpClientRequestChannel.HttpClientChannelAsyncRequest.SendRequestAsync in stress
We got _httpRequestMessage == null inside the catch (OperationCanceledException) block here:
~~~
900                 public async Task SendRequestAsync(Message message, TimeoutHelper timeoutHelper) 
901                 { 
<...>
938                         try 
939                         { 
940                             var timeoutToken = await cancelTokenTask; 
941                             timeoutToken.Register(s_cancelCts, _httpSendCts); 
942                             _httpResponseMessage = await _httpClient.SendAsync(_httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, _httpSendCts.Token); 
943 
 
944                             // As we have the response message and no exceptions have been thrown, the request message has completed it's job. 
945                             // Calling Dispose() on the request message to free up resources in HttpContent, but keeping the object around 
946                             // as we can still query properties once dispose'd. 
947                             _httpRequestMessage.Dispose(); 
948                             success = true; 
949                         } 
950                         catch (HttpRequestException requestException) 
951                         { 
952                             HttpChannelUtilities.ProcessGetResponseWebException(requestException, _httpRequestMessage, 
953                                 _abortReason); 
954                         } 
955                         catch (OperationCanceledException) 
956                         { 
957                             if (cancelTokenTask.Result.IsCancellationRequested) 
958                             { 
959                                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new TimeoutException(SR.Format( 
960                                     SR.HttpRequestTimedOut, _httpRequestMessage.RequestUri, _timeoutHelper.OriginalTimeout))); 
961                             } 
962                             else 
963                             { 
964                                 // Cancellation came from somewhere other than timeoutCts and needs to be handled differently. 
965                                 throw; 
966                             } 
967                         } 
968                         finally 
969                         { 
970                             if (!success) 
971                             { 
972                                 Abort(_channel); 
973                             } 
974                         } 
~~~
I do see us nulling out _httpRequestMessage  in Cleanup:
~~~
985                 private void Cleanup() 
986                 { 
987                     s_cancelCts(_httpSendCts); 
988 
 
989                     if (_httpRequestMessage != null) 
990                     { 
991                         var httpRequestMessageSnapshot = _httpRequestMessage; 
992                         _httpRequestMessage = null; 
993                         TryCompleteHttpRequest(httpRequestMessageSnapshot); 
994                         httpRequestMessageSnapshot.Dispose(); 
995                     } 
996                 } 
~~~
...and we call it from several places including SendRequestAsync itself.

roncain:
Problem: HttpChannelFactory.SendRequestAsync keeps the HttpRequestMessage
in a member field and uses it during async processing.  But a Fault or Abort
will null the HttpRequestMessage field.  If this happens while the
SendRequestAsync method is executing, it leads to NullReferenceException
during error handling.

Solution: keep a local variable copy of the HttpRequestMessage for use in
error handling.  Even if the original is disposed or the field set to null,
error handling only requires access to the RequestUri property.

Fixes #931

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
@@ -904,6 +904,9 @@ namespace System.ServiceModel.Channels
                     _httpClient = await _channel.GetHttpClientAsync(_to, _via, _timeoutHelper);
                     _httpRequestMessage = _channel.GetHttpRequestMessage(_via);
 
+                    // The _httpRequestMessage field will be set to null by Cleanup() due to faulting
+                    // or aborting, so use a local copy for exception handling within this method.
+                    HttpRequestMessage httpRequestMessage = _httpRequestMessage;
                     Message request = message;
 
                     try
@@ -949,7 +952,7 @@ namespace System.ServiceModel.Channels
                         }
                         catch (HttpRequestException requestException)
                         {
-                            HttpChannelUtilities.ProcessGetResponseWebException(requestException, _httpRequestMessage,
+                            HttpChannelUtilities.ProcessGetResponseWebException(requestException, httpRequestMessage,
                                 _abortReason);
                         }
                         catch (OperationCanceledException)
@@ -957,7 +960,7 @@ namespace System.ServiceModel.Channels
                             if (cancelTokenTask.Result.IsCancellationRequested)
                             {
                                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new TimeoutException(SR.Format(
-                                    SR.HttpRequestTimedOut, _httpRequestMessage.RequestUri, _timeoutHelper.OriginalTimeout)));
+                                    SR.HttpRequestTimedOut, httpRequestMessage.RequestUri, _timeoutHelper.OriginalTimeout)));
                             }
                             else
                             {


roncain:
Reviewers: this is my proposed fix and feedback is welcome. Code and windbg inspection show that we executed Abort(), then Cleanup(), which nulled a member field SendRequestAsync used in error handling.  This is difficult to repro in the stress labs, so I will see if I can induce the code path artificially in the debugger before merging.  If it is possible to generate a simple negative test that induces this condition programmatically in a single operation, I will add that.

roncain:
@mconnew, @iamjasonp @zhenlan -- your feedback would be welcome, and I need someone to sign off before I can merge. I would like to merge this and take a PR to shiproom to move this to RC2 if others agree.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -904,6 +904,9 @@ public HttpClientChannelAsyncRequest(HttpClientRequestChannel channel)
                     _httpClient = await _channel.GetHttpClientAsync(_to, _via, _timeoutHelper);
                     _httpRequestMessage = _channel.GetHttpRequestMessage(_via);
 
+                    // The _httpRequestMessage field will be set to null by Cleanup() due to faulting
+                    // or aborting, so use a local copy for exception handling within this method.
+                    HttpRequestMessage httpRequestMessage = _httpRequestMessage;
```
iamjasonp: NIT: I'd rather see the local var being set first, and then the field in the class set after, e.g.: 

```csharp

HttpRequestMessage httpRequestMessage = _channel.GethttpRequestMessage(_via); 
_httpRequestMessage = httpRequestMessage

```

More of a stylistic thing given the timings here, but ensures that we always manage to capture the httpRequestMessage before anyone else has a chance to muck with it.

iamjasonp:
LGTM. My NIT is only that, so merge at will.

PR #936, opened 2016-03-24
Adds scenario test for AV during Http async stream read

roncain:
Issue #931 was discovered by stress tests, where an aborted
Http request doing an asynchronous read would throw a
NullReferenceException.

The cause is not a race condition but instead incorrect use of
member fields in exception handling.  This PR adds a scenario test
that exactly duplicates the stress test and causes NullRefException.

This test will be merged only after #935 that contains the fix to
issue #931.  The changes are being kept separate to allow selective
cherry-picking into the RC2 branch.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/TestMockStream.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/TestMockStream.cs
@@ -1,0 +1,87 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Threading;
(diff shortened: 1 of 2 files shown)


roncain:
Reviewers: I have created PR #935 for the fix and this PR for the test.  They are separate so we can get the fix into RC2 asap yet iterate on comments on the test.  This PR will fail OuterLoop with a NullReferenceException until #935 has been merged, demonstrating it repro's the stress case.

roncain:
This is expected to show a NullReferenceException failure to demonstrate the issue exists

roncain:
The prior OuterLoop run did indeed fail for NullReferenceException in the new test.
Now that #935 has been merged, this OuterLoop run should be clean.

StephenBonikowsky:
LGTM

PR #937, opened 2016-03-24
Disable certificate revoked scenario test

roncain:
This test has been failing in ToF 100% of the time for me recently.

Now attempting to move to latest build tools, it fails 100% of the
time in OSS too.  It is a negative test that should be rewritten to
be smarter about testing after the cert has been revoked.


StephenBonikowsky:
LGTM

iamjasonp:
LGTM. 

The difference here is probably due to caching of certificates, and it feels to me that there's been a change in how CRL caching works in some code beneath us. We'll have to think about how to make this work a bit more nicely

PR #941, opened 2016-03-28
Fix async stream timeout and abort test for NET Native

### LINKED ISSUE #939: BasicHttp_Echo_Streamed_Async_Delayed_Throws_TimeoutException test fails in NET Native
The newly added functional test #936 to verify the fix for #931 has turned up another issue.
In CoreCLR, a TimeoutException is thrown, but in NET Native, a TaskCancelledException is thrown.

We need to determine which is correct and then fix the one that is not.

```
  Running Test: ServiceContractTests.BasicHttp_Echo_Streamed_Async_Delayed_Throws_TimeoutException
  Assert.True:Expected 'System.TimeoutException' to be thrown, however 'System.Threading.Tasks.TaskCanceledException' was thrown.
  Xunit.AssertTestException: Assert.True: Expected 'System.TimeoutException' to be thrown, however 'System.Threading.Tasks.TaskCanceledException' was thrown.
     at Xunit.Assert.HandleFail(String assertionName, String message) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 605
     at Xunit.Assert.Throws(Action action, Object[] CS$0$0000) in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Assertion.cs:line 76
     at ServiceContractTests.BasicHttp_Echo_Streamed_Async_Delayed_Throws_TimeoutException($ICommunicationObject[] CS$0$0000) in e:\ProjectK\src\NDP\FxCore\WcfOpen\src\System.Private.ServiceModel\tests\Scenarios\Contract\Service\ServiceContractTests.cs:line 273
     at EntryPointMain.b__7() in e:\ProjectK\src\QA\ToF\PN\x86\rel\IL\FX\Conformance\System.Private.ServiceModel\4.0.0.0\WcfOpen\Scenarios\Contract\Service\Contract.Service.Tests.main.cs:line 20
     at System.Func$1.InvokeOpenStaticThunk()
     at System.AsyncCallback.Invoke(IAsyncResult ar)
     at CoreFXTestLibrary.Internal.Runner.RunTestMethod$catch$0() in e:\ProjectK\src\QA\ToF\tests\FX\Conformance\Common\Xunit\Internal\Runner.cs:line 170
  ---- Test FAILED ---------------
```

roncain:
The regression test added to verify the fix for #931 failed on
NET Native due to the way it was written.  By aborting the channel
before timing out, the underlying NET Native HttpClient code
terminated the connection rather than cancelling the cancellation
token, so a TimeoutException was not thrown.  By reversing the
order in the test to timeout, then abort, the cancellation token
was properly cancelled, and a TimeoutException was thrown.

The test was manually verified in the debugger to determine it
still executed the code path fixed by #931.

A 2nd test as added to test only the timeout scenario without
an explicit abort.

Fixes #939

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
@@ -226,9 +226,74 @@ public static class ServiceContractTests
         Assert.True(success, "Test Scenario: BasicHttp_DefaultSettings_Echo_RoundTrips_String_Streamed_Async_WithSingleThreadedSyncContext timed-out.");
     }
 
+
     [Fact]
     [OuterLoop]
-    public static void BasicHttp_Echo_Streamed_Async_Delayed_Throws_TimeoutException()
+    public static void BasicHttp_Streamed_Async_Delayed_And_Aborted_Request_Throws_TimeoutException()
+    {
+        // This test is a regression test that verifies an issue discovered where exceeding the timeout
+        // and aborting the channel before the client's Task completed led to incorrect error handling.
+        BasicHttpBinding binding = null;
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
+        Stream stream = null;
+        int sendTimeoutMs = 3000;
+
+        try
+        {
+            // *** SETUP *** \\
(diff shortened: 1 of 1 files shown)


iamjasonp:
LGTM. Not sure why Jenkins keeps failing here, but I've built and had successful runs locally.

PR #947, opened 2016-03-30
Update manual test guide for Linux HTTPS/Kerberos tests

iamjasonp:
skip ci please

### DIFF
--- a/Documentation/manualtest-guide.md
+++ b/Documentation/manualtest-guide.md
@@ -10,14 +10,14 @@ Tests:
 
 Requirements:
 
-1. Must run on a domain machine.
+1. Must run on a domain machine
 
 Steps:
 
-1. Enable the test
-in src\System.Private.ServiceModel\tests\Scenarios\Security\TransportSecurity\Https\ClientCredentialTypeTests.cs,
-Remove  [ActiveIssue(53)] above test method WindowsAuthentication_RoundTrips_Echo()
-2. Go to the root of WCF source, run all scenario tests.
+1. Enable the test <br/>
+In src\System.Private.ServiceModel\tests\Scenarios\Security\TransportSecurity\Https\ClientCredentialTypeTests.cs,
+Remove  `[ActiveIssue(53)]` above the test method 
+2. Go to the root of WCF source, run all scenario tests
 ```
     .\build.cmd /p:WithCategories=OuterLoop
 ```
@@ -43,18 +43,19 @@ Requirements:
 
 Assumes: 
 
-* $(GitWcfRoot) is the root of the WCF Core git repo
-* $(BridgeHost) is the Bridge host machine name  
+* ${GitWcfRoot} is the root of the WCF Core git repo
+* ${BridgeHost} is the Bridge host machine name  
 
 Steps:
 
-1. Ensure the Bridge is running on a Windows machine with remote access allowed: `startBridge.cmd -allowRemote`
-2. On Linux, `pushd $(GitWcfRoot)/src/System.Private.ServiceModel/tools/setupfiles`
-3. Install the certificate file `sudo ./BridgeCertificateInstaller.sh --bridge-host $(BridgeHost) --cert-file ~/tmp/wcfca.crt`. Note that the --cert-file should be a file in the the user tmp directory and the filename must have a .crt extension
-4. Specify the Bridge host location `export BridgeHost=$(BridgeHost)`
-5. Run tests as needed, such as with `$(GitWcfRoot)/build.sh /p:WithCategories=OuterLoop`
+1. Ensure the Bridge is running on a Windows machine with remote access allowed: `startBridge.cmd -allowRemote` <br/> 
+   If running Bridge on a machine not on the same subnet as the test client machine: `startBridge.cmd -allowRemote -remoteAddresses:*` 
+2. On Linux, `pushd ${GitWcfRoot}/src/System.Private.ServiceModel/tools/setupfiles`
+3. Install the certificate file `sudo ./BridgeCertificateInstaller.sh --bridge-host ${BridgeHost} --cert-file ~/tmp/wcfca.crt`. Note that the --cert-file should be a file in the the user tmp directory and the filename must have a .crt extension
+4. Specify the Bridge host location `export BridgeHost=${BridgeHost}`
+5. Run tests as needed, such as with `${GitWcfRoot}/build.sh /p:WithCategories=OuterLoop`
 
-**Linux: NegotiateStream tests** 
+**Linux: NegotiateStream tests with ambient credentials** 
 
 Tests:
 ```csharp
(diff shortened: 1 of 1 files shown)


Documentation/manualtest-guide.md
```diff
@@ -1,17 +1,100 @@
...
+* Requires a Windows machine running the Bridge  
+
+Assumes: 
+
+* $(GitWcfRoot) is the root of the WCF Core git repo
+* $(BridgeHost) is the Bridge host machine name  
+
+Steps:
+
+1. Ensure the Bridge is running on a Windows machine with remote access allowed: `startBridge.cmd -allowRemote`
```
roncain: Also need -remoteAddresses:*, right?

Documentation/manualtest-guide.md
```diff
@@ -1,17 +1,100 @@
...
+The WCF Test Root CA certificate need to be retrieved from the Bridge and installed on the local machine before tests can be run
+
+Requirements: 
+
+* Requires a Windows machine running the Bridge  
+
+Assumes: 
+
+* $(GitWcfRoot) is the root of the WCF Core git repo
+* $(BridgeHost) is the Bridge host machine name
```
roncain: Nit: These look like MsBuild property syntax, making me think I need to set them in Environment variables for things to work.  Is that the intent?  Maybe we can adopt a syntax like {GitWcfRoot} or something that invites the user to think of it as a placeholder. Just my opinion -- don't feel obliged,

Documentation/manualtest-guide.md
```diff
@@ -1,17 +1,100 @@
...
+2. On Linux, `pushd $(GitWcfRoot)/src/System.Private.ServiceModel/tools/setupfiles`
+3. Install the certificate file `sudo ./BridgeCertificateInstaller.sh --bridge-host $(BridgeHost) --cert-file ~/tmp/wcfca.crt`. Note that the --cert-file should be a file in the the user tmp directory and the filename must have a .crt extension
+4. Specify the Bridge host location `export BridgeHost=$(BridgeHost)`
+5. Run tests as needed, such as with `$(GitWcfRoot)/build.sh /p:WithCategories=OuterLoop`
+
+**Linux: NegotiateStream tests** 
+
+Tests:
+```csharp
+   StreamingTests.NetTcp_TransportSecurity_StreamedRequest_RoundTrips_String
```
roncain: I wonder if it's time to add a trait discoverer and mark these tests.  Not required for this PR, but a list of specific tests is going to fall out of date quickly.

Documentation/manualtest-guide.md
```diff
@@ -1,17 +1,100 @@
...
+
+* $(GitWcfRoot) is the root of the WCF Core git repo
+* $(BridgeHost) is the Bridge host machine name  
+
+Steps:
+
+1. Ensure the Bridge is running on a Windows machine and as LOCAL MACHINE with `-allowRemote`
+2. On Linux, `pushd $(GitWcfRoot)/src/System.Private.ServiceModel/tools/setupfiles`
+3. Install the certificate file `sudo ./BridgeCertificateInstaller.sh --bridge-host $(BridgeHost) --cert-file ~/tmp/wcfca.crt`. Note that the --cert-file should be a file in the the user tmp directory and the filename must have a .crt extension
+4. Specify the Bridge host location `export BridgeHost=$(BridgeHost)`
```
roncain: Which script or code requires it to be an environment variable?  Just curious if we can pass it as a parameter and avoid setting environment variables.  Don't block on this question.

Documentation/manualtest-guide.md
```diff
@@ -1,17 +1,100 @@
...
 ```
+
+The WCF Test Root CA certificate need to be retrieved from the Bridge and installed on the local machine before tests can be run.
+In addition, the Bridge must be running under a LOCAL SYSTEM context
+
+Requirements: 
+
+* Requires a Windows machine running the Bridge
+* The Windows machine should be on a Windows domain 
+* A tool such as `psexec` is needed to start the Bridge as LOCAL SYSTEM
```
roncain: Could you elaborate here on starting as LOCAL SYSTEM?  Like provide an example of the script steps you would use with psexec?

Documentation/manualtest-guide.md
```diff
@@ -1,17 +1,100 @@
...
+* $(GitWcfRoot) is the root of the WCF Core git repo
+* $(BridgeHost) is the Bridge host machine name  
+
+Steps:
+
+1. Ensure the Bridge is running on a Windows machine and as LOCAL MACHINE with `-allowRemote`
+2. On Linux, `pushd $(GitWcfRoot)/src/System.Private.ServiceModel/tools/setupfiles`
+3. Install the certificate file `sudo ./BridgeCertificateInstaller.sh --bridge-host $(BridgeHost) --cert-file ~/tmp/wcfca.crt`. Note that the --cert-file should be a file in the the user tmp directory and the filename must have a .crt extension
+4. Specify the Bridge host location `export BridgeHost=$(BridgeHost)`
+5. Initialize Kerberos `kinit user@DC.DOMAIN.COM` and enter the password
```
roncain: Which password?  Is "user" meant to be the person running the command, or is that a literal?  (In other words, should we be saying 'kinit {UserName}@DC.DOMAIN.COM' and enter {UserPassword}?).  Nit: don't block on it.

roncain:
LGTM, subject to some nit level (optional) suggestions and questions.

Documentation/manualtest-guide.md
iamjasonp: Hmm, I guess I've assumed the test and server machines are on the same subnet, so this wasn't an issue. I'll add this.

Documentation/manualtest-guide.md
iamjasonp: Oops, the brackets () should be braces {} to match `bash` syntax.

Documentation/manualtest-guide.md
iamjasonp: Agreed. I didn't add the traits here because we haven't defined what they are yet. Maybe this is a good time to start discussion on it.

Documentation/manualtest-guide.md
iamjasonp: We could set it as part of build.sh, but since it's possible for someone to run tests by directly calling corerun (without going through our scripts), `export`ing a variable seems to be the most robust option to specify when providing general instructions

Documentation/manualtest-guide.md
iamjasonp: Sure, but then I think DC.DOMAIN.COM now also needs to be parameterized in such a way :smile: But sure, I agree that we should be consistent in how we surface these variables up in this doc

iamjasonp:
Closing, editing the PR comment, and reopening so that subsequent changes don't trigger CI

Documentation/manualtest-guide.md
iamjasonp: Added a note that it's not necessary if specified on the command line via build.sh

PR #964, opened 2016-03-31
Updating test to be a little clearer about what it is doing.

### LINKED ISSUE #835: TestCase is not validating anything?
ExpectedExceptionTests.TCP_ServiceCertFailedCustomValidate_Throw_Exception

Does not appear to be validating anything, based on the title it expects an exception but the catch doesn't validate anything.

So the test will pass whether an exception is thrown or not, and if it throws an exception it doesn't do an Assert on anything, there is no validation being done.

If the test were validating the Exception I would currently expect it to be failing for Issue #833 which is what the other two TCP related expected exception tests are failing for.

StephenBonikowsky:
* Also updating it to use more recent test syntax and to use Assert.Throw to verify expected exceptions.
* Fixes #835

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -491,33 +491,46 @@ public static class ExpectedExceptionTests
 
     [Fact]
     [OuterLoop]
+    // Confirm that the Validate method of the custom X509CertificateValidator is called and that an exception thrown there is handled correctly.
     public static void TCP_ServiceCertFailedCustomValidate_Throw_Exception()
     {
         string testString = "Hello";
+        NetTcpBinding binding = null;
+        EndpointAddress endpointAddress = null;
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
 
-        NetTcpBinding binding = new NetTcpBinding();
-        binding.Security.Mode = SecurityMode.Transport;
-        binding.Security.Transport.ClientCredentialType = TcpClientCredentialType.None;
-
-        EndpointAddress endpointAddress = new EndpointAddress(new Uri(Endpoints.Tcp_VerifyDNS_Address), new DnsEndpointIdentity(Endpoints.Tcp_VerifyDNS_HostName));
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -491,33 +491,46 @@ public static void TCP_ServiceCertExpired_Throw_MessageSecurityException()
...
     [Fact]
     [OuterLoop]
+    // Confirm that the Validate method of the custom X509CertificateValidator is called and that an exception thrown there is handled correctly.
     public static void TCP_ServiceCertFailedCustomValidate_Throw_Exception()
     {
         string testString = "Hello";
+        NetTcpBinding binding = null;
+        EndpointAddress endpointAddress = null;
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
```
iamjasonp: NIT: why the assignment to null, then reassignment afterwards? Just directly use 

```
var binding = new NetTcpBinding(); 
endpointAddress = new EndpointAddress(new Uri(Endpoints.Tcp_VerifyDNS_Address), new DnsEndpointIdentity(Endpoints.Tcp_VerifyDNS_HostName));  
```

etc. 
Not blocking for merge, just looks a little strange to me.

iamjasonp:

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
StephenBonikowsky: We started doing this some time back when we first decided to include the test sections // *** SETUP *** \\, // *** EXECUTE *** \\, etc...

The idea was to divide up the test into its distinct sections the first of which was to instantiate everything and then "setup" and so forth.

When I touch an old test I update it to follow this pattern, eventually we will achieve this consistent pattern across all our tests.

roncain: Yes, I tend to like the variable declaration in a non-throwable part of the code like @StephenBonikowsky is doing.  Then we can try/catch around the "work" part and still access them in the finally

roncain:
LGTM subject to passing outerloops

PR #974, opened 2016-04-01
Implementing async readahead with sync read on StreamFormatter

mconnew:
I also passed through a cancellation token from higher layers to improve timing out scenarios with chunked encoding.

### DIFF
--- a/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
+++ b/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
   <PropertyGroup>
     <Configuration Condition="'$(Configuration)'==''">$(OS)_Debug</Configuration>
@@ -21,20 +21,18 @@
     <PackageTargetRuntime Condition="'$(OSGroup)' == 'Windows_NT'">win7</PackageTargetRuntime>
     <NuGetTargetMoniker Condition="'$(TargetGroup)' == ''">.NETStandard,Version=v1.3</NuGetTargetMoniker>
   </PropertyGroup>
-
   <!-- Default configurations to help VS understand the options -->
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Unix_Debug|AnyCPU'" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Unix_Release|AnyCPU'" />
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_Debug|AnyCPU'"/>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_Release|AnyCPU'"/>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_netcore50_Debug|AnyCPU'"/>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_netcore50_Release|AnyCPU'"/>
-
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_Release|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_netcore50_Debug|AnyCPU'" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Windows_netcore50_Release|AnyCPU'" />
   <PropertyGroup Condition="'$(TargetGroup)' == 'netcore50'">
     <DefineConstants>$(DefineConstants);FEATURE_NETNATIVE</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetGroup)' != 'netcore50'">
-    <DefineConstants>$(DefineConstants);FEATURE_CORECLR;SUPPORTS_WINDOWSIDENTITY</DefineConstants> 
+    <DefineConstants>$(DefineConstants);FEATURE_CORECLR;SUPPORTS_WINDOWSIDENTITY</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition="'$(TargetsWindows)' == 'true'">
     <DefineConstants>$(DefineConstants);TARGETS_WINDOWS</DefineConstants>
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
@@ -32,7 +32,7 @@ namespace System.ServiceModel.Channels
             _encoder = factory.MessageEncoderFactory.Encoder;
         }
 
-        internal async Task<Message> ParseIncomingResponse()
+        internal async Task<Message> ParseIncomingResponse(TimeoutHelper timeoutHelper)
         {
             ValidateAuthentication();
             ValidateResponseStatusCode();
@@ -52,7 +52,7 @@ namespace System.ServiceModel.Channels
             }
             else
             {
-                message = await ReadStreamAsMessageAsync();
+                message = await ReadStreamAsMessageAsync(timeoutHelper);
             }
 
             var exception = ProcessHttpAddressing(message);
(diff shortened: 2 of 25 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -173,16 +173,16 @@ private Task<Message> ReadStreamAsMessageAsync()
...
             return ReadBufferedMessageAsync(contentStreamTask);
         }
 
-        private async Task<Message> ReadChunkedBufferedMessageAsync(Task<Stream> inputStreamTask)
+        private async Task<Message> ReadChunkedBufferedMessageAsync(Task<Stream> inputStreamTask, TimeoutHelper timeoutHelper)
         {
             try
             {
-                return await _encoder.ReadMessageAsync(await inputStreamTask, _factory.BufferManager, _factory.MaxBufferSize, _contentType);
+                return await _encoder.ReadMessageAsync(await inputStreamTask, _factory.BufferManager, _factory.MaxBufferSize, _contentType, await timeoutHelper.GetCancellationTokenAsync());
```
mconnew: I originally made this change as I was trying to get the cancellation token passed through to enable pre-reading the stream asynchronously with the timeout applied. Unfortunately this couldn't be done in the end due to needing to cross over a synchronous public api in MessageEncoder to be able to put the ReadAheadWrappingStream in the correct place for the call hierarchy to work.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReadAheadWrappingStream.cs
```diff
@@ -0,0 +1,304 @@
...
+
+            if (_previousCompletedReadTask.Result != bytesRead)
+            {
+                _previousCompletedReadTask = Task.FromResult(bytesRead);
+            }
+
+            return _previousCompletedReadTask;
+        }
+
+        private async Task<int> ReadAsyncAfterReadFromBuffer(int bytesRead, byte[] buffer, int offset, int count, CancellationToken cancellationToken)
```
roncain: Nit: possibly add comment explaining what is expected of the input parameters here.  The fact 'count' really means count-bytesRead and offset means offset+bytesRead made me look twice to understand how it is used.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReadAheadWrappingStream.cs
```diff
@@ -0,0 +1,304 @@
...
+        }
+
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                return Task.FromCanceled<int>(cancellationToken);
+            }
+
+            if (_buffers[0].Count == 0 && _buffers[1].Count == 0)
```
roncain: The decision to call base.ReadAsync() is based on the buffers both being empty.  Is there any code path that could enter here after _previousCompletedReadTask has been set but before anything is in one of the buffers?  I'm just trying to understand if there are any Asserts we can make to guarantee presumed invariants (like _previousCompletedReadTask not already being set, etc.).

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReadAheadWrappingStream.cs
```diff
@@ -0,0 +1,304 @@
...
+            }
+
+            int bytesRead = ReadFromBuffer(buffer, offset, count);
+            if (bytesRead < count)
+            {
+                _previousCompletedReadTask = ReadAsyncAfterReadFromBuffer(bytesRead, buffer, offset, count, cancellationToken);
+                return _previousCompletedReadTask;
+            }
+
+            if (_previousCompletedReadTask.Result != bytesRead)
```
roncain: Will _previousCompletedReadTask generally be completed by the time we ask for Result?  Should we use GetAwaiter().GetResult()?

roncain:
@mconnew -- are you trying to get this checked in before leaving for a week?  I'm a little nervous about the possibility of breakage.  The code is just tricky enough to take some time to study. An alternative is a collection of scenario tests that use something similar to TestMockStream (sp?) to fault-inject cancellations and eofs until we ensure we have code coverage of the corner cases.  Thoughts?

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
```diff
@@ -254,6 +255,78 @@ internal MessageBodyStream(Message message, string wrapperName, string wrapperNs
...
+                    throw TraceUtility.ThrowHelperError(new ArgumentException(SR.Format(SR.SFxInvalidStreamOffsetLength, offset + count)), _message);
+
+                try
+                {
+                    if (_reader == null)
+                    {
+                        _readAheadStream =
+                            _message.Properties[ReadAheadWrappingStream.ReadAheadWrappingStreamPropertyName] as ReadAheadWrappingStream;
+                        // Fill buffer so reading to the body contents shouldn't cause a read through to transport stream
+                        if (_readAheadStream != null)
```
roncain: Under what circumstances could this be null?  Should we assert or throw if it is?  Or is the intent that we softly use it when it is present but tolerate when it is not?  What kind of scenario test could be written to verify correct behavior in the null case?

src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj
```diff
@@ -59,4 +57,4 @@
     <TargetingPackReference Include="Windows" />
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
-</Project>
+</Project>
```
iamjasonp: NIT: Are changes to this file necessary?

mconnew: Visual Studio made these changes automatically. It's some simple formatting tidy up. If it's checked in, it means other people don't need to keep ignoring that VS made changes.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReadAheadWrappingStream.cs
mconnew: Good idea. I'll add a comment

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
mconnew: The case is really with a custom message encoder as the ReadAheadWrappingStream needs to be set in the encoder. I should still have my standalone bytestream message encoder around somewhere that we could put into test code to test this.

mconnew:
Not planning on merging this week due to being OOF for the week. This is a performance improvement (a really nice one, it gives 2X throughput on read) but not a functional fix so it can wait. Especially with the idea of adding more tests. This did pass stress tests so there shouldn't be any issues with the in contract message encoders, but I agree we should have some extra tests around this before merging.

iamjasonp:
Please do not test outerloop on this PR until it's been rebased to a recent _master_ past 8 April

mconnew:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReadAheadWrappingStream.cs
mconnew: The _previousCompletedReadTask is about avoiding creating a new Task object when completing synchronously if possible. For example, if a caller to ReadAsync asks for 1024 bytes, chances are it will ask for 1024 bytes again the next time. If this is the case, then we can avoid allocating a Task object by reusing the same one as was previously returned (if we can fulfill the request from the buffers). In this case, the buffers were empty so we need to call base.ReadAsync. There's no point trying to reuse the same Task as the base class has already provided the task. Plus it's only doable if the current operation has completed synchronously, otherwise you can't compare the result with the previous result, which won't be the case here. In this case, we just save the returned Task in case it's the same next time.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReadAheadWrappingStream.cs
mconnew: _previouslyCompletedReadTask should always be completed by the time we ask for the Result. If it hasn't, then the caller is making multiple concurrent calls to ReadAsync which could either corrupt the returned data, or cause the inner stream to throw. I'll add an assert at the beginning of the method.

src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
mconnew: I couldn't find my bytestream message encoder so instead used a slightly modified encoder from the WCF samples.

mconnew:

mconnew:

mconnew:
@roncain Can you check the latest changes and ACK if you're happy with this being merged

roncain:
LGTM

Issue #958, opened 2016-03-31
Create [ConditonalFact] conditions for tests with special requirements


iamjasonp:
We have certain classes of tests that need xunit traits defined for them. Consider this a living list of traits that we need to create:

Tests requiring:
* [x] Client certificate installation - Windows / \*nix **[Client_Certificate_Installed]**
Such tests require that the test (client) machine is provided with a certificate generated by a Certificate Authority trusted by the server. The test (client) machine must install this certificate to its user store, and then use the certificate with this thumbprint to access the WCF test service. Upon finishing the test, the cert MUST be removed from the machine. 

* [x] Client machine (Windows) domain joined **[Domain_Joined]**
For testing ambient Kerberos/NTLM credentials. Such tests require the test (client) machine is joined to an Active Directory domain or Azure Active Directory Domain Services domain. Note that this does not mean MDM joined via Azure Active Directory - that is not sufficient. If using AAD DS, there need not be an additional server machine as AAD DS provisions a domain controller capable of acting as a TGT

* [x] Client (\*nix, OS X): Kerberos ambient credential setup/destroy **[Client_Linux_Domain_Joined]** **
For testing ambient Kerberos credentials. Such tests require that the test (client) machine to have a gssapi package installed, and the tests must be able to run kinit and kdestroy under a user context. Prior to testing, kdestroy must be run, then kinit must be run and provided with valid user credentials for the target Kerberos Realm. After the test, kdestroy must be run. 

* [x] Client (\*nix, OS X): Kerberos explicit credential setup/destroy
For testing explicit Kerberos credentials. Such tests require that the test (client) machine to have a gssapi package installed, and the tests must be able to run kinit and kdestroy under a user context. Prior to testing, kdestroy must be run. The test must be supplied with valid Kerberos credentials for the target Kerberos Realm, and the test (via .NET Core) is to obtain the Kerberos ticket from the TGT. After the test, kdestroy must be run. 

* [x] Service: server machine is domain joined **[Server_Domain_Joined]**
In conjunction with clients that require Kerberos, the server must be run on a machine joined either to an Active Directory domain or an Azure AD Directory Services domain. The server/domain must provide a set of valid user credentials with which to log onto the domain and to acquire Kerberos tickets from it. 
~~We can probably always set this condition to be true in CI/Helix~~ I think we need to set up the CI server box such that it is domain joined. It will avoid many of the issues around testing that we have today where these tests are disabled. 

[...]

PR #978, opened 2016-04-04
Add support for [ConditionalFact] based on TestProperties

roncain:
We need a way to optionally run tests that are not normally
run by default.  This includes tests that have special requirements
for setup or environment.

We also need such tests to show as "skipped" rather than being
silently stripped from the test results, as happens with the
xunit -notrait option.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+namespace Infrastructure.Common
+{
+    // ConditionalWcfTest is expected to be the base class of any test
+    // class that includes [ConditionalFact] or [ConditionalTheory]. This
+    // is necessary because the conditional attributes are expected to
+    // refer to members within their test class or its base classes.
+    public class ConditionalWcfTest
+    {
+        private static bool IsConditionalTestEnabled(string testCategoryName)
+        {
+            string propertyValue = TestProperties.GetProperty(testCategoryName);
+            bool isEnabled;
+            if (!String.IsNullOrEmpty(propertyValue) && bool.TryParse(propertyValue, out isEnabled))
+            {
+                return isEnabled;
+            }
+
+            return false;
+        }
+
+        public static bool Domain_Joined()
+        {
+            return IsConditionalTestEnabled(TestProperties.Domain_Joined_PropertyName);
+        }
+
+        public static bool Root_Certificate_Installed()
+        {
+            return IsConditionalTestEnabled(TestProperties.Root_Certificate_Installed_PropertyName);
+        }
+
+        public static bool Client_Certificate_Installed()
+        {
+            return IsConditionalTestEnabled(TestProperties.Client_Certificate_Installed_PropertyName);
+        }
+
+        public static bool SPN_Available()
+        {
+            return IsConditionalTestEnabled(TestProperties.SPN_Available_PropertyName);
+        }
+
+        public static bool Kerberos_Available()
+        {
+            return IsConditionalTestEnabled(TestProperties.Kerberos_Available_PropertyName);
+        }
+
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
@@ -6,8 +6,9 @@ using System.ServiceModel;
 using System.ServiceModel.Security;
 using System.Text;
 using Xunit;
+using Infrastructure.Common;
 
-public static class Https_ClientCredentialTypeTests
+public class Https_ClientCredentialTypeTests : ConditionalWcfTest
 {
     private static string s_username;
     private static string s_password;
@@ -108,8 +109,7 @@ public static class Https_ClientCredentialTypeTests
         Assert.True(exception.Message.ToLower().Contains(paraMessage), string.Format("Expected exception message to contain: '{0}', actual: '{1}'", paraMessage, exception.Message));
     }
 
-    [Fact]
-    [ActiveIssue(270)]
+    [ConditionalFact(nameof(Domain_Joined))]
     [OuterLoop]
     public static void DigestAuthentication_RoundTrips_Echo()
     {
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/TestProperties.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/TestProperties.cs
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
 using System;
 using System.Collections.Generic;
 
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
@@ -86,6 +86,26 @@
       <NegotiateTestUpn>testuser@DOMAIN.CONTOSO.COM</NegotiateTestUpn>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(Domain_Joined)' == ''">
+      <Domain_Joined>false</Domain_Joined>
+    </PropertyGroup>
+  
+    <PropertyGroup Condition="'$(Root_Certificate_Installed)' == ''">
+      <Root_Certificate_Installed>false</Root_Certificate_Installed>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(Client_Certificate_Installed)' == ''">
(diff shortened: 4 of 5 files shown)


roncain:
Reviewers: this is a proposed approach to solving #958.  It demonstrates how one could define a new TestProperty and use [ConditionalFact] that checks it.  If we decide to take this approach, I will update the PR with whatever additional categories we need.  The output of an OuterLoop test run for this particular test looks like this:
```
                     Discovering: Security.TransportSecurity.Tests (TaskId:2904)
                     Discovered:  Security.TransportSecurity.Tests (TaskId:2904)
                     Starting:    Security.TransportSecurity.Tests (TaskId:2904)
                        Https_ClientCredentialTypeTests.DigestAuthentication_RoundTrips_Echo [SKIP] (TaskId:2904)
                           "Domain_Joined_Tests_Enabled" returned false. (TaskId:2904)
                     Finished:    Security.TransportSecurity.Tests (TaskId:2904)
                      (TaskId:2904)
                     === TEST EXECUTION SUMMARY === (TaskId:2904)
                        Security.TransportSecurity.Tests  Total: 35, Errors: 0, Failed: 0, Skipped: 1, Time: 24.016s (TaskId:2904)
  
```
I'm using this PR to discuss the approach. The use of TestProperties allows the choice to be injected into the test binaries, meaning this will work equally well locally and in tests shipped off to execute in Helix. The use of a unique test property per category allows them to be set independently without a more complex "include these" and "exclude these" mechanism.

Note that ToF does not support [ConditionalFact] but as long as we define a dummy ConditionalFactAttribute in ToF tests, the tests will build and just naturally skip the tests

Feedback is welcome.
/cc @iamjasonp @zhenlan @StephenBonikowsky @hongdai

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
```diff
@@ -6,8 +6,9 @@
 using System.ServiceModel.Security;
 using System.Text;
 using Xunit;
+using Infrastructure.Common;
 
-public static class Https_ClientCredentialTypeTests
+public class Https_ClientCredentialTypeTests : ConditionalWcfTest
```
roncain: Reviewers: ConditionalFactDiscover only works with members of the type in which it appears, including base types or nested types.  You can't specify an arbitrary helper type.  So we either need to use derivation like this or duplicate the helper methods in every test class where we use them.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
```diff
@@ -108,8 +109,7 @@ public static void BasicAuthenticationEmptyUser_throw_ArgumentException()
         Assert.True(exception.Message.ToLower().Contains(paraMessage), string.Format("Expected exception message to contain: '{0}', actual: '{1}'", paraMessage, exception.Message));
     }
 
-    [Fact]
-    [ActiveIssue(270)]
+    [ConditionalFact(nameof(Domain_Joined_Tests_Enabled))]
```
roncain: Reviewers: we have the green light to use C# 6 syntax.  CoreFx has already migrated or is migrating their [ConditionalFact] attributes to use the new "nameof".  This not only gives compile-time checking, but VS's "Find all references" works well to find every test using this condition.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
```
iamjasonp: NIT: (c) header?

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
iamjasonp: Yuck. Derivation seems like the best solution here.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Infrastructure.Common
+{
+    public class ConditionalWcfTest
+    {
+        public static bool IsTestEnabled(string testCategoryName)
```
iamjasonp: NIT: I think it might read better as "IsCategoryEnabled" or "IsConditionalTestEnabled"

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -0,0 +1,23 @@
...
+            {
+                return isEnabled;
+            }
+
+            return false;
+        }
+
+        public static bool Domain_Joined_Tests_Enabled()
+        {
+            return IsTestEnabled(TestProperties.Domain_Joined_Tests_Enabled_PropertyName);
```
iamjasonp: Question only: is there a need to cache the result of this call?

iamjasonp:
How does this work with multiple categories? For example, if something requires both Domain Join and Certificates, can we specify both, or does this mean we have another TestCategory? 

---

For ToF tests, are we just assuming that all "special" tests are going to be skipped? For example, I would think that some of the "ambient credential" tests would work if the machines are on prem and domain joined - this approach with `[ConditionalFact]` basically means "we're ok with skipping these tests in ToF". 

I don't disagree with this approach, just wondering if there's less "big hammer" way of doing this. 

--- 

Is there any case where we want these test categories to be enabled by default then _disabled_ on demand instead?

roncain:
@iamjasonp -- one can't use multiple [ConditionalFact]'s per method, so if a particular test requires more than one condition, then yes, I believe a separate static bool method would need to be added for the combination (e.g. "I need to be both on prem and domain-joined").  Alternatively, if one condition always depended on a 2nd condition, then the helper method itself could call it, something like the way our Endpoint.cs resources call on additionally required resources.

I suspect in time we may want some conditions to involve code to auto-detect whether they are on or off by default (like the canonical example "Am I running elevated?").  In that case, I think we might consider a 3-state value for the TestProperty in those cases (true, false, auto) let the test helper decide.  So my statement that all default to false is probably too strong.  I suspect that if we introduce auto-detect for some, that will be their default.  Naturally, we would want an explicit 'true' or 'false' to be able to override.

roncain:
I opened issue https://github.com/dotnet/buildtools/issues/610 to allow [ConditionalFact] to evaluate multiple conditions

roncain:
Reviewers: if there is buy-in of this approach, I would like to update this PR to adapt to review comments and add whatever conditions we know we need right now, then merge.  This will lay the foundation for other conditions to be added as we discover them.

/cc @iamjasonp @zhenlan @StephenBonikowsky @hongdai @mconnew

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
roncain: I should have added the note that this behavior matches what the 'nameof' operator does and is probably the reason it was scoped to this behavior.  CoreFx has adopted nameof for its [ConditionalFact] usage.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: I don't think so but it's a good question.  If an invalid name is passed in TestProperties will throw, and it should be obvious from the call stack which test had a problem.  Just to be safe though, I'll make that helper method private and leave only the well-formed ones public.  They won't even build if they try to use an invalid name.

roncain:
Reviewers: I would now like review feedback and to merge this once approved,  The intent of this PR is just to introduce the functionality and wire up one test to demonstrate it works.  Changing individual tests to use it is expected to come as separate PR's.  I have chosen condition names based on my read of #958 and so am looking for feedback on whether the initial set will do.

Note: if needed, we might change the default values based on OS if we think we need that. I am also hoping https://github.com/dotnet/buildtools/issues/610 is approved so we can combine conditions, but I don't want this PR blocked by that.

iamjasonp:
> so if a particular test requires more than one condition, then yes, I believe a separate static bool method would need to be added for the combination (e.g. "I need to be both on prem and domain-joined"). 

After dotnet/buildtools#619 is merged, it looks like we can specify multiple conditions, right? That alleviates any worries I have for this issue. 

> I suspect in time we may want some conditions to involve code to auto-detect whether they are on or off by default (like the canonical example "Am I running elevated?"). 

Yeah, that's probably true; that said from the test runner's perspective, I think it means that a condition is "implied auto detect" vs. "implied false"? Today it's "implied false", so no tests in any [ConditionalFact] will run unless explicitly turned on. This behaviour is probably ok today; we can probably solve this problem later. 

> if needed, we might change the default values based on OS if we think we need that.
Perhaps; it might be useful. Let's use #958 to explore this?

zhenlan:
When a test has ConditionalFact on it, it means it will not work unless certain condition(s) is/are met, so it makes sense to me that these tests are disabled/skipped by default (before we have anything auto).

The naming can be tweaked as I prefer client/server prefixes for clarity, but that can be discussed in #958 and update later.

This PR LGTM. Thanks @roncain.

PR #981, opened 2016-04-05
Add pull request service for test infrastructure

iamjasonp:
skip ci please


src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
+<%@ WebHandler Language="C#" Class="PullRequestHandler" %>
```
roncain: Question: is there a convention for adding a copyright notice to files like these?

zhenlan:
The file name (sync.ashx) suggests it will do sync only. While it is true as in current implementation, it is actually the entry point that will be called from CI, so we may need to append extra logic in the future. Why don't call it pr.ashx?

src/System.Private.ServiceModel/tools/PRService/sync.ashx
zhenlan: Yes, let's add copyright. It's just like .cs file.

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+
+//[remote "origin"]
+//	url = https://github.com/dotnet/wcf
+//	fetch = +refs/heads/*:refs/remotes/origin/*
+//	fetch = +refs/pull/*/head:refs/remotes/origin/pr/*      <-- allows us to run git checkout pr/{id} 
+
+public class PullRequestHandler : IHttpHandler
+{
+    private const string _gitExecutablePath = @"C:\Program Files (x86)\Git\bin\git.exe";
+    private const string _gitRepoPath = @"N:\git\wcf-test";
```
zhenlan: Git is installed at C:\Program Files\Git\cmd\git.exe

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+// This script assumes that the repo is set up as follows: 
+
+//[remote "origin"]
+//	url = https://github.com/dotnet/wcf
+//	fetch = +refs/heads/*:refs/remotes/origin/*
+//	fetch = +refs/pull/*/head:refs/remotes/origin/pr/*      <-- allows us to run git checkout pr/{id} 
+
+public class PullRequestHandler : IHttpHandler
+{
+    private const string _gitExecutablePath = @"C:\Program Files (x86)\Git\bin\git.exe";
```
zhenlan: Git is installed at C:\Program Files\Git\cmd\git.exe

src/System.Private.ServiceModel/tools/PRService/sync.ashx
zhenlan: We shouldn't hard code this. The root of git repo can be inferred from the relative path to PRService folder, which can be obtained by calling 
```
context.Request.PhysicalApplicationPath
```

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+        {
+            "clean -fdx",
+            "checkout master",
+            "fetch --all --prune",
+            "merge --ff-only origin/master",
+            string.Format("checkout {0}", sha)
+        };
+
+        return RunGitCommands(gitCommands, executionResult);
+    }
```
zhenlan: Let's remove SyncToSha. We can add it back when we need it in the future.

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+        {
+            // Don't try to delete "master" branch, since we're checked out to it
+            if (!branchesList[i].EndsWith("master"))
+            {
+                commands.Add(string.Format("branch -D {0}", branchesList[i]));
+            }
+        }
+
+        RunGitCommands(commands.ToArray(), executionResult);
+    }
```
zhenlan: Is CleanupMergedBranches dead code or do you forget to call it somewhere as it is actually necessary?

src/System.Private.ServiceModel/tools/PRService/sync.ashx
iamjasonp: Not on my system it isn't... but I'll modify it based on the system you've set up

src/System.Private.ServiceModel/tools/PRService/sync.ashx
iamjasonp: It's necessary for repo maintenance. 
I anticipate that it will be called either on every sync or as a scheduled task. Thoughts on where it should be done?

src/System.Private.ServiceModel/tools/PRService/sync.ashx
iamjasonp: FWIW, this is a good assumption to call out, that Git is available at this specific path

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+        bool success = false;
+
+        foreach (string gitCommand in gitCommands)
+        {
+            psi.Arguments = gitCommand;
+
+            Process p = Process.Start(psi);
+            if (!p.WaitForExit(_gitExecutionTimeoutMilliseconds))
+            {
+                executionResult.Append(string.Format("Git executable '{0}' took more than '{1}' ms to execute", _gitExecutablePath, _gitSanityExecutionTimeoutMilliseconds));
```
zhenlan: Should we log which git command caused failure here?

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+            }
+
+            success = p.ExitCode == 0;
+
+            executionResult.Append(p.StandardOutput.ReadToEnd());
+            executionResult.Append(p.StandardError.ReadToEnd());
+
+            if (!success)
+            {
+                executionResult.Append(string.Format("Git executable '{0}' exited with code '{1}', expected '{2}'", _gitExecutablePath, p.ExitCode, "0"));
```
zhenlan: Should we log which git command caused failure here?

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+                return success;
+            }
+        }
+
+        return success;
+    }
```
zhenlan: Really like how this is implemented - no any external dependency on scripts or whatsoever :clap:

src/System.Private.ServiceModel/tools/PRService/sync.ashx
zhenlan: "C:\Program Files\Git\cmd" is part of PATH environment variable on the to be used server. Do you think if we can avoid hard coding git path somehow and fail if the path to git is not set up in PATH? This way, we can avoid changing code if git is located at a different location than what we are about to hard code. Thoughts?

src/System.Private.ServiceModel/tools/PRService/sync.ashx
zhenlan: If it is necessary, we can call it on every sync (until/unless we see an issue :smile:). I would like to avoid any extra external dependency - no scheduled task.

src/System.Private.ServiceModel/tools/PRService/sync.ashx
zhenlan: I guess it's supposed to be _gitExecutionTimeoutMilliseconds instead of _gitSanityExecutionTimeoutMilliseconds? Looks like _gitSanityExecutionTimeoutMilliseconds will be dead code afterwards.

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+//	fetch = +refs/heads/*:refs/remotes/origin/*
+//	fetch = +refs/pull/*/head:refs/remotes/origin/pr/*      <-- allows us to run git checkout pr/{id} 
+
+public class PullRequestHandler : IHttpHandler
+{
+    private const string _gitExecutablePath = @"C:\Program Files (x86)\Git\bin\git.exe";
+    private const string _gitRepoPath = @"N:\git\wcf-test";
+
+    private const int _gitSanityExecutionTimeoutMilliseconds = 1000;
+    private const int _gitExecutionTimeoutMilliseconds = 60000;
```
zhenlan: Just FYI, ASP.NET has default request timeout of 110s. I am still debating myself if we should impose another timeout or just let it run until the request is aborted. I am okay with current code. Just bring this up for your consideration.

src/System.Private.ServiceModel/tools/PRService/web.config
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0"?>
```
zhenlan: I don't think your code has any dependency on 4.6 particularly (it may even work on 2.0). We potentially can skip web.config if you like smaller footprint :smile:. Your call.

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+        bool success = CheckPrerequisites(result);
+
+        if (!success)
+        {
+            context.Response.Write(result);
+            return;
+        }
+
+        // Check inputs to the script
+        if (!string.IsNullOrEmpty(pr) && VerifyPrId(pr))
```
zhenlan: NIT: if you have string.IsNullOrEmpty check in VerifyPrId, you can avoid it here.

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+        context.Response.Write(string.Format("PR = {0}<br/>", pr ?? "Unspecified"));
+
+        StringBuilder result = new StringBuilder();
+
+        bool success = CheckPrerequisites(result);
+
+        if (!success)
+        {
+            context.Response.Write(result);
+            return;
```
zhenlan: We will miss setting failure status code in this case.

src/System.Private.ServiceModel/tools/PRService/sync.ashx
```diff
@@ -0,0 +1,227 @@
...
+        else
+        {
+            context.Response.Write(string.Format("Invalid SHA or PR ID specified<br/>"));
+            success = false;
+        }
+
+        if (context.Request.IsLocal)
+        {
+            // Don't allow error results to leak to non-local machines
+            context.Response.Write(result);
```
zhenlan: Just a comment, this might be useful to see in a failure case. Let's see if we often run into this in production and then decide if we need to do something to make this information more accessible.

src/System.Private.ServiceModel/tools/PRService/sync.ashx
iamjasonp: Every sync it is :smile:

src/System.Private.ServiceModel/tools/PRService/sync.ashx
iamjasonp: Done

src/System.Private.ServiceModel/tools/PRService/sync.ashx
iamjasonp: Changed logic

src/System.Private.ServiceModel/tools/PRService/web.config
iamjasonp: Will leave it in here, if only because it's easier to set debug=true here

src/System.Private.ServiceModel/tools/PRService/sync.ashx
iamjasonp: I ran into an issue with a 60s timeout, actually, turns out that a sync can take a long, long time, either due to slow connection or necessity to pull a lot of files. I kind of like having the timeout here so that we get the execution result when an error happens. 

Based on your feedback for 110s ASP.NET request timeout, I upped the one here to 100s.

src/System.Private.ServiceModel/tools/PRService/sync.ashx
iamjasonp: Yep, the idea was that we could find out quickly and bail if git sanity didn't work out. I changed and removed.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -0,0 +1,217 @@
...
+        {
+            // Check inputs to the script
+            uint pr;
+            if (!string.IsNullOrEmpty(prString) && uint.TryParse(prString, out pr))
+            {
+                success = SyncToPr(pr, result);
+            }
+            else
+            {
+                context.Response.Write(string.Format("Invalid PR ID specified<br/>"));
```
roncain: Nit: it looks like string.Format is no longer needed

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -0,0 +1,217 @@
...
+        } 
+        
+        if (!success)
+        {
+            context.Response.StatusCode = 400;
+        }
+
+        if (context.Request.IsLocal)
+        {
+            // Don't allow error results to leak to non-local machines
```
roncain: Question only: what are the consequences of this IsLocal test for our potential PR CI's?  Does this means we won't see error information necessary to diagnose what went wrong?

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -0,0 +1,217 @@
...
+        // Do a sanity execute of the git executable 
+
+        try
+        {
+            string[] gitCommands = new string[]
+            {
+                "status"
+            };
+
+            RunGitCommands(gitCommands, executionResult);
```
roncain: Nit: check success return from the 'git status'?  If that fails, I don't believe we will see an exception and CheckPrerequisites will return true.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -0,0 +1,217 @@
...
+            else
+            {
+                context.Response.Write(string.Format("Invalid PR ID specified<br/>"));
+                success = false;
+            }
+        } 
+        
+        if (!success)
+        {
+            context.Response.StatusCode = 400;
```
roncain: 400 is correct only for a client error, like a bad PR.  Anything else (git failure, repo path problem, etc.) should be 500.  Can we make a CheckPrerequisites failure set status code to 500 and only set the 400 inside the block specifically parsing and syncing the PR?  That would at least alert us from a CI machine we are talking to an improperly configured machine or something more serious needs attention.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -0,0 +1,217 @@
...
+
+    private bool SyncToPr(uint pr, StringBuilder executionResult)
+    {
+        string[] gitCommands = new string[]
+        {
+            "clean -fdx",
+            "checkout master",
+            "fetch --all --prune",
+            "merge --ff-only origin/master",
+            string.Format("fetch origin pull/{0}/head:pr/{0}", pr),
```
roncain: Something to consider in future -- technically any failure here should be a 500 -- except for the fetch of the PR, which would likely be due to a bogus PR#.  I know, I can hear the argument we won't call this service with a bogus PR, but if we go there and say requests are always well formed and correct, then 400 would never be a correct response.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: IsLocal basically is 'am I running localhost'. So yes, IsLocal would suppress this debug spew when being called from CI. 

zhenlan had mentioned this as a potential issue as well. What are your thoughts? I kind of want to keep this output hidden mostly because I don't really want our internal implementations to be shown in its gory detail in CI output. The trade off here is that when something _does_ go wrong, it should be easy enough to go to that CI box and change pr.ashx to show this debug information as needed. 

thoughts?

roncain: Personally I'd prefer to spew everything that went wrong and find a way to protect the PRService from being used by anything other than the CI Jobs, especially while we're kicking the tires on it.  But the securing of PRService can wait till we need it.

iamjasonp: Alright - I got rid of the IsLocal check, as I think the discussions we've had err on the side of "show more things" vs. "keep things private"

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: Good catch - thanks!

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: For now, anything that is in the CheckPrerequisites and CleanRepo steps are now returning 500

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: True... assuming that nobody touches the enlistment though, we should be reasonably safe on the steps here, especially when coupled with the Cleanup step introduced earlier. 

For now though, let's keep errors here as 400s - it would take some refactoring to match commands with what error code they should produce.

iamjasonp:
Reviewers - I pushed out a latest version but _have not_ yet tested it (my main test machine seems unreachable), so there may be some syntax errors that are uncaught.

roncain:
LGTM, but I don't know how to buddy-test it.  @iamjasonp -- your call on when to merge subject to any further feedback from @zhenlan and your ability to test it.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -0,0 +1,230 @@
...
+        } 
+        
+        if (!success)
+        {
+            // it's most likely that the client supplied a bad PR ID, but there could potentially 
+            // be other errors in SyncToPr. For now, return 400
+            context.Response.StatusCode = 400;
+        }
+
+        context.Response.Write(HttpUtility.HtmlEncode(result));
```
iamjasonp: Per previous feedback, I've removed the check for the request.IsLocal here.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -0,0 +1,230 @@
...
+        bool success = CheckPrerequisites(result);
+
+        if (!success)
+        {
+            // CheckPrerequisites failed, this is a server issue
+            context.Response.StatusCode = 500;
+            context.Response.Write(HttpUtility.HtmlEncode(result));
+            return;
+        }
+        else
```
zhenlan: Nit: I know there are many ways to do this, but just to follow the way you do it with CheckPrerequisities, if you remove the else and repeat the same pattern as below in the next a few blocks, it will improve the readability of this code a lot.
```
success = DoWork();
if (!success) {
    context.Response.StatusCode = 500;
    context.Response.Write(HttpUtility.HtmlEncode(result));  
    return;
}

success = DoAnotherWork();
if (!success) {
    context.Response.StatusCode = 500;
    context.Response.Write(HttpUtility.HtmlEncode(result));  
    return;
}
```

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: We don't care git output in success case. You may consider to move this line into the ```if (!success)``` clause above.

zhenlan:
I see we are trying to create some protocols here - which HTTP status code to return for what type of errors. Please don't unless/until necessary. Do we differentiate different HTTP status codes on the caller side (CI)? My understanding is we don't - the caller should only care if the returned HTTP status code is 200 for success; otherwise it's a failure for whatever code. Keeping the logic of this code simple so we will less frequently need to make a change in the future is part of the goal.

My recommendation: I don't care it's 500 or 400, but please use only one across the board for all failure cases.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: This is kind of exactly the reason I suggest we use one code for all failure cases. When you introduce more error codes, the logic naturally becomes more complex if you want to get it right.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -0,0 +1,230 @@
...
+using System.Diagnostics;
+using System.Web;
+using System.IO;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Web;
+
+public class PullRequestHandler : IHttpHandler
+{
+    private const string _gitExecutablePath = @"C:\Program Files\Git\bin\git.exe";
```
zhenlan: Any thoughts about my [earlier comment](https://github.com/dotnet/wcf/pull/981/commits/f682c603c43ae9a6efe188e8aa78855cc8660f64#r58649092)? If not, this should be hard coded as **C:\Program Files\Git\cmd\git.exe** for the to be used server.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -0,0 +1,230 @@
+ // Copyright (c) Microsoft. All rights reserved.
+ // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+ 
+<%@ WebHandler Language="C#" Class="PullRequestHandler" %>
+
+// This script is called to prepare a local repo for testing
+// ASSUMES: 
+//   - Git is installed on the system to "C:\Program Files\Git\bin\git.exe"
```
zhenlan: Related to my [question below](https://github.com/dotnet/wcf/pull/981/files#r58778632), but it needs to be updated either way.

src/System.Private.ServiceModel/tools/PRService/web.config
zhenlan: Sounds good. We can also use it to bump up ASP.NET default request timeout (executionTimeout) if we see timeout becomes an issue in production.

roncain:
I disagree about using a single error code but not enough to block progress on this PR.  We legitimately have 2 types of errors possible here -- client and server-side.  The Http protocol is very clear 4xx means client and 5xx means server.  I could live with 5xx but I think it's a mistake for a service like this to depart from the Http protocol.

zhenlan:
This can go much deeper if we want to do it right, for example, 401 for auth failure, 404 for unexisting PR, 408 for time out and so on (I know, which code to use may be even debatable). On top of this, ASP.NET and IIS will have other 4xx and 5xx codes for failure cases too. My question is "does the caller need different error codes?". My answer is "no", which is the reason I suggest us not to go down this path because it takes much more effort to get it right while giving us little value back.

I am not against using either 5xx or 4xx code at all. If we want to differentiate our own error from errors of ASP.NET and IIS, we can even pick an unique 5xx or 4xx code.

roncain:
I recommend we obey Http protocol and return 400 for obvious client request mistakes. And return 500 for obvious server problems. There is no need to differentiate further. But 400 suggests CI  Job problems, and 500 suggests PrService configuration problems. That is useful information to debug failed CI runs. CI job should report anything other than 200 and let the dev investigate.

iamjasonp:
I actually prefer having two error cases in this case, with the loosened definitions of:

* _Likely to be_ a server issue (500) 
* _Likely to be_ a client issue (400)

I was thinking hard before actually making that change as (as @zhenlan mentioned) as it slightly complicates the logic of the PRService. 

The reason why I want both is because it helps with Jenkins error reporting. We use this thing called the "Failure Cause Analyzer" in Jenkins to give human readable instructions to the person who's running the job. 

As it's likely that any git.exe output will be substantially similar in this case, the error code is something we give as a hint as to what we tell the user. 

For example (taking some current examples): 
* NuGet restore timed out. This can occur when the machine is under high load. It's a known NuGet issue, please retry the job.	
* Disk has run out of space and cannot get latest changes. Please contact netciadmin@microsoft.com.

The first is an error that's likely to have been a transient issue and you can retry your job, the second is unrecoverable and therefore you should abandon all hope until someone takes a look :smile: 

I _do not_ propose we go and fine-tune the 50x or 40x error codes unless we need to; rather, we leave it as it stands today for usability reasons as we start using this in CI.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: Thanks - I missed your earlier comment on this. I'll look into this.

iamjasonp:
I propose we move forward - that is, hard code to what you suggest above, and then we can take this improvement (looking into PATH, as well as logic simplifications) as next-step changes. 

Then we can start turning on this piece - @zhenlan will spin up the server pieces and I can hook Jenkins in.

This way we can test how PRService behaves when it's changed

zhenlan:
@iamjasonp is the owner of this PR. I respect you to make the final call :smile:. I agree to move forward, so LGTM.

Just to show it is not easy to get HTTP status code right: what status code do you expect from request like "pr.ashx?pr=\<script\>"? Is this client's fault or server's? :wink:

iamjasonp:
I tried _exacty_ that actually. ASP.NET will block the call with a 400 :smile:

iamjasonp:
FYI: http://go.microsoft.com/fwlink/?LinkID=212874

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: For now I want it to be verbose so I can look at how git.exe behaves when I put it into CI. 
Once we've verified things work and we know it's stable, then I will suppress as we had in previous versions of this PR

PR #986, opened 2016-04-07
Fix pr.ashx returning false negatives running git

iamjasonp:
RunGitCommands() would return false in the case where the method was passed zero commands to run. Now, if it's passed with a null or empty array, we will bail out early and return true to prevent the rest of the script from failing

skip ci please 

//cc @zhenlan


zhenlan:
LGTM

Issue #985, opened 2016-04-07
Improvements to PRService: leftover work from #981


iamjasonp:
Leftover work items from #981:

* [x] Improve git executable location detection by leveraging the system %PATH% variable, if exists
* [x] Aesthetics: change logic in the !success case to be more simple and consistent https://github.com/dotnet/wcf/pull/981#discussion_r58769501
* [x] Suppress git.exe output on success case https://github.com/dotnet/wcf/pull/981#discussion_r58815094
* [x] Allow PRService to handle branches other than "master" 
* [x] Secure the service by limiting the branches and PRs that can be used
* [x] Secure the service... for real #1054
* [ ] Auto-updating PRService upon commit to master / upon update of a tag

iamjasonp:
> Auto-updating PRService upon commit to master / upon update of a tag

Considered out of scope for now, we rarely do PRService updates and it's easy enough just to do a git merge/checkout

PR #932, opened 2016-03-23
Modify netci.groovy to include new sync-pr scripts

iamjasonp:

PR #987, opened 2016-04-07
Add ability to use PRService on branches

iamjasonp:
This replaces the earlier SHA functionality. We'll need this to be able to check out feature or "master" branches for testing, for example. 

Note that we currently by design limit ourselves to allowing only "master" branch checkout. 

skip ci please

### DIFF
--- a/src/System.Private.ServiceModel/tools/PRService/pr.ashx
+++ b/src/System.Private.ServiceModel/tools/PRService/pr.ashx
@@ -11,9 +11,11 @@
 //   - 
 //   - The Git repo is set up as follows: 
 //     [remote "origin"]
-//	       url = https://github.com/dotnet/wcf
-//	       fetch = +refs/heads/*:refs/remotes/origin/*
-//	       fetch = +refs/pull/*/head:refs/remotes/origin/pr/*      <-- allows us to run git checkout pr/{id} 
+//         url = https://github.com/dotnet/wcf
+//         fetch = +refs/heads/*:refs/remotes/origin/*
+//         fetch = +refs/pull/*/head:refs/remotes/origin/pr/*      <-- allows us to run git checkout pr/{id} 
+//
+// We also support checking out branches, but we limit the the branches we can check out
 
 using System;
 using System.Collections.Generic;
@@ -21,8 +23,6 @@ using System.Diagnostics;
 using System.Web;
 using System.IO;
 using System.Text;
-using System.Text.RegularExpressions;
-using System.Web;
 
 public class PullRequestHandler : IHttpHandler
 {
@@ -32,13 +32,11 @@ public class PullRequestHandler : IHttpHandler
 
     public void ProcessRequest(HttpContext context)
     {
-        _gitRepoPath = context.Request.PhysicalApplicationPath + "..\\..\\..\\..";
-        string prString = context.Request.QueryString["pr"];
+        // Check for prerequisites needed before this script can execute
 
-        context.Response.Write(string.Format("PR = {0}<br/>", HttpUtility.HtmlEncode(prString) ?? "Unspecified"));
+        _gitRepoPath = context.Request.PhysicalApplicationPath + "..\\..\\..\\..";
 
         StringBuilder result = new StringBuilder();
-
         bool success = CheckPrerequisites(result);
 
         if (!success)
@@ -48,10 +46,8 @@ public class PullRequestHandler : IHttpHandler
             context.Response.Write(HttpUtility.HtmlEncode(result));
             return;
         }
-        else  
-        {
-            success = CleanupMergedBranches(result); 
-        }
+
+        success = CleanupMergedBranches(result);
 
         if (!success)
         {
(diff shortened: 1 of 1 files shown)


iamjasonp:
Rudimentarily tested on my own test box and confirm that I can sync to master but no other branches.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -33,9 +35,11 @@ public class PullRequestHandler : IHttpHandler
     public void ProcessRequest(HttpContext context)
     {
         _gitRepoPath = context.Request.PhysicalApplicationPath + "..\\..\\..\\..";
+        string branchString = context.Request.QueryString["branch"];
```
roncain: Question only: I believe QueryString[] returns null if the given query parameter is not present.  Is that correct?  And if so, is HtmlEncode and other uses robust against nulls?

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -60,26 +62,35 @@ public class PullRequestHandler : IHttpHandler
...
+        {
+            // We only want to support limited branches as we only use this for PR purposes
+            switch (branchString)
             {
-                context.Response.Write("Invalid PR ID specified<br/>");
-                success = false;
+                case "master":
+                    success = SyncToBranch(branchString, result);
+                    break;
+                default:
```
roncain: Sidebar: should we open an issue for handling other branches or add it to #985?

roncain:
LGTM

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: I'll add an issue to it #985

zhenlan:
I still don't completely understand what is the scenario where we need to call pr.ashx?branch=master. Can you please help me to understand?

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -137,22 +148,36 @@ public class PullRequestHandler : IHttpHandler
         return RunGitCommands(gitCommands, executionResult);
     }
 
+    private bool SyncToBranch(string branch, StringBuilder executionResult)
+    {
+        string[] gitCommands = new string[]
+        {
+            "clean -fdx",
+            "checkout master",
```
zhenlan: Since you asked people to pass in `branch` as a parameter to the function, please use it :smile:.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -60,26 +62,35 @@ public class PullRequestHandler : IHttpHandler
...
+        else if (!string.IsNullOrEmpty(branchString))
+        {
+            // We only want to support limited branches as we only use this for PR purposes
+            switch (branchString)
             {
-                context.Response.Write("Invalid PR ID specified<br/>");
-                success = false;
+                case "master":
+                    success = SyncToBranch(branchString, result);
+                    break;
```
zhenlan: If you support other branches in the future, do you expect people calling a function other than `SyncToBranch`? If the answer is "no", then here should not be a `switch` but really a `if (branchstring=="master")` clause for today and it will go away when you support branches in general.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -60,26 +62,35 @@ public class PullRequestHandler : IHttpHandler
...
-            // Check inputs to the script
-            uint pr;
-            if (!string.IsNullOrEmpty(prString) && uint.TryParse(prString, out pr))
-            {
-                success = SyncToPr(pr, result);
-            }
-            else
+            success = SyncToPr(pr, result);
+        }
+        else if (!string.IsNullOrEmpty(branchString))
```
zhenlan: For a request like `pr.ashx?pr=123&branch=1.0.0-rc2`, we will actually check out pr=123 from master branch anyway and ignore branch=1.0.0-rc2. This does not feel right. I see what you really mean is that `pr` and `branch` should not be specified at the same time.

My recommendation: just like how we implement any other APIs, move all the parameter validation logic to the beginning of this function including null/empty check, tryparse, no-coexistence, master branch only and so on. This avoids calling `CheckPrerequisites` and `CleanupMergedBranches` to just find out the input parameters are invalid. This will also make your code here much nicer just like any other blocks you had.
```
success = pr>0 ? SyncToPr(pr, result) : SyncToBranch(branchString, result);
if (!success) {
    //... ...
}
```

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: No, we wouldn't ever use it here. This is for the merge step from origin/master, and hence this will never vary.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: I think that's just style preference. 

In the future this will go away once we decide what branches we want to support (probably all of them) and a better way to validate input.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: I think there's a misunderstanding on how branches work: 

* You would never attempt to checkout a PR from "a branch", because the PR implies the branch already
* When you check out a branch, you necessarily do not specify a PR number because you are asking to sync to the branch HEAD
* When you specify both, I could either output an error, or just make the best guess assumption. In this case, I assume PR take precedence
* Our CI systems will _never_ specify both

I also disagree here about `CheckPrerequisites`. All the prereqs are a necessary condition for this script to work, so I don't care about the input if (say) git is not available or the filesystem is not available. Furthermore, the later implementation of checking for branch valid input is going to rely on the script being able to run `git branch -r` and checking against a list of valid branches on `origin`. 

CleanupMergedBranches might be avoidable with invalid input, but I view CleanupMergedBranhes as an extension of CheckPrerequisites. All assumptions about checkouts being successful are essentially invalid if I cannot assume a clean branch. Hence, CleanupMergedBranches should stay where it is.

iamjasonp: I could certainly do something about the non-coexistence though. However, given that our assumption is that only known callers (e.g., CI system) will call this service, I have assumed that we own the callers and they wouldn't do anything silly like specifying both PR and branch.

iamjasonp:
> I still don't completely understand what is the scenario where we need to call pr.ashx?branch=master. Can you please help me to understand?

You'll need this when you want to sync an underlying repo to the HEAD of a branch, for example, when the timed twice-daily outerloop CI builds run, or the upon-checkin CI builds run

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: I see. Thanks for the explanation. I missed line 159.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: If you do make that assumption, the whole `if (!string.IsNullOrEmpty(branchString))` can be saved (and probably many other things like the validation of "master") :smile:. Maybe I did not use a good example, this code will work for `pr.ashx?pr=bogus&branch=master`.

Validating input parameters at beginning (you decide where is the right place) is a good practice many people like to do and it helps to make code after that simpler. I am just trying to see how we can make this code reliable as we both know we want to avoid updating this code once it's in action, but I will not block you from merging if you don't think so.

iamjasonp: I agree that I _usually_ would validate right at the beginning, yes (see all the code I write for WCF itself :smile:). This time I feel like verifying prerequisites trumps.

zhenlan:
Thanks for the explanation. Saw your usage in pull #932.

PR #993, opened 2016-04-11
Change quotation in netci.groovy

iamjasonp:
WCF netci.groovy was previously using ''' as a quoting mechanism, but that prevented variables from being evaluated. Change the ''' to "


iamjasonp:
@dotnet-bot test ci please

PR #994, opened 2016-04-11
Remove latest dependences build from netci.groovy

iamjasonp:
... but for now, keep the command around as a comment in case we need it for
the near future. We can remove the comment later should we find that we haven't
needed it for a long time


iamjasonp:
@dotnet-bot test ci please

iamjasonp:
@hongdai - any concerns with removing this?

zhenlan:
Given the packing change due to netstandard support, we will always depend on latest packages and let netstandard decide which part of assets of a package should be used. This latest dependency build should not be needed anymore.

This PR LGTM.

hongdai:
LGTM.

PR #995, opened 2016-04-11
Init Check in IIS Hosted WCF Test Service

hongdai:
* It includes all the sources needed for the IIS hosted WCF Test Services
* The web config file define service activation information.


src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
```diff
@@ -0,0 +1,95 @@
+<?xml version="1.0" encoding="utf-8"?>
+ <!--Copyright (c) Microsoft Corporation.  All Rights Reserved.-->
+<configuration>
+  <system.diagnostics>
+    <sources>
+      <source name="System.ServiceModel" switchValue="Information, ActivityTracing" propagateActivity="true">
+        <listeners>
+          <add name="traceListener" type="System.Diagnostics.XmlWriterTraceListener" initializeData="c:\log\Traces.svclog" />
```
zhenlan: Can we avoid hard coding path like this because when we have multiple instances of WCF service hosted on one machine they will be overwritten by each other? Not sure if we can use relative path and save trace log in bin folder or something alike.

hongdai: I believe the same trace files can be shared by multiple service instances, i.e. they will co-exist and can be filtered out by process IDs. 

Shall I change the default value from Information to Warning to avoid the trace file growing too fast?

hongdai: I will also investigate whether I can use relative path.

zhenlan: If the same trace file can be shared by multiple service instances, it will be fine. Yes, relative path will be nicer, eg., [path to root of repo]\..\WcfServiceLog\traces.svclog. This is not blocking.

During the setup we need to remember to grant corresponding app pool write permission to this folder.

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
```diff
@@ -0,0 +1,95 @@
...
+        <add name="UrlRoutingModule" type="System.Web.Routing.UrlRoutingModule,             System.Web.Routing, Version=4.0.0.0,             Culture=neutral,             PublicKeyToken=31BF3856AD364E35" />
+    </modules>
+   <handlers>
+    <add name="UrlRoutingHandler" preCondition="integratedMode" verb="*" path="UrlRouting.axd" type="System.Web.HttpForbiddenHandler,            System.Web, Version=4.0.0.0, Culture=neutral,            PublicKeyToken=b03f5f7f11d50a3a" />
+   </handlers>
+    <directoryBrowse enabled="true" />
+        <staticContent>
+            <clientCache cacheControlMode="NoControl" />
+        </staticContent>
+  </system.webServer>
```
zhenlan: NIT: indentation

Curious, what made us need this?

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
```diff
@@ -0,0 +1,95 @@
...
+    </serviceHostingEnvironment>
+  </system.serviceModel>
+  <system.webServer>
+     <modules runAllManagedModulesForAllRequests="true">
+        <add name="UrlRoutingModule" type="System.Web.Routing.UrlRoutingModule,             System.Web.Routing, Version=4.0.0.0,             Culture=neutral,             PublicKeyToken=31BF3856AD364E35" />
+    </modules>
+   <handlers>
+    <add name="UrlRoutingHandler" preCondition="integratedMode" verb="*" path="UrlRouting.axd" type="System.Web.HttpForbiddenHandler,            System.Web, Version=4.0.0.0, Culture=neutral,            PublicKeyToken=b03f5f7f11d50a3a" />
+   </handlers>
+    <directoryBrowse enabled="true" />
```
zhenlan: Disable directory browsing if we don't need it.

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
```diff
@@ -0,0 +1,95 @@
...
+        <add factory="WcfService.WebSocketHttpsRequestReplyTextBufferedTestServiceHostFactory" service="WcfService.WSRequestReplyService" relativeAddress="~\WebSocketHttpsRequestReplyTextBuffered.svc" />
+        <add factory="WcfService.WebSocketHttpsDuplexBinaryBufferedTestServiceHostFactory" service="WcfService.WSDuplexService" relativeAddress="~\WebSocketHttpsDuplexBinaryBuffered.svc" />
+        <add factory="WcfService.WebSocketHttpsDuplexTextBufferedTestServiceHostFactory" service="WcfService.WSDuplexService" relativeAddress="~\WebSocketHttpsDuplexTextBuffered.svc" />
+      </serviceActivations>
+    </serviceHostingEnvironment>
+  </system.serviceModel>
+  <system.webServer>
+     <modules runAllManagedModulesForAllRequests="true">
+        <add name="UrlRoutingModule" type="System.Web.Routing.UrlRoutingModule,             System.Web.Routing, Version=4.0.0.0,             Culture=neutral,             PublicKeyToken=31BF3856AD364E35" />
+    </modules>
```
zhenlan: Trying to understand what made us need url routing?

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
hongdai: From my investigation:
1. When the same file name specified, the subsequent service instances will get a trace file name prefix with a Guid automatically. I.E we can use the same file name, in the run time, a different file name will be automatically generated per service instance.
2. Relative to the root of repo is impossible to predefine in the config file, and it's also not recommended to put log file under vroot for the security reason.

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
zhenlan: Okay. I think this is probably needed because of ICtlService.GetCrl.

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
zhenlan: I see. This (ClientCache) is probably needed because of test.crl download. Just wondering why we don't use cacheControlMode="DisableCache" to disable cache explicitly. This is not blocking.

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
zhenlan: In the case of CI, there are multiple repos. It will be hard to tell which trace belongs to which service when trace files are only differentiated by prefixed GUIDs. I was wondering if something like this will work: `initialzeData="..\..\..\bin\WcfServicelog\traces.svclog"`, which is supposed to point to `repo-root\bin` and `bin` folder is one of git ignored folders.

If above does not work (or you don't like it for any reason), can you please just update the hard coded path to `c:\WCFTest\log\traces.svclog`? Because there is another hard coded path `c:\WCFTest\test.crl` in `Util.cs`, putting all external resources/dependencies in one root folder (`c:\WCFTest`) will make setup easier.

> Shall I change the default value from Information to Warning to avoid the trace file growing too fast?

Warning may be the right level of logging as most of the time we care only when there are problems, but I don't feel strongly in any case.

zhenlan:
This LGTM subject to the conclusion of the [hard coded path discussion](https://github.com/dotnet/wcf/pull/995#discussion_r59308816).

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
hongdai: I think the relative path defined in web.config is probably relative to w3wp.exe location, not the vroot. I will give initialzeData="..\..\..\bin\WcfServicelog\traces.svclog" a try.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/MahojongTypes.cs
```diff
@@ -0,0 +1,240 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
iamjasonp: I seemed to recall we were trying to purge these Mahjong tests - are we still running them?

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/MahojongTypes.cs
```diff
@@ -0,0 +1,240 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace WcfService
+{
+    public class MahojongTypes
```
iamjasonp: NIT for much later: "mahjong" not "mahojong" :smile:

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfRestartService.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
iamjasonp: Should these tests still exist? I really don't like the idea of killing IIS to test things

Not blocking

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
hongdai: I've just verified the relative path was in fact relative to the vroot! The only issue is the bin directory is not created when we don't need to build the src. In such case, we will need to create the bin directory after git sync.

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
hongdai: I will fix the indent. 
I disabled the cache from UI, not sure why it's shown up as "No
control" instead of "DisableCache", I will update it manually.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/MahojongTypes.cs
hongdai: I'd suggest we don't purge things in the init checkin as it's time consuming to decide what are really not being used. The only certain way is to do a test run.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfRestartService.cs
hongdai: This test has been changed to kill a dedicate servicehost instance instead of the whole process some time ago:-)

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
hongdai: Agreed.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/MahojongTypes.cs
hongdai: Agreed :-)

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfRestartService.cs
iamjasonp: Does this require some special setup in the IIS world?

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfService.cs
```diff
@@ -0,0 +1,406 @@
...
+
+            ServiceHost host = new ServiceHost(typeof(WcfRestartService));
+            host.AddServiceEndpoint(typeof(IWcfRestartService), binding, localHost + path);
+            host.Open();
+
+            // Add the ServiceHost instance to a static dictionary so that it can be used by restart service operation to close the ServiceHost
+            WcfRestartService.serviceHostDictionary.Add(guid, host);
+
+            // Return the unique endpoint for this ServiceHost instance of the WcfRestartService
+            return "http://[HOST]" + path;
```
iamjasonp: Happened to spot this - is `[HOST]` a string to be replaced?

iamjasonp:
Consider running a CodeFormatter run over all the files you're (re-)committing. There are a number of formatting issues but I haven't pointed them out in the code itself

iamjasonp:
LGTM for initial commit. We can iterate on this as we go along

iamjasonp:

Let's try a small test to see the deployment in action ;)

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
zhenlan: Just checked, the *.svclog itself is actual on the .gitignore file. This means you can put the .svclog file anywhere and git will not bother it. I was hoping WCF tracing would create the folder if it did not exist. If this is not the case, you can put .svclog at the root of the repo or any other existing folders. You decide.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/ComplexCompositeType.cs
```diff
@@ -0,0 +1,275 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.Serialization;
+using System.Text;
+
+namespace WcfService
+{
+    [DataContract(Namespace = "http://www.contoso.com/wcfnamespace")]
+    public class ComplexCompositeType : IEquatable<ComplexCompositeType>
```
roncain: What is your plan for files that need to be shared by IIS, SelfHost and test client?  At one point I proposed a separate folder (SharedTestTypes?) reserved for service interface types and data types that both server and client side need to use.  If we can collect them into a common folder, it would be easier to link to them.  Not strictly blocking for this PR, but we need a solution by the time SelfHost and Client code PR's start arriving.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfDuplexService.cs
```diff
@@ -0,0 +1,84 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.ServiceModel;
+using System.Threading.Tasks;
+
+namespace WcfService
+{
+    [ServiceBehavior(AddressFilterMode = AddressFilterMode.Any)]  
+    public class WcfDuplexService : IWcfDuplexService, IWcfDuplexService_DataContract, IWcfDuplexService_Xml
```
roncain: Not required for this checkin, but I think it would be easier to maintain if we had a separate folder for services (testservices?) like you do for testhosts.  We can discuss or rearrange after checkin.  I'm thinking something like App_code\testtypes, App_code\testservices, and App_Code\testhosts.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/Util.cs
```diff
@@ -0,0 +1,166 @@
...
+        Stream GetCrl(string serialNum);
+    }
+
+    public class CrlService : ICrlService
+    {
+        public Stream GetCrl(string serialNum)
+        {
+            // The test.crl is generated by the cert util tool and will not expire until the root cert expires.
+            // We should investigate if we can generate it in the run time. This is not a blocking issue.
+            string downloadFilePath = @"c:\\WCFTest\\test.crl";
```
roncain: Add this to the list of hard-coded paths we need to discuss/revisit.  Not blocking for this checkin.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/AuthenticationResourceHelper.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Linq;
+using System.ServiceModel;
+
+namespace WcfService
+{
+    public static class AuthenticationResourceHelper
```
roncain: Not required for checkin, but we probably need to make a sweep and remove "Resource" from names.  That's a hold-over from Bridge naming

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/TestServiceHostBase.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (debugBehavior == null)
+            {
+                debugBehavior = new ServiceDebugBehavior();
+                this.Description.Behaviors.Add(debugBehavior);
+            }
+            debugBehavior.IncludeExceptionDetailInFaults = true;
+            //Now that we've populated the ServiceDescription, we can reach into it
+            //and do interesting things (in this case, we'll add an instance of
+            //ServiceMetadataBehavior if it's not already there.
+            ServiceMetadataBehavior mexBehavior = this.Description.Behaviors.Find<ServiceMetadataBehavior>();
```
roncain: What is the purpose of adding this mexBehavior to all our service hosts?
I didn't think we had any way to access mex info from the client. 
Also, we don't support named pipes

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/RequestLoggingServiceBehavior.cs
```diff
@@ -0,0 +1,138 @@
...
+
+           The constructor takes a delegate of type RequestLoggingServiceBehavior.LogRequest. There is a simple static logger that can
+           be used that's available as RequestLoggingServiceBehavior.SimpleConsoleLogger and is used in the example above. As the logger
+           is a delegate, you can add multiple loggers and they will be called one after the other. To add more loggers, use the AddLogger
+           method.
+           */
+        private LogRequest _logCallback;
+
+        public delegate void LogRequest(OperationContext operationContext, ref Message message, bool authenticated);
+        public RequestLoggingServiceBehavior() { }
```
roncain: I know we're not pruning on this iteration, but if you're keeping track, I don't think this is being used.  We can keep it for now.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/UtilTestServiceHost.cs
```diff
@@ -0,0 +1,31 @@
+﻿//  Copyright (c) Microsoft Corporation.  All Rights Reserved.
+using System;
+using System.ServiceModel;
+using System.ServiceModel.Activation;
+using System.ServiceModel.Channels;
+
+namespace WcfService
+{
+    public class UtilTestServiceHostFactory : ServiceHostFactory
```
roncain: Not strictly required for checkin, but Util.cs should be renamed and UtilTestServiceHost should be renamed.  The name is too general and doesn't describe what it does.

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
roncain: Good conversation but do we need to get svclog tracing working on this PR?   It feels like a lower-priority issue to deal with later when everything is running.

---
In reply to: [59334362](https://github.com/dotnet/wcf/pull/995#discussion_r59334362) [](ancestors = 59334362)

roncain:
LGTM for initial commit, but please make some work items for what suggestions we don't change in this PR.  I think that includes hard-coded paths, the folder structure to separate service implementation types and shared server/client types, and name fixups (like Util.cs, *Resource*).

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
hongdai: Agree it's lower priority. But since we have reached the conclusion, I will just put the .svclog to the root of the repro.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfRestartService.cs
hongdai: Don't know. This test does not fail in my trial runs, it's either disabled for other issues, or does not require special setup.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfService.cs
hongdai: Don't know the answer - I copied the file from the current service, no changes.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/ComplexCompositeType.cs
hongdai: All *.cs files are shared by IIS and SelfHost. The SelfHost will include AppCode folder. 

For service interface types and data types between service and client, it's duplicated in the current test infrastructure and we haven't discussed if we should redesign. I think this should be tracked by a separate issue and discuss the design over there.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfDuplexService.cs
hongdai: I maintain the same folder structure as the current WCF service, just replace endpoint resource with testhosts. We can discuss rearrange them later on.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/Util.cs
hongdai: I also will open a issue to track the investigation of generate the test.crl in the runtime, thus get rid of hard coded file path.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/AuthenticationResourceHelper.cs
hongdai: Agreed. We can do it once we have everything in place.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/RequestLoggingServiceBehavior.cs
hongdai: I thing pruning on existing service code is out of scope currently. I will open a issue to track the cleaning work.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/ComplexCompositeType.cs
roncain: I think we've already discussed this and said code duplication was to be avoided.  I agree we don't want to clean up old mistakes in this PR, but this PR already is a fundamental redesign of the location of these kinds of files. Don't block this PR on this discussion, but I recommend when the client PR becomes available we don't duplicate code.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/TestServiceHostBase.cs
hongdai: It was for testing purpose, I can test the service is there by simply hitting the mex endpoint for http. I'll keep it for http only.

hongdai:
Given I have LGTM, I'll merge once the innerloop passed.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/TestServiceHostBase.cs
zhenlan: Just to provide the extra data - mex is useful for our ASR tooling work. We will want to make sure our ASR tool can work well with all services that are used by our runtime testing.

PR #996, opened 2016-04-11
Update buildtools, update use of test-runtime

mellinoe:
Newer versions of buildtools are going to change how the test runtime is used and restored. Instead of having a "TestRuntimeProjectJson" property which is used to restore and deploy test assets, each individual test project must now reference a test-runtime it wants to use. This lets us do a couple things:

* Resolve all nuget dependencies at once. Previously, assets from the test-runtime and the test project were resolved separately. These assets were merged very haphazardly and this resulted in various issues due to conflicts being resolved incorrectly.
* More control over test-runtime references.

Goes with https://github.com/dotnet/buildtools/pull/616 and https://github.com/dotnet/corefx/pull/7651

@weshaggard , @roncain, @StephenBonikowsky


mellinoe:
Note: This will fail until a new version of buildtools is published. There's currently some issues with signing, so buildtools hasn't been published at all today. Once that is resolved, I'll update the PR.

weshaggard:
LGTM

src/System.Private.ServiceModel/tests/Common/Infrastructure/tests/project.json
```diff
@@ -5,11 +5,21 @@
...
     "dnxcore50": {
       "imports": "portable-net45+win8"
     }
+  },
+  "runtimes": {
+    "win7-x64": {},
+    "win7-x86": {},
+    "ubuntu.14.04-x64": {},
+    "osx.10.10-x64": {}
   }
```
zhenlan: Any reason this does not include centos, rhel and debian? Even if we include those, I am not sure it's an exhaustive list. Is there a better we don't have to explicitly list all the runtimes (assuming answer is "no" but to confirm)?

mellinoe: For projects that are going to be published for runtime usage, we need to list all of the runtimes they may be used on. Currently, we are relying on the "--infer-runtimes" feature of the CLI, which creates a lock file for your current runtime (that is my understanding of it, anyways). That feature is going away, so we will need to list all of the runtimes regardless.

mellinoe: As for why it doesn't include centos, rhel, and debian: I'm not sure. I just included the runtimes that I already saw in this repo, which didn't include those three. We can augment it if those runtimes are tested on.

roncain: We are testing on centos, rhel and debian nightly, so I think it's appropriate to add them.  @mellinoe is that something you can add to this PR?

iamjasonp: +1. I've been in the process of adding these to CI as well (matching whatever corefx has in Jenkins). Thanks!

mellinoe:
Updated the list of test runtime ID's. These should match what corefx has now, including centos, rhel, and debian.

roncain:
LGTM. Thanks for doing this @mellinoe .  We've done some private testing and intend to merge this as soon as we see https://github.com/dotnet/corefx/pull/7651 is accepted and merged (just in case @weshaggard or @ericstj have any requested changes to that PR).

mellinoe:
@roncain I just merged the corefx PR (after chatting with @ericstj). Let me know if you found anything in the private testing. I needed to clean up some build warnings in corefx caused by the new MSBuild task, but I didn't see any additional ones in WCF that needed cleanup. But there may be some additional configurations I didn't hit with different warnings.

roncain:
Thanks @mellinoe -- I'll merge.  No issues were found in private testing.

zhenlan:
LGTM postmortem. Thanks @mellinoe.

PR #1000, opened 2016-04-12
Disable the debug mode compilation of WCF service

zhenlan:
Debug mode could take way too long and slow down our start up unnecessarily.

### DIFF
--- a/src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
+++ b/src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
@@ -11,9 +11,7 @@
     </sources>
   </system.diagnostics>
   <system.web>
-    <compilation debug="true">
-</compilation>
-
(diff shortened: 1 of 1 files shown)


roncain:
LGTM subject to 2 questions -- will this adversely affect any dev debugging scenarios?  I'm thinking of things like debug-only Asserts and pdb's.  And do we have any numbers on the slowdown you describe?

zhenlan:
>will this adversely affect any dev debugging scenarios? I'm thinking of things like debug-only Asserts and pdb's.

No, Devs can change debug to true as needed in their own hosted server for debugging purpose, but it does not mean we should deploy debug=true in our production.

> And do we have any numbers on the slowdown you describe?

No, I don't have numbers, but not having debug=true in production is general recommendation from ASP.NET. [ScottGu said so](http://weblogs.asp.net/scottgu/442448) too :smile:.

PR #1002, opened 2016-04-12
Init checkin of client changes

hongdai:
* Allow Test property ServiceUri.
* Note that we in the self hosted case, we only allow localhost, not multiple
service instances(localhost/wcfservice1).
* Change all endpoints point to the new services.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
@@ -29,7 +29,7 @@ namespace Infrastructure.Common
         private const string ClientCertificateSubject = "WCF Client Certificate";
         private const string ClientCertificatePassword = "test"; // this needs to be kept in sync with the Bridge configuration 
 
-        public static string LocalCertThumbprint { get; private set; }
+        public static string LocalCertThumbprint { get; set; }
 
         // Dictionary of certificates installed by CertificateManager
         // Keyed by the Thumbprint of the certificate
@@ -246,7 +246,7 @@ namespace Infrastructure.Common
 
         // Adds the given certificate to the given store unless it is
         // already present.  Returns 'true' if the certificate was added.
-        private static bool AddToStoreIfNeeded(StoreName storeName,
+        public static bool AddToStoreIfNeeded(StoreName storeName,
                                                StoreLocation storeLocation,
                                                X509Certificate2 certificate)
         {
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
@@ -105,7 +105,11 @@
     <PropertyGroup Condition="'$(Kerberos_Available)' == ''">
       <Kerberos_Available>false</Kerberos_Available>
     </PropertyGroup>
-  
+
+    <PropertyGroup Condition="'$(ServiceUri)' == ''">
+      <ServiceUri >localhost</ServiceUri >
+    </PropertyGroup>
+
   <!--
      GeneratedTestPropertiesFileName:
      The full path of the C# file that will be generated at build time to
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
@@ -1,49 +1,183 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+using System.ServiceModel;
+using System.Security.Cryptography.X509Certificates;
+
 using Infrastructure.Common;
 
 public static partial class Endpoints
 {
+    private static X509Certificate2 rootCert;
+    private static X509Certificate2 clientCert;
+    private static object rootCertLock = new object();
+    private static object clientCertLock = new object();
+    //Install Root certificate, this is required for all https test
+    private static void EnsureRootCertificateInstalled()
+    {
+        lock (rootCertLock)
+        {
+            if (rootCert == null)
+            {
+                BasicHttpBinding basicHttpBinding = new BasicHttpBinding();
+                ChannelFactory<IUtil> factory = null;
+                IUtil serviceProxy = null;
+                try
+                {
+                    factory = new ChannelFactory<IUtil>(basicHttpBinding, new EndpointAddress(ServiceUtil_Address));
+                    serviceProxy = factory.CreateChannel();
+                    rootCert = new X509Certificate2(serviceProxy.GetRootCert(false));
+
+                    if (rootCert == null)
+                    {
+                        //throw
+                        throw new Exception("Failed to obtain root cert from the server");
+                    }
+
+                    BridgeClientCertificateManager.InstallCertificateToRootStore(rootCert);
+                }
+                finally
+                {
+                    ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+                }
+            }
+        }
+    }
+
+    //Install client certificate
+    private static void EnsureLocalClientCertifciateInstalled()
+    {
+        EnsureRootCertificateInstalled();
+        lock (clientCertLock)
+        {
+            if (clientCert == null)
+            {
+                BasicHttpBinding basicHttpBinding = new BasicHttpBinding();
+                ChannelFactory<IUtil> factory = null;
+                IUtil serviceProxy = null;
+                try
+                {
+                    factory = new ChannelFactory<IUtil>(basicHttpBinding, new EndpointAddress(ServiceUtil_Address));
+                    serviceProxy = factory.CreateChannel();
+                    clientCert = new X509Certificate2(serviceProxy.GetClientCert(false), "test", X509KeyStorageFlags.PersistKeySet);
+                    if (clientCert == null)
+                    {
+                        //throw
+                        throw new Exception("Failed to obtain client cert from the server");
+                    }
+
+                    BridgeClientCertificateManager.AddToStoreIfNeeded(StoreName.My, StoreLocation.CurrentUser, clientCert);
+                    BridgeClientCertificateManager.LocalCertThumbprint = clientCert.Thumbprint;
+                }
+                finally
+                {
+                    ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+                }
+            }
+        }
+    }
+
+    private static string GetServiceHostName()
+    {
+        return TestProperties.GetProperty(TestProperties.BridgeHost_PropertyName);
+    }
+
+    private static bool IISHosted
+    {
+        get
+        {
+            if (TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName).ToLower()==("localhost"))
+            {
+                return false;
+            };
+
+            return true;
+        }
+    }
+
+    private static Uri BuildBaseUri(string protocol)
+    {
+        var builder = new UriBuilder();
+        builder.Host = TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName);
+        builder.Scheme = protocol;
+
+        if (!IISHosted)
+        {
+            switch (protocol)
+            {
+                case "http":
+                    builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeHttpPort_PropertyName));
+                    break;
+                case "ws":
+                    builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeWebSocketPort_PropertyName));
+                    builder.Scheme = "http";
+                    break;
+                case "https":
+                     builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeHttpsPort_PropertyName));
+                    break;
+                case "wss":
+                    builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeSecureWebSocketPort_PropertyName));
+                    builder.Scheme = "https";
+                    break;
+                case "net.tcp":
+                    builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeTcpPort_PropertyName));
+                    break;
+                default:
+                    break;
+            }
+        }
+
+        return builder.Uri;
+    }
+    public static string GetEndpointAdddress(string endpoint, string protocol = "http")
+    {
+        return string.Format(@"{0}/{1}", BuildBaseUri(protocol), endpoint);
+    }
+
+    public static string ServiceUtil_Address
+    {
+        get { return GetEndpointAdddress("Util.svc//Util"); }
+    }
     // HTTP Addresses
     public static string DefaultCustomHttp_Address
     {
-        get { return BridgeClient.GetResourceAddress("WcfService.TestResources.DefaultCustomHttpResource"); }
+        get { return GetEndpointAdddress("DefaultCustomHttp.svc//default-custom-http"); }
     }
 
     public static string HttpBaseAddress_Basic
     {
-        get { return BridgeClient.GetResourceAddress("WcfService.TestResources.BasicHttpResource"); }
+        get { return GetEndpointAdddress("BasicHttp.svc//Basic"); }
     }
 
     public static string HttpBaseAddress_NetHttp
     {
-        get { return BridgeClient.GetResourceAddress("WcfService.TestResources.NetHttpResource"); }
+        get { return GetEndpointAdddress("NetHttp.svc//NetHttp"); }
     }
 
     public static string HttpSoap11_Address
     {
-        get { return BridgeClient.GetResourceAddress("WcfService.TestResources.HttpSoap11Resource"); }
+        get { return GetEndpointAdddress("HttpSoap11.svc//http-soap11"); }
     }
 
     public static string HttpSoap12_Address
     {
-        get { return BridgeClient.GetResourceAddress("WcfService.TestResources.HttpSoap12Resource"); }
+        get { return GetEndpointAdddress("HttpSoap12.svc//http-soap12"); }
     }
 
     public static string HttpBinary_Address
     {
-        get { return BridgeClient.GetResourceAddress("WcfService.TestResources.HttpBinaryResource"); }
+        get { return GetEndpointAdddress("HttpBinary.svc//http-binary"); }
     }
 
     public static string NetHttpWebSocketTransport_Address
     {
-        get { return BridgeClient.GetResourceAddress("WcfService.TestResources.WebSocketTransportResource"); }
+        get { return GetEndpointAdddress("WebSocketTransport.svc//http-requestreplywebsockets-transportusagealways", protocol:"ws"); }
     }
 
     public static string NetHttpDuplexWebSocket_Address
     {
-        get { return BridgeClient.GetResourceAddress("WcfService.TestResources.DuplexWebSocketResource"); }
+        get { return GetEndpointAdddress("DuplexWebSocket.svc//http-defaultduplexwebsockets", protocol: "ws"); }
     }
     
     public static string HttpProtocolError_Address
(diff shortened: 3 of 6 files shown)


src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
 public static partial class Endpoints
 {
+    private static X509Certificate2 rootCert;
+    private static X509Certificate2 clientCert;
+    private static object rootCertLock = new object();
+    private static object clientCertLock = new object();
+    //Install Root certificate, this is required for all https test
+    private static void EnsureRootCertificateInstalled()
+    {
+        lock (rootCertLock)
```
zhenlan: NIT: null check of `rootCert` before taking a lock.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+            }
+        }
+    }
+
+    //Install client certificate
+    private static void EnsureLocalClientCertifciateInstalled()
+    {
+        EnsureRootCertificateInstalled();
+        lock (clientCertLock)
+        {
```
zhenlan: NIT: null check of `clientCert` before taking a lock.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+                {
+                    ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+                }
+            }
+        }
+    }
+
+    private static string GetServiceHostName()
+    {
+        return TestProperties.GetProperty(TestProperties.BridgeHost_PropertyName);
```
zhenlan: Is this supposed to be changed to `ServiceUri_PropertyName`?

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+    }
+
+    private static bool IISHosted
+    {
+        get
+        {
+            if (TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName).ToLower()==("localhost"))
```
zhenlan: Just to point out technically, ServiceUri==localhost does not necessarily mean the service is not IIS hosted. This will work in our current plan of lab/CI setup. Hopefully we can find a better way in the future, or even better if we don't have to differentiate IIS or self hosted at all.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+        var builder = new UriBuilder();
+        builder.Host = TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName);
+        builder.Scheme = protocol;
+
+        if (!IISHosted)
+        {
+            switch (protocol)
+            {
+                case "http":
+                    builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeHttpPort_PropertyName));
```
zhenlan: Can we just use the same ports (and the standard ports) for both IIS hosted and self hosted services? It will make machine setup much easier. We can do this as a follow up item.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
hongdai: Nope. Bridge host name is just the service full qualified host name. It's used by some of the tests.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
hongdai: I originally had a test property called IISHosted. If there is a requirement for "localhost" with IIS hosted, we could add this test property back. The drawback is one more test property to set.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
hongdai: I don't see a blocking issue to use the same ports. I'll open a follow up issue.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
zhenlan: So do you still need to pass BridgeHost in addition to ServiceUri in new infrastructure? If this is the case, I think we should parse host name (if that's what you need) from ServiceUri instead (trim everything after first '/' or '\').

---
In reply to: [59492053](https://github.com/dotnet/wcf/pull/1002#discussion_r59492053) [](ancestors = 59492053)

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
zhenlan: Agree. Let's not invent something new yet. If needed, we can do smarter detection.

zhenlan:
Other than some NITs for your consideration this LGTM.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
```diff
@@ -105,7 +105,11 @@
     <PropertyGroup Condition="'$(Kerberos_Available)' == ''">
       <Kerberos_Available>false</Kerberos_Available>
     </PropertyGroup>
-  
+
+    <PropertyGroup Condition="'$(ServiceUri)' == ''">
+      <ServiceUri >localhost</ServiceUri >
```
iamjasonp: 1. Is this necessarily correct to have the ServiceUri be this by default? 
2. Should we split out the hostname and basepath like we talked about earlier?

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
iamjasonp: Do we really need a lock here? 

1. The call to `serviceProxy.GetRootCert(false)` should be idempotent during one test run. 
2. BridgeClientCertificateManager will take the appropriate locks and do the appropriate checks to ensure the cert is only installed once per system

src/System.Private.ServiceModel/tests/Common/Scenarios/ServiceInterfaces.cs
```diff
@@ -502,3 +502,13 @@ public interface IWcfAspNetCompatibleService
...
 }
+
+[ServiceContract]
+public interface IUtil
+{
+    [OperationContract]
+    byte[] GetClientCert(bool exportAsPem);
+
+    [OperationContract]
+    byte[] GetRootCert(bool exportAsPem);
```
iamjasonp: NIT: spell out the name, e.g., `GetServiceRootCertificate`

src/System.Private.ServiceModel/tests/Common/Scenarios/ServiceInterfaces.cs
```diff
@@ -502,3 +502,13 @@ public interface IWcfAspNetCompatibleService
     [OperationContract(Action = "http://tempuri.org/IWcfService/EchoTimeAndSetCookie", ReplyAction = "http://tempuri.org/IIWcfService/EchoTimeAndSetCookieResponse")]
     string EchoTimeAndSetCookie(string name);
 }
+
+[ServiceContract]
+public interface IUtil
+{
+    [OperationContract]
+    byte[] GetClientCert(bool exportAsPem);
```
iamjasonp: NIT: spell out name, e.g. `GetMachineClientCertificate` or `GetUserClientCertificate` (there are two types of client certs we may need to use)

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+                IUtil serviceProxy = null;
+                try
+                {
+                    factory = new ChannelFactory<IUtil>(basicHttpBinding, new EndpointAddress(ServiceUtil_Address));
+                    serviceProxy = factory.CreateChannel();
+                    rootCert = new X509Certificate2(serviceProxy.GetRootCert(false));
+
+                    if (rootCert == null)
+                    {
+                        //throw
```
iamjasonp: Delete. It's super obvious that you're throwing an exception here.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+    private static X509Certificate2 clientCert;
+    private static object rootCertLock = new object();
+    private static object clientCertLock = new object();
+    //Install Root certificate, this is required for all https test
+    private static void EnsureRootCertificateInstalled()
+    {
+        lock (rootCertLock)
+        {
+            if (rootCert == null)
+            {
```
iamjasonp: We call `EnsureRootCertificateInstalled()` in multiple places. 

If the root cert fails the first time, by https://github.com/dotnet/wcf/pull/1002/files#r59512010 bullet 1, it should fail every time we call this. Hence, we should use some kind of fail fast mechanism to fail all further calls to EnsureRootCertificateInstalled() in a particular run. 

We can't detect this condition currently because we are only keying off if `rootCert == null`

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+                BasicHttpBinding basicHttpBinding = new BasicHttpBinding();
+                ChannelFactory<IUtil> factory = null;
+                IUtil serviceProxy = null;
+                try
+                {
+                    factory = new ChannelFactory<IUtil>(basicHttpBinding, new EndpointAddress(ServiceUtil_Address));
+                    serviceProxy = factory.CreateChannel();
+                    rootCert = new X509Certificate2(serviceProxy.GetRootCert(false));
```
iamjasonp: NIT: consider using named params here to spell out what's false here - `GetRootCert(exportAsPem: false)`

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+                    {
+                        //throw
+                        throw new Exception("Failed to obtain root cert from the server");
+                    }
+
+                    BridgeClientCertificateManager.InstallCertificateToRootStore(rootCert);
+                }
+                finally
+                {
+                    ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
```
iamjasonp: I'm a _little_ weirded out that we're using a "ScenarioTestHelper" in a file where we're not doing any scenario tests. 
Not blocking.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
iamjasonp: ditto on whether or not we need to take a lock for obtaining the clientCertificate.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+                }
+            }
+        }
+    }
+
+    //Install client certificate
+    private static void EnsureLocalClientCertifciateInstalled()
```
iamjasonp: Typo: EnsureLocalClientCertif**ciate**Installed

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+
 using Infrastructure.Common;
 
 public static partial class Endpoints
 {
+    private static X509Certificate2 rootCert;
+    private static X509Certificate2 clientCert;
+    private static object rootCertLock = new object();
+    private static object clientCertLock = new object();
+    //Install Root certificate, this is required for all https test
```
iamjasonp: NIT: not only all https tests, but all transport security tests

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
iamjasonp: If someone's dev machine has IIS running and we try to selfhost on port 80 and 443, isn't that an issue?

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
iamjasonp: NIT: use `string.Compare(TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName, "localhost", StringComparison.OrdinalIgnoreCase)`

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+                    break;
+                case "https":
+                     builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeHttpsPort_PropertyName));
+                    break;
+                case "wss":
+                    builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeSecureWebSocketPort_PropertyName));
+                    builder.Scheme = "https";
+                    break;
+                case "net.tcp":
+                    builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeTcpPort_PropertyName));
```
iamjasonp: NIT: could use ushort.Parse instead to detect errors in passed-in port numbers.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+                    break;
+                default:
+                    break;
+            }
+        }
+
+        return builder.Uri;
+    }
+    public static string GetEndpointAdddress(string endpoint, string protocol = "http")
```
iamjasonp: My opinion on this: given the nature of the parameters being passed, it would read much more clearly in subsequent usages if you explicitly called out the protocol ("http") rather than made it an optional argument. 

For example 
`GetEndpointAdddress("WebSocketHttpDuplexTextBuffered.svc//WebSocketHttpDuplexTextBufferedResource", protocol: "ws");`

reads better to me as 

`GetEndpointAdddress("ws", "WebSocketHttpDuplexTextBuffered.svc//WebSocketHttpDuplexTextBufferedResource");`

Similarly, 
`GetEndpointAdddress("NetHttp.svc//NetHttp")`

doesn't call out the protocol unless i look up the call to GetEndpointAddress, but 

`GetEndpointAdddress("http", "NetHttp.svc//NetHttp")` 

does

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+            {
+                return false;
+            };
+
+            return true;
+        }
+    }
+
+    private static Uri BuildBaseUri(string protocol)
+    {
```
iamjasonp: I would Contract.Assert(protocol != null) here

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+        var builder = new UriBuilder();
+        builder.Host = TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName);
+        builder.Scheme = protocol;
```
iamjasonp: This should be moved into the switch statement. The `default:` case should throw.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+                    break;
+                case "wss":
+                    builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeSecureWebSocketPort_PropertyName));
+                    builder.Scheme = "https";
+                    break;
+                case "net.tcp":
+                    builder.Port = int.Parse(TestProperties.GetProperty(TestProperties.BridgeTcpPort_PropertyName));
+                    break;
+                default:
+                    break;
```
iamjasonp: The `default:` case should throw, since you'd be trying to create a Uri without a protocol.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+                ChannelFactory<IUtil> factory = null;
+                IUtil serviceProxy = null;
+                try
+                {
+                    factory = new ChannelFactory<IUtil>(basicHttpBinding, new EndpointAddress(ServiceUtil_Address));
+                    serviceProxy = factory.CreateChannel();
+                    clientCert = new X509Certificate2(serviceProxy.GetClientCert(false), "test", X509KeyStorageFlags.PersistKeySet);
+                    if (clientCert == null)
+                    {
+                        //throw
```
iamjasonp: elide. It's obvious you're throwing

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -53,390 +187,394 @@ public static string HttpProtocolError_Address
...
     }
 
     // HTTPS Addresses
     public static string Https_BasicAuth_Address
     {
         get
         {
-            BridgeClientCertificateManager.InstallRootCertificateFromBridge();
-            return BridgeClient.GetResourceAddress("WcfService.TestResources.BasicAuthResource");
+            EnsureLocalClientCertifciateInstalled();
```
iamjasonp: I think you need the root cert here not client cert.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -53,390 +187,394 @@ public static string HttpProtocolError_Address
...
     }
 
     public static string Https_NtlmAuth_Address
     {
         get
         {
-            BridgeClientCertificateManager.InstallRootCertificateFromBridge();
-            return BridgeClient.GetResourceAddress("WcfService.TestResources.HttpsNtlmResource");
+            EnsureLocalClientCertifciateInstalled();
+            return GetEndpointAdddress("HttpsNtlm.svc//https-ntlm.svc//");
```
iamjasonp: Why the extra slashes at end?

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -53,390 +187,394 @@ public static string HttpProtocolError_Address
...
         }
     }
 
     public static string Https_NtlmAuth_Address
     {
         get
         {
-            BridgeClientCertificateManager.InstallRootCertificateFromBridge();
-            return BridgeClient.GetResourceAddress("WcfService.TestResources.HttpsNtlmResource");
+            EnsureLocalClientCertifciateInstalled();
```
iamjasonp: Root vs. client cert

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -53,390 +187,394 @@ public static string HttpProtocolError_Address
...
         }
     }
 
     public static string Tcp_RevokedServerCertResource_HostName
     {
         get
         {
-            return BridgeClient.GetResourceHostName("WcfService.TestResources.TcpRevokedServerCertResource");
+            return GetServiceHostName();
         }
```
iamjasonp: Question only: how are we going to be testing this scenario?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
roncain: Micro nit: remove extra spaces in "<ServiceUri >"

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClientCertificateManager.cs
```diff
@@ -29,7 +29,7 @@ public static class BridgeClientCertificateManager
         private const string ClientCertificateSubject = "WCF Client Certificate";
```
roncain: Not blocking -- remove "Bridge" from file and class names at some point.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
roncain: Not blocking:  For future consideration we might let the first test hit and report the failure and then set a value so the remaining [ConditionalFact] tests depending on the root cert show as Skipped due to no root cert. This will satisfy @iamjasonp 's suggestion not to attempt doing it again if it fails once.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+
+        return builder.Uri;
+    }
+    public static string GetEndpointAdddress(string endpoint, string protocol = "http")
+    {
+        return string.Format(@"{0}/{1}", BuildBaseUri(protocol), endpoint);
+    }
+
+    public static string ServiceUtil_Address
+    {
```
roncain: Not blocking: we need a better name for the service interface, class and endpoint name.  Maybe CertificateService_Address.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+        get { return GetEndpointAdddress("Util.svc//Util"); }
+    }
```
roncain: Not blocking: (but the debt is increasing if we don't fix these soon) -- need a better name, like CertificateService.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
roncain: Not blocking: note to capture for future thinking -- perhaps we should make the [ConditionalFact] for "Root_Certificate_Installed" be the one to call EnsureRootCertificateInstalled.  This would eliminate the need to try to keep both Endpoint.cs and the tests in sync regarding what they depend on.

---
In reply to: [59565735](https://github.com/dotnet/wcf/pull/1002#discussion_r59565735) [](ancestors = 59565735)

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -1,49 +1,183 @@
...
+    //Install Root certificate, this is required for all https test
+    private static void EnsureRootCertificateInstalled()
+    {
+        lock (rootCertLock)
+        {
+            if (rootCert == null)
+            {
+                BasicHttpBinding basicHttpBinding = new BasicHttpBinding();
+                ChannelFactory<IUtil> factory = null;
+                IUtil serviceProxy = null;
```
roncain: Not blocking: I'll stop pointing out this name needs to be fixed, preferrably before its use spreads too far.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
roncain: +1 about parsing if possible for the tests.  If we need 2 properties, it should no longer be called "BridgeHost"

---
In reply to: [59492624](https://github.com/dotnet/wcf/pull/1002#discussion_r59492624) [](ancestors = 59492624,59492053)

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
roncain: Not blocking: All test properties with Bridge need to be renamed at some point.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
hongdai: Some tests need fully qualified server host name even it's localhost. So parsing is difficult. Another way is to get from the server, I have the method in the IUtil interface.

hongdai:
Since I have one LGTM, I will go ahead merge to unblock the progress.

I have opened a issue https://github.com/dotnet/wcf/issues/1006 to track addressing the comments in the future iterations.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
hongdai: IIS only has restriction on https port. For TCP port, we can have as many certs as we want.

iamjasonp:
No objection to merge now and fix later... I think one of my messages didn't get through. Sorry if that wasn't clear :smile:

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
roncain: FYI: I opened #1007 to consider moving this logic to [ConditionalFact].

PR #1008, opened 2016-04-13
Add placeholders for PCL profiles

ericstj:
This is the WCF version of https://github.com/dotnet/corefx/pull/7665

We need this in ASAP because the WCF repo is broken in internal builds
since it is built with the CoreFx version of buildtools.


weshaggard:
LGTM

ericstj:
Merging now to unblock the internal builds.

roncain:
Thanks for the quick turnaround @ericstj .

PR #1011, opened 2016-04-14
Init check in self host service

hongdai:
* Also add missing TcpDefault service activation. It has caused several security
tests failures.

### DIFF
--- a/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/TcpDefaultResourceTestServiceHost.cs
+++ b/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/TcpDefaultResourceTestServiceHost.cs
@@ -16,11 +16,11 @@ namespace WcfService
     }
     public class TcpDefaultResourceTestServiceHost : TestServiceHostBase<IWcfService>
     {
-        protected override string Address { get { return "tcp-nosecurity"; } }
+        protected override string Address { get { return "tcp-default"; } }
 
         protected override Binding GetBinding()
         {
-            return new NetTcpBinding(SecurityMode.None) { PortSharingEnabled = false };
+            return new NetTcpBinding() { PortSharingEnabled = false };
         }
 
         public TcpDefaultResourceTestServiceHost(Type serviceType, params Uri[] baseAddresses)
--- a/src/System.Private.ServiceModel/tools/SelfHostedWcfService/App.config
+++ b/src/System.Private.ServiceModel/tools/SelfHostedWcfService/App.config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup>
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.1" />
+    </startup>
+</configuration>
--- a/src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
+++ b/src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
@@ -1,0 +1,282 @@
+﻿using System;
+using System.ServiceModel;
+using System.ServiceModel.Description;
+using System.ServiceModel.Web;
+using WcfService;
+
+namespace SelfHostedWCFService
+{
+    public class SelfHostedWCFService
+    {
+        static int httpPort = 8081;
+        static int httpsPort = 44285;
+        static int tcpPort = 809;
+        static int websocketPort = 8083;
+        static int websocketsPort = 8084;
+
+        static private void GetPorts()
+        {
+            httpPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("httpPort")) ? httpPort : int.Parse(Environment.GetEnvironmentVariable("httpPort"));
+            httpsPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("httpsPort")) ? httpsPort : int.Parse(Environment.GetEnvironmentVariable("httpsPort"));
+            tcpPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("tcpPort")) ? tcpPort : int.Parse(Environment.GetEnvironmentVariable("tcpPort"));
+            websocketPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("websocketPort")) ? websocketPort : int.Parse(Environment.GetEnvironmentVariable("websocketPort"));
+            websocketsPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("websocketsPort")) ? websocketsPort : int.Parse(Environment.GetEnvironmentVariable("websocketsPort"));
+        }
+
+        static void Main()
+        {
+            GetPorts();
+
+            string httpBaseAddress = string.Format(@"http://localhost:{0}", httpPort);
+            string httpsBaseAddress = string.Format(@"https://localhost:{0}", httpsPort);
+            string tcpBaseAddress = string.Format(@"net.tcp://localhost:{0}", tcpPort);
+            string websocketBaseAddress = string.Format(@"http://localhost:{0}", websocketPort);
+            string websocketsBaseAddress = string.Format(@"https://localhost:{0}", websocketsPort);
+
+            //Do not need to catch exceptions and dispose service hosts as the process will terminate
+            Uri crlUrl = new Uri(string.Format("http://localhost/CrlService.svc", httpPort));
+            WebServiceHost host = new WebServiceHost(typeof(CrlService), crlUrl);
+            WebHttpBinding binding = new WebHttpBinding();
+            host.AddServiceEndpoint(typeof(ICrlService), binding, "");
+            ServiceDebugBehavior stp = host.Description.Behaviors.Find<ServiceDebugBehavior>();
+            stp.HttpHelpPageEnabled = false;
+            host.Open();
+
+            Uri[] utilTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/Util.svc", httpBaseAddress))};
+            UtilTestServiceHost utilTestServiceHostServiceHost = new UtilTestServiceHost(typeof(WcfService.Util), utilTestServiceHostbaseAddress);
+            utilTestServiceHostServiceHost.Open();
+
+            Uri[] basicAuthTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/BasicAuth.svc", httpsBaseAddress))};
+            BasicAuthTestServiceHost basicAuthTestServiceHostServiceHost = new BasicAuthTestServiceHost(typeof(WcfService.WcfUserNameService), basicAuthTestServiceHostbaseAddress);
+            basicAuthTestServiceHostServiceHost.Open();
+
+            Uri[] basicHttpsTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/BasicHttps.svc", httpsBaseAddress))};
+            BasicHttpsTestServiceHost basicHttpsTestServiceHostServiceHost = new BasicHttpsTestServiceHost(typeof(WcfService.WcfService), basicHttpsTestServiceHostbaseAddress);
+            basicHttpsTestServiceHostServiceHost.Open();
+
+            Uri[] basicHttpTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/BasicHttp.svc", httpBaseAddress))};
+            BasicHttpTestServiceHost basicHttpTestServiceHostServiceHost = new BasicHttpTestServiceHost(typeof(WcfService.WcfService), basicHttpTestServiceHostbaseAddress);
+            basicHttpTestServiceHostServiceHost.Open();
+
+            Uri[] defaultCustomHttpTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/DefaultCustomHttp.svc", httpBaseAddress))};
+            DefaultCustomHttpTestServiceHost defaultCustomHttpTestServiceHostServiceHost = new DefaultCustomHttpTestServiceHost(typeof(WcfService.WcfService), defaultCustomHttpTestServiceHostbaseAddress);
+            defaultCustomHttpTestServiceHostServiceHost.Open();
+
+            Uri[] duplexTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/Duplex.svc", tcpBaseAddress))};
+            DuplexTestServiceHost duplexTestServiceHostServiceHost = new DuplexTestServiceHost(typeof(WcfService.WcfDuplexService), duplexTestServiceHostbaseAddress);
+            duplexTestServiceHostServiceHost.Open();
+
+            Uri[] duplexCallbackTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/DuplexCallback.svc", tcpBaseAddress))};
+            DuplexCallbackTestServiceHost duplexCallbackTestServiceHostServiceHost = new DuplexCallbackTestServiceHost(typeof(WcfService.DuplexCallbackService), duplexCallbackTestServiceHostbaseAddress);
+            duplexCallbackTestServiceHostServiceHost.Open();
+
+            Uri[] duplexChannelCallbackReturnTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/DuplexChannelCallbackReturn.svc", tcpBaseAddress))};
+            DuplexChannelCallbackReturnTestServiceHost duplexChannelCallbackReturnTestServiceHostServiceHost = new DuplexChannelCallbackReturnTestServiceHost(typeof(WcfService.DuplexChannelCallbackReturnService), duplexChannelCallbackReturnTestServiceHostbaseAddress);
+            duplexChannelCallbackReturnTestServiceHostServiceHost.Open();
+
+            Uri[] duplexCallbackDataContractComplexTypeTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/DuplexCallbackDataContractComplexType.svc", tcpBaseAddress))};
+            DuplexCallbackDataContractComplexTypeTestServiceHost duplexCallbackDataContractComplexTypeTestServiceHostServiceHost = new DuplexCallbackDataContractComplexTypeTestServiceHost(typeof(WcfService.WcfDuplexService), duplexCallbackDataContractComplexTypeTestServiceHostbaseAddress);
+            duplexCallbackDataContractComplexTypeTestServiceHostServiceHost.Open();
+
+            Uri[] duplexCallbackXmlComplexTypeTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/DuplexCallbackXmlComplexType.svc", tcpBaseAddress))};
+            DuplexCallbackXmlComplexTypeTestServiceHost duplexCallbackXmlComplexTypeTestServiceHostServiceHost = new DuplexCallbackXmlComplexTypeTestServiceHost(typeof(WcfService.WcfDuplexService), duplexCallbackXmlComplexTypeTestServiceHostbaseAddress);
+            duplexCallbackXmlComplexTypeTestServiceHostServiceHost.Open();
+
+            Uri[] httpBinaryTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpBinary.svc", httpBaseAddress))};
+            HttpBinaryTestServiceHost httpBinaryTestServiceHostServiceHost = new HttpBinaryTestServiceHost(typeof(WcfService.WcfService), httpBinaryTestServiceHostbaseAddress);
+            httpBinaryTestServiceHostServiceHost.Open();
+
+            Uri[] httpDigestNoDomainTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpDigestNoDomain.svc", httpBaseAddress))};
+            HttpDigestNoDomainTestServiceHost httpDigestNoDomainTestServiceHostServiceHost = new HttpDigestNoDomainTestServiceHost(typeof(WcfService.WcfService), httpDigestNoDomainTestServiceHostbaseAddress);
+            httpDigestNoDomainTestServiceHostServiceHost.Open();
+
+            Uri[] httpsClientCertificateTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpsClientCertificate.svc", httpsBaseAddress))};
+            HttpsClientCertificateTestServiceHost httpsClientCertificateTestServiceHostServiceHost = new HttpsClientCertificateTestServiceHost(typeof(WcfService.WcfService), httpsClientCertificateTestServiceHostbaseAddress);
+            httpsClientCertificateTestServiceHostServiceHost.Open();
+
+            Uri[] httpsDigestTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpsDigest.svc", httpsBaseAddress))};
+            HttpsDigestTestServiceHost httpsDigestTestServiceHostServiceHost = new HttpsDigestTestServiceHost(typeof(WcfService.WcfService), httpsDigestTestServiceHostbaseAddress);
+            httpsDigestTestServiceHostServiceHost.Open();
+
+            Uri[] httpsNtlmTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpsNtlm.svc", httpsBaseAddress))};
+            HttpsNtlmTestServiceHost httpsNtlmTestServiceHostServiceHost = new HttpsNtlmTestServiceHost(typeof(WcfService.WcfService), httpsNtlmTestServiceHostbaseAddress);
+            httpsNtlmTestServiceHostServiceHost.Open();
+
+            Uri[] httpSoap11TestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpSoap11.svc", httpBaseAddress))};
+            HttpSoap11TestServiceHost httpSoap11TestServiceHostServiceHost = new HttpSoap11TestServiceHost(typeof(WcfService.WcfService), httpSoap11TestServiceHostbaseAddress);
+            httpSoap11TestServiceHostServiceHost.Open();
+
+            Uri[] httpsSoap11TestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpsSoap11.svc", httpsBaseAddress))};
+            HttpsSoap11TestServiceHost httpsSoap11TestServiceHostServiceHost = new HttpsSoap11TestServiceHost(typeof(WcfService.WcfService), httpsSoap11TestServiceHostbaseAddress);
+            httpsSoap11TestServiceHostServiceHost.Open();
+
+            Uri[] httpsSoap12TestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpsSoap12.svc", httpsBaseAddress))};
+            HttpsSoap12TestServiceHost httpsSoap12TestServiceHostServiceHost = new HttpsSoap12TestServiceHost(typeof(WcfService.WcfService), httpsSoap12TestServiceHostbaseAddress);
+            httpsSoap12TestServiceHostServiceHost.Open();
+
+            Uri[] httpSoap12TestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpSoap12.svc", httpBaseAddress))};
+            HttpSoap12TestServiceHost httpSoap12TestServiceHostServiceHost = new HttpSoap12TestServiceHost(typeof(WcfService.WcfService), httpSoap12TestServiceHostbaseAddress);
+            httpSoap12TestServiceHostServiceHost.Open();
+
+            Uri[] httpsWindowsTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/HttpsWindows.svc", httpsBaseAddress))};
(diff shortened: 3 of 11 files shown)


src/System.Private.ServiceModel/tools/SelfHostedWcfService/dir.targets
```diff
@@ -0,0 +1,77 @@
...
+    <DownloadFile FileName="$(NuGetToolPath)"
+                  Address="https://www.nuget.org/nuget.exe" />
+
+    <PropertyGroup>
+      <NugetRestoreCommand>"$(NuGetToolPath)" restore -PackagesDirectory $(PackageOutputDir)</NugetRestoreCommand>
+    </PropertyGroup>
+    
+    <Exec Command="$(NugetRestoreCommand)" StandardOutputImportance="Low" WorkingDirectory="$(SolutionDir)" />
+  </Target>
+</Project>
```
iamjasonp: By any chance, is there a higher-level dir.targets we can reference to get the nuget restore functionality?

src/System.Private.ServiceModel/tools/SelfHostedWcfService/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Web.Administration" version="7.0.0.0" targetFramework="net451" userInstalled="true" />
+</packages>
```
iamjasonp: NIT: newline

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
```diff
@@ -0,0 +1,49 @@
...
+        goto :EnvSet
+    )
+
+    if defined VS120COMNTOOLS (
+        call "%VS120COMNTOOLS%\VsDevCmd.bat"
+        goto :EnvSet
+    )
+
+    echo Error: %~nx0 requires Visual Studio 2013 or 2015.
+    echo        Please see https://github.com/dotnet/wcf/wiki/Developer-Guide for build instructions.
```
iamjasonp: We don't use the wiki at all, can you please update the link?

src/System.Private.ServiceModel/tools/SelfHostedWcfService/SelfHostedWCFService.csproj
```diff
@@ -0,0 +1,74 @@
...
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+ <Import Project="$(MSBuildThisFileDirectory)\dir.targets" />
+  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
+       Other similar extension points exist, see Microsoft.Common.targets.
+  <Target Name="BeforeBuild">
+  </Target>
+  <Target Name="AfterBuild">
+  </Target>
+  -->
+</Project>
```
iamjasonp: NIT: newline

iamjasonp:
By the way we've branched, shouldn't the SelfHostWcfService go into the infrastructure branch? The TCP change makes sense in the master branch

Secondly, would prefer you to split the SelfHostWcfService commit from the tcp service activation commit for easier git blame - but not a big deal for this time, just keep it in mind for the future

LGTM

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("SelfHostDriver")]
```
zhenlan: Be consistent: SelfHostedWCFService

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("SelfHostDriver")]
```
zhenlan: SelfHostedWCFService

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
```diff
@@ -0,0 +1,49 @@
...
+goto :eof
+
+:AfterBuild
+
+echo.
+:: Pull the build summary from the log file
+findstr /ir /c:".*Warning(s)" /c:".*Error(s)" /c:"Time Elapsed.*" "%_buildlog%"
+echo Build Exit Code = %BUILDERRORLEVEL%
+
+exit /b %BUILDERRORLEVEL%
```
zhenlan: endlocal was missing

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
```diff
@@ -0,0 +1,49 @@
...
+
+    echo Error: %~nx0 requires Visual Studio 2013 or 2015.
+    echo        Please see https://github.com/dotnet/wcf/wiki/Developer-Guide for build instructions.
+    exit /b 1
+)
+
+:EnvSet
+
+:: Log build command line
+set _buildproj1=%~dp0..\SelfHostedWcfService\SelfHostedWCFService.sln
```
zhenlan: NIT: odd to see a suffix '1' and technically this is _buildsln :)

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
iamjasonp: Endlocals are typically implicit in cmd files so I think this performs correctly, but I agree with @zhenlan that I prefer to see it explicitly. 

https://technet.microsoft.com/en-us/library/bb490898.aspx

src/System.Private.ServiceModel/tools/SelfHostedWcfService/App.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup>
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.1" />
+    </startup>
+</configuration>
```
iamjasonp: NIT: newline

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -0,0 +1,282 @@
...
+            WebSocketHttpsDuplexBinaryBufferedTestServiceHost webSocketHttpsDuplexBinaryBufferedTestServiceHostServiceHost = new WebSocketHttpsDuplexBinaryBufferedTestServiceHost(typeof(WcfService.WSDuplexService), webSocketHttpsDuplexBinaryBufferedTestServiceHostbaseAddress);
+            webSocketHttpsDuplexBinaryBufferedTestServiceHostServiceHost.Open();
+
+            Uri[] webSocketHttpsDuplexTextBufferedTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/WebSocketHttpsDuplexTextBuffered.svc", websocketsBaseAddress))};
+            WebSocketHttpsDuplexTextBufferedTestServiceHost webSocketHttpsDuplexTextBufferedTestServiceHostServiceHost = new WebSocketHttpsDuplexTextBufferedTestServiceHost(typeof(WcfService.WSDuplexService), webSocketHttpsDuplexTextBufferedTestServiceHostbaseAddress);
+            webSocketHttpsDuplexTextBufferedTestServiceHostServiceHost.Open();
+
+            Console.WriteLine("All service hosts have started.");
+            Console.WriteLine("Press <ENTER> to terminate the self service Host.");
+            Console.ReadLine();
```
iamjasonp: Question only: what's the error message that spews out if something doesn't start correctly? I think it should be easily identifiable which one failed, right?

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -0,0 +1,282 @@
...
+        static int websocketPort = 8083;
+        static int websocketsPort = 8084;
+
+        static private void GetPorts()
+        {
+            httpPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("httpPort")) ? httpPort : int.Parse(Environment.GetEnvironmentVariable("httpPort"));
+            httpsPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("httpsPort")) ? httpsPort : int.Parse(Environment.GetEnvironmentVariable("httpsPort"));
+            tcpPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("tcpPort")) ? tcpPort : int.Parse(Environment.GetEnvironmentVariable("tcpPort"));
+            websocketPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("websocketPort")) ? websocketPort : int.Parse(Environment.GetEnvironmentVariable("websocketPort"));
+            websocketsPort = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("websocketsPort")) ? websocketsPort : int.Parse(Environment.GetEnvironmentVariable("websocketsPort"));
```
iamjasonp: Not blocking: I feel like names like 'httpPort' in the context of an environment variable seem a little bit vague. Perhaps 'wcfService_httpPort' or something similar? 

Also, as before, consider using `ushort.Parse` to capture some errors on port numbers

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
```diff
@@ -0,0 +1,49 @@
...
+set _buildpostfix=^> "%_buildlog%"
+
+call :build %*
+
+:: Build
+set _buildprefix=
+set _buildpostfix=
+call :build %*
+
+goto :AfterBuild
```
zhenlan: I may not follow... but looks like the 4 lines above will never run and this script ends up just display the build command you plan to run.

Normally, I will do it this way... hopefully it will be easier to read

```
set cmd=msbuild "%_buildproj1%" /nologo /maxcpucount /verbosity:minimal /nodeReuse:false /fileloggerparameters:Verbosity=diag;LogFile="%_buildlog%";Append %*
echo %cmd%
%cmd%> "%_buildlog%"
```

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -0,0 +1,282 @@
...
+
+        static void Main()
+        {
+            GetPorts();
+
+            string httpBaseAddress = string.Format(@"http://localhost:{0}", httpPort);
+            string httpsBaseAddress = string.Format(@"https://localhost:{0}", httpsPort);
+            string tcpBaseAddress = string.Format(@"net.tcp://localhost:{0}", tcpPort);
+            string websocketBaseAddress = string.Format(@"http://localhost:{0}", websocketPort);
+            string websocketsBaseAddress = string.Format(@"https://localhost:{0}", websocketsPort);
```
iamjasonp: Is this _always_ "localhost"? What if I want to run a selfhost service on a machine that's not localhost? We can solve this later though.

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
```diff
@@ -0,0 +1,49 @@
+@echo off
```
zhenlan: This script should live in tools/scripts folder as test asset structure we discussed.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+[assembly: AssemblyCopyright("Copyright ©  2016")]
```
iamjasonp: From what i see in CoreFx, there's no Copyright attribute in AssemblyInfo.cs, but the standard copyright notice appears at the top of AssemblyInfo.cs

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+// Setting ComVisible to false makes the types in this assembly not visible 
+// to COM components.  If you need to access a type in this assembly from 
+// COM, set the ComVisible attribute to true on that type.
+[assembly: ComVisible(false)]
+
+// The following GUID is for the ID of the typelib if this project is exposed to COM
+[assembly: Guid("89ed8209-1446-49e7-b588-70de0e9634ee")]
```
iamjasonp: NIT: remove, we're not using COM here :)

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -0,0 +1,282 @@
+﻿using System;
```
iamjasonp: NIT: (c) notice

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
zhenlan: Apparently I stayed too late :smile: - it will run, but I still recommend what I put there for easier reading.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/TcpDefaultResourceTestServiceHost.cs
```diff
@@ -16,11 +16,11 @@ protected override ServiceHost CreateServiceHost(Type serviceType, Uri[] baseAdd
     }
```
iamjasonp: I can't add a comment on the correct line, but the copyright notice needs to be updated to the one standard one we use for .NET Core projects

src/System.Private.ServiceModel/tools/SelfHostedWcfService/SelfHostedWCFService.csproj
```diff
@@ -0,0 +1,74 @@
...
+    <Reference Include="System.ServiceModel.Activation" />
+    <Reference Include="System.ServiceModel.Activities" />
+    <Reference Include="System.ServiceModel.Channels" />
+    <Reference Include="System.ServiceModel.Web" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
```
iamjasonp: NIT for later: clean up the list of referenced assemblies; I'm pretty sure some of these references aren't used.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/SelfHostedWCFService.csproj
```diff
@@ -0,0 +1,74 @@
...
+  <ItemGroup>
+    <Compile Include="..\IISHostedWcfService\**\*.cs" />
```
iamjasonp: I think @roncain may have mentioned directory structure before; this structure is a bit confusing as we're having to depend on IISHostedWcfService in SelfHostedWcfService. 

Not blocking for this merge.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/packages.config
iamjasonp: Question only: Are we using M.W.Administration? If not, can we get rid of it? I don't see references to it in files in this PR, but it might be necessary in the IISHostedWcfService (another reason why I'm not that fond of this cross-directory dependency) :smile:

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
iamjasonp: ... for later, perhaps you could update this link from wherever you copied from too :) 
Not blocking, we can take the change later

src/System.Private.ServiceModel/tools/SelfHostedWcfService/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Web.Administration" version="7.0.0.0" targetFramework="net451" userInstalled="true" />
```
zhenlan: It's very possible I missed it somewhere... who is using Microsoft.Web.Administration?

src/System.Private.ServiceModel/tools/SelfHostedWcfService/SelfHostedWCFService.csproj
zhenlan: Not blocking. We should make all cs files from IISHostedWcfService as <Link> to this project, so source code will show nicely when the SelfHostedWcfService project is opened in VS. They can be organized in any way we want in this project but without messing up app_code folder structure.

zhenlan:
> By the way we've branched, shouldn't the SelfHostWcfService go into the infrastructure branch?

The SelfHostWcfService is a new project. It will not break anything existing. The infrastructure branch is created to do work that will break existing things before we are ready. It is the plan that all service code goes to master directly.

iamjasonp:
> The infrastructure branch is created to do work that will break existing things before we are ready. It is the plan that all service code goes to master directly.

Yep, I recall this as part of the plan and have no objections - it just seemed more natural to me that we'd keep all these changes logically in one place and then merge once. 

Further, what's going to happen here is that if one wants to test the `infrastructure` branch against a SelfHostedWcfService, then you'll have to have two repos on the local box to test against (one synced to infrastructure, one to master)

Anyway, not blocking :smile:

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -0,0 +1,282 @@
...
+
+namespace SelfHostedWCFService
+{
+    public class SelfHostedWCFService
+    {
+        static int httpPort = 8081;
+        static int httpsPort = 44285;
+        static int tcpPort = 809;
+        static int websocketPort = 8083;
+        static int websocketsPort = 8084;
```
zhenlan: Will it work if we just use the standard ports for all above when IIS is enabled on the machine (but self and IIS hosted will be differentiated by URL)? Maybe create an issue to look into later.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
roncain: I don't like reinventing a new way to specify and access ports.  TestProperties does this better.  Why didn't you use a known working system for this? Please open an issue to share TestProperties and to parse command line arguments into them (as Bridge.exe does).

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -0,0 +1,282 @@
...
+            //Do not need to catch exceptions and dispose service hosts as the process will terminate
+            Uri crlUrl = new Uri(string.Format("http://localhost/CrlService.svc", httpPort));
+            WebServiceHost host = new WebServiceHost(typeof(CrlService), crlUrl);
+            WebHttpBinding binding = new WebHttpBinding();
+            host.AddServiceEndpoint(typeof(ICrlService), binding, "");
+            ServiceDebugBehavior stp = host.Description.Behaviors.Find<ServiceDebugBehavior>();
+            stp.HttpHelpPageEnabled = false;
+            host.Open();
+
+            Uri[] utilTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/Util.svc", httpBaseAddress))};
```
roncain: For later... I'd like to find a way to make it easier to add new services and hosts that both self-host and IIS-host use. I know the activation models are different, but today I need to touch multiple files to add a service and host, and it's not obvious if I forgot to add an entry to the right file.  Also, I think we want to make the hosts take generic parameters, not typeof.  Something to think about after this is all checked in.

roncain:
LGTM -- agree with @iamjasonp to keep moving and open issues to resolve after we have this working.

roncain:
Regarding which branch this goes into -- I thought the plan was that non-breaking new test code would move into master opportunistically and we would leave existing master code alone till we merged branches.  But if code like this doesn't start from Infrastructure, how do we test it?  I'd prefer to see new code originate in Infrastructure and get cherry-picked into master.  Otherwise I think we need to FI this code back down into Infrastructure to make progress there.  Not blocking, but worth clarifying as we move ahead.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
hongdai: You can set environment variables to override the ports.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
hongdai: You will get an unhandled exception with call stacks, thus it's obvious what's going wrong.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/dir.targets
hongdai: I didn't find one.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/packages.config
hongdai: The CRL service.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/TcpDefaultResourceTestServiceHost.cs
hongdai: I will fix it.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/packages.config
hongdai: You are right it's not needed in the self hosted service. I will remove it.

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
hongdai: Will fix it.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Properties/AssemblyInfo.cs
hongdai: Will fix.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/SelfHostedWCFService.csproj
roncain: Agree with @Zhenlan -- which is why I have proposed making unique folders for testhosts, testservices and testtypes under App_code.  Then the self-host project can include **\*.cs from those folders and get a good layout in VS.  We don't want the self-host .csproj or client test projects to need to include or exclude individual files by name.  Not blocking, but by not solving it early, our debt in this area will increase as we go.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
hongdai: Hmm, this needs more discussion. I'll open an issue.

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
hongdai: This is corefx build.cmd style that we have been following to minimize merge difference. Changing the style introduce one more difference. I think the drawback outweigh the easier reading :-)

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
hongdai: Currently it satisfies the known requirements. If a new requirement comes up, we will need to make changes.

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
hongdai: Ok.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Properties/AssemblyInfo.cs
hongdai: Will fix it.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/packages.config
roncain: What is the CRL restful service and where does it live?  Why is it unique to self-host and not IIS host? It's not clear to me why one hosting model has special services the other doesn't have (or why a hosting project has dependencies needed by a service).  Not blocking, but it would reduce questions if we heard a brief plan where this is going.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
roncain: Not blocking, but let's record this for later discussion.  This "localhost" dependency breaks the dev cross-machine scenario using self-host.  Jason and I use that every day.  I believe the "known requirements" should include not disabling scenarios we actively use today. (For the record, finding and fixing all the dependencies on localhost was quite costly last time we did this for the Bridge.  It's better to avoid hard-coding it in new code.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
hongdai: Will open a issue.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
hongdai: A remote client should be able to access a WCF service started as "localhost", just use the fully qualified domain name to access, as long as the firewall does not block the ports.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/packages.config
hongdai: Both Self-host and IIS host have CRL restful service. The difference is IIS host require some extra dependencies.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
iamjasonp: Awesome - just double check that the endpoint name shows up (i think it does)

hongdai:
I agree It's easier for testing purpose if new code originate from Infrastructure branch.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Properties/AssemblyInfo.cs
iamjasonp: Again, not a blocker today

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
iamjasonp: Hmm, I don't remember if that's a correct statement or not, because I recall from the depths of my memory that listening on localhost was binding to the loopback adapter instead of the network adapter - hence it was listening literally only on 127.x.x.x. +1 Let's investigate this later

hongdai:
I'll go ahead merge since I have LGTM and address the comments in later iterations. Issues will be opened.

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
zhenlan: Unlike build scripts, we have no plan to sync this script with anything in the future. This is our script to maintain. It's okay to "borrow" code but no reason to let "bad" code spread. I know you have merged, but I feel strongly about this one (and similar situations in the future) :smiling_imp:.

src/System.Private.ServiceModel/tools/setupfiles/BuildWCFSelfHostedService.cmd
zhenlan: Thanks @iamjasonp. Good to know :smile:

src/System.Private.ServiceModel/tools/SelfHostedWcfService/packages.config
zhenlan: I still saw `Microsoft.Web.Administration` in the merged code. Please open issue to clean this up. Since this is the only package dependency we have, this will eliminate everything related to package restore too.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/SelfHostedWCFService.csproj
zhenlan: Just to add a note, I was suggesting to add cs files from IISHostedWcfService as **\<Link\>** to this project. I put the comment in codeflow and GitHub markup ate the word "\<Link\>" :smile:. I've also fixed it inline now.

PR #1013, opened 2016-04-14
Add support for building rc2 branch in CI

iamjasonp:
* Add support for building rc2 and infrastructure branch in CI
* ~~Add PR Number guard to prevent PRService from breaks. When syncing to PR numbers lower than 987, the PRService will be wiped from the PRService endpoint and be put into an unrecoverable state.~~

~~Add a guard here to prevent syncing to a PR number < 987 and return HTTP/200 so that CI can succeed.~~

~~This should not happen during the normal course of events. In any case, PRs prior to 987 will not hit the hosted WcfService, so this no-op will not affect the outcome of any tests before that~~

(this will not prevent all breaks, but at least prevents one class of them)

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -24,7 +24,7 @@ def targetNugetRuntimeMap = ['OSX' : 'osx.10.10-x64',
                              'CentOS7.1' : 'centos.7-x64',
                              'OpenSUSE13.2' : 'ubuntu.14.04-x64',
                              'RHEL7.2': 'rhel.7-x64']
-def branchList = ['master', 'pr']
+def branchList = ['master', 'pr', 'rc2', 'infrastructure']
 def osShortName = ['Windows 10': 'win10',
                    'Windows 7' : 'win7',
                    'Windows_NT' : 'windows_nt',
@@ -39,6 +39,7 @@ def osShortName = ['Windows 10': 'win10',
 def static getFullBranchName(def branch) {
     def branchMap = ['master':'*/master',
         'rc2':'*/release/1.0.0-rc2',
+        'infrastructure':'*/infrastructure',
         'pr':'*/master']
     def fullBranchName = branchMap.get(branch, null)
     assert fullBranchName != null : "Could not find a full branch name for ${branch}"
@@ -49,6 +50,9 @@ def static getJobName(def name, def branchName) {
     def baseName = name
     if (branchName == 'rc2') {
         baseName += "_rc2"
+    }
+    else if (branchName == "infrastructure") {
+        baseName += "_infrastructure"
     }
     return baseName
 }
@@ -66,13 +70,20 @@ class WcfUtilities
     
     // Outerloop jobs for WCF Core require an external server reference
     // This should be run 
-    def addWcfOuterloopTestServiceSync(def job, String os, boolean isPR) { 
+    def addWcfOuterloopTestServiceSync(def job, String os, String branch, boolean isPR) { 
+
+        // Exclude rc2 branch, since that branch will not have the sync scripts in
+        if (branch.toLowerCase().contains("rc2")) {
+            return 
+        }
+
         wcfRepoSyncServiceCount++
 
         def operation = isPR ? "pr" : "branch"
 
         job.with { 
             parameters {
+                stringParam('WcfServiceUrl', "http://wcfcoresrv2.cloudapp.net/WcfService${wcfRepoSyncServiceCount}", 'Wcf OuterLoop Test Service Uri')
                 stringParam('WcfRepoSyncServiceUrl', "http://wcfcoresrv2.cloudapp.net/PRService${wcfRepoSyncServiceCount}/pr.ashx", 'Wcf OuterLoop Test PR Service Uri')
             }
         }
@@ -110,7 +121,7 @@ branchList.each { branchName ->
         label('windows-elevated')
     }
     
-    wcfUtilities.addWcfOuterloopTestServiceSync(newJob, os, isPR)
+    wcfUtilities.addWcfOuterloopTestServiceSync(newJob, os, branchName, isPR)
     
     newJob.with {
         steps {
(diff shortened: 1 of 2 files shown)


iamjasonp:
@dotnet-bot test ci please

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -86,7 +86,20 @@ public class PullRequestHandler : IHttpHandler
...
+            if (pr < 987)
+            {
+                context.Response.Write(string.Format("PR specified, '{0}' is not supported by this script<br/>", HttpUtility.HtmlEncode(pr)));
+                context.Response.Write("Treating this as a no-op<br/>");
+                success = true;
+            }
+            else
+            {
+                success = SyncToPr(pr, result);
+            }
```
zhenlan: We only have one PR at this moment that is active and lower than 987. This sounds like a overkill to me.

iamjasonp: It's easy to discover the URI for the service and hit the URI in a browser. 

It doesn't solve the problem entirely but it reduces (slightly) the ways in which someone could screw up the PRService. Other ways are being investigated too.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -99,6 +112,8 @@ public class PullRequestHandler : IHttpHandler
             {
                 case "master":
                 case "origin/master":
+                case "origin/release/1.0.0-rc2":
+                case "origin/infrastructure":
```
zhenlan: Why don't we just open it up for all branches and let it fail if a branch does not exist? I really want to minimize the churn to PRService.

iamjasonp: * We have branches release/1.0.0-rc1 and StressTests which will cause the PR service to be nuked from the test server box. These should not be accessible by anyone and should return 400. 
* CI will test release/1.0.0-rc2 branch periodically and we need that to return 200 (the alternative is to churn the netci.groovy file, which is harder to deal with, but will still not guard against someone hitting the URI directly). We can get rid of this once we stop periodically testing the rc2 branch. 
* CI will test infrastructure branch periodically - we want that to work until we merge. 

Unfortunately, the way this service is deployed, we cannot avoid either a white or blacklist; one of these will exist.

iamjasonp:
@dotnet-bot test ci please

iamjasonp:
@dotnet-bot test ci please

iamjasonp:
@dotnet-bot test ci please

iamjasonp:
@dotnet-bot test ci please

iamjasonp:
Spot checked the new _rc2 and _infrastructure builds, and they look ok.

iamjasonp:
In case anyone's counting, that's 36 outerloops that we've generated with this particular groovy :scream:

roncain:
Very very nice @iamjasonp !

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -99,10 +99,15 @@ public class PullRequestHandler : IHttpHandler
             {
                 case "master":
                 case "origin/master":
+                case "origin/infrastructure":
                     success = SyncToBranch(branchString, result);
                     break;
+                case "origin/release/1.0.0-rc2":
+                    context.Response.Write(string.Format("Branch name specified, '{0}' will not be synced by this service<br/>", HttpUtility.HtmlEncode(branchString)));
+                    success = true;
```
roncain: Question only: how does this "succeed" if it does not sync?  I probably don't understand what is temporary and what will happen in the future.  For example, we have a PR against RC2 right now (and may again).  Is it the plan to allow CI to test these?

netci.groovy
```diff
@@ -66,27 +70,31 @@ class WcfUtilities
...
-                steps {
-                    batchFile(".\\src\\System.Private.ServiceModel\\tools\\scripts\\sync-pr.cmd ${operation} %WcfRepoSyncServiceUrl%")
-                }           
-            }
-        } 
-        else {
-            job.with { 
-                steps {
-                   shell("HOME=\$WORKSPACE/tempHome ./src/System.Private.ServiceModel/tools/scripts/sync-pr.sh ${operation} \$WcfRepoSyncServiceUrl")
+        // Exclude rc2 branch, since that branch will not have the sync scripts in
```
roncain: Observation only: this highlights an area we need to consider, namely what can CI assume it can invoke on the PR machine, and where do they live?  I think we now have adequate proof we have an issue to solve (ex: PR's before #xxx could blow up the PR machine, RC2 needs special treatment, etc.) We don't need to solve it now, but I think PR service may need to rely on some kind of bootstrap script executing outside the enlistment that is "always there" and which can delegate into scripts inside the enlistment if they exist.  For later discussion...

roncain:
LGTM.  My comments are for future thought, not blocking.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: I'm viewing the HTTP/200 as "the PR service did what it needed to do for a PR to start". In this case, it's a no-op, since the PR service, if synced to the rc2 branch, would essentially nuke itself. 

Actually upon further thinking, we probably don't need this case, because the sync-pr scripts couldn't be called anyway (they don't exist either).

iamjasonp:
As this PR isn't too urgent, I'll wait for the queues to slow down (past 1900 PDT) before doing the merge so that I can confirm stability without affecting others' work.

netci.groovy
iamjasonp: There's a bunch of considerations here: 

* PRs before xxx could blow up a PRService 
* Syncing to rc2, rc1, etc. could blow up the machine
* ... and there are numerous other scenarios that will blow up the machine and require manual intervention which I don't care to enumerate here for now. 

In short, yes, I think there's a lot to consider for later discussion :smile:

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: Correct. Whatever in RC1 or RC2 should just use whatever testing mechanism when they were forked. @iamjasonp please make sure this is the case.

zhenlan: In case I did not make it clear - we don't want to enable any new testing for rc2. Please undo the changes about rc2.

zhenlan:
The change to unblock "infrastructure" branch CI testing LGTM.

I expect the rest of changes (related to rc2) will be undone.

iamjasonp:
> I expect the rest of changes (related to rc2) will be undone.

No, not all the changes will be undone; the netci.groovy changes will be kept

iamjasonp:
@dotnet-bot test ci please

iamjasonp:
@dotnet-bot test ci please

PR #1017, opened 2016-04-14
Fix Default TCP endpoint in client

hongdai:
* It should be tcp-default, not the no security one.
* A one line of code change that will fixes several security tests.


zhenlan:
LGTM

PR #1018, opened 2016-04-14
April 14 Script Sync

StephenBonikowsky:
First commit syncs build-tests.cmd and related changes.
Second commit adopts the new test .builds files for building and running test projects.
Third commit takes updates to changes to shared files from last 24 hours and updates all the .sh files.
   - The .sh files have not been tested on Linux.
   - Even though we do not use run-test.sh I synced it, we may have to at some point.

### DIFF
--- a/build-tests.cmd
+++ b/build-tests.cmd
@@ -1,0 +1,33 @@
+@if "%_echo%" neq "on" echo off
+setlocal EnableDelayedExpansion
+
+set buildTests=build-tests.log
+set binclashLoggerDll=%~dp0Tools\net45\Microsoft.DotNet.Build.Tasks.dll
+set binclashlog=%~dp0binclash.log
+echo Running build-tests.cmd %* > %buildTests%
+
+set options=/nologo /maxcpucount /v:minimal /clp:Summary /nodeReuse:false /flp:v=detailed;Append;LogFile=%buildTests% /l:BinClashLogger,%binclashLoggerDll%;LogFile=%binclashlog%
+set allargs=%*
+
+if /I [%1] == [/?] goto Usage
+if /I [%1] == [/help] goto Usage
+
+REM ensure that msbuild is available
+echo Running init-tools.cmd
+call %~dp0init-tools.cmd
+
+echo msbuild.exe %~dp0src\tests.builds !options! !allargs! >> %buildTests%
+call msbuild.exe %~dp0src\tests.builds !options! !allargs!
+if NOT [%ERRORLEVEL%]==[0] (
+  echo ERROR: An error occurred while building the tests, see %buildTests% for more details.
+  exit /b
+)
+
+echo Done Building tests.
+exit /b
+
+:Usage
+echo.
+echo Builds the tests that are in the repository.
+echo No option parameters.
+exit /b
--- a/build.sh
+++ b/build.sh
@@ -7,7 +7,7 @@ usage()
     echo "native - optional argument to build the native code"
     echo "The following arguments affect native builds only:"
     echo "BuildArch can be: x64, x86, arm, arm64"
-    echo "BuildType can be: Debug, Release"
+    echo "BuildType can be: debug, release"
     echo "clean - optional argument to force a clean build."
     echo "verbose - optional argument to enable verbose build output."
     echo "clangx.y - optional argument to build using clang version x.y."
@@ -15,6 +15,7 @@ usage()
     echo "cross - optional argument to signify cross compilation,"
     echo "      - will use ROOTFS_DIR environment variable if set."
     echo "skiptests - skip the tests in the './bin/*/*Tests/' subdirectory."
+    echo "generateversion - if building native only, pass this in to get a version on the build output."
     echo "cmakeargs - user-settable additional arguments passed to CMake."
     exit 1
 }
@@ -99,6 +100,25 @@ prepare_native_build()
     if [ $__VerboseBuild == 1 ]; then
         export VERBOSE=1
     fi
+
+    # If managed build is supported, then generate version
+    if [ $__buildmanaged == true ]; then
+        __generateversionsource=true
+    fi
+
+    # Ensure tools are present if we will generate version.c
+    if [ $__generateversionsource == true ]; then
+        $__scriptpath/init-tools.sh
+    fi
+
+    # Generate version.c if specified, else have an empty one.
+    __versionSourceFile=$__scriptpath/bin/obj/version.c
+    if [ $__generateversionsource == true ]; then
+        $__scriptpath/Tools/corerun $__scriptpath/Tools/MSBuild.exe "$__scriptpath/build.proj" /t:GenerateVersionSourceFile /p:NativeVersionSourceFile=$__scriptpath/bin/obj/version.c /p:GenerateVersionSourceFile=true /v:minimal
+    else
+        __versionSourceLine="static char sccsid[] __attribute__((used)) = \"@(#)No version information produced\";"
+        echo $__versionSourceLine > $__versionSourceFile
+    fi
 }
 
 build_managed()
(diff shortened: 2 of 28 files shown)


build.sh
```diff
@@ -108,7 +128,7 @@ build_managed()
     __binclashlog=$__scriptpath/binclash.log
     __binclashloggerdll=$__scriptpath/Tools/Microsoft.DotNet.Build.Tasks.dll
 
-    $__scriptpath/Tools/corerun $__scriptpath/Tools/MSBuild.exe "$__buildproj" /nologo /verbosity:minimal "/fileloggerparameters:Verbosity=normal;LogFile=$__buildlog" "/l:BinClashLogger,$__binclashloggerdll;LogFile=$__binclashlog" /t:Build /p:ConfigurationGroup=$__BuildType /p:OSGroup=$__BuildOS /p:SkipTests=$__SkipTests /p:COMPUTERNAME=$(hostname) /p:USERNAME=$(id -un) /p:TestNugetRuntimeId=$__TestNugetRuntimeId $__UnprocessedBuildArgs
+    $__scriptpath/Tools/corerun $__scriptpath/Tools/MSBuild.exe "$__buildproj" /m /nologo /verbosity:minimal "/fileloggerparameters:Verbosity=normal;LogFile=$__buildlog" "/l:BinClashLogger,$__binclashloggerdll;LogFile=$__binclashlog" /t:Build /p:ConfigurationGroup=$__BuildType /p:OSGroup=$__BuildOS /p:SkipTests=$__SkipTests /p:COMPUTERNAME=$(hostname) /p:USERNAME=$(id -un) /p:TestNugetRuntimeId=$__TestNugetRuntimeId $__UnprocessedBuildArgs
```
roncain: At one point we had an issue building projects in parallel with /m.  It was a race condition so hard to repro.  I'm okay trying to go back to /m but be aware if OTG builds randomly fail in ServiceModel, this could be the cause.  I think the binclasher detector may have helped us find the conflicting project builds. I'd say "yes, let's try this" but if so it might be wise to wait for Monday or be ready to watch OTG builds tonight.

dir.props
```diff
@@ -106,16 +111,23 @@
...
+  <PropertyGroup Condition="'$(BuildTestsAgainstPackages)' == 'true'">
+      <GeneratedProjectJsonDir>$(ObjDir)generated</GeneratedProjectJsonDir>
+  </PropertyGroup>
+
   <!-- list of directories to perform batch restore -->
   <ItemGroup>
     <DnuRestoreDir Include="$(MSBuildThisFileDirectory)/src" />
     <DnuRestoreDir Include="$(MSBuildThisFileDirectory)/pkg" />
+    <DnuRestoreDir Condition="'$(BuildTestsAgainstPackages)' == 'true'" Include="$(GeneratedProjectJsonDir)" />
   </ItemGroup>
```
roncain: Ooo -- I like the sound of this.  @StephenBonikowsky -- do you know offhand if this allows us to build our current tests against packages, or will we need to change them?  Just a question -- don't block on it.

run-test.cmd
```diff
@@ -0,0 +1,17 @@
...
+:: %1 is the path to the tests\<OSConfig> folder
+
+pushd %1
+
+FOR /D %%F IN (*.Tests) DO (
+	IF EXIST %%F\dnxcore50 (
+		pushd %%F\dnxcore50
+		@echo "corerun.exe xunit.console.netcore.exe %%F.dll -xml testResults.xml -notrait category=outerloop -notrait category=failing -notrait category=nonwindowstests -notrait Benchmark=true"
+		corerun.exe xunit.console.netcore.exe %%F.dll -xml testResults.xml -notrait category=outerloop -notrait category=failing -notrait category=nonwindowstests -notrait Benchmark=true
+		popd
```
roncain: Question only -- @StephenBonikowsky , do you know if this works in our repo?  I wish it was smarter about parameterizing OuterLoop and failing traits, but we can offer that suggestion later.  Don't block on this answer.

run-test.sh
```diff
@@ -29,7 +29,7 @@ usage()
     echo "                                      default: <repo_root>/bin/Product/<OS>.x64.<ConfigurationGroup>"
```
roncain: We're planning on not using this any more.  @iamjasonp -- do you know if it is used in any remaining Jenkins jobs?  To test this will require manually running on a Linux box.  Personally I think it would be safer not to checkin this file unless something is broken.  If we use it, changes could break us.  If we don't use it, we should nuke it.

src/System.Private.ServiceModel/tests/Scenarios/ServiceModel.Scenario.Tests.builds
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <ItemGroup>
+    <Project Include="Binding\Custom\Binding.Custom.Tests.csproj" />
+    <Project Include="Binding\Http\Binding.Http.Tests.csproj" />
```
roncain: Can we change all these into a single <Project Include=**\*.csproj />?  I'd prefer not to have to maintain a list of tests whenever we add a new test project.

run-test.sh
StephenBonikowsky: Yeah, discussed this in sync-up today. Corefx made changes to it involving the new test .builds files, we don't use it now but might we at some point? If we keep it I wouldn't expect to test it until such time as we needed it.

src/System.Private.ServiceModel/tests/Scenarios/ServiceModel.Scenario.Tests.builds
StephenBonikowsky: Yeah, that makes sense, should have done it that way to start with.

roncain:
LGTM, subject to your own opinion on testing.  I can't test TFS right now, so we'd need to watch OTG builds if this is merged today.  I can't test Linux right now due to VM issues. My comments about the project include are optional and can be done later.

run-test.cmd
StephenBonikowsky: It works but as you point out it doesn't run OuterLoop tests.

dir.props
StephenBonikowsky: It is supposed to but I haven't tried it yet, this is more like "adding support to do this", we now need to test it and see what else needs to be done.

StephenBonikowsky:
Amended the previous commit with...
   - Simplifying the ServiceModel.Scenario.Tests.builds file to pick up all *.csproj files below it.
   - Based on concerns removed any changes to run-test.sh

StephenBonikowsky:
Well it looks like run-test.sh is being used by Jenkins somewhere since all the non-Windows checks failed after I reverted my sync of the file. :cry: 

Gonna re-sync it.

StephenBonikowsky:

StephenBonikowsky:
More info, the build process is somehow picking up projects under the tools directory and so we are seeing failures on non-Windows for...

`error MSB4801: The task factory "CodeTaskFactory" could not be loaded because this version of MSBuild does not support it. `

We used to explicitly exclude these projects but that has changed, need to find out how to exclude these projects.

StephenBonikowsky:
Added back some logic to exclude `"System.Private.ServiceModel\tests\Scenarios\SelfHostWcfService\WcfService.csproj"` from building.

StephenBonikowsky:
The file `src/System.Private.ServiceModel/tools/dir.targets` added in Hong's PR #1019 has a reference to "CodeTaskFactory", this was intended to be used by `CertificateGenerator.csproj` which should not need to be built in Linux environments. However, the `tools\test\CertificateCleanup\CertificateCleanup.csproj` does need to be built everywhere as it is used by tests and it pulls in the dir.targets file with this line...
`<Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
`
Commenting out this line and re-trying to see if it works as a short-term solution.

StephenBonikowsky:
That didn't work, the project needs to inherit "...\wcf\src\dir.targets" or it gets other errors such as...
`error MSB4057: The target "BuildAndTest" does not exist in the project.`
Adding an explicit path as a very temporary workaround.
`Import Project="\..\..\..\..\dir.targets" /`

StephenBonikowsky:
That didn't work either, couldn't find the dir.targets
Instead opened PR #1019 to move the dir.targets and dir.props into the CertificateGenerator folder.

StephenBonikowsky:
@dotnet-bot test this please

StephenBonikowsky:

iamjasonp:
@StephenBonikowsky - just fyi, I'm the other pair of eyes taking a look at this...

iamjasonp:
Question: was there a reason why the CodeTaskFactory was included? I didn't think it was needed for the Certificate-anything...

build-tests.cmd
```diff
@@ -0,0 +1,33 @@
...
+)
+
+echo Done Building tests.
+exit /b
+
+:Usage
+echo.
+echo Builds the tests that are in the repository.
+echo No option parameters.
+exit /b
```
iamjasonp: NIT: newline. But disregard if it breaks your diff.

build.sh
```diff
@@ -99,6 +100,25 @@ prepare_native_build()
...
+    fi
+
+    # Generate version.c if specified, else have an empty one.
+    __versionSourceFile=$__scriptpath/bin/obj/version.c
+    if [ $__generateversionsource == true ]; then
+        $__scriptpath/Tools/corerun $__scriptpath/Tools/MSBuild.exe "$__scriptpath/build.proj" /t:GenerateVersionSourceFile /p:NativeVersionSourceFile=$__scriptpath/bin/obj/version.c /p:GenerateVersionSourceFile=true /v:minimal
+    else
+        __versionSourceLine="static char sccsid[] __attribute__((used)) = \"@(#)No version information produced\";"
+        echo $__versionSourceLine > $__versionSourceFile
+    fi
```
iamjasonp: I have a feeling that this is for native only... but i guess it's no big deal for us to have if it isn't causing problems.

global.json
```diff
@@ -0,0 +1,3 @@
+{
+  "projects": [ "./src", "./src/Common" ]
+}
```
iamjasonp: ditto nit on newline.

run-test.sh
iamjasonp: I say we keep it if the cost to do so is not too high. We may need to use this for certain OSes where we have trouble building in the native environment, and have to build-then-zip from Windows.

weshaggard:
Please cc my on these type of infrastructure syncs as I want to make sure I keep a pulse on these type of changes so I can try push for changes to minimize the need for such syncs.

StephenBonikowsky:
Will do thanks @weshaggard

PR #1025, opened 2016-04-15
Harvest internal package deps

ericstj:
These are no longer needed with the latest buildtools.  They'll be
harvested just like other dependencies. /cc @roncain @StephenBonikowsky


roncain:
LGTM. 
We should probably try to coordinate this merge with the matching one in CoreFx https://github.com/dotnet/corefx/pull/7772 so we are both using the same build tools version when it TFS builds.

 @StephenBonikowsky -- any concerns or does this affect the current sync work you are doing with CoreFx scripts?

ericstj:
Shouldn't impact sync scenario.  LMK if it does.

PR #1026, opened 2016-04-15
Apply [ConditionalFact] to some tests

roncain:
This PR creates some new conditional evaluators that can be used by [ConditionalFact] and applies them to the tests that currently need them.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
@@ -2,6 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+
 namespace Infrastructure.Common
 {
     // ConditionalWcfTest is expected to be the base class of any test
@@ -10,42 +14,152 @@ namespace Infrastructure.Common
     // refer to members within their test class or its base classes.
     public class ConditionalWcfTest
     {
-        private static bool IsConditionalTestEnabled(string testCategoryName)
+        private static Dictionary<string, bool> _evaluatedConditions = new Dictionary<string, bool>(StringComparer.OrdinalIgnoreCase);
+        private static object _evaluationLock = new object();
+
+        // Returns 'true' or 'false' for the given condition name.
+        // There are several levels of precedence to evaluate the condition.
+        // The value is cached after being evaluated the first time.
+        // Precedence order for evaluation is:
+        //  1. Environment variable, if set and is 'true' or 'false'.
+        //  2. TestProperties, if set and is 'true' or 'false'
+        //  3. detection func, if specified
+        //  4. If none of the above, 'false'
+        private static bool GetConditionValue(string conditionName, Func<bool> detectFunc = null)
         {
-            string propertyValue = TestProperties.GetProperty(testCategoryName);
-            bool isEnabled;
-            if (!String.IsNullOrEmpty(propertyValue) && bool.TryParse(propertyValue, out isEnabled))
+            // Lock to evaluate once only
+            lock (_evaluationLock)
             {
-                return isEnabled;
+                bool result = false;
+                if (_evaluatedConditions.TryGetValue(conditionName, out result))
+                {
+                    return result;
+                }
+
+                bool evaluatedResult = false;
+
+                // Highest precedence: environment variable if set and can be parsed
+                string value = Environment.GetEnvironmentVariable(conditionName);
+                bool parsedValue = false;
+                if (!String.IsNullOrWhiteSpace(value) && bool.TryParse(value, out parsedValue))
+                {
+                    result = parsedValue;
+                    evaluatedResult = true;
+                }
+
+                // Next precedence: TestProperties if present and can be parsed
+                else if (TestProperties.PropertyNames.Contains(conditionName))
+                {
+                    value = TestProperties.GetProperty(conditionName);
+                    if (!String.IsNullOrWhiteSpace(value) && bool.TryParse(value, out parsedValue))
+                    {
+                        result = parsedValue;
+                        evaluatedResult = true;
+                    }
+                }
+
+                // Next precedence: optional runtime detection func
+                if (!evaluatedResult && detectFunc != null)
+                {
+                    result = detectFunc();
+                    evaluatedResult = true;
+                }
+
+                // Final precedence: false is default
+                if (!evaluatedResult)
+                {
+                    result = false;
+                }
+
+                _evaluatedConditions[conditionName] = result;
+                return result;
+            }
+        }
+
+        private static bool Is_Server_Local()
+        {
+            // Temporary workaround to detect whether this test is using localhost
+            // for either new or old test infrastructure.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            // to use centralized helper method.
+            string host = null;
+            if (TestProperties.PropertyNames.Contains("ServiceUri"))
+            {
+                host = TestProperties.GetProperty("ServiceUri");
+            }
+            else if (TestProperties.PropertyNames.Contains("BridgeHost"))
+            {
+                host = TestProperties.GetProperty("BridgeHost");
             }
 
-            return false;
+            if (String.IsNullOrWhiteSpace(host))
+            {
+                return false;
+            }
+
+            int index = host.IndexOf("localhost", 0, StringComparison.OrdinalIgnoreCase);
+            return index >= 0;
+        }
+        
+        private static bool Is_Windows()
+        {
+            return RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
         }
 
         public static bool Domain_Joined()
         {
-            return IsConditionalTestEnabled(TestProperties.Domain_Joined_PropertyName);
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(Domain_Joined),
+                                     Is_Server_Local);
         }
 
         public static bool Root_Certificate_Installed()
         {
-            return IsConditionalTestEnabled(TestProperties.Root_Certificate_Installed_PropertyName);
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this to test only when solving https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(Root_Certificate_Installed),
+                                     Is_Server_Local);
         }
 
         public static bool Client_Certificate_Installed()
         {
-            return IsConditionalTestEnabled(TestProperties.Client_Certificate_Installed_PropertyName);
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this to test only when solving https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(Client_Certificate_Installed),
+                                     Is_Server_Local);
         }
 
         public static bool SPN_Available()
         {
-            return IsConditionalTestEnabled(TestProperties.SPN_Available_PropertyName);
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(SPN_Available),
+                                     Is_Server_Local);
         }
 
         public static bool Kerberos_Available()
         {
-            return IsConditionalTestEnabled(TestProperties.Kerberos_Available_PropertyName);
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(Kerberos_Available),
+                                     Is_Server_Local);
         }
 
+        public static bool Windows_Authentication_Available()
+        {
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(Windows_Authentication_Available),
+                                     Is_Server_Local);
+        }
+
+        public static bool NTLM_Available()
+        {
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(NTLM_Available),
+                                     Is_Server_Local);
+        }
     }
 }
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
@@ -86,24 +86,37 @@
       <NegotiateTestUpn>testuser@DOMAIN.CONTOSO.COM</NegotiateTestUpn>
     </PropertyGroup>
 
+    <!-- Convention for TestProperties values to use by [ConditionalFact]:
+      'true' = yes, the condition is satisfied
+      'false' = no, the condition is not satisfied
+      empty string = self-detect at runtime if possible, else assume 'false'
+    -->
     <PropertyGroup Condition="'$(Domain_Joined)' == ''">
-      <Domain_Joined>false</Domain_Joined>
+      <Domain_Joined></Domain_Joined>
     </PropertyGroup>
   
     <PropertyGroup Condition="'$(Root_Certificate_Installed)' == ''">
-      <Root_Certificate_Installed>false</Root_Certificate_Installed>
+      <Root_Certificate_Installed></Root_Certificate_Installed>
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(Client_Certificate_Installed)' == ''">
-      <Client_Certificate_Installed>false</Client_Certificate_Installed>
+      <Client_Certificate_Installed></Client_Certificate_Installed>
     </PropertyGroup>
   
     <PropertyGroup Condition="'$(SPN_Available)' == ''">
-      <SPN_Available>false</SPN_Available>
+      <SPN_Available></SPN_Available>
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(Kerberos_Available)' == ''">
-      <Kerberos_Available>false</Kerberos_Available>
+      <Kerberos_Available></Kerberos_Available>
+    </PropertyGroup>
+    
+     <PropertyGroup Condition="'$(NTLM_Available)' == ''">
+      <NTLM_Available></NTLM_Available>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(Windows_Authentication_Available)' == ''">
+      <Windows_Authentication_Available></Windows_Authentication_Available>
     </PropertyGroup>
   
   <!--
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.targets
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.targets
@@ -43,6 +43,8 @@ namespace Infrastructure.Common
         public static readonly string Client_Certificate_Installed_PropertyName = "Client_Certificate_Installed"%3B
         public static readonly string SPN_Available_PropertyName = "SPN_Available"%3B
         public static readonly string Kerberos_Available_PropertyName = "Kerberos_Available"%3B
+        public static readonly string Windows_Authentication_Available_PropertyName = "Windows_Authentication_Available"%3B
+        public static readonly string NTLM_Available_PropertyName = "NTLM_Available"%3B
                 
         static partial void Initialize(Dictionary<string, string> properties)
         {
@@ -71,6 +73,8 @@ namespace Infrastructure.Common
             properties["Client_Certificate_Installed"] = "$(Client_Certificate_Installed)"%3B
             properties["SPN_Available"] = "$(SPN_Available)"%3B
             properties["Kerberos_Available"] = "$(Kerberos_Available)"%3B
+            properties["Windows_Authentication_Available"] = "$(Windows_Authentication_Available)"%3B
+            properties["NTLM_Available"] = "$(NTLM_Available)"%3B
         }
     }
 }
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/TestProperties.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/TestProperties.cs
@@ -16,7 +16,7 @@ namespace Infrastructure.Common
         // partial method in generated code.
         private static Lazy<Dictionary<String, String>> _properties = new Lazy<Dictionary<String, String>>(() =>
         {
-            Dictionary<String, String> properties = new Dictionary<String, String>();
+            Dictionary<String, String> properties = new Dictionary<String, String>(StringComparer.OrdinalIgnoreCase);
             Initialize(properties);
             return properties;
         });
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/Binding.Custom.Tests.csproj
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/Binding.Custom.Tests.csproj
@@ -32,6 +32,10 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Targets>Build;DebugSymbolsProjectOutputGroup</Targets>
     </ProjectReference>
+    <ProjectReference Include="$(WcfInfrastructureCommonProj)">
+      <Project>{AFD69665-89A3-42AE-A32E-AB2CBBE6EE7E}</Project>
+      <Name>Infrastructure.Common</Name>
+    </ProjectReference>
     <ProjectReference Include='$(WcfScenarioTestCommonProj)'>
       <Project>{9098B41C-9C2E-4FC6-B7D8-FC3411736A22}</Project>
       <Name>ScenarioTests.Common</Name>
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.cs
@@ -8,11 +8,12 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
+using Infrastructure.Common;
 
-public static class CustomBindingTests
+public class CustomBindingTests : ConditionalWcfTest
 {
     // Client and Server bindings setup exactly the same using default settings.
-    [Fact]
+    [ConditionalFact(nameof(Root_Certificate_Installed))]
     [OuterLoop]
     public static void DefaultSettings_Https_Text_Echo_RoundTrips_String()
     {
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/Client.ClientBase.Tests.csproj
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/Client.ClientBase.Tests.csproj
@@ -29,6 +29,10 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Targets>Build;DebugSymbolsProjectOutputGroup</Targets>
     </ProjectReference>
+    <ProjectReference Include="$(WcfInfrastructureCommonProj)">
+      <Project>{AFD69665-89A3-42AE-A32E-AB2CBBE6EE7E}</Project>
+      <Name>Infrastructure.Common</Name>
+    </ProjectReference>
     <ProjectReference Include='$(WcfScenarioTestCommonProj)'>
       <Project>{9098B41C-9C2E-4FC6-B7D8-FC3411736A22}</Project>
       <Name>ScenarioTests.Common</Name>
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
@@ -5,11 +5,12 @@ using System;
 using System.ServiceModel;
 using System.Threading.Tasks;
 using Xunit;
+using Infrastructure.Common;
 
-public static class DuplexClientBaseTests
+public class DuplexClientBaseTests : ConditionalWcfTest
 {
 
-    [Fact]
+    [ConditionalFact(nameof(Root_Certificate_Installed))]
     [OuterLoop]
     public static void DuplexClientBaseOfT_OverHttp_Call_Throws_InvalidOperation()
     {
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/Client.ExpectedExceptions.Tests.csproj
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/Client.ExpectedExceptions.Tests.csproj
@@ -32,6 +32,10 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Targets>Build;DebugSymbolsProjectOutputGroup</Targets>
     </ProjectReference>
+    <ProjectReference Include="$(WcfInfrastructureCommonProj)">
+      <Project>{AFD69665-89A3-42AE-A32E-AB2CBBE6EE7E}</Project>
+      <Name>Infrastructure.Common</Name>
+    </ProjectReference>
     <ProjectReference Include='$(WcfScenarioTestCommonProj)'>
       <Project>{9098B41C-9C2E-4FC6-B7D8-FC3411736A22}</Project>
       <Name>ScenarioTests.Common</Name>
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -14,7 +14,10 @@ using Xunit;
 using System.Threading;
 using System.IO;
 
-public static class ExpectedExceptionTests
+using Infrastructure.Common;
+
+
+public class ExpectedExceptionTests : ConditionalWcfTest
 {
     [Fact]
     [OuterLoop]
@@ -491,7 +494,7 @@ public static class ExpectedExceptionTests
         }
     }
 
-    [Fact]
+    [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed))]
     [OuterLoop]
     // Confirm that the Validate method of the custom X509CertificateValidator is called and that an exception thrown there is handled correctly.
     public static void TCP_ServiceCertFailedCustomValidate_Throw_Exception()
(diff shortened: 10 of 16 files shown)


roncain:
Reviewers -- this is for an early look and an opportunity to comment.  I need to rebase and make a few enhancements before testing it more broadly.

roncain:
@iamjasonp and @StephenBonikowsky -- I am interested in your early thoughts about the approach here.  Conditions that appear in [ConditionalFact] are all defined in ConditionalWcfTest as methods. Any condition can be explicitly set true or false via an environment variable.  But if not set, it falls back to auto-detection logic, which admittedly is not fully flushed out.  The goal of the first check in is to have a full run with no failures on any platform.  Instead, you will see the tests that were skipped and which condition was not met for that test to run.  The ActiveIssue condition I added is an experiment to see how this feels in practice.  I can make that test run by setting its condition to true in an environment variable, which allows me to debug a single active issue at a time without modifying source code.

I need to merge this into the infrastructure changes happening the first of next week, so I'd appreciate review comments.  There will be follow-on PR's as this matures.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
```diff
@@ -86,26 +86,39 @@
...
     </PropertyGroup>
 
+    <!-- Convention for TestProperties values to use by [ConditionalFact]:
+      'true' = yes, the condition is satisfied
+      'false' = no, the condition is not satisfied
+      empty string = self-detect at runtime if possible, else assume 'false'
+    -->
     <PropertyGroup Condition="'$(Domain_Joined)' == ''">
-      <Domain_Joined>false</Domain_Joined>
+      <Domain_Joined></Domain_Joined>
```
roncain: Reviewers -- a blank value now means "auto detect please".  Some auto-detection is over-simplified to get this to work across our platforms, but it will mature into real tests at some point.

StephenBonikowsky:
Design looks good considering we have to work within the ConditionalFact requirements.
I don't like having to create a new ActiveIssue function for each new issue, I realize why you are doing it but still don't like it.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
+                propertyValue = TestProperties.GetProperty(conditionName);
+                if (string.IsNullOrWhiteSpace(propertyValue))
+                {
+                    // Empty string means auto-detect if possible, else false
+                    return (detectFunc == null) ? false : detectFunc();
+                }
+                else
+                {
+                    bool conditionResult = false;
+                    if (!bool.TryParse(propertyValue, out conditionResult))
```
iamjasonp: Feels like we have a tri-state here, "true", "false", "auto (default)" 

I think that if the value is not parsable as "true" or "false", we should probably fall back to "auto (default)" rather than assume false. My thinking: "the value you specified was invalid, so we'll ignore your input". 

All this said, I'm ok either way

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
+            {
+                host = TestProperties.GetProperty("BridgeHost");
+            }
+             if (String.IsNullOrWhiteSpace(host))
+            {
+                return false;
+            }
+            int index = host.IndexOf("localhost", 0, StringComparison.OrdinalIgnoreCase);
+            return index >= 0;
+        }
```
iamjasonp: Question only: TestProperties.PropertyNames will deal case sensitivity, right?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
+        // 'true' or 'false', it will be used as is.  If the value is
+        // empty, and 'detectFunc' is specified, it will be invoked to
+        // evaluate the condition at runtime.  The value is cached, so
+        // 'detectFunc' will be called only once.
+        private static bool GetConditionValueFromTestProperties(string conditionName, Func<bool> detectFunc = null)
+        {
+            string propertyValue;
+            return GetConditionValue(conditionName, () =>
+            {
+                propertyValue = TestProperties.GetProperty(conditionName);
```
iamjasonp: Question only: will TestProperties.GetProperty deal with case (in)sensitivity?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
 
         public static bool Domain_Joined()
         {
-            return IsConditionalTestEnabled(TestProperties.Domain_Joined_PropertyName);
+            // If TestProperties does not specify true/false explicitly, we need to autodetect whether
+            // the client and server are domain joined.
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValueFromTestProperties(TestProperties.Domain_Joined_PropertyName,
+                                                       Detect_Server_Is_Local);
```
iamjasonp: A heuristic I've used in the past for "is domain joined" 

* Is running Windows
* LOGONSERVER is not equal to COMPUTERNAME
* USERDOMAIN is not equal to COMPUTERNAME 

When the LOGONSERVER and COMPUTERNAME are the same, it means you're logging in as a local user (you don't have access to domain resources) or you are not part of a domain. This fails when you are logged in on the domain controller itself. 

The second check will help prevent the above case, as few environments will name a machine the same as a domain. 

That said, it should be fairly rare that anyone is actually doing work on a DC machine.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
-            return IsConditionalTestEnabled(TestProperties.Client_Certificate_Installed_PropertyName);
+            // Give precedence to TestProperties, but if the value is unspecified, go ahead and
+            // try to install the local cert now.  Failure auto-detects it cannot be done and skips
+            // subsequent tests.
+            return GetConditionValueFromTestProperties(TestProperties.Client_Certificate_Installed_PropertyName,
+                                                       () =>
+                                                       {
+                                                           try
+                                                           {
+                                                               BridgeClientCertificateManager.InstallLocalCertificateFromBridge();
```
iamjasonp: I'm a little wary of doing the work to install certs here... feels to me that: 

* if we can automatically do it, then detection isn't necessary
* if there are certain conditions precluding the "do it automatically" action, then we should condition on that instead. 

So maybe `Client_Certificate_Installed` and `Root_Certificate_Installed` need only be conditioned on "Is_Elevated" or something similar

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
-            return IsConditionalTestEnabled(TestProperties.Kerberos_Available_PropertyName);
+            // If TestProperties does not specify true/false explicitly, we need to autodetect whether
+            // Kerberos is available.
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValueFromTestProperties(TestProperties.Kerberos_Available_PropertyName,
+                                                        Detect_Server_Is_Local);
+        }
+
+        public static bool Windows_Authorization_Available()
```
zhenlan: Do you mean windows authentication or is this something else?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
zhenlan: Question only: what's the difference between this and not having this PropertyGroup at all?

zhenlan:
This may worth some design discussion. My intuitive expectation of where a condition is read is in this order

1. TestProperty
2. Environment variable
3. Auto detection
4. False

Basically, if I explicitly specify a condition through TestProperty, I will expect it will be respected regardless. If the condition is not specified explicitly as TestProperty, it's up to the environment. If no one specify the condition through either TestProperty or environment variable, we will try to be smart in code (auto detection) if we can. If we cannot auto detect, the condition will be false.

Is this something reasonable? Current proposed code has 1. and 2. in reversed order.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
zhenlan: +1. This will be a better way to implement "is domain joined", but this can only detect if client machine is domain joined. Is "server domain joined" part of this condition or is it separate?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
zhenlan: @iamjasonp I have a different opinion. If a condition is specified explicitly through either TestProperty or environment variable but the value is invalid, we should always use false instead of trying to be smart (auto detection). This will make sure we have consistent results and also help to discover the invalid value (most likely unintentionally).

I think the problem here is we don't have an agreement on design yet. I suggest us to conclude [this design discussion](https://github.com/dotnet/wcf/pull/1026#issuecomment-211263175) before proceeding.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
     public class ConditionalWcfTest
     {
-        private static bool IsConditionalTestEnabled(string testCategoryName)
+        private static ConcurrentDictionary<string, bool> _evaluatedConditions = new ConcurrentDictionary<string, bool>();
+
+        // Returns 'true' or 'false' for the given condition name.
+        // If this is the first call and 'detectFunc' is specified, it will be
+        // called to detect whether the condition is true or false.  If not
+        // specified, it will default to false. The value is cached after first call.
+        private static bool GetConditionValue(string conditionName, Func<bool> detectFunc = null)
```
zhenlan: The existence of `GetConditonValue(...)` and `GetConditionValueFromTestProperties(...)` are very confusing. I recommend us to conclude [this design discussion](https://github.com/dotnet/wcf/pull/1026#issuecomment-211263175) and merge these two functions to be one that always follows the same searching order.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
```diff
@@ -86,26 +86,39 @@
...
+      <SPN_Available></SPN_Available>
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(Kerberos_Available)' == ''">
-      <Kerberos_Available>false</Kerberos_Available>
+      <Kerberos_Available></Kerberos_Available>
     </PropertyGroup>
-  
+    
+     <PropertyGroup Condition="'$(NTLM_Available)' == ''">
```
zhenlan: Client or server?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
```diff
@@ -86,26 +86,39 @@
...
-      <Client_Certificate_Installed>false</Client_Certificate_Installed>
+      <Client_Certificate_Installed></Client_Certificate_Installed>
     </PropertyGroup>
   
     <PropertyGroup Condition="'$(SPN_Available)' == ''">
-      <SPN_Available>false</SPN_Available>
+      <SPN_Available></SPN_Available>
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(Kerberos_Available)' == ''">
```
zhenlan: Client or server?

zhenlan:
I know there are some discussions about using [ConditionalFact] to replace [ActiveIssue], I am aware of there are some pros and cons in either option, but I am not fully convinced this the right move we should take yet. Even if we do move, we probably should have some design discussion about what's the right approach to do it. @roncain are you aware of any more official recommendation between using [ConditionalFact] and [ActiveIssue]?

I think this probably will be too big a turn for this PR to take. I recommend for this PR to stick with using [ActiveIssue] before we conclude the discussion of above.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
zhenlan: In an ideal design, a detection logic should _not_ change the environment. It's like a getter should only return the state/property of an object instead of changing the object.

I understand we don't have a better option here due to the lack of an extension point in current test system where we can call setup functions before any tests starting to run. We should move functions like this to setup if we find a way to call them before any tests start in the future.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
+        {
+            // If TestProperties does not specify true/false explicitly, we need to autodetect whether
+            // NTLM is available.
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValueFromTestProperties(TestProperties.NTLM_Available_PropertyName,
+                                                        Detect_Server_Is_Local);
+        }
+
+        public static bool WebSockets_Supported()
```
zhenlan: This eventually will become a discussion about what should be a ConditionalFact vs. ActiveIssue. In my understanding, the fact that websocket does not work on *nix is bug, which is very different from windows stream security tests will not work without some special setup (but there is no product bug).

I suggest us not to do this change in this PR (ie. continue using ActiveIssue for WebSocket for now) until we conclude [this discussion](https://github.com/dotnet/wcf/pull/1026#issuecomment-211284415).

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
+        {
+            return GetConditionValue(nameof(WebSockets_Supported), Detect_Is_Windows);
+        }
+
+        // Active issues
+
+        // Remove this method after 979 is fixed.
+        // Set the environment variable ActiveIssue_979_Fixed=true to override this default.
+        public static bool ActiveIssue_979_Fixed()
```
zhenlan: Similar suggestion as I had for WebSocket above.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -169,10 +170,10 @@ public static void WebSocket_Http_Duplex_BinaryStreamed()
         }
     }
 
-    [Fact]
+    [ConditionalFact(nameof(WebSockets_Supported), nameof(Root_Certificate_Installed))]
     [OuterLoop]
 #if !FEATURE_NETNATIVE
-    [ActiveIssue(420, PlatformID.AnyUnix)]
+    //[ActiveIssue(420, PlatformID.AnyUnix)]
```
zhenlan: The ActiveIssue 420 applies to AnyUnix only. Wouldn't the ConditionalFact WebSockets_Support will disable this test everywhere?

There are many other similar cases like this in this PR. I will not point them out one by one, but I think you get the idea.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
zhenlan: Just to point out that, while we are using concurrent dictionary, [MSDN](https://msdn.microsoft.com/en-us/library/ee378677(v=vs.110).aspx) said the func could be called multiple times (concurrently?). So depending on how reliable our cert install code is, tests may not work as nicely as we hope.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
iamjasonp: "Server is domain joined" is a necessary condition for Kerberos, so I think it'll fall into that category.

iamjasonp:
> 1. TestProperty
> 2. Environment variable
> 3. Auto detection
> 4. False

I think the order currently implemented, that is, 

1. Env var
2. TestProperty
3. (and so on) 

is correct to me. I want to be able to build once and be able to modify the env var as necessary to vary my testing conditions without rebuilding.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.targets
```diff
@@ -71,6 +73,8 @@ namespace Infrastructure.Common
             properties["Client_Certificate_Installed"] = "$(Client_Certificate_Installed)"%3B
             properties["SPN_Available"] = "$(SPN_Available)"%3B
             properties["Kerberos_Available"] = "$(Kerberos_Available)"%3B
+            properties["Windows_Authorization_Available"] = "$(Windows_Authorization_Available)"%3B
```
iamjasonp: I think you mean "authentication" not "authorization"?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
+                                                           try
+                                                           {
+                                                               BridgeClientCertificateManager.InstallRootCertificateFromBridge();
+                                                               return true;
+                                                           }
+                                                           catch
+                                                           {
+                                                               return false;
+                                                           }
+                                                       });
```
iamjasonp: Can we fix the indenting? Seems a little far to the right :smile:

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +15,191 @@ namespace Infrastructure.Common
...
+            string host = null;
+            if (TestProperties.PropertyNames.Contains("ServiceUri"))
+            {
+                host = TestProperties.GetProperty("ServiceUri");
+            }
+            else if (TestProperties.PropertyNames.Contains("BridgeHost"))
+            {
+                host = TestProperties.GetProperty("BridgeHost");
+            }
+             if (String.IsNullOrWhiteSpace(host))
```
iamjasonp: NIT: spacing

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
iamjasonp: NIT: you could host locally using the hostname of the machine, in theory, and it'd still be "local".  But for a temporary heuristic, I think we're okay. 

Maybe Is_Localhost might better describe your condition better.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
iamjasonp: For me personally - it's nice to see what properties are settable in one place

iamjasonp:
Most of the things I have are NIT-level or consider-level comments, so you have my LGTM.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
iamjasonp: If we're looking for consistent results and error discovery, rather than being tolerant of bad input, we should hard throw instead of ignoring or guessing - both "guessing" and "ignoring" behaviours are strictly wrong.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
roncain: Worth discussing in design review.  Initially they were all here because they all defaulted to 'false'. Now they are here to specify that (1) they exist to be set, and (2) they default to auto-detect. I left them here primarily because of the reason @iamjasonp stated -- visibility.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: Unfortunately, the point at which this is parsed is in xunit test discovery.  Throwing there will probably bring down the entire xunit run.  There is also no way currently to deliver a custom message (like "Conditions(s) not met: 'mumble' is not a valid Boolean value for Root_Certificate_Installed"), so there is no good way to report the issue.

I think defaulting to false is the best approach because it will clearly say "Conditions not met: Root_Certificate_Installed", and if you know you tried to set it to 'true' then the only explanation is that you must have done it wrong.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: I think we need an unambiguous test infrastructure method we can call for this.  The logic doesn't belong here.  Bear in mind, part of this initial checkin is bounded by making some simple heuristic decisions for logic we probably want supported by the test infrastructure.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: Good point -- it isn't today and I will make it be case-insensitive.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: I would like to distinguish between the [ConditionalFact] that returns the decision and having test infrastructure code that does the actual work.  I would prefer to see methods like "Is_Service_Local" and "Is_Domain_Joined" be in one place and re-used by any code that needs it.  My interpretation of this initial checkin was to become concrete about all the *conditions* we need to branch on, and then to provide solid implementations later.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: Oops, good catch.  Will change to Windows Authentication.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: A couple observations worth discussing.
1. Should detection do work? Initially I thought 'no' but then I discovered we had multiple ways of stating a dependency ([ConditionFact(nameof(Root_Certificate_Installed)] and in Endpoints.cs where we do EnsureRootCertificateInstalled on specific endpoints).  So after some thought, I concluded it was fragile to state this dependency using 2 different mechanisms in 2 different places, and [ConditionalFact] became the ideal single place to state a dependency.  Given that, it needed the right to do work that could change the environment.  If we decide this is incorrect, then I want to revisit our use of test helper methods like EnsureRootCertificateInstalled, and we need to have test infrastructure support to allow us to ask "Is the Root Certificate installed?"

Also, bear in mind that we don't have the pre-run setup logic we need to make all the setup happen before tests, so having detection do work is necessary.  I am happy to make detectors only detect once we have infrastructure support to do the work in advance.  This is the initial checkin, like all the rest of the infrastructure changes.

2. Good catch on concurrent dictionary.  I have to say I find CD's design counter-intuitive in this regard. I can convert to a normal dictionary and a lock to prevent this multiple func invoke.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: I agree we can combine into a single method.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
roncain: Honestly, I was waiting for requested feedback on https://github.com/dotnet/wcf/issues/958 to settle on the conditions.  Example, when do we need to distinguish between root and client cert (do you always need to require root if you are testing client certs)?  When do we need to distinguish between client and server?  Do we need to distinguish between domain-joined server but not a domain-joined client.

This PR was a way to say "here they are -- tell me what I'm missing".

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: I can remove the [ActiveIssue] example to avoid complicating the discussion.  It was intended to foster this very discussion 😀 .

But I think we need to have some clarity between what is a condition and what is an issue. WebSockets on *nix is a good example.  If we mark them only [ActiveIssue], they silently don't run.  If we mark them with [ConditionalFact], then we see "Condition(s) not met: web sockets supported".  My personal opinion is that [ConditionaFact] should be used for these scenarios:

- This test environment is not setup to allow feature X to be tested

- Feature X is not supported in this environment, and never will be

- Feature X is not currently supported in this environment (and an active issue is tracking it)

When there is a legitimate product failure (ex: wrong exception from streaming error) then I think it is a true ActiveIssue, and we need to settle on how that is represented.  We could invent our own [ActiveIssueFact(999)] is this helps distinguish

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
roncain: No, currently the WebSockets_Supported condition checks the platform and returns true for Windows, so it would run.  This is a perfect example where the websockets tests need to say "I need web socket support to run".  The alternative is listing the same one or more active issues on each web socket test which is less clear and duplicated information.

This gets to the heart of what [ConditionalFact] means.  To me it means "this test requires X to run" where X can be a feature, platform, machine configuration, environment variable, etc.  If there is one or more active issues preventing feature X from running, I think they should be listed in the implementation of the condition (e.g. WebSockets_Supported).  I can make that clearer in the code.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
roncain: Do we need to distinguish at the [ConditionalFact] level?
Again, waiting for closure on the questions in #958

roncain:
Thanks @iamjasonp.  It wasn't clear to me which issues raised by @zhenlan or @StephenBonikowsky were blocking comments.  I will address the obvious and wait for further feedback or discussion before merging.  Reviewer request: as with the recent infrastructure PR's, it is very useful if you identify which comments are non-blocking for the initial check-in.  I would like to merge whatever subset we agree on for an initial checkin and then refine on follow-up PR's, as I think we are doing with all the other infrastructure PR's.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
roncain: Regarding an earlier comment from @iamjasonp -- [ConditionalFact] is meant to be used to detect things available *to* the client, not necessarily *on* the client.  I fully expect some of these will involve calls to helper methods that may attempt to contact the utility WCF service to ask questions.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
zhenlan: > I'm under the assumption that our current [ConditionalFact] discussion is based only on things that can be entirely decided on the client

ConditionalFact is not limited to local conditions. It's our auto detection that can tell only local conditions for the most of the time.

>  I would prefer to see methods like "Is_Service_Local" and "Is_Domain_Joined" be in one place and re-used by any code that needs it.

So sounds like the condition you really mean is something like "Is_Client_Domain_Joined". For tests that will work only if both client and remote server are domain joined, they will need an additional condition, something like "Is_Server_Domain_Joined". The condition "Is_Server_Domain_Joined" can potentially be auto detected if "Is_Service_Local" is true.

> My interpretation of this initial checkin was to become concrete about all the conditions we need to branch on, and then to provide solid implementations later.

Agreed. I think current discussion _is_ about what are concrete conditions for domain joined scenarios. I did not bother implementation too much, although I prefer to see most auto detection logic to be more restrict (return false if cannot be sure).

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
zhenlan: I will not be against throwing, but what @roncain said makes sense to me.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
zhenlan: TestProperties.props is not the place to look for the complete list of TestProperties. It is the place to provide default value if a TestProperty has one. TestProperties.targets is the place to look for complete list. If our build does not require every TestProperty to show up in TestProperties.props, sooner or later people will need to look TestProperties.targets for the ultimate truth anyway.

mconnew:
I would be inclined to check in rather than spend more time debating things. As long as the approach is flexible to change behavior on those points of debate such as order of precedence of test property vs environment variable without a massive rewrite, we should prefer moving forwards even if it isn't the exact solution we might end up with.

mconnew:
LGTM

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
zhenlan: As I mentioned I understand it is because we don't have a pre-run setup logic in current testing systems (TOF or Helix), I am okay your current proposal as a temporary solution for us.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +14,198 @@ namespace Infrastructure.Common
...
+
+                // Next precedence: optional runtime detection func
+                if (!evaluatedResult && detectFunc != null)
+                {
+                    result = detectFunc();
+                    evaluatedResult = true;
+                }
+
+                // Final precedence: false
+                else if (!evaluatedResult && detectFunc != null)
```
zhenlan: Did you mean to have `else` only (no `if`)?

zhenlan:
> I want to be able to build once and be able to modify the env var as necessary to vary my testing conditions without rebuilding.

If this is what you want, you should not specify TestProperty at the build time and then you are free to use environment variable later. If I use TestProperty at build time, I expect it will be respected (enforced).

@roncain since this the foundation of ConditionalFact we are going to use. Other design or scripts will depend on this logic (expectation), I would like us to conclude this in the first round.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
zhenlan: Even if server has NTLM enabled, some Linux distros require particular packages for NTLM to work from client side. This is why my question. Yes, we should finish the design on #985. This is not blocking.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
zhenlan: Agreed. We need to finish the design on #985 (I am looking at you @iamjasonp 😄). This is not blocking.

roncain:
> If this is what you want, you should not specify TestProperty at the build time and then you are free to use environment variable later. If I use TestProperty at build time, I expect it will be respected (enforced).
> 
> @roncain since this the foundation of ConditionalFact we are going to use. Other design or scripts will depend on this logic (expectation), I would like us to conclude this in the first round.
> 

We want to retain the current semantics of TestProperties (and follow them in [ConditionalFact]), namely:

- At build time, TestProperties are set, either from environment variables, msbuild command line properties, or their defaults.  These values are code-gen'ed into the source code and compiled into the test binaries.

- At runtime *unless you say otherwise* these are the values used to run the test.  The way you "say otherwise" is through environment variable.

This went through several iterations and has proven to be the most useful pattern in practice.  It is important to remember that the environment to *run* the tests is usually not the environment where they were *built*.  We want to the test binaries to arrive at the run environment fully configured for TestProperties, but we want the run environment freedom to override without rebuilding.

This PR is simply re-using this convention for [ConditionalFact], and I believe it is the correct convention

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
zhenlan: Then the condition is really whether it's Windows instead of whether websocket is supported. I think you are right "this gets to the heart of what [ConditionalFact] means". I created #1034 to track this discussion. I prefer we don't change websocket tests just yet, but I will not block on this.

zhenlan:
> This went through several iterations and has proven to be the most useful pattern in practice.

Great. If this is the most useful pattern, I think we can conclude the discussion on this subject. We are cleared on this.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -10,42 +14,198 @@ namespace Infrastructure.Common
...
+            int index = host.IndexOf("localhost", 0, StringComparison.OrdinalIgnoreCase);
+            return index >= 0;
+        }
+        
+        private static bool Is_Windows()
+        {
+            return RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+        }
+
+        private static bool Is_NET_Native()
```
zhenlan: Not seen this is used anywhere. Maybe we can include it when we actually use it?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/testproperties.props
roncain: We can discuss this further, but the convention to date (prior to ConditionalFact] has been that testproperties.props is the "go to" place to see both the list of properties *and their default values*.  I prefer not to reverse that convention in this PR related to ConditionalFact.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
roncain: I've decided to remove from this PR all attempts to remove or replace [ActiveIssue].  Once we close on #1034, we can apply those decisions.

roncain:

roncain:
Reviewers: I am going to merge a subset of this PR to be able to move forward with infrastructure integration. I have removed from the PR:
 - All changes related to [ActiveIssue] -- we can address those later
 - The [ConditionalFact] "detectors" that used the Bridge to install certificates

Issue #1015, opened 2016-04-14
Use TestProperties and to parse command line arguments to configure ports


hongdai:
Pull request https://github.com/dotnet/wcf/pull/1011 recommended using TestProperties and to parse command line arguments to configure ports for self hosted service.

We need to evaluate if it's needed to configure ports as there is a debate to use the same port IIS uses.

hongdai:
The as is test infrastructure has been running OK. Further improvement is not needed for now.

PR #1027, opened 2016-04-15
Init checkin StartWCFService.cmd

hongdai:
* also move BuildWCFSelfHostedService.cmd to scripts dir.
* I have tested it on the cross machine. Only some websocket and security tests failed due to other reasons.


src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,50 @@
...
+    echo ERROR: An error occurred while building the Certificate generator.
+    set __EXITCODE=%ERRORLEVEL%
+    goto :done
+  )
+
+
+REM Config Certs
+REM we need the direcotry to save the test.crl file. We are investigate a way to get rid of it
+REM
+md c:\wcftest
```
roncain: Please make a note to find a way to remove these hard coded paths.   Not blocking for this merge.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,50 @@
...
+pushd ..\..\..\..\bin\Wcf\tools\CertificateGenerator
+call CertificateGenerator.exe
```
roncain: It would be better to base the path on %~dp0 rather than assume the user was executing from this folder and use this relative path.  Not blocking but please fix at some point.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,50 @@
...
+  )
+
+powershell -NoProfile -ExecutionPolicy unrestricted .\ConfigHttpsPort.ps1
+if NOT [%ERRORLEVEL%]==[0] (
+    echo Warning: An error occurred while configuration https port.
+  )
+
+REM Start the self hosted WCF Test Service
+pushd ..\..\..\..\bin\Wcf\tools\SelfHostedWcfService
+call SelfHostedWcfService.exe
```
roncain: Same comment about %~dp0

roncain: Not blocking for now, but it should pass %* to SelfHostedWcfService.exe, which will probably parse TestProperties.  This would allow us to wire StartWcfService.cmd directly into build.cmd and flow msbuild properties all the way down.  This is how the bridge plumbing works today.

roncain:
LGTM.  My review comments are not blocking and can be dealt with later.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
roncain: I forgot to ask -- what happens if we call this from a non-elevated window?  We're still establishing the rules whether we need to run elevated.  The Bridge.exe project shows how to make an exe that runs elevated from a non-elevated window if we decide this is necessary.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,53 @@
+@echo off 
+setlocal
+
+REM 
+REM Start WCF Self Hosted Service on local machine
+REM
+
+REM Build tools
+call BuildWCFSelfHostedService.cmd
```
zhenlan: add %~dp0

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,53 @@
...
+if NOT [%ERRORLEVEL%]==[0] (
+    echo ERROR: An error occurred while building WCF Self hosted Service.
+    set __EXITCODE=%ERRORLEVEL% 
+    goto :done
+  )
+call BuildCertUtil.cmd
```
zhenlan: add %~dp0

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,53 @@
...
+pushd ..\..\..\..\bin\Wcf\tools\CertificateGenerator
+call CertificateGenerator.exe
+set __EXITCODE=%ERRORLEVEL%
+popd
+if NOT [%__EXITCODE%]==[0] (
+    echo Warning: An error occurred while running certificate generator.
+  )
+
+REM skip the checking of error code as OpenFirewallPorts can return different error code even it's successful
+call OpenFirewallPorts.cmd
```
zhenlan: add %~dp0

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,53 @@
...
+
+powershell -NoProfile -ExecutionPolicy unrestricted .\ConfigHttpsPort.ps1
```
zhenlan: add %~dp0

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -0,0 +1,18 @@
...
+
+REM 
+REM Start Open all ports WCF test services use
+REM
+
+netsh advfirewall firewall add rule name="_WCF Test Server PortHttp" dir=in action=allow profile=any localport=8081 protocol=tcp
+netsh advfirewall firewall add rule name="_WCF Test Server PortHttps" dir=in action=allow profile=any localport=44285 protocol=tcp
+netsh advfirewall firewall add rule name="_WCF Test Server PortTcp" dir=in action=allow profile=any localport=809 protocol=tcp
+netsh advfirewall firewall add rule name="_WCF Test Server PortWebSocket" dir=in action=allow profile=any localport=8083 protocol=tcp
+netsh advfirewall firewall add rule name="_WCF Test Server PortWebSockets" dir=in action=allow profile=any localport=8084 protocol=tcp
```
zhenlan: If our setup hard coded the ports, it will be much less meaningful for the ports to be configurable for the self hosted wcf service. We have issue #1015 tracking the port configuration issue. We should include setup as part of the consideration.

zhenlan:
Please fix the `%~dp0` issue before merging; otherwise, the script will be too fragile to run. Other than that this PR LGTM.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,53 @@
...
+REM we need the direcotry to save the test.crl file. We are investigate a way to get rid of it
+REM
+md c:\wcftest
+pushd ..\..\..\..\bin\Wcf\tools\CertificateGenerator
+call CertificateGenerator.exe
+set __EXITCODE=%ERRORLEVEL%
+popd
+if NOT [%__EXITCODE%]==[0] (
+    echo Warning: An error occurred while running certificate generator.
+  )
```
iamjasonp: Should this script bail out if %__EXITCODE% != 0?

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,53 @@
...
+
+REM skip the checking of error code as OpenFirewallPorts can return different error code even it's successful
+call OpenFirewallPorts.cmd
+
+powershell -NoProfile -ExecutionPolicy unrestricted .\ConfigHttpsPort.ps1
+if NOT [%ERRORLEVEL%]==[0] (
+    echo Warning: An error occurred while configuration https port.
+  )
```
iamjasonp: Ditto - we should exit the script if ERRORLEVEL != 0

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -0,0 +1,53 @@
+@echo off 
+setlocal
+
+REM 
+REM Start WCF Self Hosted Service on local machine
```
iamjasonp: I feel like this should echo rather than be a REM.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -0,0 +1,18 @@
...
+
+set __EXITCODE=%ERRORLEVEL%
+
+endlocal
+exit /b %__EXITCODE% 
+
```
iamjasonp: NIT: We should have a corresponding "CleanupFirewallPorts.cmd" as well.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -0,0 +1,18 @@
...
+REM Start Open all ports WCF test services use
+REM
+
+netsh advfirewall firewall add rule name="_WCF Test Server PortHttp" dir=in action=allow profile=any localport=8081 protocol=tcp
+netsh advfirewall firewall add rule name="_WCF Test Server PortHttps" dir=in action=allow profile=any localport=44285 protocol=tcp
+netsh advfirewall firewall add rule name="_WCF Test Server PortTcp" dir=in action=allow profile=any localport=809 protocol=tcp
+netsh advfirewall firewall add rule name="_WCF Test Server PortWebSocket" dir=in action=allow profile=any localport=8083 protocol=tcp
+netsh advfirewall firewall add rule name="_WCF Test Server PortWebSockets" dir=in action=allow profile=any localport=8084 protocol=tcp
+
+set __EXITCODE=%ERRORLEVEL%
```
iamjasonp: This _EXITCODE will only be for the final netsh command - that is, we're not testing the errorlevel for any of the previous commands, so we won't know what failed, and we're not bailing early if any of the previous rules are unsuccessful. 

Making matters slightly worse, someone who now looks in their firewall rules will see some of the rules created but not others,  probably making an assumption that all the rules were created just fine. 

I would prefer an all or nothing approach, which either means that we clean up all of the ones we tried to create if unsuccessful. At the very least we should report on them and note that _any_ of them failing will cause the script to bail out

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -0,0 +1,18 @@
+@echo off 
+setlocal
+
+REM 
+REM Start Open all ports WCF test services use
```
iamjasonp: echo vs. REM

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -0,0 +1,18 @@
...
+REM
+
```
iamjasonp: I think you need to do some kind of elevated prompt check too, as all the netsh commands will fail if you run this while not elevated. A trick I've used before is to run the command "net session" which requires admin rights to execute properly. 

somthing like: 

```
net session > nul
if not %ERRORLEVEL% equ 0 (
    echo this script must be run elevated.  
    exit /b 1
)

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
iamjasonp: PS. The "all or nothing" approach is not blocking; the EXITCODE issue should be fixed prior to merge

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
roncain: Just a side note for consideration -- not blocking of this PR.  The existing Bridge self-elevated so that it could do operations like this.  It is possible for the dev today to run from a non-elevated window (using the Bridge) and still succeed for most tests.  We may also shortly add a [ConditionalFact("elevated")] to allow bypassing tests requiring elevation. This current script can only be run elevated, or it will fail (right now silently), which is a loss of functionality. And it also means build.cmd will need to be run elevated too. Personally I believe we want a self-elevating exe to do the 'netsh' work, as we did with the Bridge.  It is also the place where we would parse the ports from command line variables or environment variables, as we did with the Bridge.

hongdai: I'd prefer to avoid exe for readability reason. Let's investigate if there is a way to self elevating in a script.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: Agreed. I'll add it.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: Will add %*.
 Call from a non-elevated window is not supported now as we are still establishing the rules. If we decide we want to run non-elevated, we can add the support.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: I would like to get feedback on this. I currently don't bail out for cert and firewall failures as those tests do not require certificates would still pass.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: Ditto. I would like a vote on this one. Shall we exit for the cert configuration failure?

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
hongdai: Agreed.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
hongdai: This is because I don't know what's the expected successful code for netsh. It return NON zero even it succeeded. I will need some time to figure out all the successful non zero codes. So the exit code here is not meaningful at the time being - we cannot bail out. I will add a comment here.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
hongdai: For the elevation check, I will add it AFTER I enable checking exit code for the OpenFireWallPorts script.

roncain: Not blocking, but let's discuss the exe/script issue before investigating doing in a script.  The self-elevation and msbuild property parsing is a solved issue in the Bridge we can easily copy and paste. There is no need to reinvent the same code in script.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
roncain: Great area for later discussion regarding [ConditionalFact].  If we can't install firewall rules. it would be great for those tests to be skipped with "Condition(s) not met: firewall rule for port NNN installed".  This could involve setup scripts leaving "breadcrumb" files that [ConditionalFact] can access.  Definitely not needed for this PR!  I do agree in principal that if certain parts of setup fail, we should not bail out entirely, but should leave a clear log of what went wrong and potentially run the tests that can run.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
roncain: IMHO, I'd prefer to log everything that goes wrong to a known log file and keep going.  At the end, check that "anything" went wrong and issue an ECHO about it.   This is how clean.cmd works, and it seems to be a good compromise.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
roncain: Thanks.  Regarding elevated vs non-elevated, let's agree to keep that topic alive.  I run both elevated and non-elevated today, so these changes are losing features I already use.  I don't think it is correct to expect users doing build.cmd (which will presumably call this) must run elevated.

hongdai: If we want to add the support, we can call startWcfService elevated. The build.cmd can remain not elevated.

PR #1028, opened 2016-04-15
Get service FQDN from the server instead of using BridgeHost

hongdai:
* We do not need to specify BridgeHost if we get it from the server.
* The service side already has the GetFQDN method, no need to update the service.


zhenlan:
LGTM

zhenlan:
GitHub does not allow me to comment on it, but this [line](https://github.com/dotnet/wcf/pull/1028/files#diff-5413d874ad35f97354d88cef0af3e152R453) also needs to be updated.

roncain:
LGTM

zhenlan:
@dotnet-bot test this please

zhenlan:
Interesting... how could this pass the windows build while Linux build can actually catch the compilation error!

PR #1029, opened 2016-04-15
Change the way we identify if it's IISHosted

hongdai:
* The previous method does not work for Self host cross machine.


zhenlan:
LGTM

zhenlan:
Just realized this could AV if serviceUri is not specified. We need to fix this later.

PR #1030, opened 2016-04-15
Update version to 24015-00 using new PS script.

StephenBonikowsky:
* Checking-in a simpler version of a corefx script to update our repo to use the latest package version.
* Updating from rc3-23931-00 to rc3-24015-00


StephenBonikowsky:
Missed some new tests, rebasing and updating.

iamjasonp:
Buddy built and builds in TFS for me, LGTM

UpdateDependencies.ps1
```diff
@@ -0,0 +1,52 @@
+﻿#
```
roncain: Nit: I would have preferred the ps1 and the version changes to be separate commits within this PR.  It makes cherry picking easier if you don't have to take both.  No need to change for this PR.

src/Common/test-runtime/LatestDependencies/project.json
```diff
@@ -1,74 +0,0 @@
-{
```
roncain: Question only: did the automatic tool delete this.  Was it manually deleted?  Don't block on the question.

roncain:
LGTM

src/Common/test-runtime/LatestDependencies/project.json
StephenBonikowsky: I deleted it, Hong said it is no longer used or needed.

PR #1036, opened 2016-04-18
Script to setup WCF test services hosted on IIS for WCF Core testing

zhenlan:
Script to Setup WCF test services hosted on IIS for the testing of WCF on .NET Core

Usage: WcfTestServerSetup Id [/c]
   Id: The Id of a WCF repo and its associated IIS hosted services to be created.
       The Id will be appended to the name of all assets to be created.
       Master: if Id is 'Master', additional IIS hosted services such as PRService
       will be created to serve as central services for other WCF repos.
   /c: If specified, this will clean up any existing setup of Id instead of creating new.

Examples:
   WcfTestServerSetup 42
   :Create a WCF repo named 'wcf42' and IIS hosted services such as 'WcfService42'
   WcfTestServerSetup 42 /c
   :Clean up WCF repo 'wcf42' and all associated IIS hosted services such as 'WcfService42'

### DIFF
--- a/src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
@@ -1,0 +1,191 @@
+@echo off
+setlocal
+
+set Id=%1
+set IdMaster=Master
+set WcfTestDir=c:\WCFTest
+set RepoHome=c:\git
+set CurrentRepo=%RepoHome%\wcf%Id%
+set MasterRepo=%RepoHome%\wcf%IdMaster%
+set PRServiceName=PRService%Id%
+set WcfServiceName=WcfService%Id%
+set PRServiceMaster=PRService%IdMaster%
+set WcfServiceMaster=WcfService%IdMaster%
+set LogFile="%~dp0%~n0.log"
+set CertExpirationInDays=99
+set ExitCode=0
+set cmd=
+
+if '%1'=='' goto :Usage
+if '%1'=='/?' goto :Usage
+if '%1'=='-?' goto :Usage
+if '%1'=='/help' goto :Usage
+if '%1'=='-help' goto :Usage
+
+:: Make sure IIS appcmd.exe is accessible
+set appcmd=%windir%\System32\inetsrv\appcmd.exe
+if NOT EXIST %appcmd% (
+    echo Could not find: %appcmd%.
+    echo Please make sure IIS is installed.
+    goto :Failure
+)
+
+:: Clean up any existing setup
+echo Deleting WCF repo at %CurrentRepo% if exists and associated application pools and services hosted on IIS...
+if exist %CurrentRepo% rmdir /s /q %CurrentRepo%
+if exist %WcfTestDir% if /I '%MasterRepo%'=='%CurrentRepo%' rmdir /s /q %WcfTestDir%
+%appcmd% delete app "Default Web Site/%PRServiceName%" >nul
+%appcmd% delete apppool %PRServiceName% >nul
+%appcmd% delete app "Default Web Site/%WcfServiceName%" >nul
+%appcmd% delete apppool %WcfServiceName% >nul
+echo Clean up done.
+if /I '%2'=='/c' goto :Done
+
+:: Make sure git.exe is available
+where git.exe /Q
+if ERRORLEVEL 1 (
+    echo Could not find: git.exe.
+    echo Please make sure git is installed and its location is part of PATH environment.
+    goto :Failure
+)
+
+:: Clone a WCF repo
+set cmd=git clone https://github.com/dotnet/wcf %CurrentRepo%
+call %cmd%
+if ERRORLEVEL 1 goto :Failure
+pushd %CurrentRepo%
+call :Run git config --add origin.fetch "+refs/pull/*/head:refs/remotes/origin/pr/*"
+if ERRORLEVEL 1 goto :Failure
+popd
+
+:: Create a new application pool and an application for the WCF test service
+echo Create IIS application pool: %WcfServiceName%
+call :Run %appcmd% add apppool /name:%WcfServiceName% /processModel.IdentityType:LocalSystem /managedRuntimeVersion:v4.0 /managedPipelineMode:Integrated
+if ERRORLEVEL 1 goto :Failure
+echo Create IIS application: %WcfServiceName%
+call :Run %appcmd% add app /site.name:"Default Web Site" /path:/%WcfServiceName% /physicalPath:%CurrentRepo%\src\System.Private.ServiceModel\tools\IISHostedWcfService /applicationPool:%WcfServiceName% /enabledProtocols:"http,net.tcp"
+if ERRORLEVEL 1 goto :Failure
+
+:: Grant app pool %WcfServiceName% "Read and Execute" access to the IISHostedWcfService and its subdirectories
+echo Grant app pool %WcfServiceName% "Read and Execute" access to the IISHostedWcfService and its subdirectories
+call :Run icacls %CurrentRepo%\src\System.Private.ServiceModel\tools\IISHostedWcfService /grant:r "IIS APPPOOL\%WcfServiceName%":RX /T /Q
+if ERRORLEVEL 1 goto :Failure
+
+:: Setup PR service only if this is a master repo
+:: Uncomment line below when we switch to use one central PRService
+REM if /I NOT '%MasterRepo%'=='%CurrentRepo%' goto :SkipPRService
+
+:: Create a new application pool and an application for the PR service
+echo Create IIS applicaton pool: %PRServiceName%
+call :Run %appcmd% add apppool /name:%PRServiceName% /managedRuntimeVersion:v4.0 /managedPipelineMode:Integrated
+if ERRORLEVEL 1 goto :Failure
+echo Create IIS applicatoin: %PRServiceName%
+call :Run %appcmd% add app /site.name:"Default Web Site" /path:/%PRServiceName% /physicalPath:%CurrentRepo%\src\System.Private.ServiceModel\tools\PRService -applicationPool:%PRServiceName%
+if ERRORLEVEL 1 goto :Failure
+
+:: Grant app pool %PRServiceName% "Modify" access to %CurrentRepo% and its subdirectories
+echo Grant app pool %PRServiceName% "Modify" access to %CurrentRepo% and its subdirectories
+call :Run icacls %CurrentRepo% /grant:r "IIS APPPOOL\%PRServiceName%":M /T /Q
+if ERRORLEVEL 1 goto :Failure
+:: Uncomment below and delete above when we switch to central PR service
+REM echo Grant app pool %PRServiceName% "Read and Execute" access to the PRService and its subdirectories
+REM call :Run icacls %CurrentRepo%\src\System.Private.ServiceModel\tools\PRService /grant:r "IIS APPPOOL\%PRServiceName%":RX /T /Q
+REM if ERRORLEVEL 1 goto :Failure
+
+:SkipPRService
+
+:: Grant master PR service's app pool access to current repo if master PRService exists
+if /I '%MasterRepo%'=='%CurrentRepo%' goto :SkipGrantAccess
+if NOT EXIST %MasterRepo% goto :SkipGrantAccess
+%appcmd% list apppool %PRServiceMaster% >nul
+if ERRORLEVEL 1 goto :SkipGrantAccess
+echo Grant master PRServer app pool: %PRServiceMaster% "Modify" access to %CurrentRepo% and its sudirectories
+call :Run icacls %CurrentRepo% /grant:r "IIS APPPOOL\%PRServiceMaster%":M /T /Q
+if ERRORLEVEL 1 goto :Failure
+
+:SkipGrantAccess
+
+:: Install certificates if %WcfTestDir% does not exist; otherwise, assume certs are installed already
+if EXIST %WcfTestDir% goto :SkipCertInstall
+
+:: Make sure there is an HTTPS binding in IIS site "Default Web Site"
+%appcmd% list site "Default Web Site" /text:* | findstr /i bindings | findstr /i https >nul
+if ERRORLEVEL 1 (
+    echo Add an HTTPS binding to "Default Web Site"
+    call :Run %appcmd% set site "Default Web Site" /+bindings.[protocol='https',bindingInformation='*:443:']
+    if ERRORLEVEL 1 goto :Failure
+)
+
+:: Use current repo for cert gen if master repo does not exist
+if NOT EXIST %MasterRepo% (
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
+@echo off
+setlocal
+
+set Id=%1
```
roncain: Nit: I know setlocal is used, but I'd prefer if these environment variables were prefixed with underscore so it is clear when reading later areas of the script that you are referring only to local environment variables.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
...
+set IdMaster=Master
+set WcfTestDir=c:\WCFTest
+set RepoHome=c:\git
+set CurrentRepo=%RepoHome%\wcf%Id%
```
roncain: I'm a little surprised we get away with using % and not having EnableDelayedExpansion turned on.  I've often run into problems composing environment variables from other environment variables set in the same script.  Obviously it works for you -- I'm just observing I can't explain why.  Example, the other shared CoreFx scripts are moving to EnableDelayedExpansion and using ! instead of %.  Not blocking, just an observation.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
+@echo off
+setlocal
+
+set Id=%1
+set IdMaster=Master
+set WcfTestDir=c:\WCFTest
+set RepoHome=c:\git
```
roncain: Question: could we infer RepoHome from %_dp0?  Or is it the intent that this CMD will be moved out of the repo and executed somewhere else?

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
...
+    echo Could not find: %appcmd%.
+    echo Please make sure IIS is installed.
+    goto :Failure
+)
+
+:: Clean up any existing setup
+echo Deleting WCF repo at %CurrentRepo% if exists and associated application pools and services hosted on IIS...
+if exist %CurrentRepo% rmdir /s /q %CurrentRepo%
+if exist %WcfTestDir% if /I '%MasterRepo%'=='%CurrentRepo%' rmdir /s /q %WcfTestDir%
+%appcmd% delete app "Default Web Site/%PRServiceName%" >nul
```
roncain: Nit: consider echoing the commands and stdout to the log file to help understand what took place while this script executed.  Not blocking.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: It's latter. When you create the first repo on a machine, there is no repo that contains this script. You have to copy it from somewhere else. It's chicken and egg problem 😄.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: I know delayed expansion (by using !) is often necessary for loops. Everything works fine here.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: The app (and app pool) actually may or may not exist. It's a best effort clean up. A failed clean up does not mean much. But you will see failures when you create them later if there are any problems.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
...
+:: Make sure git.exe is available
+where git.exe /Q
+if ERRORLEVEL 1 (
+    echo Could not find: git.exe.
+    echo Please make sure git is installed and its location is part of PATH environment.
+    goto :Failure
+)
+
+:: Clone a WCF repo
+set cmd=git clone https://github.com/dotnet/wcf %CurrentRepo%
```
iamjasonp: NIT : would prefer repo path to be set at the top of the file somewhere for visibility. Not blocking

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
iamjasonp: Can you make a comment to that effect that it's best effort in the file itself?

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
...
+:: Clean up any existing setup
+echo Deleting WCF repo at %CurrentRepo% if exists and associated application pools and services hosted on IIS...
+if exist %CurrentRepo% rmdir /s /q %CurrentRepo%
+if exist %WcfTestDir% if /I '%MasterRepo%'=='%CurrentRepo%' rmdir /s /q %WcfTestDir%
+%appcmd% delete app "Default Web Site/%PRServiceName%" >nul
+%appcmd% delete apppool %PRServiceName% >nul
+%appcmd% delete app "Default Web Site/%WcfServiceName%" >nul
+%appcmd% delete apppool %WcfServiceName% >nul
+echo Clean up done.
+if /I '%2'=='/c' goto :Done
```
iamjasonp: NIT: I tend to like to do command parsing in one go and then set the appropriate variables in one place, and not rely on %1, %2... etc. later in the script. Helps with input checking and allows for passing args in any order as the script grows in scope / functionality

Not blocking

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
...
+    echo Please make sure git is installed and its location is part of PATH environment.
+    goto :Failure
+)
+
+:: Clone a WCF repo
+set cmd=git clone https://github.com/dotnet/wcf %CurrentRepo%
+call %cmd%
+if ERRORLEVEL 1 goto :Failure
+pushd %CurrentRepo%
+call :Run git config --add origin.fetch "+refs/pull/*/head:refs/remotes/origin/pr/*"
```
iamjasonp: Question only: why are you not using the `set cmd, call %cmd%` convention you're using three lines above?

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
...
+if NOT '%cmd%'=='' (
+    echo.
+    echo Failed to run:
+    echo %cmd%
+)
+if EXIST %LogFile% type %LogFile%
+
+:Done
+exit /b %ExitCode%
+endlocal
```
iamjasonp: NIT: newline

iamjasonp:
LGTM, just a few small questions/nits.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
roncain: Nit and not blocking: a failed clean *may* not mean much for the problems you expect.  The reason I recommend logging is for the failures you didn't.  I still think it would be a good idea for this script to not suppress stdout and let the person who ran it decide what was important.  Your call.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
roncain: +1.  I also still prefer to add an underscore to make it clear in other parts of the script you are using local env vars.  Not blocking but good practice.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
...
+
+:: Create a new application pool and an application for the PR service
+echo Create IIS applicaton pool: %PRServiceName%
+call :Run %appcmd% add apppool /name:%PRServiceName% /managedRuntimeVersion:v4.0 /managedPipelineMode:Integrated
+if ERRORLEVEL 1 goto :Failure
+echo Create IIS applicatoin: %PRServiceName%
+call :Run %appcmd% add app /site.name:"Default Web Site" /path:/%PRServiceName% /physicalPath:%CurrentRepo%\src\System.Private.ServiceModel\tools\PRService -applicationPool:%PRServiceName%
+if ERRORLEVEL 1 goto :Failure
+
+:: Grant app pool %PRServiceName% "Modify" access to %CurrentRepo% and its subdirectories
```
roncain: Why not full access?  If git syncs, we should anticipate its automated GC to delete files at unexpected times.  When we sync or clean, we should expect files to be both deleted and created.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
...
+    set WcfServiceMaster=%WcfServiceName%
+)
+
+echo Build CertificateGenerator tool
+call :Run %MasterRepo%\src\System.Private.ServiceModel\tools\scripts\BuildCertUtil.cmd
+if ERRORLEVEL 1 goto :Failure
+
+echo Run CertificateGenerator tool. This will take a little while...
+md %WcfTestDir%
+set certGen=%MasterRepo%\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe
```
roncain: Nit, not blocking -- like Jason I like to see env vars set at the top.  Easier to see and maintain.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
```diff
@@ -0,0 +1,191 @@
...
+:: TODO: Grant all existing app pools named WCFService# 'Read' access to %WcfTestDir%. This is not needed for now
+
+:SkipCertInstall
+
+:: Grant app pool %WcfServiceName% "Read" access to %WcfTestDir% and its subdirectories
+echo Grant app pool %WcfServiceName% "Read" access to %WcfTestDir% and its subdirectories
+call :Run icacls %WcfTestDir% /grant:r "IIS APPPOOL\WcfService%Id%":R /T /Q
+if ERRORLEVEL 1 goto :Failure
+
+:: Clean up log file if everything worked perfectly
```
roncain: Personally I'd rather be able to see the log file even on a successful run (but it doesn't need to be echoed like it is for failure).  Your call.

roncain:
LGTM, subject to answers to the questions about the ACL levels granted the repo

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: I can change it to full if you can show me GC needs full access. I basically only grant minimum required access for all parts and I _think_ M is all git needs.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: To be clear, I don't support logging by this script at all. I just use a file to temporarily save output of each command currently running and show it when it's a failure. Even if I don't delete it, you only see the output from last command.

Supporting logging involves fairly amount of work. I don't have plan to do it and I don't think we should spend time to do it.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: > Can you make a comment to that effect that it's best effort in the file itself?

I do have "Deleting WCF repo at %CurrentRepo% _if exists_..." 😄 

> I still think it would be a good idea for this script to not suppress stdout

I prefer a cleaner output in successful case. Do you really want to see big ERROR like below while nothing is wrong? I don't.
```
C:\Windows\System32\inetsrv>appcmd delete app "blah"
ERROR ( message:Cannot find APP object with identifier "blah". )
```

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: Sure, will get https://github.com/dotnet/wcf set at the beginning.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: >  I tend to like to do command parsing in one go 

I actually thought about this too because that's normally I like to do too. For the case, the code will be changed to something like below. Kind of feel it's a little over kill 😄 

```
if /I '%2'=='/c' set clean=true
... many lines later...
if /I '%clean%'=='true' goto :Done
```

> allows for passing args in any order as the script grows in scope / functionality

This will require more work. I told myself if we need to pass one more argument, I will do it ... and I will move the %2 too 😄.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: This command does not give us any interesting output in successful case and I will show details when anything goes wrong by doing `call :Run ...`. In three lines above, however, `git clone` shows very useful information in successful case: what's the folder it's cloning into, what the progress in percentage etc, so I decide to just let user see what's going on.

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
iamjasonp: It _might_ be necessary to grant full access to the .git directory, but this is speculation on my part. We can try running git gc --aggressive to see if it fails, but this might not have results in our repos as they are all relatively new.

zhenlan: Let's wait and see then.

iamjasonp: Agreed

src/System.Private.ServiceModel/tools/scripts/WcfTestServerSetup.cmd
zhenlan: Will do

zhenlan:
Merge this round. I will send another PR to support central PRService only (after it is validated in real) and address any feedback I have agreed here.

PR #1049, opened 2016-04-19
Fix StartWcfService.cmd to reference exes relative to itself

roncain:
Modified StartWcfService.cmd to locate its built executables
relative to itself rather than the current working directory.

Without this, running the script from any other folder will fail
to start the self-hosted service.

### DIFF
--- a/src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
@@ -24,10 +24,9 @@ REM Config Certs
 REM we need the direcotry to save the test.crl file. We are investigate a way to get rid of it
 REM
 md c:\wcftest
-pushd ..\..\..\..\bin\Wcf\tools\CertificateGenerator
-call CertificateGenerator.exe
+call %~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe %*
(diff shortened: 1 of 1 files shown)


hongdai:
LGTM.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -24,7 +24,7 @@ REM Config Certs
 REM we need the direcotry to save the test.crl file. We are investigate a way to get rid of it
 REM
 md c:\wcftest
-pushd ..\..\..\..\bin\Wcf\tools\CertificateGenerator
+pushd %~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator
 call CertificateGenerator.exe
```
zhenlan: Why do we bother pushd? it worked well for me if I just do
```
%~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe
```

hongdai: I thought it's the best practice to go to that directory before calling a script/tool. If it's not needed, we can remove the push/pop and thus reduce the lines:-)

roncain: Agreed -- updated to run directly from the folder.  Will merge shortly.

PR #1050, opened 2016-04-19
Add package baseline

ericstj:
Package baseline is now defined in individual repos instead of
buildtools.

Pick up CoreFx's baseline from the Microsoft.Private.PackageBaseline
package and add a WCF specific baseline.

I've primed MyGet with the required version of this package.

Needs: https://github.com/dotnet/buildtools/pull/646


ericstj:
Removed WIP tag, this is ready to be merged.  @roncain can you have a look?

Packaging.props
```diff
@@ -16,8 +16,18 @@
     <PackagePlatform Condition="'$(PackagePlatform)' == ''">$(Platform)</PackagePlatform>
```
weshaggard: orthogonal but we should also use CoreFxExpectedPrerelease for the Microsoft.NETCore.Platforms version above in this file.

roncain:

roncain:
OuterLoop failures in CI appear to be due to problems in our test infrastructure.  I verified manually I can run OuterLoops on Windows.

/cc @StephenBonikowsky for his input on the changes.

StephenBonikowsky:
This change means that Packaging.props diverges significantly from the version in corefx, is it not possible to have a unified version that does the right thing based on the repo?

weshaggard:
@StephenBonikowsky I just chatted with @ericstj about the differences here and he is going to work on factoring these as an extension so the core packaging.props file will remain the same across corefx and wcf. As a secondary change he is also going to clean up the runtime.json difference as well.

StephenBonikowsky:
That's great, thank you!

ericstj:
Refactored this to make the diff for packaging.props identical between the two branches.

Packaging.props
```diff
@@ -18,6 +18,8 @@
     <NativePackagePath>$(MSBuildThisFileDirectory)src/Native/pkg</NativePackagePath>
   </PropertyGroup>
 
+  <Import Project="pkg/baseline/baseline.props" />
```
weshaggard: Can we make this conditional on existence? Again I'm thinking of other repo's like CLI or CoreCLR that might not have a unique baselining requirement.

weshaggard:
Other then the one comment looks good to me.

dir.props
```diff
@@ -90,6 +87,9 @@
     <CoreFxExpectedPrerelease>rc3-24015-00</CoreFxExpectedPrerelease>
   </PropertyGroup>
 
+  <!-- Import packaging props -->
+  <Import Project="$(MSBuildThisFileDirectory)Packaging.props"/>
+
```
StephenBonikowsky: Does this import need to be dropped lower in the ordering, if so does it also need to be done in corefx?

ericstj: Yes since it will consume CoreFxExpectedPrerelease.  I can make that change in CoreFx as well.

StephenBonikowsky:
LGTM

dir.props
ericstj: Made this change.

ericstj:
@dotnet-bot test  Innerloop OSX Release Build and Test

roncain:
LGTM.  Doing one CI OuterLoop pass now that we have fixed our infrastructure.

roncain:

roncain:
LGTM. I'll merge if I don't hear any objections.  Any objections @StephenBonikowsky or @ericstj ?

ericstj:
We need to make sure the corefx change mereges as well, otherwise this will break OTG when it mirrors.

roncain:
Thanks @ericstj .  Do you recommend that whoever merges the CoreFx changes would merge this one instantly after that?  I'm okay with anyone doing the merge when they think the timing is right.

ericstj:
Yes, let's merge at the same time as CoreFx.  I can push the button.

PR #1038, opened 2016-04-19
Test CI only

iamjasonp:

PR #1054, opened 2016-04-20
Consolidate various PRServices into one endpoint

iamjasonp:
Rather than have multiple PRServices (one for each repo) on the PR service machine, centralize them into one endpoint, and instead pass the repo id. This has the benefit of allowing only one copy of the service to be maintained, having fewer points of entry, having one consistent service across all repos, and finally, preventing accidental syncs from killing the service and requiring manual intervention

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -62,7 +62,7 @@ def branch = GithubBranchName
 // **************************
 class WcfUtilities
 {
-    def wcfRepoSyncServiceCount = 0 
+    def wcfPRServiceCount = 0 
     
     // Outerloop jobs for WCF Core require an external server reference
     // This should be run 
@@ -73,27 +73,28 @@ class WcfUtilities
             return 
         }
 
-        wcfRepoSyncServiceCount++
+        wcfPRServiceCount++
 
         def operation = isPR ? "pr" : "branch"
 
         job.with { 
             parameters {
-                stringParam('WcfServiceUri', "wcfcoresrv2.cloudapp.net/WcfService${wcfRepoSyncServiceCount}", 'Wcf OuterLoop Test Service Uri')
-                stringParam('WcfRepoSyncServiceUri', "http://wcfcoresrv2.cloudapp.net/PRService${wcfRepoSyncServiceCount}/pr.ashx", 'Wcf OuterLoop Test PR Service Uri')
+                stringParam('WcfServiceUri', "wcfcoresrv2.cloudapp.net/WcfService${wcfPRServiceCount}", 'Wcf OuterLoop Test Service Uri')
+                stringParam('WcfPRServiceUri', "http://wcfcoresrv2.cloudapp.net/PRServiceMaster/pr.ashx", 'Wcf OuterLoop Test PR Service Uri')
+                stringParam('WcfPRServiceId', "${wcfPRServiceCount}", 'Wcf OuterLoop Test PR Service Id')
             }
         }
         if (os.toLowerCase().contains("windows")) {
             job.with { 
                 steps {
-                    batchFile(".\\src\\System.Private.ServiceModel\\tools\\scripts\\sync-pr.cmd ${operation} %WcfRepoSyncServiceUri%")
+                    batchFile(".\\src\\System.Private.ServiceModel\\tools\\scripts\\sync-pr.cmd %WcfPRServiceId% ${operation} %WcfPRServiceUri%")
                 }           
             }
         } 
         else {
             job.with { 
                 steps {
-                   shell("HOME=\$WORKSPACE/tempHome ./src/System.Private.ServiceModel/tools/scripts/sync-pr.sh ${operation} \$WcfRepoSyncServiceUri")
+                   shell("HOME=\$WORKSPACE/tempHome ./src/System.Private.ServiceModel/tools/scripts/sync-pr.sh \$WcfPRServiceId ${operation} \$WcfPRServiceUri")
                 }
             }
         }
--- a/src/System.Private.ServiceModel/tools/PRService/pr.ashx
+++ b/src/System.Private.ServiceModel/tools/PRService/pr.ashx
@@ -19,92 +19,134 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Specialized;
+using System.Configuration;
 using System.Diagnostics;
-using System.Web;
 using System.IO;
 using System.Text;
+using System.Web;
 
 public class PullRequestHandler : IHttpHandler
 {
-    private const string _gitExecutablePath = @"C:\Program Files\Git\cmd\git.exe";
-    private static string _gitRepoPath;
-    private const int _gitExecutionTimeoutMilliseconds = 100000;
+    private string _gitExecutablePath;
+    private string _gitRepositoriesBasePath;
+    private TimeSpan _gitExecutionTimeout;
+
+    private bool _configurationRead = false;
+
+    private const string _gitExecutablePathConfigurationKeyString = "GitExecutablePath";
+    private const string _gitRepositoriesBasePathConfigurationKeyString = "GitRepositoriesBasePath";
+    private const string _gitExecutionTimeoutConfigurationKeyString = "GitExecutionTimeout";
+
+    private const string _gitRepoPathTemplate = @"{0}\wcf{1}";
 
     public void ProcessRequest(HttpContext context)
     {
-        // Check for prerequisites needed before this script can execute
-
-        _gitRepoPath = context.Request.PhysicalApplicationPath + "..\\..\\..\\..";
-
         StringBuilder result = new StringBuilder();
-        bool success = CheckPrerequisites(result);
 
+        // Read configuration
+        bool success = ReadConfiguration(result);
         if (!success)
         {
-            // CheckPrerequisites failed, this is a server issue
+            // ReadConfiguration failed, this is a server issue 
             context.Response.StatusCode = 500;
-            context.Response.Write(HttpUtility.HtmlEncode(result));
+            context.Response.Write(HttpUtility.HtmlEncode(result.ToString()));
             return;
         }
 
-        success = CleanupMergedBranches(result);
-
-        if (!success)
-        {
-            // CleanupMergedBranches failed, this is a server issue
-            context.Response.StatusCode = 500;
-            context.Response.Write(HttpUtility.HtmlEncode(result));
-            return;
-        }
-
-        // Parameter verification
+        // Check inputs 
+        // We expect one of the following to be specified: 
+        // - A numerical repo ID
+        // - A branch name OR a PR number (not both)
+        //
+        // We will use Git to validate and verify that the branch exists if specified
 
         success = false;
 
+        string repoIdString = context.Request.QueryString["id"];
         string branchString = context.Request.QueryString["branch"];
         string prString = context.Request.QueryString["pr"];
+
+        uint repoId;
+        if (string.IsNullOrWhiteSpace(repoIdString) || !uint.TryParse(repoIdString, out repoId))
+        {
+            context.Response.StatusCode = 400;
+            context.Response.Write(string.Format("The client ID specified, '{0}', is invalid. Please specify a valid 'id' in the request query string <br/>", repoIdString ?? "unspecified"));
+            return;
+        }
 
         if (string.IsNullOrWhiteSpace(prString) && string.IsNullOrWhiteSpace(branchString))
         {
             context.Response.StatusCode = 400;
-            context.Response.Write("No PR or branch specified");
+            context.Response.Write("No PR or branch specified. Specify either a 'pr' or 'branch' in the request query string <br/>");
             return;
         }
 
         if (!string.IsNullOrWhiteSpace(prString) && !string.IsNullOrWhiteSpace(branchString))
         {
             context.Response.StatusCode = 400;
-            context.Response.Write("PR and branch specified; please specify only one of them to sync to <br/>");
+            context.Response.Write("Both 'pr' and 'branch' specified; please specify only one of them to sync to <br/>");
             return;
         }
 
-        context.Response.Write(string.Format("PR = {0}<br/>", HttpUtility.HtmlEncode(prString) ?? "Unspecified"));
-        context.Response.Write(string.Format("branch = {0}<br/>", HttpUtility.HtmlEncode(branchString) ?? "Unspecified"));
+        context.Response.Write(string.Format("pr = {0} <br/>", HttpUtility.HtmlEncode(prString) ?? "unspecified"));
+        context.Response.Write(string.Format("branch = {0} <br/>", HttpUtility.HtmlEncode(branchString) ?? "unspecified"));
+
+        // Check for prerequisites needed before this script can execute
+        string gitRepoPath = GetRepoPath(repoId);
+        success = CheckPrerequisites(gitRepoPath, result);
+
+        if (!success)
+        {
+            // CheckPrerequisites failed, this is a server issue
+            context.Response.StatusCode = 500;
+            context.Response.Write(HttpUtility.HtmlEncode(result.ToString()));
+            return;
+        }
+
+        // Cleanup all branches except master from the repo - this shouldn't fail, and if it does, there's something wrong with the repo
+        success = CleanupBranches(gitRepoPath, result);
+
+        if (!success)
+        {
+            // CleanupMergedBranches failed, this is a server issue
+            context.Response.StatusCode = 500;
+            context.Response.Write(HttpUtility.HtmlEncode(result.ToString()));
+            return;
+        }
 
         // Do stuff
+        success = false;
 
         uint pr;
-        if (!string.IsNullOrEmpty(prString) && uint.TryParse(prString, out pr))
+        // At this point we are guaranteed to have either a 'pr' or 'branch' but not both
+        if (!string.IsNullOrEmpty(prString))
         {
-            success = SyncToPr(pr, result);
-        }
-        else if (!string.IsNullOrEmpty(branchString))
-        {
-            // We only want to support limited branches as we only use this for PR purposes
-            // This should be opened up later once we decide we need to (e.g., for testing release branches), but for now, 
-            // whitelist what we need
-            // for example, we could use the output from "git branch -r" to figure out what branches are valid
-
-            switch (branchString)
+            if (uint.TryParse(prString, out pr))
             {
-                case "master":
-                case "origin/master":
-                case "origin/infrastructure":
-                    success = SyncToBranch(branchString, result);
-                    break;
-                default:
-                    context.Response.Write(string.Format("Branch name specified, '{0}' is not supported by this service<br/>", HttpUtility.HtmlEncode(branchString)));
-                    break;
+                success = SyncToPr(pr, gitRepoPath, result);
+            }
+        }
+        else
+        {
+            // Validate input here, because we don't want someone passing in an arbitrary branch
+            // Obtain a list of valid remote branches and allow sync to one of these only
+
+            string[] branches;
+            success = GetBranches(gitRepoPath, true, out branches, result);
+
+            if (success)
+            {
+                success = false;
+                for (int i = 0; i < branches.Length; i++)
+                {
+                    // git branches are case sensitive 
+                    if (string.Equals(branchString, branches[i]))
+                    {
+                        success = SyncToBranch(branchString, gitRepoPath, result);
+                        break;
+                    }
+                }
             }
         }
 
@@ -113,14 +155,14 @@ public class PullRequestHandler : IHttpHandler
             // it's most likely that the client supplied a bad PR ID, but there could potentially 
             // be other errors in SyncToPr or SyncToBranch. For now, return 400 in this path
             context.Response.StatusCode = 400;
-            context.Response.Write(string.Format("Invalid PR ID, '{0}' or branch, '{1}' specified <br/>", HttpUtility.HtmlEncode(prString), HttpUtility.HtmlEncode(branchString)));
+            context.Response.Write(string.Format("Invalid 'pr', '{0}' or 'branch', '{1}' specified. Please specify a valid 'pr' or 'branch' <br/>", HttpUtility.HtmlEncode(prString), HttpUtility.HtmlEncode(branchString)));
         }
 
-        context.Response.Write(HttpUtility.HtmlEncode(result));
+        context.Response.Write(HttpUtility.HtmlEncode(result.ToString()));
     }
 
     // Check prerequisites for whether or not we can run
-    private bool CheckPrerequisites(StringBuilder executionResult)
+    private bool CheckPrerequisites(string gitRepoPath, StringBuilder executionResult)
     {
         if (!File.Exists(_gitExecutablePath))
         {
@@ -128,9 +170,9 @@ public class PullRequestHandler : IHttpHandler
             return false;
         }
 
-        if (!Directory.Exists(_gitRepoPath))
+        if (!Directory.Exists(gitRepoPath))
         {
-            executionResult.AppendFormat("Could not find or insufficient permissions to access the git repo at path '{0}'", _gitRepoPath);
+            executionResult.AppendFormat("Could not find or insufficient permissions to access the git repo at path '{0}'", gitRepoPath);
             return false;
         }
 
@@ -144,19 +186,24 @@ public class PullRequestHandler : IHttpHandler
                 "status"
             };
 
-            success = RunGitCommands(gitCommands, executionResult);
+            success = RunGitCommands(gitCommands, gitRepoPath, executionResult);
         }
         catch (Exception ex)
         {
             executionResult.AppendFormat("Exception caught while attempting to start '{0}'", _gitExecutablePath);
             executionResult.Append(ex.ToString());
-            return success;
         }
 
         return success;
     }
 
-    private bool SyncToPr(uint pr, StringBuilder executionResult)
+    private string GetRepoPath(uint repoId)
+    {
+        // Note: assumes configuration is read. 
+        return string.Format(_gitRepoPathTemplate, _gitRepositoriesBasePath, repoId);
+    }
+
+    private bool SyncToPr(uint pr, string gitRepoPath, StringBuilder executionResult)
     {
         string[] gitCommands = new string[]
         {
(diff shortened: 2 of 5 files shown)


src/System.Private.ServiceModel/tools/scripts/sync-pr.cmd
```diff
@@ -60,15 +66,17 @@ if "%3"=="" (
         )
     )
 ) else (
-    set __BRANCH_OR_PR_ID=%3
+    set __BRANCH_OR_PR_ID=%4
     echo   [%~n0] WARNING: This script should usually only be called only from the context of a GitHub Pull Request
-    echo   [%~n0] Branch or PR ID overridden: '%3'
+    echo   [%~n0] Branch or PR ID overridden: '%4'
```
iamjasonp: There's some expansion funniness, hence why this has to stay as %4 and not _BRANCH_OR_PR_ID. Basically the variable is only defined after exiting the if.

src/System.Private.ServiceModel/tools/scripts/sync-pr.cmd
```diff
@@ -60,15 +66,17 @@ if "%3"=="" (
...
 
-set __REQUEST_URL=%__SYNC_HOST_URL%?%__OPERATION_MODE%=%__BRANCH_OR_PR_ID%
+REM Disregard the chevrons and quotataions in the next line - it's to output the correct ampersand character to PowerShell
+REM We have to dupliate the request URL below for output and for PowerShell because of Powershell's crazy escape sequences interacting 
+REM badly with cmd's crazy and conflicting escape sequences
 
-echo   [%~n0] Making call to '%__REQUEST_URL%'
-powershell -NoProfile -ExecutionPolicy unrestricted -Command "(New-Object Net.WebClient).DownloadString('%__REQUEST_URL%'); 
+echo   [%~n0] Making call to %__SYNC_HOST_URL%?repo=%__REPO_ID%^&%__OPERATION_MODE%=%__BRANCH_OR_PR_ID%
+powershell -NoProfile -ExecutionPolicy unrestricted -Command "(New-Object Net.WebClient).DownloadString('%__SYNC_HOST_URL%?repo=%__REPO_ID%&%__OPERATION_MODE%=%__BRANCH_OR_PR_ID%');"
```
iamjasonp: Reviewers: This was super frustrating to figure out, but I think I got it. If anyone else has a better idea on how to do this, I would be much obliged.

src/System.Private.ServiceModel/tools/scripts/sync-pr.cmd
```diff
@@ -15,31 +15,37 @@ echo.
...
     echo   sync-url    - URL on remote server for PR synchronization
     echo   branch-name - branch name to sync to
     echo   pr-number   - PR number to sync to
     echo.
-    echo   Example:  %~n0 branch http://wcfci-sync-server/PRService1/pr.ashx 
-    echo   Example:  %~n0 branch http://wcfci-sync-server/PRService1/pr.ashx master
-    echo   Example:  %~n0 pr http://wcfci-sync-server/PRService1/pr.ashx 
-    echo   Example:  %~n0 pr http://wcfci-sync-server/PRService1/pr.ashx 404
+    echo   If the branch-name or pr-number are left blank, then the script will use 
+    echo   the GIT_BRANCH or ghprbPullId environment variable depending on the operation mode
```
iamjasonp: Reviewers: This just keeps comments in sync with sync-pr.sh

iamjasonp:
Aside - I can't wait for bash to be available in Windows and that version of Windows deployed on these machines :smile:

iamjasonp:
@zhenlan - this is ready to go and service tested where the repoID is 1. However you will need to modify permissions to allow RW to all repos by the PRServiceMaster user - currently it doesn't seem like the access rights work correctly for all repos

The repos that need modifications show Git complaining of the following: 
`fatal: unable to write new index file Git executable  'C:\Program Files\Git\cmd\git.exe' exited with code '128'`, Repo 10 is showing this behaviour

src/System.Private.ServiceModel/tools/scripts/sync-pr.cmd
roncain: Question only: could you clarify what you wanted ideas on improving? I'm not sure what the "this" was in the question. Don't block on it.

iamjasonp: "this" refers to the REQUEST_URL. The escaping sequences when combining cmd and PS are nightmarish when you need to escape the & symbol...

iamjasonp: Basically - how to simplify it such that I only need one %REQUEST_URI% variable or otherwise avoid repeating the URI

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -273,6 +349,62 @@ public class PullRequestHandler : IHttpHandler
         return success;
     }
 
+    // Read configuration
+    // Precedence is: 
+    // 1. Environment variable
+    // 2. AppSettings in web.config
+    // 
+    private bool ReadConfiguration(StringBuilder result)
```
roncain: Nit: not blocking, but could add comment saying what 'bool' return and StringBuilder parameter are used for.  I interpret 'result' to be the error result whenver it fails.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -19,92 +19,135 @@
...
+            context.Response.StatusCode = 400;
+            context.Response.Write(string.Format("The repo ID specified, '{0}', is invalid <br/>", repoString ?? "unspecified"));
+            return;
+        }
+
         if (string.IsNullOrWhiteSpace(prString) && string.IsNullOrWhiteSpace(branchString))
         {
             context.Response.StatusCode = 400;
-            context.Response.Write("No PR or branch specified");
+            context.Response.Write("No PR or branch specified. Specify either a PR or branch <br/>");
```
roncain: Nit: could expand error message to tell them how to do this, such as name of query parameter.  Not blocking or required.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -19,92 +19,135 @@
...
-                    success = SyncToBranch(branchString, result);
-                    break;
-                default:
+                success = false;
+                for (int i = 0; i < branches.Length; i++)
+                {
+                    // git branches are case sensitive 
+                    if (string.Equals(branchString, branches[i]))
+                    {
+                        success = SyncToBranch(branchString, gitRepoPath, result);
```
roncain: Failure here plus the break will lead people to think the branch is unsupported.  It would be better to write 'result' into the response and to use a different error message than the one below.

roncain:
LGTM subject to nits (your call) and coordinating the merge with the changes in PR #1036.

netci.groovy
```diff
@@ -80,20 +80,20 @@ class WcfUtilities
         job.with { 
             parameters {
                 stringParam('WcfServiceUri', "wcfcoresrv2.cloudapp.net/WcfService${wcfRepoSyncServiceCount}", 'Wcf OuterLoop Test Service Uri')
-                stringParam('WcfRepoSyncServiceUri', "http://wcfcoresrv2.cloudapp.net/PRService${wcfRepoSyncServiceCount}/pr.ashx", 'Wcf OuterLoop Test PR Service Uri')
+                stringParam('WcfRepoSyncServiceUri', "http://wcfcoresrv2.cloudapp.net/PRService/pr.ashx", 'Wcf OuterLoop Test PR Service Uri')
```
zhenlan: It's `http://wcfcoresrv2.cloudapp.net/PRServiceMaster/pr.ashx`

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -19,92 +19,135 @@
...
+        // Check inputs 
+        // We expect one of the following to be specified: 
+        // - A numerical repo ID
+        // - A branch name OR a PR number (not both)
+        //
+        // We will use Git to validate and verify that the branch exists if specified
 
         success = false;
 
+        string repoString = context.Request.QueryString["repo"];
```
zhenlan: The parameter `repo` is the API surface of PRService. In the perspective of PRService, it asks callers to identify themselves so the PRService can differentiate. Although we use different repo's behind the scene, we should not expose it on the API surface. Can you change this to `id` please?

zhenlan:
> @zhenlan - this is ready to go and service tested where the repoID is 1. However you will need to modify permissions to allow RW to all repos by the PRServiceMaster user - currently it doesn't seem like the access rights work correctly for all repos

The CI server is already set up to work in both modes (individual and central PRService). PRServiceMaster app pool has M access to all other repos. The error you saw is caused by another issue that caused the index file lost the permission. I've made a fix and updated all repos on CI server. Please give it a try. Once this PR is in, I will update CI server to stop supporting individual PRService.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -19,92 +19,135 @@
...
+    {
+        StringBuilder result = new StringBuilder();
 
+        // Read configuration
+        bool success = ReadConfiguration(result);
         if (!success)
         {
-            // CleanupMergedBranches failed, this is a server issue
+            // ReadConfiguration failed, this is a server issue 
             context.Response.StatusCode = 500;
```
zhenlan: I just throw ideas here, you decide if you want to take it or not...

After playing with PRService in the past a few days, many times I found it was hard to tell a 500 or a 400 are from IIS, ASP.NET or us. I wish we could throw a unique 500/400 number so I could tell more easily who failed the request.

netci.groovy
iamjasonp: Thanks! good catch :)

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: I like the idea. I know that IIS emits 400.x to differentiate different issues. We could probably do 400.xx (maybe .20, to give us some distance from what IIS emits). Will play with this.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: I was struggling with this myself, 'id' seemed a bit vague ("id of what?" from a caller perspective), and 'repo' seemed a bit specific (exposes the underlying implementation). I was toying around with 'clientid' which is really what this is about (we're identifying the client who will be calling the service, effectively) 

I'll use 'id' as you recommended unless you have further thoughts on this.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -19,92 +19,135 @@
...
         // Do stuff
+        success = false;
 
         uint pr;
         if (!string.IsNullOrEmpty(prString) && uint.TryParse(prString, out pr))
         {
-            success = SyncToPr(pr, result);
+            success = SyncToPr(pr, gitRepoPath, result);
         }
         else if (!string.IsNullOrEmpty(branchString))
```
zhenlan: Since you have already validated all input parameters, you can change code to below to be simpler. Less code, happier developer 😄 
```
if (!string.IsNullOrEmpty(prString) {
    if (uint.TryParse(prString, out pr))  {
        success = syncToPr(pr, gitRepoPath, result);
    }
}
else {
 ...
}
```

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -19,92 +19,135 @@
...
-                case "origin/master":
-                case "origin/infrastructure":
-                    success = SyncToBranch(branchString, result);
-                    break;
-                default:
+                success = false;
+                for (int i = 0; i < branches.Length; i++)
+                {
+                    // git branches are case sensitive 
+                    if (string.Equals(branchString, branches[i]))
```
zhenlan: Meant to be case sensitive?

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: Good call. The correct error message would be something like 'branch not found in the repo'. I'll play with the message a bit. 

'result' gets output later on in the !success block

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: Yep. GIt branches operate in a case-sensitive manner - e.g., try "git checkout StressTests" vs "git checkout stresstests"

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -156,7 +199,13 @@ public class PullRequestHandler : IHttpHandler
         return success;
```
zhenlan: The `return success` in `catch {...}` above can be removed (GitHub does not allow me to put comment there).

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: The output on L159 is probably the right message. L147's error message isn't needed

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -156,7 +199,13 @@ public class PullRequestHandler : IHttpHandler
...
     }
 
-    private bool SyncToPr(uint pr, StringBuilder executionResult)
+    private string GetRepoPath(uint repoId)
```
zhenlan: Although we use numbers on CI server as Ids, please don't assume id can only be unsigned number. It's perfectly fine the id is "Tom" or "Mary". We actually already have a special Id that is "Master". Please allow it to be just a string. It will only be valid if you can find "Tom" or Mary" on disk (and you have permission to it); otherwise, it is invalid anyway even if I give you a number.

iamjasonp: If we allow an arbitrary string, we will have to start parsing it for special characters such as .. / \ > < (etc) to prevent injections and whatnot. While I _think_ the current code is hardened against such attacks, I'd rather limit inputs to as narrow a subset as possible to shut the door as much as possible 

Our current design calls for numeric 'id' s, so I would rather leave this to be a uint. Let us solve that problem when we make an explicit design decision to open this up to have non-numeric ids.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: Done and slightly happier :smile:

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: ack

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: Dang it, just tried the substatus codes, and they don't get surfaced.

iamjasonp: I would have preferred to use 400.xxx, but unfortunately, this sub status code doesn't seem to be surfaced. 

So what we can do is to the StatusDescription to surface up a better message. Instead of just "bad request" we could reply with "bad request - PRService <blahblahblah>" if we need to. 

This is probably best done as a separate small discussion - changing the error codes and descriptions is trivial later on. #1059 created to track this work. I will leave the status code return values as-is for now and take #1059 as an improvement to the PRService.

src/System.Private.ServiceModel/tools/PRService/web.config
```diff
@@ -8,7 +8,21 @@
...
+    <!--
+      These appSettings provide the default environment under which we execute
+      These should be changed to reflect what is correct on the local system, or 
+      specified as a system-wide environment variable. 
+      
+      Environment variables take precedence over what is specified here
+    -->
+    <add key="GitExecutablePath" value="C:\Program Files\Git\cmd\git.exe"/>
+    <add key="GitRepositoriesBasePath" value="C:\git"/>
+    <add key="GitExecutionTimeout" value="00:01:00"/>
```
iamjasonp: Reviewers: i changed the total execution timeout to be 300s, but each git command should only take 60s or less (we aren't building anything). 

We want the timeout to happen on the command rather than relying on the overall ASP.NET timeout because if we hit the ASP.NET timeout, all output is via context.Response will be suppressed. The way the code is written, a git command timing out will spew out an error message and tell you which git command caused the problem

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -182,31 +231,58 @@ public class PullRequestHandler : IHttpHandler
...
+        else
+        {
+            success = RunGitCommands(new string[] { "branch --list" }, gitRepoPath, branchesString);
+        }
+
+        executionResult.Append(branchesString);
+
+        if (success)
+        {
+            branches = branchesString.ToString().Split(new char[] { '\n', '\r', ' ', '*' }, StringSplitOptions.RemoveEmptyEntries);
```
zhenlan: Question: do you expect branch=**origin/**master now from querystring?

iamjasonp: Yep. This is passed as the branch name when CI launches

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -273,6 +349,62 @@ public class PullRequestHandler : IHttpHandler
...
+    {
+        if (_configurationRead)
+        {
+            // No locking needed, reading this in twice, even on multiple concurrent threads, doesn't do harm
+            return true;
+        }
+
+        NameValueCollection appSettings = ConfigurationManager.AppSettings;
+
+        string gitExecutablePath = Environment.GetEnvironmentVariable(_gitExecutablePathConfigurationKeyString);
```
zhenlan: More code more chances for failure. I really want to cut PRService code to bare minimum :)

Will reading environment variable cause more permission to be needed by the app pool?

iamjasonp: Doesn't seem like it. I deployed the code to PRServiceMaster already (not committed) and this code seems to work fine.

src/System.Private.ServiceModel/tools/scripts/sync-pr.sh
```diff
@@ -67,26 +68,33 @@ show_banner
...
+    if [ "$2" == "branch" ] || [ "$2" == "pr" ]; then 
+        __operation_mode=$2
     else 
         show_usage
         exit 1
     fi 
 fi 
 
-if [ -z "$2" ]; then 
+if [ -z "$3" ]; then
```
zhenlan: If you do this at the beginning, will it save you most code to check $1 and $2 and show usage?

src/System.Private.ServiceModel/tools/scripts/sync-pr.cmd
zhenlan: I think we can tolerate this amount of duplication :)

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: Sounds good!

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: Eventually, we have to rely on ACL to enforce the security, which is what we already did, but it's not a bad idea to have some sanity validation like rejecting anything contains .., / and \\.

I do have a user scenario related to private Helix run. I want `hr.ashx?id=Tom&pr=123` to work when Tom try to submit a private Helix run. This change is blocking my scenario 😄.

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: So do I need to do the same from browser? When should I use origin/master vs. upstream/master? Trying to same some typing... :)

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: Just make sure I understand... you are saying this works in current setup/permission level?

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: Agreed. Let's do that in a separate (small) discussion.

src/System.Private.ServiceModel/tools/scripts/sync-pr.sh
iamjasonp: Changed to do one check :)

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: It works today

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: "origin" is done by convention - it's the immediate repo that your local repo is linked to. 
"upstream" is a convention that specifies that your immediate repo was forked from somewhere else. 

In this case, origin refers to https://github.com/dotnet/wcf, so you will always be referring to origin. For the purposes of PRService and how it is set up, you will never set up "upstream"

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: I share the same feeling.. couldn't think of anything better so ended up just 'id' in #1036 too :)

src/System.Private.ServiceModel/tools/PRService/pr.ashx
iamjasonp: Why would you create a private repo in this scenario? The implication of doing this is that someone will be making direct edits to the PRService server to do a private run. 

IMO, one is much better off creating a dev branch and asking the PR service to key off id={somenumber}&branch=origin/dev/tom. It is much more friction-free than having one of our devs log into the box and creating a repo.

iamjasonp: I'm willing to take named repos (as opposed to numerical) as an item to discuss for later, which probably will tie into a discussion on how we get Helix to work for private runs. 

Given the discussions we've had and scenarios offered have always stated clearly that we will use numbered repos only, I stand by my decision to implement as shown here. 

Unless there's something really wrong with how this is implemented for numbered repos, I propose leaving this discussion for later and taking this as an item to discuss.

iamjasonp:
@dotnet-bot test ci please

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: Okay then... as long as it does not ask for more permission :)

src/System.Private.ServiceModel/tools/PRService/pr.ashx
zhenlan: Yes, let's discuss that separately.

zhenlan:
LGTM

iamjasonp:
Thanks @zhenlan. 

Merging shortly - still spot checking a couple more things, but otherwise will merge and then test the system

iamjasonp:
@dotnet-bot test ci please

netci.groovy
```diff
@@ -73,27 +73,28 @@ class WcfUtilities
...
-                    batchFile(".\\src\\System.Private.ServiceModel\\tools\\scripts\\sync-pr.cmd ${operation} %WcfRepoSyncServiceUri%")
+                    batchFile(".\\src\\System.Private.ServiceModel\\tools\\scripts\\sync-pr.cmd %WcfPRServiceId% ${operation} %WcfPRServiceUri%")
                 }           
             }
         } 
         else {
             job.with { 
                 steps {
-                   shell("HOME=\$WORKSPACE/tempHome ./src/System.Private.ServiceModel/tools/scripts/sync-pr.sh ${operation} \$WcfRepoSyncServiceUri")
+                   shell("HOME=\$WORKSPACE/tempHome ./src/System.Private.ServiceModel/tools/scripts/sync-pr.sh \$WcfPRServiceId ${operation} \$WcfPRServiceUri")
```
iamjasonp: @zhenlan - post your LGTM, I made a change here to make WcfPRServiceId a parameter to prevent magic numbers from showing up in the command lines, which is why you're seeing a lot of test ci please.

iamjasonp:
There was one additional change I made to netci.groovy to better expose the WcfPRServiceId. It's minor and pointed out here: https://github.com/dotnet/wcf/pull/1054/files#r60546497

@dotnet-bot test ci please

netci.groovy
```diff
@@ -73,27 +73,28 @@ class WcfUtilities
             return 
         }
 
-        wcfRepoSyncServiceCount++
+        wcfPRServiceCount++
```
zhenlan: I kind of have the question for a while but never asked so I just ask explicitly so I don't worry anymore :). With this approach, do we guarantee each CI job is always assigned the same id?

iamjasonp: So that's a little complicated: 

* For each run of the CI build queue generator, each queue is guaranteed to have exactly one number assigned to it. It will not vary for the lifetime of this CI build queue... 
* When the generator is run again, *these queue numbers may change* 
* However, because the ordering of the queue generation is deterministic, the generated number will be the same for a given queue unless there's a change in, say, the supported branch/OS/buildconfiguration list

Does this explanation help? (and alleviate your fears :smile:)

iamjasonp: So in short, over the course of a typical day/week where no netci.groovy changes are being made, the wcfPRServiceId will remain the same.

zhenlan: Thanks @iamjasonp for the explanation. This is very helpful to understand what I can expect!

iamjasonp:
Spot checks done - but I now need to merge to make sure it works for real. Wish me luck :smile:

iamjasonp:
FYI - test occurring at #1038

PR #1062, opened 2016-04-21
Re-enable temporarily disabled WebSockets tests

### LINKED ISSUE #1043: Fix WebSockets tests to not use Bridge test helper
Many of the WebSockets tests now fail under the new test infrastructure because they are checking whether the WCF services are running locally using (to be) deprecated Bridge helper methods.

To resolve this issue:
- Find all [ActiveIssue]'s that refer to this issue number
- Replace the call to ScenarioTestHelpers.BridgeIsLocalHost() with a call to a new helper that does this test against the new test infrastructure (uses ServiceUri rather than BridgeHost).
- Remove the [ActiveIssues] and retest with the new test infrastructure

roncain:
Moving to the new test infrastructure made these tests
fail due to using the BridgeHost property to determine
if they were running as localhost. We disabled them
temporarily during infrastructure integration.

The fix is to use the ServiceUri instead of BridgeHost.

Fixes #1043

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/ScenarioTestHelpers.cs
@@ -31,10 +31,13 @@ public static class ScenarioTestHelpers
         }
     }
 
-    public static bool BridgeIsLocalHost()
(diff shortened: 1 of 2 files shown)


roncain:

iamjasonp:
LGTM

PR #1067, opened 2016-04-21
Move setup and cleanup steps to msbuild.

StephenBonikowsky:
* Where before we did setup and cleanup steps directly in build.cmd now adding a generic "build.override.targets" hook to build.proj from which we can call any custom scripts to do whatever we want.
* Including example code to show how to create a setup and cleanup Target.
* Minor change to build.cmd for corefx diff reason.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -104,7 +104,6 @@ set _binclashLoggerDll=%~dp0Tools\net45\Microsoft.DotNet.Build.Tasks.dll
 set _binclashlog=%~dp0binclash.log
 set _buildprefix=echo
 set _buildpostfix=^> "%_buildlog%"
-
 call :build %__args%
 
 :: Build
--- a/build.proj
+++ b/build.proj
@@ -114,4 +114,7 @@
     <RemoveDir Directories="$(BinDir)" />
   </Target>
 
+  <!-- Hook that can be used to insert custom build tasks to the build process such as setup and/or cleanup tasks -->
+  <Import Project="build.override.targets" Condition="Exists('build.override.targets')" />
+
 </Project>


StephenBonikowsky:
After talking to @zhenlan we realized that we may need to add a condition to the import of this file in the build.proj based on being on a Windows machine, I'll amend with this change before final merge.

StephenBonikowsky:
To be clear the content of build.override.targets is just to help @hongdai know how to apply the scripts she has been working on, her PR should replace it with real content.

build.override.targets
```diff
@@ -0,0 +1,19 @@
...
+    <ReadLinesFromFile File="$(MyLogFile)" >
+      <Output TaskParameter="Lines" PropertyName="ReturnCode"/>
+    </ReadLinesFromFile>
+    <!-- This PropertyGroup may not be needed for the WcfCleanup Target to access ReturnCode -->
+    <PropertyGroup>
+      <SetupReturnCode>$(ReturnCode)</SetupReturnCode>
+    </PropertyGroup>
+  </Target>
+
+  <Target Name="WcfCleanup" Condition="'$(WithCategories)' == 'OuterLoop'" AfterTargets="Test">
```
roncain: We'll need a smarter test to know if we are running OuterLoop for these 2 reasons:
1. WithCategories can be a semi-colon separated list
2. You can choose OuterLoop with /p:OuterLoop=true (I recently learned this)
3. You can trigger an OuterLoop by setting xunit traits directly on the command line.
We'll need to investigate when and how buildtools converts $(WithCategories) and $(OuterLoop) into xunit trait arguments to be accurate. But for now, I think we can get by with tests that look like this (my syntax may be off):
```
Condition="$(WithCategories.Contains('OuterLoop')) or '$(OuterLoop)' == 'true'"
```

build.override.targets
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Target Name="WcfSetup" Condition="'$(WithCategories)' == 'OuterLoop'" BeforeTargets="Test">
+    <Exec Command="$(MySetupScript) > $(MyLogFile)" />
```
roncain: Where are MySetupScript, MyLogFile and MyCleanupScript initialized?  We should probably choose better names (if we own the names), like WcfSetupScript.  We should probably also add to the Condition that these properties are not empty.

build.override.targets
```diff
@@ -0,0 +1,19 @@
...
+    <ReadLinesFromFile File="$(MyLogFile)" >
```
roncain: Why do we need to read from the log?  Was this something CoreFx requested?  The Exec task exposes ExitCode directly if that's all we need.  See https://msdn.microsoft.com/en-us/library/x8zx72cd.aspx

roncain:
A couple questions for @StephenBonikowsky ...
1. Why restrict only to Windows?  Granted, we will *use* this only on Windows, but why wouldn't we also want to allow setup/cleanup scripts to occur using .sh files?
2. If the content of the override targets is just an example for @hongdai , why would we merge this PR with "placeholder" content that does actual work?  I think it would be better to merge a PR that has only the basic framework you have here but leave out the placeholder code you expect to be replaced.

build.override.targets
roncain: Thanks for the clarification @StephenBonikowsky -- I won't code review the parts you expect to be replaced.

build.override.targets
StephenBonikowsky: Sounds good.

build.override.targets
StephenBonikowsky: This is just an example.

StephenBonikowsky:
> 1. Why restrict only to Windows? Granted, we will use this only on Windows, but why wouldn't we also want to allow setup/cleanup scripts to occur using .sh files?

My earlier statement wasn't accurate, we wouldn't use the Windows condition for importing the file, we would add it as a condition to any Target that should only run on Windows, it seemed from speaking with @zhenlan that this might be the case with the setup for the new infrastructure, maybe he can clarify this point.

> 2. If the content of the override targets is just an example for @hongdai , why would we merge this PR with "placeholder" content that does actual work? I think it would be better to merge a PR that has only the basic framework you have here but leave out the placeholder code you expect to be replaced. 

Then I'll just remove the build.override.targets from this PR entirely as its content is completely based on what Hong needs to do. I'll send her what I had via e-mail.

build.override.targets
StephenBonikowsky: Because if the ExitCode is anything other than 0 it is a build error. I wasn't able to take a non-0 code and set it to a property without it still throwing a build error.

build.proj
```diff
@@ -31,6 +31,9 @@
   <Import Project="$(ToolsDir)CodeCoverage.targets" Condition="Exists('$(ToolsDir)CodeCoverage.targets')" />
   <Import Project="$(ToolsDir)PerfTesting.targets" Condition="Exists('$(ToolsDir)PerfTesting.targets') and '$(Performance)' == 'true'"/>
 
+  <!-- Hook that can be used to insert custom build tasks to the build process such as setup and/or cleanup tasks -->
+  <Import Project="build.override.targets" Condition="Exists('build.override.targets')" />
+
```
iamjasonp: Question only: does this mean the build.override.targets must exist in the same directory as the build.proj, and therefore there is one build.override.targets for our repo, or can each child project that references this build.proj have their own override.targets?

build.cmd
```diff
@@ -104,7 +104,6 @@ set _binclashLoggerDll=%~dp0Tools\net45\Microsoft.DotNet.Build.Tasks.dll
 set _binclashlog=%~dp0binclash.log
 set _buildprefix=echo
 set _buildpostfix=^> "%_buildlog%"
-
```
zhenlan: NIT: Normally I prefer an empty line for clearance between defining vars and doing work, but if this is to sync with the build.cmd in corefx, I can live with it 😄.

zhenlan:
LGTM

roncain:
LGTM
I really like this approach of using msbuild overrides.  It is general purpose enough to be used for other scenarios as they arise.  Thanks @StephenBonikowsky for demonstrating sample usage that we can refactor and use for starting and stopping our self-host service.

build.cmd
StephenBonikowsky: Yeah, its just to get a clean diff.

build.proj
StephenBonikowsky: One build.override.targets for the repo.

StephenBonikowsky:
Amended it based on final corefx feedback that just moved the Import to the bottom of the build.proj.

PR #1070, opened 2016-04-22
Disable concurrent builds in code coverage

iamjasonp:
Copies a change for outerloops (see Line 215) to prevent ~~current~~ concurrent builds from happening in code coverage, at it also relies on outerloops with the WcfService

@dotnet-bot test ci please

EDIT: current -> concurrent

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -133,6 +133,13 @@ branchList.each { branchName ->
     // Archive results
     Utilities.addArchival(newJob, '**/coverage/*,msbuild.log')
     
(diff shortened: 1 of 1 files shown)


iamjasonp:
@dotnet-bot test ci please

roncain:
LGTM.
Question only (not blocking) -- I resume 'current' in the title s/b 'concurrent', right?  I didn't fully understand from the comment what was happening, but agree we should propagate this temporary fix we already made elsewhere

zhenlan:
LGTM

iamjasonp:
@roncain - you're right, it should read "concurrent" not "current". Edited inline

PR #1083, opened 2016-04-25
Integrate dev experience to build

hongdai:
* Hook Start and clean up service into build
* If the build starts the service, it will clean up the service as well.
* If the service already started, the build will not restart and clean up.
* The start service will be elevated.
* Need feedback on whether "BuildAllProjects" target is the best target to ensure
set up get run before tests start.
* I have removed all ENDLOCAL as it's not needed.
* Each script is designed to handle a particular set up and clean up so that the scripts can share between Lab and Dev experiences.
* We ping the service in the set up to ensure the service is fully started.
* All clean up errors are treated as non fatal. Need feedback on it.

### DIFF
--- a/build.override.targets
+++ b/build.override.targets
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0"  xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <TraversalBuildDependsOn>
+      WcfSetup;
+      $(TraversalBuildDependsOn);
+      WcfCleanup;
+    </TraversalBuildDependsOn>
+  </PropertyGroup>
+  <!--if ServiceUri is set, the service has been already started on another machine-->
+  <Target Name="WcfSetup" Condition="($(WithCategories.Contains('OuterLoop')) or '$(OuterLoop)' == 'true') and ('$(ServiceUri)' == '') and $(OSEnvironment) == 'Windows_NT'" >
+    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\StartWCFSelfHostedSvc.cmd" ContinueOnError="true">
+      <Output TaskParameter="ExitCode" PropertyName="WcfSetupErrorCode"/>
+    </Exec>
+    <Message Text="WcfSetupErrorCode: $(WcfSetupErrorCode)" />
+  </Target>
+
+  <Target Name="WcfCleanUp" Condition="$(WcfSetupErrorCode)!='-1' and $(WcfSetupErrorCode)!=''" >
+    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\CleanUpWCFSelfHostedSvc.cmd" ContinueOnError="true">
+    </Exec>
+  </Target>
+</Project>
--- a/src/System.Private.ServiceModel/tools/CertificateGenerator/Program.cs
+++ b/src/System.Private.ServiceModel/tools/CertificateGenerator/Program.cs
@@ -51,11 +51,37 @@ namespace CertUtil
 
             RemoveCertificatesFromStore(StoreName.Root, StoreLocation.LocalMachine);
             RemoveCertificatesFromStore(StoreName.Root, StoreLocation.CurrentUser);
+
+            RemoveCertificatesFromStore(StoreName.TrustedPeople, StoreLocation.LocalMachine);
         }
 
-        static void Main(string[] args)
+        static void Usage()
+        {
+            Console.WriteLine("Supported argument is -Uninstall");
+            Console.WriteLine("                      -help");
+        }
+        static int Main(string[] args)
         {
             ApplyAppSettings();
+
+            if (args.Length > 0)
+            {
+                if (string.Compare(args[0], "-Uninstall", true) == 0)
+                {
+                    UninstallAllCerts();
+                    return 0;
+                }
+                else if (string.Compare(args[0], "-help", true) == 0)
+                {
+                    Usage();
+                    return 0;
+                }
+                else
+                {
+                    Usage();
+                    return 1;
+                }
+            }
 
             UninstallAllCerts();
 
@@ -153,6 +179,8 @@ namespace CertUtil
 
             //Create CRL and save it
             File.WriteAllBytes(s_CrlFileLocation, certificateGenerate.CrlEncoded);
+
+            return 0;
 
         }
 
--- a/src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
+++ b/src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
@@ -34,14 +34,6 @@ namespace SelfHostedWCFService
             string websocketsBaseAddress = string.Format(@"https://localhost:{0}", websocketsPort);
 
             //Do not need to catch exceptions and dispose service hosts as the process will terminate
-            Uri crlUrl = new Uri(string.Format("http://localhost/CrlService.svc", httpPort));
-            WebServiceHost host = new WebServiceHost(typeof(CrlService), crlUrl);
-            WebHttpBinding binding = new WebHttpBinding();
-            host.AddServiceEndpoint(typeof(ICrlService), binding, "");
-            ServiceDebugBehavior stp = host.Description.Behaviors.Find<ServiceDebugBehavior>();
-            stp.HttpHelpPageEnabled = false;
-            host.Open();
-
             Uri[] utilTestServiceHostbaseAddress = new Uri[]  { new Uri(string.Format("{0}/Util.svc", httpBaseAddress))};
             UtilTestServiceHost utilTestServiceHostServiceHost = new UtilTestServiceHost(typeof(WcfService.Util), utilTestServiceHostbaseAddress);
             utilTestServiceHostServiceHost.Open();
@@ -286,9 +278,26 @@ namespace SelfHostedWCFService
             WebSocketHttpsDuplexTextBufferedTestServiceHost webSocketHttpsDuplexTextBufferedTestServiceHostServiceHost = new WebSocketHttpsDuplexTextBufferedTestServiceHost(typeof(WcfService.WSDuplexService), webSocketHttpsDuplexTextBufferedTestServiceHostbaseAddress);
             webSocketHttpsDuplexTextBufferedTestServiceHostServiceHost.Open();
 
+            //Start the crlUrl service last as the client use it to ensure all services have been started
+            Uri crlUrl = new Uri(string.Format("http://localhost/CrlService.svc", httpPort));
+            WebServiceHost host = new WebServiceHost(typeof(CrlService), crlUrl);
+            WebHttpBinding binding = new WebHttpBinding();
+            host.AddServiceEndpoint(typeof(ICrlService), binding, "");
+            ServiceDebugBehavior serviceDebugBehavior = host.Description.Behaviors.Find<ServiceDebugBehavior>();
+            serviceDebugBehavior.HttpHelpPageEnabled = false;
+            host.Open();
+
             Console.WriteLine("All service hosts have started.");
-            Console.WriteLine("Press <ENTER> to terminate the self service Host.");
-            Console.ReadLine();
+            do
+            {
+                Console.WriteLine("Type <Exit> to terminate the self service Host.");
+                string input = Console.ReadLine();
+                if (string.Compare(input, "exit", true) == 0)
+                {
+                    return;
+                }
+            } while (true);
+
         }
     }
 }
--- a/src/System.Private.ServiceModel/tools/scripts/BuildCertUtil.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/BuildCertUtil.cmd
@@ -37,5 +37,4 @@ echo.
 findstr /ir /c:".*Warning(s)" /c:".*Error(s)" /c:"Time Elapsed.*" "%_buildlog%"
 echo Build Exit Code = %BUILDERRORLEVEL%
 
-endlocal
 exit /b %BUILDERRORLEVEL%
--- a/src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
@@ -1,0 +1,33 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+echo Remove Https configuration
+
+net session >nul 2>&1
+set __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
+    echo Error: this script must be called in an elevated window
+	goto :done
+)
+
+call :RemoveHttps 0.0.0.0:8084
+
+call :RemoveHttps 0.0.0.0:44285
+
+call :RemoveHttps 0.0.0.0:443
+
+goto :done
+
+:RemoveHttps
+netsh http delete sslcert ipport=%1
+REM If the rules are not there, we get a errorleve 1, thus errorlevel 1 should be ignored
+set _ErrorLevel=%ERRORLEVEL%
+if NOT [%_ErrorLevel%]==[0] if NOT [%_ErrorLevel%]==[1] (
+    set __EXITCODE=%_ErrorLevel%
+    echo WARNING: An error occurred while deleting sslcert from port %1
+  )
+goto :eof
+
+:done
+
+exit /b %__EXITCODE%
--- a/src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
@@ -1,20 +1,51 @@
 @echo off 
 setlocal
 
-REM 
 echo Start Open all ports WCF test services use
-REM
 
-netsh advfirewall firewall add rule name="_WCF Test Server PortHttp" dir=in action=allow profile=any localport=8081 protocol=tcp
-netsh advfirewall firewall add rule name="_WCF Test Server PortHttps" dir=in action=allow profile=any localport=44285 protocol=tcp
-netsh advfirewall firewall add rule name="_WCF Test Server PortTcp" dir=in action=allow profile=any localport=809 protocol=tcp
-netsh advfirewall firewall add rule name="_WCF Test Server PortWebSocket" dir=in action=allow profile=any localport=8083 protocol=tcp
-netsh advfirewall firewall add rule name="_WCF Test Server PortWebSockets" dir=in action=allow profile=any localport=8084 protocol=tcp
+SET __EXITCODE=0
 
-REM we do not currently check the error level for the netsh advfirewall command. Will need to investigate the expected error
-REM code in all successful cases and add the check. Will also need to add remove the firewall rules script for clean up
+net session >nul 2>&1
 set __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
+    echo Error: this script must be called in an elevated window
+	goto :done
+)
 
-endlocal
-exit /b %__EXITCODE% 
+netsh advfirewall firewall add rule name="_WCF Test Server PortHttp"  dir=in action=allow profile=any localport=8081 protocol=tcp
+set __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
+  echo Error: error adding rule name _WCF Test Server PortHttp
+  goto :done
+  )
 
+netsh advfirewall firewall add rule name="_WCF Test Server PortHttps"  dir=in action=allow profile=any localport=44285 protocol=tcp
+set __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
+  echo Error: error adding rule name _WCF Test Server PortHttps
+  goto :done
+  )
+
+netsh advfirewall firewall add rule name="_WCF Test Server PortTcp"  dir=in action=allow profile=any localport=809 protocol=tcp
+set __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
+  echo Error: error adding rule name _WCF Test Server PortTcp
+  goto :done
+  )
+
+netsh advfirewall firewall add rule name="_WCF Test Server PortWebSocket"  dir=in action=allow profile=any localport=8083 protocol=tcp
+set __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
+  echo Error: error adding rule name _WCF Test Server PortWebSocket
+  goto :done
+  )
+
+netsh advfirewall firewall add rule name="_WCF Test Server PortWebSockets"  dir=in action=allow profile=any localport=8084 protocol=tcp
+set __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
+  echo Error: error adding rule name _WCF Test Server PortWebSockets
+  goto :done
+  )
+
+:done
+exit /b %__EXITCODE%
--- a/src/System.Private.ServiceModel/tools/scripts/PingWCFService.ps1
+++ b/src/System.Private.ServiceModel/tools/scripts/PingWCFService.ps1
@@ -1,0 +1,19 @@
+﻿param (
+    [Parameter(Mandatory=$true)]
+    [string] $Url
+)
+$request = New-Object System.Net.WebClient
+For ($i=0; $i -le 10; $i++)
+{
+    try
+    {
+        $request.DownloadFile($Url, "temp.crl")
+    }
+    catch
+    {
+        Start-Sleep -s 5
+        continue;
+    }
+    exit 0;
+}
+exit 1;
--- a/src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
@@ -1,0 +1,41 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+echo Start Remove all ports WCF test services use
+
+SET __EXITCODE=0
+
+net session >nul 2>&1
+set __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
+    echo Error: this script must be called in an elevated window
+	goto :done
+)
+
+call :DeleteRule "_WCF Test Server PortHttp"
+
+call :DeleteRule "_WCF Test Server PortHttps"
+
+call :DeleteRule "_WCF Test Server PortTcp"
+
+call :DeleteRule "_WCF Test Server PortWebSocket"
+
+call :DeleteRule "_WCF Test Server PortWebSockets"
+
+goto :done
+
+:DeleteRule
+netsh advfirewall firewall delete rule name=%1
+set _ErrorLevel=%ERRORLEVEL%
+REM errorlevel 1 means the rule is not found, thus we should ignore it.
+if NOT [%_ErrorLevel%]==[0] if NOT [%_ErrorLevel%]==[1] (
+    echo WARNING: An error occurred while removing firewall rule %1.
+    set __EXITCODE=%_ErrorLevel%
+  )
+
+goto:eof
+
+:done
+
+exit /b %__EXITCODE%
+
--- a/src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvcDoWork.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvcDoWork.cmd
@@ -1,50 +1,50 @@
-@echo off 
+@if "%_echo%" neq "on" echo off
 setlocal
 
-REM 
-echo Start WCF Self Hosted Service on local machine
-REM
+REM This script should always be called from StartWCFSelfHostedSvc.cmd
+
+echo Do the actual work to start WCF self hosted service
 
 REM Build tools
 call %~dp0BuildWCFSelfHostedService.cmd
-if NOT [%ERRORLEVEL%]==[0] (
+SET __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
     echo ERROR: An error occurred while building WCF Self hosted Service.
-    set __EXITCODE=%ERRORLEVEL% 
     goto :done
   )
 call %~dp0BuildCertUtil.cmd
-if NOT [%ERRORLEVEL%]==[0] (
+set __EXITCODE=%ERRORLEVEL%
+if NOT [%__EXITCODE%]==[0] (
     echo ERROR: An error occurred while building the Certificate generator.
-    set __EXITCODE=%ERRORLEVEL%
     goto :done
   )
 
-
 REM Config Certs
 REM we need the direcotry to save the test.crl file. We are investigate a way to get rid of it
-REM
 md c:\wcftest
-call %~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe %*
-set __EXITCODE=%ERRORLEVEL%
-
-if NOT [%__EXITCODE%]==[0] (
+REM Certificate configuration errors are all non fatal currently because we non cert tests will still pass
+%~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe
+if NOT [%ERRORLEVEL%]==[0] (
     echo Warning: An error occurred while running certificate generator.
   )
 
-REM skip the checking of error code as OpenFirewallPorts can return different error code even it's successful
 call %~dp0OpenFirewallPorts.cmd
+if NOT [%ERRORLEVEL%]==[0] (
+    echo Warning: An error occurred while running certificate generator.
+  )
 
 powershell -NoProfile -ExecutionPolicy unrestricted %~dp0ConfigHttpsPort.ps1
 if NOT [%ERRORLEVEL%]==[0] (
     echo Warning: An error occurred while configuration https port.
   )
 
+REM
 REM Start the self hosted WCF Test Service
-call %~dp0..\..\..\..\bin\Wcf\tools\SelfHostedWcfService\SelfHostedWcfService.exe %*
+call %~dp0..\..\..\..\bin\Wcf\tools\SelfHostedWcfService\SelfHostedWcfService.exe
 set __EXITCODE=%ERRORLEVEL%
 
+:Cleanup
+call %~dp0CleanUpWCFSelfHostedSvc.cmd
 :done
 
-endlocal
-exit /b %__EXITCODE% 
-
+exit /b %__EXITCODE%
--- a/src/System.Private.ServiceModel/tools/scripts/CleanUpWCFSelfHostedSvc.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/CleanUpWCFSelfHostedSvc.cmd
@@ -1,0 +1,46 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+REM
+echo Cleanup test services
+REM  Note that we always return successful code in clean up routine because clean up errors are not fatal
+
+TASKKILL /F /IM SelfHostedWCFService.exe
+REM error 128 will be returned if The process not found
+if NOT [%ERRORLEVEL%]==[0]  if NOT [%ERRORLEVEL%]==[128] (
+	echo Warning: An error occurred while killing the SelfHostedWCFService.exe.
+	)
+
+If NOT exist %~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe (
+	call %~dp0BuildCertUtil.cmd
+	set __EXITCODE=%ERRORLEVEL%
+	if NOT [%__EXITCODE%]==[0] (
+		echo Warning: An error occurred while building the Certificate generator.
+	 )
+)
+
+REM We will open a separate elevated window to do clean up only if the current window is not elevated
+REM This is to improve the debugging experience
+net session >nul 2>&1
+if NOT [%ERRORLEVEL%]==[0] (
+    set runelvated=%~dp0RunElevated.vbs
+)
+
+call %runelvated% %~dp0CleanUpHttpsPort.cmd
+if NOT [%ERRORLEVEL%]==[0] (
+	echo Warning: An error occurred while removing https port.
+	)
+
+call %runelvated% %~dp0RemoveFirewallPorts.cmd
+if NOT [%ERRORLEVEL%]==[0] (
+	echo Warning: An error occurred while removing Firewall ports.
+	)
+
+call %runelvated% %~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe -Uninstall
+if NOT [%ERRORLEVEL%]==[0] (
+	echo Warning: An error occurred while removing test certificates.
+	)
+
+:done
+
+exit /b 0
--- a/src/System.Private.ServiceModel/tools/scripts/RunElevated.vbs
+++ b/src/System.Private.ServiceModel/tools/scripts/RunElevated.vbs
@@ -1,0 +1,5 @@
+Set objShell = CreateObject("Shell.Application")
+dim fso,currentDirectory
+Set fso = CreateObject("Scripting.FileSystemObject")
+currentDirectory = fso.GetAbsolutePathName(".")
+objShell.ShellExecute WScript.Arguments(0), currentDirectory,"", "runas", 1
--- a/src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvc.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvc.cmd
@@ -1,0 +1,33 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+REM Script return code
+REM -1: Already running
+REM 0: Started the test service
+
+Echo Set Up test services
+set __EXITCODE=0
+
+REM If it's already running, we will not set up/clean up service
+tasklist /FI "imagename eq SelfHostedWCFService.exe" | findstr /i /C:"SelfHostedWCFService.exe" 1>nul
+if [%ERRORLEVEL%]==[0] (
+    set __EXITCODE=-1
+    goto :PingService
+)
+
+REM run the start wcf service elevated
+start /D %~dp0 RunElevated.vbs StartWCFSelfHostedSvcDoWork.cmd
+
+:PingService
+REM the serial number is not used in the service side. Thus it can be any number
+call powershell -NoProfile -ExecutionPolicy unrestricted %~dp0pingWcfService.ps1 "http://localhost/CrlService.svc/GetCrl?serialNum=b52"
+SET _ERRORLEVEL=%ERRORLEVEL%
+if [%_ERRORLEVEL%] NEQ [0] (
+    echo Error: failed to ping WCF Test Service.
+	set __EXITCODE=%_ERRORLEVEL%
+    goto :done
+)
+
+:done
+
+exit /b %__EXITCODE%


build.override.targets
```diff
@@ -0,0 +1,15 @@
...
+<Project ToolsVersion="4.0"  xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--if ServiceUri is set, the service has been already started on another machine-->
+  <Target Name="WcfSetup" Condition="($(WithCategories.Contains('OuterLoop')) or '$(OuterLoop)' == 'true') and ('$(ServiceUri)' == '') and $(OS.Contains('Windows'))" BeforeTargets="BuildAllProjects">
+    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\WCFCustomSetup.cmd" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="WcfSetupOutput" />
+    </Exec>
+    <Message Text="WcfSetupOutput: $(WcfSetupOutput)" />
+  </Target>
+
+  <Target Name="WcfCleanUp" Condition="($(WithCategories.Contains('OuterLoop')) or '$(OuterLoop)' == 'true') and ('$(ServiceUri)' == '') and $(WcfSetupOutput.Contains('WCFServiceCleanupNeeded')) and $(OS.Contains('Windows'))" AfterTargets="Test">
```
roncain: Can you use the ExitCode from the WcfSetup Exec call instead of checking console output?  We went through multiple iterations with the Bridge version to simplify it to use only the exit code (0 = already running, 1 = started).  I think the use of a magic string inside the console output could be fragile and not obvious if people modify that script.

src/System.Private.ServiceModel/tools/CertificateGenerator/Program.cs
```diff
@@ -53,10 +53,24 @@ static void UninstallAllCerts()
...
         }
 
-        static void Main(string[] args)
+        static int Main(string[] args)
         {
             ApplyAppSettings();
 
+            if (args.Length > 0)
+            {
+                if (string.Compare(args[0], "-Uninstall", true) == 0)
```
roncain: Nit: I prefer case-insensitive compares for arguments.

hongdai: Agreed. Will fix it.

src/System.Private.ServiceModel/tools/CertificateGenerator/Program.cs
```diff
@@ -53,10 +53,24 @@ static void UninstallAllCerts()
...
+                {
+                    UninstallAllCerts();
+                    return 0;
+                }
+                else
+                {
+                    Console.WriteLine(string.Format("Argument {0} is not supported, the only supported argument is -Uninstall", args[0]));
```
roncain: Nit: I would prefer having a Usage() method that shows valid usage of this command.  Then any problem can call it and exit.  Also consider supporting -? or -help to call Usage().

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -274,6 +266,15 @@ static void Main()
...
+            WebServiceHost host = new WebServiceHost(typeof(CrlService), crlUrl);
+            WebHttpBinding binding = new WebHttpBinding();
+            host.AddServiceEndpoint(typeof(ICrlService), binding, "");
+            ServiceDebugBehavior stp = host.Description.Behaviors.Find<ServiceDebugBehavior>();
+            stp.HttpHelpPageEnabled = false;
+            host.Open();
+
             Console.WriteLine("All service hosts have started.");
             Console.WriteLine("Press <ENTER> to terminate the self service Host.");
             Console.ReadLine();
```
roncain: Nit: in use, we found requiring only ENTER made it too easy to kill the Bridge.exe window by mistake.   As long as you are touching this file, I would prefer that the user must type "exit" or something like that.  We found this out the hard way before, so I would like to learn from that.

build.override.targets
hongdai: What if the script failed with error code 1? 

If we want to use the exit code, i'd suggest(-1 = already running, 0=started) and treat both -1 and 0 exit code as successful. Currently anything other than 0 will fail the setup and the build.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
```diff
@@ -0,0 +1,34 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+REM
+echo Remove Https configuration
+REM Note that we always return successful code in clean up routine because clean up errors are not fatal
+
+netsh http delete sslcert ipport=0.0.0.0:8084
```
roncain: It is time to start making ports configurable.  This is now the 2nd cmd file with hardcoded ports.  This comment is not blocking for the initial merge, but please open an issue to revisit this.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -1,10 +1,12 @@
-@echo off 
+@if "%_echo%" neq "on" echo off
 setlocal
 
-REM 
+REM
 echo Start WCF Self Hosted Service on local machine
 REM
 
+set WCFServiceCleanup=true
```
roncain: If I have already started the self-host server I think this script should exit immediately.  Can we call the ping powershell script first?  Not blocking for checkin.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -19,14 +21,12 @@ if NOT [%ERRORLEVEL%]==[0] (
     goto :done
   )
 
-
 REM Config Certs
 REM we need the direcotry to save the test.crl file. We are investigate a way to get rid of it
 REM
 md c:\wcftest
-call %~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe %*
+%~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe
```
roncain: Nit: you removed 'call' here but not in the other places.  We should be consistent.  Not blocking.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -39,12 +39,14 @@ if NOT [%ERRORLEVEL%]==[0] (
...
 
+REM
 REM Start the self hosted WCF Test Service
-call %~dp0..\..\..\..\bin\Wcf\tools\SelfHostedWcfService\SelfHostedWcfService.exe %*
+REM
+call %~dp0..\..\..\..\bin\Wcf\tools\SelfHostedWcfService\SelfHostedWcfService.exe
 set __EXITCODE=%ERRORLEVEL%
 
+:Cleanup
+call %~dp0WCFCustomCleanup.cmd WCFServiceCleanupNeeded
```
roncain: What is the purpose of the argument passed to WCFCustomCleanup.cmd?  As far as I can tell, the environment variable is always set to 'true' and the CMD does not look at arguments (plus you are not passing "true" but instead the string "WcfServiceCleanupNeeded" as %1).  I recommend deleting this argument for now.

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
```diff
@@ -0,0 +1,36 @@
...
+echo Cleanup test services
+REM  Note that we always return successful code in clean up routine because clean up errors are not fatal
+
+TASKKILL /F /IM SelfHostedWCFService.exe
+REM error 128 will be returned if The process not found
+if NOT [%ERRORLEVEL%]==[0]  if NOT [%ERRORLEVEL%]==[128] (
+	echo Warning: An error occurred while killing the SelfHostedWCFService.exe.
+	)
+
+call %~dp0CleanUpHttpsPort.cmd
```
roncain: Not blocking for review -- the reason I want to deal with making ports configurable is that we are accumulating debt in all the places that should forward those parameters, like here.

src/System.Private.ServiceModel/tools/scripts/WCFCustomSetup.cmd
```diff
@@ -0,0 +1,34 @@
...
+    goto :PingService
+)
+
+REM run the start wcf service elevated
+start /D %~dp0 RunElevated.vbs StartWCFService.cmd
+Echo WCFServiceCleanupNeeded
+
+
+:PingService
+call powershell -NoProfile -ExecutionPolicy unrestricted %~dp0pingWcfService.ps1 "http://localhost/CrlService.svc/GetCrl?serialNum=b52"
```
roncain: What is serialNum=52 for?  Recommend adding a comment.

build.override.targets
roncain: I recommend we use ExitCode. It took several iterations with the Bridge startup to settle on the 0=running, 1=started convention.  It doesn't matter if we use -1 or 1, but I think it makes the "contract" between the scripts very obvious to anyone reading the script or the .targets file and does not require any magic strings anywhere.

src/System.Private.ServiceModel/tools/scripts/WCFCustomSetup.cmd
```diff
@@ -0,0 +1,34 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+REM
+Echo Set Up test services
+REM
+
+REM ServiceUri is set, then the service is not local hosted
+if NOT "%ServiceUri%"=="" (
```
roncain: I strongly recommend we do not require ServiceUri to be an environment variable.  It should be passed to this script from the override .targets file via $(ServiceUri).  Until we make ports configurable, it could just be %1 for now, but ultimately we probably want WcfCustomSetup.cmd /ServiceUri=XXX /HttpPort=YYY /HttpsPort=ZZZ, etc.

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
```diff
@@ -0,0 +1,39 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+REM
+echo Start Remove all ports WCF test services use
+REM
+
+SET __EXITCODE=0
+
+netsh advfirewall firewall delete rule name="_WCF Test Server PortHttp"
```
roncain: Nit: this could be made simpler by replacing :CheckError with a :DeleteRule that calls netsh and checks the error. It would eliminate duplication of strings and look like
```
call :DeleteRule "_WCF TestServer PortHttp"
call :DeleteRule "_WCF TestServer PortHttps"
etc
```

build.override.targets
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0"  xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--if ServiceUri is set, the service has been already started on another machine-->
+  <Target Name="WcfSetup" Condition="($(WithCategories.Contains('OuterLoop')) or '$(OuterLoop)' == 'true') and ('$(ServiceUri)' == '') and $(OS.Contains('Windows'))" BeforeTargets="BuildAllProjects">
```
roncain: You asked a great question about whether BuildAllProjects is the right "before" target.  Another possibility is using the mechanism already being used by dir.traversal.targets, something like this:
```
  <PropertyGroup>
    <TraversalBuildDependsOn>
      WcfSetup;
      $(TraversalBuildDependsOn);
    </TraversalBuildDependsOn>
<PropertyGroup>
```
Another possibility is to say "Test" is the target you run before (to be consistent with being the target you run after) and to set an MsBuild property in this file to indicate "I have been called before".  If you did that, I think it would give a better experience for the /t:BuildAndTest that some of us use to run just one test project at a time.

Let's not block on the decision here but consider what we might do after the merge.

/cc @weshaggard for thoughts about where/how is the right place to hook .

weshaggard: For these I would follow the example at (https://github.com/dotnet/wcf/blob/master/build.proj#L59) and use TraversalBuildDependsOn, so you can ensure it gets run in the correct order.

build.override.targets
```diff
@@ -0,0 +1,15 @@
...
+    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\WCFCustomSetup.cmd" ConsoleToMSBuild="true">
```
weshaggard: I have to ask what is the reasoning for using a bunch of custom cmd/ps1/vbs scripts for this setup/clean-up work? That is going to make it more difficult to support these things on x-plat. I would suggest moving them into msbuild and perhaps even a custom build task to make the logic work x-plat.

build.override.targets
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0"  xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--if ServiceUri is set, the service has been already started on another machine-->
```
weshaggard: I would factor out all those custom conditions into properties so you can share them across the targets and also make reading them easier.

build.override.targets
roncain: At the moment, the self-hosted WCF services all rely on the full framework, and therefore can run only on Windows (ex: there is no NET Core version of the WCF server-side code).

build.override.targets
```diff
@@ -0,0 +1,15 @@
...
+  <Target Name="WcfSetup" Condition="($(WithCategories.Contains('OuterLoop')) or '$(OuterLoop)' == 'true') and ('$(ServiceUri)' == '') and $(OS.Contains('Windows'))" BeforeTargets="BuildAllProjects">
+    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\WCFCustomSetup.cmd" ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="WcfSetupOutput" />
+    </Exec>
+    <Message Text="WcfSetupOutput: $(WcfSetupOutput)" />
+  </Target>
+
```
weshaggard: For the cleanup you seem to hook AfterTargets=Test but for the setup you did BeforeTargets=BuildAllTargets. I suspect you want to try and scope these onto the same target, otherwise they might not be in-sync.

build.override.targets
weshaggard: I suspect it is worth adding that as part of the condition as well then.

build.override.targets
hongdai: What I need is to run setup before test running and clean up after test has run.

I have tried BeforeTargets=Test for the setup target,  however the tests got run before setup was called. So I'm out of idea what the same target I could use. Could you suggest one?

build.override.targets
hongdai: I have added the condition $(OS.Contains('Windows'))" . Is it sufficient?

build.override.targets
hongdai: My problem is How do I guarantee the Setup is called before testing started to run. I haven't found a good target to depend on.

build.override.targets
hongdai: I will make the change.

src/System.Private.ServiceModel/tools/CertificateGenerator/Program.cs
hongdai: Agreed.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
hongdai: Agreed.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: The ports will be passed from CMD to CMD via test properties. We have an issue opened to change self host service ports. Making ports configurable will be considered when we solve the issue.

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
hongdai: Agreed.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: I will remove the argument. It's not needed - I forgot to remove it.

src/System.Private.ServiceModel/tools/scripts/WCFCustomSetup.cmd
hongdai: I will remove this check then.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: If self-host server already started, WCFCustomSetup(who calls this script) will ping and exit.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: It's removed by accident. I'll add the call back. The call to an exe does not make any difference, but it makes difference for cmd scripts.

src/System.Private.ServiceModel/tools/scripts/WCFCustomSetup.cmd
hongdai: I will add a comment.

build.override.targets
weshaggard: Keep in mind that what ends up happening as part of a traversal project here is that the BuildAllProjects target runs the default target on each child project and that default target is BuildAndTest so while it might be confusing the BuildAllProjects actually makes each project do a BuildAndTest individually, so it encapsulates the full build and test run. The BuildAndTest target in build.proj itself pretty much only does a build as you can see the Test target is empty https://github.com/dotnet/wcf/blob/master/dir.traversal.targets#L247. So what I think you essentially want is something like like:
```
    <TraversalBuildDependsOn>
      WcfSetup;
      $(TraversalBuildDependsOn);
      WcfCleanup;
    </TraversalBuildDependsOn>
```

hongdai: @weshaggard It worked. I will make the change. Thanks for the suggestion!

build.override.targets
weshaggard: I would just use ``$(OSEnvironment)'=='Windows_NT'`

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -274,6 +266,15 @@ static void Main()
             WebSocketHttpsDuplexTextBufferedTestServiceHost webSocketHttpsDuplexTextBufferedTestServiceHostServiceHost = new WebSocketHttpsDuplexTextBufferedTestServiceHost(typeof(WcfService.WSDuplexService), webSocketHttpsDuplexTextBufferedTestServiceHostbaseAddress);
             webSocketHttpsDuplexTextBufferedTestServiceHostServiceHost.Open();
 
+            //Start the crlUrl service last as the client use it to ensure all services have been started
+            Uri crlUrl = new Uri(string.Format("http://localhost/CrlService.svc", httpPort));
```
zhenlan: Not blocking but we should really create a separate ping service.

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
```diff
@@ -0,0 +1,39 @@
...
+call :CheckError %ERRORLEVEL% "_WCF Test Server PortHttp"
```
zhenlan: Question: will we stop cleaning up the rest if the first one failed?

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -39,12 +39,14 @@ if NOT [%ERRORLEVEL%]==[0] (
     echo Warning: An error occurred while configuration https port.
   )
 
+REM
 REM Start the self hosted WCF Test Service
-call %~dp0..\..\..\..\bin\Wcf\tools\SelfHostedWcfService\SelfHostedWcfService.exe %*
+REM
+call %~dp0..\..\..\..\bin\Wcf\tools\SelfHostedWcfService\SelfHostedWcfService.exe
```
zhenlan: Question: Will this script pause here and continue to run when SelfHostedWcfService process either exits gracefully (by pressing enter) or is killed by taskkill?

zhenlan:
Thanks @hongdai for putting this together and using our new extension points!

I think you have all necessary pieces. My feedback is more on structure change, so I put them here together.

----------------
WCFCustomSetup.cmd
- Change the name to something like "StartWCFServiceElevated.cmd" and move the rest of logic to "StartWCFService.cmd". This script will be useful by itself even if not in the context of build.cmd, so I hope we can give it a name more meaningful.

StartWCFService.cmd owns the decision if a new instance of self-hosted WCF service should be started
- Move the logic of instance detection and ping service from "WCFCustomSetup.cmd" into "StartWCFService.cmd" and remove "ServiceUri" check.
- It should return the right exit code to indicate if it actually starts a new instance and "StartWCFServiceElevated.cmd" will pass along this exit code

WCFCustomCleanup.cmd
- Change the name to something like "CleanupSelfHostedWCFService.cmd" so we can easily tell what it is for. A meaningful name helps to use it even outside of the context of build.cmd.

build.override.target owns the decision if it should call starting self-hosted service by checking "ServiceUri" (in additon to Outerloop and Windows). 
- Change it to call "StartWCFServiceElevated.cmd" instead of "WCFCustomSetup.cmd"
- Change it to call "CleanupSelfHostedWCFService.cmd" instead of "WCFCustomCleanup.cmd"
- Change the condition of calling cleanup scrip solely depending on the exit code of "StartWCFServiceElevated.cmd", ie., the cleanup will only be called if the "StartWCFServiceElevated.cmd" is called and it asks cleanup to happen.

---------
Other thoughts:

If you think people will always call StartWCFServiceElevated.cmd, you can consider renaming it to StartWCFService.cmd and renaming the original StartWCFService.cmd to something like StartWCFServiceDoWork.cmd.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -274,6 +266,15 @@ static void Main()
...
+            WebServiceHost host = new WebServiceHost(typeof(CrlService), crlUrl);
+            WebHttpBinding binding = new WebHttpBinding();
+            host.AddServiceEndpoint(typeof(ICrlService), binding, "");
+            ServiceDebugBehavior stp = host.Description.Behaviors.Find<ServiceDebugBehavior>();
```
iamjasonp: NIT: name `stp` - isn't apparent what it stands for

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -274,6 +266,15 @@ static void Main()
             WebSocketHttpsDuplexTextBufferedTestServiceHost webSocketHttpsDuplexTextBufferedTestServiceHostServiceHost = new WebSocketHttpsDuplexTextBufferedTestServiceHost(typeof(WcfService.WSDuplexService), webSocketHttpsDuplexTextBufferedTestServiceHostbaseAddress);
             webSocketHttpsDuplexTextBufferedTestServiceHostServiceHost.Open();
```
iamjasonp: Suggestion only: I almost feel like all these Uri/ServiceHost pairs should go into a Tuple<Uri, ServiceHost> pair and then you can loop over them to open.

src/System.Private.ServiceModel/tools/scripts/BuildCertUtil.cmd
```diff
@@ -37,5 +37,4 @@ echo.
 findstr /ir /c:".*Warning(s)" /c:".*Error(s)" /c:"Time Elapsed.*" "%_buildlog%"
 echo Build Exit Code = %BUILDERRORLEVEL%
 
-endlocal
```
iamjasonp: Question only: How does removing the `endlocal` impact `call`ing of a batch file from another? We don't `call BuildCertUtil.cmd` today, but will it produce weird environments if we omit `endlocal`?

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
```diff
@@ -0,0 +1,34 @@
...
+
+goto :done
+
+:CheckError
+SET _ErrorLevel=%1
+if NOT [%_ErrorLevel%]==[0] if NOT [%_ErrorLevel%]==[1] (
+    echo %2
+    set __EXITCODE=%_ErrorLevel%
+  )
+goto:eof
```
iamjasonp: i think you're missing a space between "goto" and ":eof"

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
```diff
@@ -0,0 +1,34 @@
...
+
+:done
+
+exit /b %__EXITCODE%
+
+
+
+
+
```
iamjasonp: NIT: why so many newlines?

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
```diff
@@ -0,0 +1,34 @@
...
+REM Note that we always return successful code in clean up routine because clean up errors are not fatal
+
+netsh http delete sslcert ipport=0.0.0.0:8084
+call :CheckError %ERRORLEVEL% "WARNING: An error occurred while deleting sslcert from port 8084."
+
+netsh http delete sslcert ipport=0.0.0.0:44285
+call :CheckError %ERRORLEVEL% "WARNING: An error occurred while deleting sslcert from port 44285."
+
+netsh http delete sslcert ipport=0.0.0.0:443
+call :CheckError %ERRORLEVEL% "WARNING: An error occurred while deleting sslcert from port 443."
```
iamjasonp: If this rule doesn't already exist, does this mean we'll get warnings? If so, seems weird that a cleanup step would output errors if you try to clean up something that didn't exist.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
```diff
@@ -0,0 +1,34 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+REM
+echo Remove Https configuration
+REM Note that we always return successful code in clean up routine because clean up errors are not fatal
```
iamjasonp: You should be checking if the script is being run elevated. checking whether or not `net session` exits with errorlevel 0 should be a sufficient indicator of whether or not you're elevated.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -5,16 +5,32 @@ REM
 echo Start Open all ports WCF test services use
 REM
```
iamjasonp: elide extra REM

iamjasonp: This script needs to check for elevation, and bail out if we're not running elevated

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
```diff
@@ -0,0 +1,34 @@
...
+call :CheckError %ERRORLEVEL% "WARNING: An error occurred while deleting sslcert from port 44285."
+
+netsh http delete sslcert ipport=0.0.0.0:443
+call :CheckError %ERRORLEVEL% "WARNING: An error occurred while deleting sslcert from port 443."
+
+goto :done
+
+:CheckError
+SET _ErrorLevel=%1
+if NOT [%_ErrorLevel%]==[0] if NOT [%_ErrorLevel%]==[1] (
```
iamjasonp: Question: What does exit code 1 for netsh mean?

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -5,16 +5,32 @@ REM
...
-REM we do not currently check the error level for the netsh advfirewall command. Will need to investigate the expected error
-REM code in all successful cases and add the check. Will also need to add remove the firewall rules script for clean up
-set __EXITCODE=%ERRORLEVEL%
+goto :done
 
-endlocal
-exit /b %__EXITCODE% 
+:CheckError
+SET _ErrorLevel=%1
+if NOT [%_ErrorLevel%]==[0] if NOT [%_ErrorLevel%]==[1] (
```
iamjasonp: Question: What does exit code 1 for netsh mean? Is it okay to ignore an error here?

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -5,16 +5,32 @@ REM
...
+    echo WARNING: An error occurred while adding firewall rule name %2.
+    set __EXITCODE=%_ErrorLevel%
+  )
```
iamjasonp: Should you be exiting if the ERRORLEVEL is not 0 or 1

src/System.Private.ServiceModel/tools/scripts/PingWCFService.ps1
```diff
@@ -0,0 +1,19 @@
+﻿param (
+    [Parameter(Mandatory=$true)]
+    [string] $Url
+)
+$request = New-Object System.Net.WebCLient
```
iamjasonp: NIT: typo on WebC**L**ient

src/System.Private.ServiceModel/tools/scripts/PingWCFService.ps1
```diff
@@ -0,0 +1,19 @@
...
+For ($i=0; $i -le 10; $i++)
+{
+    try
+    {
+        $request.DownloadFile($Url, "temp.crl")
```
iamjasonp: Why are we trying to download "temp.crl" here when the purpose of this script is to ping the service to see if it's alive?

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
```diff
@@ -0,0 +1,39 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+REM
+echo Start Remove all ports WCF test services use
+REM
```
iamjasonp: NIT: remove redundant REMs from file.

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
```diff
@@ -0,0 +1,39 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
```
iamjasonp: Check for elevation for this script

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
```diff
@@ -0,0 +1,39 @@
...
+call :CheckError %ERRORLEVEL% "_WCF Test Server PortWebSocket"
+
+netsh advfirewall firewall delete rule name="_WCF Test Server PortWebSockets"
+call :CheckError %ERRORLEVEL% "_WCF Test Server PortWebSockets"
+
+goto :done
+
+:CheckError
+SET _ErrorLevel=%1
+if NOT [%_ErrorLevel%]==[0] if NOT [%_ErrorLevel%]==[1] (
```
iamjasonp: I notice this pattern a lot, but it's not apparent to me in any of the scripts what ERRORLEVEL == 1 is. 0 is pretty standard, but if 1 is something special in netsh, it'd be great to put some comments here to explain why 1 is an acceptable errorlevel to ignore when calling netsh.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
```diff
@@ -19,14 +21,12 @@ if NOT [%ERRORLEVEL%]==[0] (
...
 REM we need the direcotry to save the test.crl file. We are investigate a way to get rid of it
 REM
 md c:\wcftest
-call %~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe %*
+%~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe
 set __EXITCODE=%ERRORLEVEL%
-
 if NOT [%__EXITCODE%]==[0] (
     echo Warning: An error occurred while running certificate generator.
   )
```
iamjasonp: Line 34 (I can't comment there) 
Could you elaborate here or in OpenFirewallPorts.cmd under what scenarios you'll get a non-zero error code?

src/System.Private.ServiceModel/tools/scripts/WCFCustomSetup.cmd
```diff
@@ -0,0 +1,34 @@
...
+
+REM If it's already running, we will not set up/clean up service
+tasklist /FI "imagename eq SelfHostedWCFService.exe" | findstr /i /C:"SelfHostedWCFService.exe" 1>nul
+if [%ERRORLEVEL%]==[0] (
+    goto :PingService
+)
+
+REM run the start wcf service elevated
+start /D %~dp0 RunElevated.vbs StartWCFService.cmd
+Echo WCFServiceCleanupNeeded
```
iamjasonp: What's this echo for?

src/System.Private.ServiceModel/tools/scripts/WCFCustomSetup.cmd
iamjasonp: I feel like the CRL service being co-opted for the PingService feels strange to me...

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
```diff
@@ -0,0 +1,36 @@
...
+	)
+
+:done
+
+exit /b 0
+
+
+
+
+
```
iamjasonp: NIT: lots of extra newlines

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
```diff
@@ -0,0 +1,36 @@
...
+if NOT [%ERRORLEVEL%]==[0] (
+	echo Warning: An error occurred while removing https port.
+	)
+
+call %~dp0RemoveFirewallPorts.cmd
+if NOT [%ERRORLEVEL%]==[0] (
+	echo Warning: An error occurred while removing Firewall ports.
+	)
+
+call %~dp0..\..\..\..\bin\Wcf\tools\CertificateGenerator\CertificateGenerator.exe -Uninstall
```
iamjasonp: Might be worth checking if this exists first before trying to run it; if someone did a partial build of the WCF repo then it's possible that the CertificateGenerator has not been created yet.

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
```diff
@@ -0,0 +1,36 @@
+@if "%_echo%" neq "on" echo off
```
iamjasonp: Who do we expect to be calling WcfCustomCleanup.cmd and WcfCustomSetup,cmd? It seems like you've assumed that this will be called only via MSBuild during a full build. If this is the case, I recommend either renaming the script to something that describes it (say, WcfServiceMSBuildCleanup.cmd, WcfServiceMSBuildSetup.cmd) or do some kind of check at the tops of the scripts to check it's running in the context of MSBuild. 

Not blocking - but as I read through the script it's apparent to me that you've assumed certain conditions for this script to work (e.g., that the CertificateGenerator is around)

iamjasonp:
> All clean up errors are treated as non fatal. Need feedback on it.

I think this depends on your confidence level for each particular cleanup step and what we determine as "normal". I see completion for cleanup happening in two paths: 

1. Expected path (normal + expected error cases), which are considered okay; at the end of execution, my machine is in the state I expect it in. 
2. Abnormal paths (unexpected error cases) which are not considered okay; these errors leave the machine in a bad or unknown state. 

For example, in an earlier case, you check the error code taskkill and ensure that if 0 or 128, then treat it as not a problem. However, there might be (rare) cases where those error codes aren't returned, in which case, we probably want to report it. 

Ditto with netsh and error code 0 / 1. We should be able to determine what "normal" errors are (e.g., firewall rule never existed) vs. true cleanup errors (e.g., running as non-elevated which causes the command to fail).

It would be good to comment on when an exit code is ignored by scripts and not though :smile: Documentation on, say, netsh exit codes is sparse.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
roncain: I agree we should find a better structure for these pairs in a new PR.  It would also offer the option to display them to the console. For future thought.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
roncain: No the ports won't be passed via TestProperties from CMD to CMD.  CMD files do not have access to the MsBuild properties unless we pass them explicitly. We can deal with this when we make ports configurable.  I'm just observing not doing this soon is accumulating more debt as we go, so we may want to increase the priority of that discussion.

src/System.Private.ServiceModel/tools/scripts/WCFCustomSetup.cmd
roncain: If you remove the check, how will this code know it should exit due to not running as localhost?  If you check that condition from another place and only call this when $(ServiceUri) is empty, then this should be renamed WcfSetupSelfHostedWcfService.cmd or something like that.  WcfCustomSetup doesn't describe what it does.  The rename is not blocking but worth considering.

build.override.targets
hongdai: These custom conditions only shared in WCF setup and clean up targets. I will make them into properties if there is a need to share with more targets in the future.

build.override.targets
hongdai: Agreed. Will make the change.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
hongdai: Ok. I will consider it in the future.

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
hongdai: Agreed. Will rename it.

build.override.targets
weshaggard: That is up to you guys but do keep in mind it isn't just about sharing it is about readability as well as the conditions are very log right now and I cannot even see the full line without scrolling.

src/System.Private.ServiceModel/tools/scripts/BuildCertUtil.cmd
hongdai: It should not. The documentation said it would implicitly call endlocal to each setlocal when a batch file exit.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: Agreed. Will add the check.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: If the rule does not exit, the errorlevel is 1. We will NOT get the warning.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: Means the rule does not exist. I should NOT have added it to the OpenFireWallPorts though. Will remove it.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: Good catch! Will fix it.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: Not intentionally. Will remove them.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
hongdai: Ok.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
hongdai: exit code 1 is OK for removing firewall port, not adding it. I'll remove it here.

src/System.Private.ServiceModel/tools/scripts/PingWCFService.ps1
hongdai: Will be considered in the future.

src/System.Private.ServiceModel/tools/scripts/PingWCFService.ps1
hongdai: Oh, will fix it.

src/System.Private.ServiceModel/tools/scripts/WCFCustomSetup.cmd
hongdai: The WCFSetup target has a condition on $(ServiceUri) is empty. So WCFCustomSetup will not get run if it's NOT localhost.
Agree to rename it to be more specific.

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
hongdai: Nope. Clean up is non fatal errors. It will just issue a warning.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: Yes. It will pause here.

roncain: Is pausing (i.e. blocking execution) at this point the right thing to do?  The EnsureBridgeRunning used to launch a new process.  I found it awkward to have to open a 2nd window for this.

src/System.Private.ServiceModel/tools/scripts/PingWCFService.ps1
roncain: Agree to consider for the future -- incidentally, having a ping address I could hit using 'curl' from Linux or Fiddler was instrumental in testing whether the services were available.  I'll open an issue for it

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
hongdai: I'd prefer to continue and let the caller decide if they want to continue.

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
hongdai: Agreed.

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
hongdai: ok.

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
hongdai: I will add comments.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: I will explain why "1" sometimes are successful code. But I cannot explain non-zero as there could be too many error conditions.

src/System.Private.ServiceModel/tools/scripts/WCFCustomSetup.cmd
hongdai: Will be removed as we have decided to use exit code to decide if clean up is needed.

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
hongdai: Will remove them.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: This is because we want to have the opportunity to do graceful cleanup.

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
hongdai: It will product an error if the file is not found. Not much difference.

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
hongdai: I originally designed to be only called from build.cmd. Now I think Dev can use it to launch startwcfservice.cmd. 

In the clean up, it's usually the best effort - output error as warning, then continue.

What are other conditions you observed other than assume CertificateGenerator is around?

hongdai:
I will comment on all non zero exit code that ignored by scripts.
+ 1 on The netsh documentation on error code :smile:

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
zhenlan: Thanks @hongdai. Will a taskkill of the exe resume this script?

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
roncain: My personal opinion is that the dev should be able to run StartWcfService from a clean enlistment and have it build and start.  That is how StartBridge worked.  As far as allowing the dev to call other scripts by hand, I'd support that only if necessary to get a good experience with VS.  For example, can I just do StartWcfService and then go run tests from VS?  Or would I also need to run other scripts?  My most common scenario is:
```
   build /p:SkipTests=true
   StartBridge (or StartWcfService)
```
then I open VS and select the test I want run.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
iamjasonp: Can you make a comment to that effect that ERRORLEVEL 1 == "rule doesn't exist" please? 0 is well documented; any other one isn't really.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
iamjasonp: Ok. Just be warned that __EXITCODE may take on more than one value during the course of this script, if multiple commands have errors. As long as we're not differentiating between them in the calling script, it should be fine.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -5,16 +5,32 @@ REM
...
+goto:eof
```
iamjasonp: NIT: space needed between goto and :eof, but it doesn't seem to be affecting script processing.

src/System.Private.ServiceModel/tools/scripts/PingWCFService.ps1
iamjasonp: +1 for future consideration

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
iamjasonp: I admittedly haven't run this script on my box yet, but I did notice that one dependency as mentioned above. I do agree with Ron's comment though that running this command should mean that it'll start the WCF service without fail except under exceptional circumstances. 

I ran into similar pitfalls with Linux certificate installers and the like where I depended on, but didn't explicitly state or check for, such dependencies. It makes the experience a bit more fluid and nicer for the dev

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
hongdai: Nope it will not. But it will not affect set up as set up does not assume a clean ENV.

src/System.Private.ServiceModel/tools/scripts/WCFCustomCleanup.cmd
hongdai: You will be able to do the same, just replace StartBridge with StartWCFSelfHostedSvc.cmd, will will call StartWCFSelfHostedSvcDoWork.cmd in another window, so it does not steal your window.

src/System.Private.ServiceModel/tools/scripts/StartWCFService.cmd
zhenlan: Sounds good. I just want to understand if the rest of script will ever be executed if the exe is taskkilled. In this case, the cleanup code in this script will be skipped, but that's not a problem because the reason that exe is taskkilled is due to another cleanup call from build.override.targets.

build.override.targets
```diff
@@ -0,0 +1,23 @@
...
+  </PropertyGroup>
+  <!--if ServiceUri is set, the service has been already started on another machine-->
+  <Target Name="WcfSetup" Condition="($(WithCategories.Contains('OuterLoop')) or '$(OuterLoop)' == 'true') and ('$(ServiceUri)' == '') and $(OSEnvironment) == 'Windows_NT'" >
+    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\StartWCFSelfHostedSvc.cmd">
+      <Output TaskParameter="ExitCode" PropertyName="WcfSetupErrorCode"/>
+    </Exec>
+    <Message Text="WcfSetupErrorCode: $(WcfSetupErrorCode)" />
+  </Target>
+
+  <Target Name="WcfCleanUp" Condition="$(WcfSetupErrorCode)=='0'" >
```
zhenlan: I did not see `WcfSetupErrorCode` is initialized. If target `WcfSetup` did not run as its conditions did not meet, what's the value of `WcfSetupErrorCode`? - we should skip `WcfCleanUp` target in this case.

src/System.Private.ServiceModel/tools/CertificateGenerator/Program.cs
```diff
@@ -53,10 +53,34 @@ static void UninstallAllCerts()
...
+                }
+                else if (string.Compare(args[0], "-help", true) == 0)
+                {
+                    Usage();
+                    return 0;
+                }
+                else
+                {
+                    Usage();
+                    return -1;
```
zhenlan: Can we return 1 instead of -1, because script code like `if ERRORLEVEL 1 goto :Failure` will fail to capture failure like this as it can only capture exit code that is equal or greater than 1?

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
```diff
@@ -0,0 +1,32 @@
...
+
+call :RemoveHttps 0.0.0.0:44285
+
+call :RemoveHttps 0.0.0.0:443
+
+goto :done
+
+:RemoveHttps
+netsh http delete sslcert ipport %1
+//If the rules are not there, we get a errorleve 1, thus errorlevel 1 should be ignored
```
zhenlan: Does cmd support `//` as comment?

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -1,20 +1,36 @@
...
+
+call :AddRule "_WCF Test Server PortWebSockets" dir=in action=allow profile=any localport=8084 protocol=tcp"
+
+goto :done
+
+:AddRule
+netsh advfirewall firewall add rule name=%1
+if NOT [%_ErrorLevel%]==[0] (
+    echo WARNING: An error occurred while adding firewall rule %1.
+    set __EXITCODE=%_ErrorLevel%
```
zhenlan: Just to point out the last caller of this function will set (overwrite) the exit code for the whole script. Should we fail the whole script on first failure instead?

src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvcDoWork.cmd
```diff
@@ -1,50 +1,54 @@
-@echo off 
+@if "%_echo%" neq "on" echo off
 setlocal
 
-REM 
-echo Start WCF Self Hosted Service on local machine
-REM
+REM Script return code
+REM -1: alreay running
```
zhenlan: This is not applicable in this script anymore.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
hongdai: I treat it as warning currently. I'm ok to fail the entire scripts and exit. Will make the change.

src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvcDoWork.cmd
hongdai: Correct. I forgot to remove it.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: Nope. I'll change it.

src/System.Private.ServiceModel/tools/CertificateGenerator/Program.cs
hongdai: Agreed. I'll change it to 1.

build.override.targets
hongdai: By default, WcfSetupErrorCode is empty string "". The WcfCleanUp won't run if WcfSetupErrorCode is not explicitly set to string 0.

zhenlan: Sounds good. Thanks @hongdai for confirmation.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -1,20 +1,37 @@
...
+call :AddRule "_WCF Test Server PortWebSockets" dir=in action=allow profile=any localport=8084 protocol=tcp"
+
+goto :done
+
+:AddRule
+netsh advfirewall firewall add rule name=%1
+if NOT [%_ErrorLevel%]==[0] (
+    echo Error: An error occurred while adding firewall rule %1.
+    set __EXITCODE=%_ErrorLevel%
+	goto :done
```
zhenlan: I am just not sure if this will jump to the outside of current function. You may like to verified. Otherwise, this sounds good to me.

zhenlan:
LGTM. Thanks @hongdai.

build.override.targets
```diff
@@ -0,0 +1,23 @@
...
+  <PropertyGroup>
+    <TraversalBuildDependsOn>
+      WcfSetup;
+      $(TraversalBuildDependsOn);
+      WcfCleanup;
+    </TraversalBuildDependsOn>
+  </PropertyGroup>
+  <!--if ServiceUri is set, the service has been already started on another machine-->
+  <Target Name="WcfSetup" Condition="($(WithCategories.Contains('OuterLoop')) or '$(OuterLoop)' == 'true') and ('$(ServiceUri)' == '') and $(OSEnvironment) == 'Windows_NT'" >
+    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\StartWCFSelfHostedSvc.cmd">
```
roncain: You need to add ContinueOnError="true" to the Exec line for the manually started WCF service scenario to work.  Here is what happened when I started the WCF test service first and then tried to build:
```
c:\OSS\wcf>build /p:WithCategories=OuterLoop
Tools are already initialized.
[ 4:45:53.43] Building Native Libraries...
[ 4:45:54.91] Building Managed Libraries...
  Set Up test services
c:\OSS\wcf\build.override.targets(13,5): error MSB3073: The command "c:\OSS\wcf\\src\System.Private.ServiceModel\tools\scripts\StartWCFSelfHostedSvc.cmd" exited with code -1. [c:\OSS\wcf\build.proj]

c:\OSS\wcf\build.override.targets(13,5): error MSB3073: The command "c:\OSS\wcf\\src\System.Private.ServiceModel\tools\scripts\StartWCFSelfHostedSvc.cmd" exited with code -1. [c:\OSS\wcf\build.proj]

    0 Warning(s)
    1 Error(s)
Time Elapsed 00:00:04.13
[ 4:45:59.75] Build Exit Code = 1

c:\OSS\wcf>
```

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
```diff
@@ -1,20 +1,37 @@
...
+call :AddRule "_WCF Test Server PortTcp" dir=in action=allow profile=any localport=809 protocol=tcp"
+
+call :AddRule "_WCF Test Server PortWebSocket" dir=in action=allow profile=any localport=8083 protocol=tcp"
+
+call :AddRule "_WCF Test Server PortWebSockets" dir=in action=allow profile=any localport=8084 protocol=tcp"
+
+goto :done
+
+:AddRule
+netsh advfirewall firewall add rule name=%1
```
roncain: When I run build.cmd to start everything, I see the following error for every firewall port.  Are they expected?  These errors are come from netsh.
```
Error: An error occurred while adding firewall rule "_WCF Test Server PortHttp".
One or more essential parameters were not entered.
Verify the required parameters, and reenter them.
```

roncain: I see the problem -- you need to change %1 to %* to get the rest of the arguments.

---
In reply to: [61415751](https://github.com/dotnet/wcf/pull/1083#discussion_r61415751) [](ancestors = 61415751)

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
```diff
@@ -0,0 +1,39 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+echo Start Remove all ports WCF test services use
+
+SET __EXITCODE=0
+
+net session > NUL
+if NOT [%ERRORLEVEL%]==[0] ( (
```
roncain: There are too many open parentheses here.  Unless I manually remove this one open paren, running this command yields:
```
c:\OSS\wcf\src\System.Private.ServiceModel\tools\scripts>RemoveFirewallPorts.cmd
Start Remove all ports WCF test services use
The syntax of the command is incorrect.
```

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
roncain: I opened issue #1089 for making a ping service later.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
```diff
@@ -0,0 +1,32 @@
+@if "%_echo%" neq "on" echo off
+setlocal
+
+echo Remove Https configuration
+
+net session > NUL
+if NOT [%_ErrorLevel%]==[0] (
```
roncain: When this script runs on my machine (in an elevated CMD window), I get this error, and the https ports are not removed:
```
Remove Https configuration
Error: this script must be called in an elevated window
```
I think the error is using an underscore in %_ErrorLevel%.  When I remove the underscore, it gets past this point.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
```diff
@@ -0,0 +1,32 @@
...
+call :RemoveHttps 0.0.0.0:8084
+
+call :RemoveHttps 0.0.0.0:44285
+
+call :RemoveHttps 0.0.0.0:443
+
+goto :done
+
+:RemoveHttps
+netsh http delete sslcert ipport %1
```
roncain: I think you need an "=" between ippprt and %1.  When I run it, I see:
```
WARNING: An error occurred while deleting sslcert from port 0.0.0.0:8084
The parameter is incorrect.
```
When the script is done, 'netsh http show sslcert' shows me these ports are still assigned.

When I add the "=" it (and fix the %_ErrorLevel%) this script runs successfully.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
roncain: Actually, there are other issues, including unmatched quotes.  Here is how I rewrote it and it seems to work:
```
call :AddRule "_WCF Test Server PortHttp" 8081

call :AddRule "_WCF Test Server PortHttps" 44285

call :AddRule "_WCF Test Server PortTcp" 809

call :AddRule "_WCF Test Server PortWebSocket" 8083

call :AddRule "_WCF Test Server PortWebSockets" 8084

goto :done

:AddRule
echo netsh advfirewall firewall add rule name=%1 dir=in action=allow profile=any localport=%2 protocol=tcp
netsh advfirewall firewall add rule name=%1 dir=in action=allow profile=any localport=%2 protocol=tcp
if NOT [%ErrorLevel%]==[0] (
    echo Error: An error occurred while adding firewall rule %*.
    set __EXITCODE=%ErrorLevel%
 goto :done
  )
goto:eof

```


---
In reply to: [61416253](https://github.com/dotnet/wcf/pull/1083#discussion_r61416253) [](ancestors = 61416253,61415751)

roncain:
Thanks for all the cleanup work @hongdai !
LGTM, subject to fixing the CMD files as I mentioned in the comments.  Post-merge, we should manually verify that all the cleanup scripts really do cleanup (i.e. 'netsh http show sslcert', looking at certmgr, looking at firewall. etc.)

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
hongdai: You are right it's not jump out of the outside of current function. I'll need to find another way.

build.override.targets
hongdai: You're right that I need to add ContinueOnError="true" to allow manually started WCF service.  But it means we will ignore any set up error.

src/System.Private.ServiceModel/tools/scripts/OpenFirewallPorts.cmd
hongdai: The only problem is goto :done does not exit from the function. It will still continue on a adding rule failure. I'll have to go back to my original way(without using a function).

src/System.Private.ServiceModel/tools/scripts/RemoveFirewallPorts.cmd
hongdai: That's right. I need to remove one.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: I forgot to set _ErrorLevel.  I'll set __ExitCode directly to avoid two statements.

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: Again I missed set up _ErrorLevel here and "=". I used build.cmd to test it, somehow msbuild.log does not show any errors. Looks like I need to test each script independently.

build.override.targets
roncain: If we detect the exit code is other than the ones we expect (0 or -1), then I believe we can use the Error task to log a message and fail the build.  https://msdn.microsoft.com/en-us/library/8b08t3s4.aspx .

src/System.Private.ServiceModel/tools/scripts/CleanUpHttpsPort.cmd
hongdai: @roncain Thanks for testing and catching all the script failures! I will do extensive testing before merge.

hongdai:
In the third iteration:
* I have fixed several script errors.
* Make a few improvement so that any errors will show up in build log if it's
running from an Admin window.
* I have tested all the scenarios to my best.

I will merge once the CI check passes.

hongdai:

iamjasonp:
LGTM2. thanks @hongdai!

PR #1086, opened 2016-04-25
Implements Ssl security upgrade for NetTcp for UWP apps

### LINKED ISSUE #833: PlatformNotSupportedException: SslStreamSecurityUpgradeInitiator.InInitiateUpgradeAsync
|Failing Test Cases |
|--------------------------|
|**Tcp_ClientCredentialTypeTests.SameBinding_SecurityModeTransport_ClientCredentialTypeCertificate_EchoString**|
|**Tcp_ClientCredentialTypeTests.TcpClientCredentialType_Certificate_With_ServerAltName_EchoString**|
|**IdentityTests.VerifyServiceIdentityMatchDnsEndpointIdentity**|
|**IdentityTests.ServiceIdentityNotMatch_Throw_MessageSecurityException**|
|**ExpectedExceptionTests.TCP_ServiceCertExpired_Throw_MessageSecurityException**|
|**ExpectedExceptionTests.TCP_ServiceCertRevoked_Throw_MessageSecurityException**|

### LINKED ISSUE #1057: TcpClientCredentialType_Certificate_EchoString fails with new self-hosted WCF service
Start the new self-hosted WCF services and run OuterLoops to see this failure.  If you don't see it, please try it on a different machine or VM.  My net.tcp port sharing service and listener adapter were both confirmed running when this failed.

```
                 c:\OSS\wcf\src\System.Private.ServiceModel\tests\Scenarios\Security\TransportSecurity\Https\ClientCredentialTypeTests.cs(157,0): at Https_ClientCredentialTypeTests.NtlmAuthentication_RoundTrips_Echo()
            Tcp_ClientCredentialTypeTests.TcpClientCredentialType_Certificate_EchoString [FAIL]
               System.ServiceModel.EndpointNotFoundException : There was no endpoint listening at net.tcp://localhost:809//TcpTransportSecuritySslClientCredentialTypeCertificate.svc//tcp-server-ssl-security-clientcredentialtype-certificate that could accept the message. This is often caused by an incorrect address or SOAP action. See InnerException, if present, for more details.
               Stack Trace:
                  c:\OSS\wcf\src\System.Private.ServiceModel\src\System\ServiceModel\Channels\FramingChannels.cs(469,0): at System.ServiceModel.Channels.ConnectionUpgradeHelper.<DecodeFramingFaultAsync>d__0.MoveNext()
```

mconnew:
Fixes #833 
Fixes #1057 
Also fixed the test service configuration for one of the NetTcp client credential endpoints.


mconnew:

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs.orig
```diff
@@ -0,0 +1,328 @@
...
+            factory.Close();
+        }
+        finally
+        {
+            // *** ENSURE CLEANUP *** \\
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+        }
+    }
+
+<<<<<<< 1624dfdcd3339492aecdeda3c8ecabc2f184d45d
```
iamjasonp: I think you missed a three-way merge somewhere :smile:

iamjasonp: ... or at least a merge conflict

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs.orig
```diff
@@ -0,0 +1,328 @@
...
+    [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed))]
+#if FEATURE_NETNATIVE
+    [ActiveIssue(833)] // Not supported in NET Native
+#endif
+=======
+    [Fact]
+>>>>>>> Implements Ssl security upgrade for NetTcp for UWP apps
```
iamjasonp: Ditto

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs.orig
roncain: This is a cs.orig file -- probably it was intended to be deleted before committing?  Renamed?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
```diff
@@ -190,7 +190,7 @@ protected override void ShutdownCore(TimeSpan timeout)
             catch (ObjectDisposedException objectDisposedException)
             {
                 Exception exceptionToThrow = ConvertObjectDisposedException(objectDisposedException, TransferOperation.Undefined);
-                if (object.ReferenceEquals(exceptionToThrow, objectDisposedException))
```
roncain: Questions: Why remove object from ReferenceEquals?  Did an automated tool do that?  CoreFx uses object.ReferenceEquals.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
```diff
@@ -258,6 +259,10 @@ protected override void ShutdownCore(TimeSpan timeout)
                     throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(exceptionToThrow);
                 }
             }
+            catch (OperationCanceledException)
+            {
+                throw;
```
roncain: Question: why is this new catch and rethrow necessary?  Bug fix or change in the code paths allowed by other changes in this PR? Should it AbortWrite()?  Not blocking -- just questions.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
```diff
@@ -497,6 +505,10 @@ protected override void TraceSocketReadStop(int bytesRead, bool async)
...
```
roncain: Same questions about why this is needed and whether we need to AbortWrite().  Not blocking, just questions.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -303,139 +284,6 @@ protected override void OnEndOpen(IAsyncResult result)
         }
     }
 
-    internal class SslStreamSecurityUpgradeAcceptor : StreamSecurityUpgradeAcceptorBase
```
roncain: Not blocking -- it would be helpful if you could comment in this PR why some of these types are being deleted.  It's a little difficult to grock the whole picture from a PR.

src/System.Private.ServiceModel/tools/IISHostedWcfService/Web.config
```diff
@@ -63,6 +63,7 @@
         <add factory="WcfService.TcpTransportSecuritySslCustomCertValidationTestServiceHostFactory" service="WcfService.WcfService" relativeAddress="~\TcpTransportSecuritySslCustomCertValidation.svc" />
         <add factory="WcfService.TcpTransportSecurityStreamedTestServiceHostFactory" service="WcfService.WcfService" relativeAddress="~\TcpTransportSecurityStreamed.svc" />
         <add factory="WcfService.TcpTransportSecurityWithSslTestServiceHostFactory" service="WcfService.WcfService" relativeAddress="~\TcpTransportSecurityWithSsl.svc" />
+        <add factory="WcfService.TcpTransportSecuritySslClientCredentialTypeCertificateFactory" service="WcfService.WcfService" relativeAddress="~\TcpTransportSecuritySslClientCredentialTypeCertificate.svc" />
```
roncain: I think name no longer matches after you renamed the types.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -566,16 +414,210 @@ internal override void Close(TimeSpan timeout)
...
+            if (_clientToken != null)
+            {
+                foreach (var extension in _clientToken.Certificate.Extensions)
+                {
+                    var attachmentExtension =
+                        extension as X509CertificateInitiatorClientCredential.X509UwpCertificateAttachmentExtension;
+                    if (attachmentExtension != null && attachmentExtension.AttachedCertificate != null)
+                    {
+                        clientCertificate = attachmentExtension.AttachedCertificate;
+                        break;
```
roncain: Question: can there be only one certificate that matches these conditions?  What happens if there are more?  Is the first the right one?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -566,16 +414,210 @@ internal override void Close(TimeSpan timeout)
...
+                    }
+                }
+            }
+
+            Contract.Assert(clientCertificate != null, "Missing UWP Certificate as an attachment to X509Certificate2");
```
roncain: Should this instead be a throw?  Or is it unconditionally impossible for this situation to occur?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -566,16 +414,210 @@ internal override void Close(TimeSpan timeout)
...
+                // lives in the callback for CoreClr but needs to be pulled into this method for UWP.
+                EndpointAddress remoteAddress = RemoteAddress;
+                if (remoteAddress.Identity == null && remoteAddress.Uri != Via)
+                {
+                    remoteAddress = new EndpointAddress(Via);
+                }
+                EndpointIdentity identity;
+                if (!_parent.IdentityVerifier.TryGetIdentity(remoteAddress, out identity))
+                {
+                    SecurityTraceRecordHelper.TraceIdentityVerificationFailure(identity, null, GetType());
```
roncain: Nit: could used named parameter pattern here so we know what these parameters mean (ex: the null).  Not blocking.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -566,16 +414,210 @@ internal override void Close(TimeSpan timeout)
...
+                // Verify that we didn't negotiate a protocol lower than the binding configuration specified. No need to check Tls12 
+                // as it will only be negotiated if Tls12 was actually specified. 
+                var negotiatedProtectionLevel = rtStreamSocket.Information.ProtectionLevel;
+                if ((negotiatedProtectionLevel == SocketProtectionLevel.Tls11 && (_parent.SslProtocols & SslProtocols.Tls11) == SslProtocols.None) ||
+                    (negotiatedProtectionLevel == SocketProtectionLevel.Tls10 && (_parent.SslProtocols & SslProtocols.Tls) == SslProtocols.None))
+                {
+                    // Need to dispose StreamSocket as normally SslStream wouldn't end up in a usable state in this situation. As
+                    // post-upgrade validation is required in UWP, the connection needs to be Dispose'd to ensure it isn't used.
+                    rtStreamSocket.Dispose();
+                    throw new SecurityNegotiationException(negotiatedProtectionLevel.ToString());
```
roncain: Should we introduce a new resource here and produce a more meaningful message?  I know this is UWP and we won't see the resource content, but the resource name might offer better clues to the failure.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
```diff
@@ -167,9 +167,6 @@ public static void SameBinding_SecurityModeTransport_ClientCredentialTypeCertifi
     }
 
     [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed))]
```
roncain: When you run this in ToF (which is of course necessary to verify the fix), you will have to temporarily make this [Fact], because ToF does not understand [ConditionalFact] ad would otherwise skip this test.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
```diff
@@ -217,9 +214,6 @@ public static void TcpClientCredentialType_Certificate_EchoString()
...
```
roncain: Ditto -- needs to be [Fact] when run manually in ToF.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs.orig
mconnew: It got left behind by Beyond Compare I think. I'll remove it.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
mconnew: We still have a lot of this. references in our code so whenever I'm touching any code which has a this. in it, I do the cleanup. I used resharper to remove redundant this. qualifiers and it removed this too. Do you want me to put this object. back in?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
mconnew: This was a bug fix for an issue I found while implementing this. If the sendCts.Token has been cancelled, the writeTask will throw an OperationCanceledException when the result is fetched. This was getting caught by the general Exception catch block presuming it was an exception from the UWP StreamSocket class. If this does get thrown, abortWrite will not get set to false so in the finally block AbortWrite() will get called.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
mconnew: UpgradeAcceptor's are used by Listener's, which we don't have in this code base, we only have client channel so only need upgrade initiators.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
mconnew: In the ClientCertificate.SetCertificate method, we throw if we don't find exactly 1 certificate.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
mconnew: For UWP apps, we don't allow directly setting an X509Certificate2, the only allowed way is using the SetCertificate method to fetch it from the store. If an X509Certificate2 is found, then we expect a UWP Certificate to also be found and throw if that's not the case. So the only way we can get an X509Certificate2 is if an UWP Certificate has also been found and attached.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
mconnew: The failure here is we were unable to negotiate a common SSL protocol. Any suggestions on verbage?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
roncain: We should not use resharper to fix up the code.  The way we have done this in the past (per CoreFx recommendations) is https://github.com/dotnet/codeformatter/releases .  It was invented to make all the dotnet code follow the same conventions.  It might cause other changes in other files, which we wouldn't want in this PR.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
mconnew: Already done. The ConditionalFact's were added after I had already verified the fix.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
mconnew: I'll create a resharper coding style config which matches that. As an aside, do you think that would be worth checking in? Resharper supports per solution config files so it would mean anyone loading our solution would get the same coding style automatically. The codeformatter tool is a lot more work to use on an everyday basis, but also does a lot more such as the copyrights.

roncain: I don't think we want to introduce a 2nd way to reformat code, though I see your point about being lighter weight.  Not all of us use Resharper, so maybe we should chat separately about that.  I do think we should avoid any changes that CodeFormatter will immediately reverse, or the next person will take the hit.  Not blocking for merge if cleaning up post-merge makes more sense.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
roncain: For UWP, the name of the resource itself should also be informative.  Maybe call the resource SSL_Protocol_Negotiation_Failed and make the resource contain "SSL protocol negotiation failed.  Requested '{0}' but was offered '{1}'." -- or something like that?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/RTSocketConnection.cs
mconnew: I don't think CodeFormatter would revert the removal of object. I think we should do a round of running CodeFormatter on our code base as a separate task. It's been a while.

roncain: Agree -- let's do a CodeFormatter pass over everything in a different PR.

roncain:
LGTM, subject to the tests passing against NET Native in ToF (which you said you verified).

roncain:
@dotnet-bot test outerloop windows_nt

mconnew:
@dotnet-bot test outerloop windows_nt

mconnew:

PR #1094, opened 2016-04-27
Add new custom binding test to eliminate lab test failure reports

roncain:
Adds a single new CustomBinding test that uses Http rather than
the existing Https test. Done for 2 reasons:
  1. Slightly better coverage
  2. The existing Https test has [ConditionalFact] and is the only test in
     this project.  When a lab run in Helix or ToF chooses not to run
     this test, they both report failure due to having no tests to run
     in the project.  Having a 2nd test eliminates these spurious error
     reports.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.cs
@@ -12,7 +12,44 @@ using Infrastructure.Common;
 
 public class CustomBindingTests : ConditionalWcfTest
 {
-    // Client and Server bindings setup exactly the same using default settings.
+    // Http: Client and Server bindings setup exactly the same using default settings.
+    [Fact]
+    [OuterLoop]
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.cs
```diff
@@ -12,7 +12,44 @@
 
 public class CustomBindingTests : ConditionalWcfTest
 {
-    // Client and Server bindings setup exactly the same using default settings.
+    // Http: Client and Server bindings setup exactly the same using default settings.
```
roncain: Reviewers: this test is just a duplicate of the Https test that was already here (below).  It differs in using Http rather than Https and therefore does not depend on a [ConditionalFact].

roncain:

roncain:
@hongdai or @StephenBonikowsky -- could one of you review so I can merge this?  Thanks.

StephenBonikowsky:
LGTM

PR #1096, opened 2016-04-27
Add initial [ConditionalFact] detectors

roncain:
Prior to this, the [ConditionalFact] condition tests were
simple heuristics.  This PR introduces runtime detectors that
to detect root and client certificates are installed as well as
whether the client is domain joined.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
@@ -10,7 +10,7 @@ using System.Text;
 
 namespace Infrastructure.Common
 {
-    public static class BridgeClient 
+    internal static class BridgeClient 
     {
         private const string EndpointResourceResponseUriKeyName = "uri";
         private const string EndpointResourceResponseFQHNKeyName = "hostname";
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
@@ -1,0 +1,94 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace Infrastructure.Common
+{
+    // This class contains helper "detector" methods to detect specific
+    // conditions about the environment in which the tests are run.
+    // They are used for [ConditionalFact] and [ConditionalTheory]
+    // tests.  They are called at most once, and the return value cached.
+    internal static class ConditionalTestDetectors
+    {
+        public static bool IsRootCertificateInstalled()
+        {
+            try
+            {
+                ServiceUtilHelper.EnsureRootCertificateInstalled();
+                return true;
+            }
+            catch
+            {
+                return false;
+            }
+        }
+
+        public static bool IsClientCertificateInstalled()
+        {
+            try
+            {
+                ServiceUtilHelper.EnsureLocalClientCertificateInstalled();
+                return true;
+            }
+            catch
+            {
+                return false;
+            }
+        }
+
+        public static bool IsClientDomainJoined()
+        {
+            // Currently non-Windows clients always return false until we have
+            // a cross-platform way to detect this.
+            if (!IsWindows())
+            {
+                return false;
+            }
+
+            // To determine whether the client is running on a domain joined machine,
+            // we use this heuristic based on well known environment variables.
+            string computerName = Environment.GetEnvironmentVariable("COMPUTERNAME");
+            string logonServer = Environment.GetEnvironmentVariable("LOGONSERVER");
+            string userDomain = Environment.GetEnvironmentVariable("USERDOMAIN");
+
+            return !String.Equals(computerName, logonServer, StringComparison.OrdinalIgnoreCase) &&
+                   !String.Equals(computerName, userDomain, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public static bool IsServerDomainJoined()
+        {
+            // Requires solution to https://github.com/dotnet/wcf/issues/1095
+            // Till then, heuristic is that running localhost == domain joined if client is domain joined
+            return IsServerLocalHost() && IsClientDomainJoined();
+        }
+
+        public static bool IsWindows()
+        {
+            return RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+        }
+
+        // This test is meant to be used only to detect that our service URI
+        // indicates localhost.  It is not intended to be used to detect that
+        // the services are running locally but using an explicit host name.
+        public static bool IsServerLocalHost()
+        {
+            string host = TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName);
+
+            if (String.IsNullOrWhiteSpace(host))
+            {
+                return false;
+            }
+
+            int index = host.IndexOf("localhost", 0, StringComparison.OrdinalIgnoreCase);
+            return index == 0;
+        }
+
+        public static bool AreUserNameAndPasswordAvailable()
+        {
+            return !String.IsNullOrWhiteSpace(TestProperties.GetProperty(TestProperties.TestUserName_PropertyName)) &&
+                   !String.IsNullOrWhiteSpace(TestProperties.GetProperty(TestProperties.TestPassword_PropertyName));
+        }
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.InteropServices;
 
 namespace Infrastructure.Common
 {
@@ -76,58 +75,41 @@ namespace Infrastructure.Common
             }
         }
 
-        private static bool Is_Server_Local()
+        private static bool Server_Is_LocalHost()
         {
-            // Temporary workaround to detect whether this test is using localhost
-            // for either new or old test infrastructure.
-            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
-            // to use centralized helper method.
-            string host = null;
-            if (TestProperties.PropertyNames.Contains("ServiceUri"))
-            {
-                host = TestProperties.GetProperty("ServiceUri");
-            }
-            else if (TestProperties.PropertyNames.Contains("BridgeHost"))
-            {
-                host = TestProperties.GetProperty("BridgeHost");
-            }
-
-            if (String.IsNullOrWhiteSpace(host))
-            {
-                return false;
-            }
-
-            int index = host.IndexOf("localhost", 0, StringComparison.OrdinalIgnoreCase);
-            return index >= 0;
+            return GetConditionValue(nameof(Server_Is_LocalHost),
+                                     ConditionalTestDetectors.IsServerLocalHost);
         }
         
         private static bool Is_Windows()
         {
-            return RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            return GetConditionValue(nameof(Is_Windows),
+                                     ConditionalTestDetectors.IsWindows);
         }
 
         public static bool Domain_Joined()
         {
-            // Temporarily use the simple heuristic that if we are running the services locally, it is.
-            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
             return GetConditionValue(nameof(Domain_Joined),
-                                     Is_Server_Local);
+                                     () => ConditionalTestDetectors.IsClientDomainJoined() && 
+                                           ConditionalTestDetectors.IsServerDomainJoined());
         }
 
         public static bool Root_Certificate_Installed()
         {
-            // Temporarily use the simple heuristic that if we are running the services locally, it is.
-            // Refactor this to test only when solving https://github.com/dotnet/wcf/issues/1024 
             return GetConditionValue(nameof(Root_Certificate_Installed),
-                                     Is_Server_Local);
+                                     ConditionalTestDetectors.IsRootCertificateInstalled);
         }
 
         public static bool Client_Certificate_Installed()
         {
-            // Temporarily use the simple heuristic that if we are running the services locally, it is.
-            // Refactor this to test only when solving https://github.com/dotnet/wcf/issues/1024 
             return GetConditionValue(nameof(Client_Certificate_Installed),
-                                     Is_Server_Local);
+                                     ConditionalTestDetectors.IsClientCertificateInstalled);
+        }
+
+        public static bool UserName_And_Password_Available()
+        {
+            return GetConditionValue(nameof(UserName_And_Password_Available),
+                                     ConditionalTestDetectors.AreUserNameAndPasswordAvailable);
         }
 
         public static bool SPN_Available()
@@ -135,7 +117,7 @@ namespace Infrastructure.Common
             // Temporarily use the simple heuristic that if we are running the services locally, it is.
             // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
             return GetConditionValue(nameof(SPN_Available),
-                                     Is_Server_Local);
+                                     Server_Is_LocalHost);
         }
 
         public static bool Kerberos_Available()
@@ -143,7 +125,31 @@ namespace Infrastructure.Common
             // Temporarily use the simple heuristic that if we are running the services locally, it is.
             // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
             return GetConditionValue(nameof(Kerberos_Available),
-                                     Is_Server_Local);
+                                     Server_Is_LocalHost);
+        }
+
+        public static bool Server_Accepts_Certificates()
+        {
+            // Temporarily use the simple heuristic that if we are running the services locally, it does.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(Server_Accepts_Certificates),
+                                     Server_Is_LocalHost);
+        }
+
+        public static bool Basic_Authentication_Available()
+        {
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(Basic_Authentication_Available),
+                                     Server_Is_LocalHost);
+        }
+
+        public static bool Digest_Authentication_Available()
+        {
+            // Temporarily use the simple heuristic that if we are running the services locally, it is.
+            // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
+            return GetConditionValue(nameof(Digest_Authentication_Available),
+                                     Server_Is_LocalHost);
         }
 
         public static bool Windows_Authentication_Available()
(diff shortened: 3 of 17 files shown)


src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
```diff
@@ -0,0 +1,105 @@
...
+    // They are used for [ConditionalFact] and [ConditionalTheory]
+    // tests.  They are called at most once, and the return value cached.
+    internal static class ConditionalTestDetectors
+    {
+        private const string ClientCertificateSubject = "WCF Client Certificate";
+        private const string CertificateIssuer = "DO_NOT_TRUST_WcfBridgeRootCA";
+
+        public static bool IsRootCertificateInstalled()
+        {
+            return IsCertificateInStore(StoreName.Root, StoreLocation.LocalMachine, CertificateIssuer) ||
```
roncain: Reviewers: this is first cut, and feedback is welcome.  Right now the detector only detects and does not attempt to install the certificates.  We also have a duplication of issue name literals that we should find a way to share.  These changes are not expected to light up ToF or Helix tests currently using [ConditionalFact] on the new detectors.  It is necessary to find ways to install the certificates prior to running those tests.

roncain:

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
```diff
@@ -0,0 +1,105 @@
...
+                   IsCertificateInStore(StoreName.Root, StoreLocation.CurrentUser, CertificateIssuer);
+        }
+
+        public static bool IsClientCertificateInstalled()
+        {
+            return IsCertificateInStore(StoreName.My, StoreLocation.LocalMachine, CertificateIssuer) ||
+                   IsCertificateInStore(StoreName.My, StoreLocation.CurrentUser, CertificateIssuer);
+        }
```
hongdai: Need to use client certificate subject to find out whether client cert is installed.

roncain: Agreed -- I added this to commit 3

roncain:

roncain:
The OuterLoop failures correspond to tests marked with [ConditionalFact(nameof(Root_Certificate_Installed))].  This is the first time these tests are being run other than self-hosted, because prior to this PR, this [ConditionalFact] returned true only in self-hosted testing.  Now the [ConditionalFact] is actively verifying that a root cert is installed, returns true, and allows these tests to run -- and fail.  So the bug is not in the [ConditionalFact] detector itself, but probably related to the certs themselves.  The typical failure looks like this:
```
    Tcp_ClientCredentialTypeCertificateCanonicalNameTests.TcpClientCredentialType_Certificate_With_Only_CanonicalName_EchoString(endpointUri: net.tcp://wcfcoresrv2.cloudapp.net/WcfService11//TcpCertificateWithSubjectCanonicalNameLocalhost.svc//tcp-server-subject-cn-localhost-cert, shouldCallSucceed: False) [FAIL]
        System.ServiceModel.Security.SecurityNegotiationException : The X.509 certificate CN=localhost, O=DO_NOT_TRUST, OU=Created by https://github.com/dotnet/wcf; AD436762A2047B67DFBED1E25419EE20A4A929FF chain building failed. The certificate that was used has a trust chain that cannot be verified. Replace the certificate or change the certificateValidationMode. The revocation function was unable to check revocation for the certificate.
         The revocation function was unable to check revocation because the revocation server was offline.
 
```
@Zhenlan, @hongdai , @iamjasonp -- to make progress, I think we may need to mark these tests with [ActiveIssue] to proceed with this PR.  Thoughts?

roncain:

hongdai:
Make sense to mark those tests with ActiveIssue as the failures are not related to the detector.

roncain:

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            return !String.Equals(computerName, logonServer, StringComparison.OrdinalIgnoreCase) &&
+                   !String.Equals(computerName, userDomain, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public static bool IsServerDomainJoined()
+        {
+            // Requires solution to https://github.com/dotnet/wcf/issues/1095
+            // Till then, heuristic is that running localhost == domain joined
+            return IsServerLocalHost();
```
zhenlan: Should the logic be `IsServerLocalHost() && IsClientDomainJoined()`?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            string host = TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName);
+
+            if (String.IsNullOrWhiteSpace(host))
+            {
+                return false;
+            }
+
+            int index = host.IndexOf("localhost", 0, StringComparison.OrdinalIgnoreCase);
+            return index >= 0;
```
zhenlan: NIT: technically, we only expect `index==0` at the moment.

zhenlan:
I understand we have test failures if this is merged (so this has to wait a little bit). I will comment on related issues #1105 and #1108. This PR, however, LGTM (other than a couple of comments I had in CR) for us to make incremental improvement.

zhenlan:
@dotnet-bot test outerloop selfhost Windows_NT please

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
roncain: Good idea -- will do.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
roncain: Okay -- will check 0 only.

roncain:

roncain:

roncain:

roncain:
Note that the self-hosted OuterLoops pass because all the tests marked with [ConditionalFact(nameof(Client_Certificate_Installed))] are being skipped.  Either the detector logic is failing, or certificate generation really did fail in the lab.  I opened #1112 to enhance our logs so we can know what happened.

roncain:
Please do not merge until we investigate and fix #1108 and #1111, or our normal OuterLoops will fail.

roncain:
Reviewers -- apologies this PR contains a mixed bag of changes,  It turned out that some refactoring was needed to make the certificate installation code available to the detectors.  This also uncovered several dangling dependencies directly on the Bridge.  I'll squash before merging, but this collection of changes really needs to go into a single PR.

roncain:

roncain:
This PR has been updated as per the discussion about adding new [ConditionalFact]s and making the cert detectors become cert installers.  I won't merge till we get a CI pass, but I would appreciate review comments anytime.

/cc @zhenlan @hongdai @iamjasonp @StephenBonikowsky @mconnew

roncain:

roncain:

roncain:

roncain:
Reviewers: I would like to merge this as soon as possible because my other PR's will be based on it.
It passed Windows and self-host OuterLoops last night.  Jenkins had a problem today, but if we get a clean OuterLoop pass today, I would like to consider merging.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/BridgeClient.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Infrastructure.Common
 {
-    public static class BridgeClient 
+    internal static class BridgeClient
```
roncain: Reviewers: I made these internal as the 1st step in deprecating/renaming them.  We had tests that took direct dependencies on these types, and 'internal' caught all those.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
```diff
@@ -0,0 +1,101 @@
...
+
+namespace Infrastructure.Common
+{
+    // This class contains helper "detector" methods to detect specific
+    // conditions about the environment in which the tests are run.
+    // They are used for [ConditionalFact] and [ConditionalTheory]
+    // tests.  They are called at most once, and the return value cached.
+    internal static class ConditionalTestDetectors
+    {
+        private const string ClientCertificateSubject = "WCF Client Certificate";
```
roncain: Reviews: I will remove these 2 lines now that we don't use them in this code.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
```diff
@@ -0,0 +1,101 @@
...
+        public static bool IsRootCertificateInstalled()
+        {
+            try
+            {
+                ServiceUtilHelper.EnsureRootCertificateInstalled();
+                return true;
+            }
+            catch
+            {
+                return false;
```
roncain: Reviewers: I will investigate independently how we can communicate failures here back to xunit, but I would prefer not to block this PR on that investigation.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalWcfTest.cs
```diff
@@ -76,90 +75,97 @@ private static bool GetConditionValue(string conditionName, Func<bool> detectFun
...
         {
             // Temporarily use the simple heuristic that if we are running the services locally, it is.
             // Refactor this after integration to address https://github.com/dotnet/wcf/issues/1024 
             return GetConditionValue(nameof(Kerberos_Available),
-                                     Is_Server_Local);
+                                     Server_Is_LocalHost);
+        }
+
+        public static bool Server_Accepts_Certificates()
+        {
```
roncain: Reviewers: as we discussed @hongdai will investigate and provide the complete list for conditions like Basic, Digest, website-allows-certificates, etc.  For now I added the bare minimum I knew so we could mark the tests that currently fail in CI running in the lab.  I will add the others or rename these in later PR's.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/IUtil.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
roncain: Reviewers: to give some context, the IUtil interface was embedded in in ServiceInterfaces available only in the Common\Scenarios project (which depends on this Common\Infrastructure project, which would create a circular dependency).  To follow the pattern we found worked best for the Bridge "helper" service interactions, I have moved this interface as well as the certificate and endpoint helper methods into this Common\Infrastructure project.  They have not been modified -- only rehomed.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ServiceUtilHelper.cs
```diff
@@ -0,0 +1,202 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
roncain: Reviewers: the code here used to be in Endpoints.cs, which wasn't really the right home, and on which this project could not depend.  It has been moved verbatim (but renamed) from Endpoints.cs into this project.  Comments welcome, of course, but all the code in this class already existed.

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -9,191 +9,41 @@
 
 public static partial class Endpoints
 {
-    private static X509Certificate2 rootCert;
```
roncain: Reviewers: as mentioned elsewhere, this code moved verbatim into ServiceUtilHelper.cs

src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
```diff
@@ -306,88 +156,79 @@ public static string Https_BasicAuth_Address
...
-            return GetEndpointAdddress("HttpDigestNoDomain.svc//");
+            return GetEndpointAddress("HttpDigestNoDomain.svc//");
         }
     }
 
     public static string Https_NtlmAuth_Address
     {
         get
         {
-            EnsureLocalClientCertifciateInstalled();
```
roncain: Reviewers: because the [ConditionalFact] detector now calls precisely this method, I believe the right design is to remove these explicit calls from Endpoints.cs.  They would be NOP's and add some confusion if we left them here.

src/System.Private.ServiceModel/tests/Common/Scenarios/ServiceInterfaces.cs
```diff
@@ -506,15 +506,3 @@ public interface IWcfAspNetCompatibleService
     string EchoTimeAndSetCookie(string name);
 }
 
-[ServiceContract]
```
roncain: Reviewers: this moved verbatim into IUtil.cs in the Common\Infrastructure project.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Http/ClientCredentialTypeTests.cs
```diff
@@ -26,7 +26,10 @@ public static void DigestAuthentication_Echo_RoundTrips_String_No_Domain()
             binding = new BasicHttpBinding(BasicHttpSecurityMode.TransportCredentialOnly);
             binding.Security.Transport.ClientCredentialType = HttpClientCredentialType.Digest;
             factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.Http_DigestAuth_NoDomain_Address));
-            factory.Credentials.HttpDigest.ClientCredential = BridgeClientAuthenticationManager.NetworkCredential;
+
+            // https://github.com/dotnet/wcf/issues/1045 needs to supply a replacement for NetworkCredential below
```
roncain: Reviewers: this test is marked with [ActiveIssue(1045)] so was not executing.  Now we know why it failed -- it was trying to invoke Bridge logic.  Issue 1045 will be responsible for providing a new way to do this.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
```diff
@@ -19,7 +19,7 @@ static Https_ClientCredentialTypeTests()
         s_password = "wcfSaysHell0World!";
     }
 
-    [ConditionalFact(nameof(Root_Certificate_Installed))]
+    [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Basic_Authentication_Available))]
```
roncain: Reviewers: I added these new conditions only to tests known failing in CI.  This allows them to run in self-host today.  After @hondai investigates what web-site configuration pivots we need, we'll make the detectors for these real.

roncain:

roncain:
Reviewers: the Ubuntu InnerLoop failure was a package restore issue unrelated to the changes in the PR.  It has passed on other CI test requests.

StephenBonikowsky:
LGTM

roncain:

roncain:
Thanks @StephenBonikowsky .  I squashed and merged.  I'll still accept post-merge feedback.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
zhenlan: Do BasicAuthentication_* tests really need root certificate?

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
```diff
@@ -19,6 +19,7 @@ public class HttpsTests : ConditionalWcfTest
     // Server: BasicHttpsBinding default value is Soap11
     [ConditionalFact(nameof(Root_Certificate_Installed))]
     [OuterLoop]
+    [ActiveIssue(1123, PlatformID.AnyUnix)]
```
zhenlan: This should not be an ActiveIssue. The product is okay. It's the setup condition is not met that causes the test failure on AnyUnix. This test should be disabled on AnyUnix by stronger condition.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
```diff
@@ -0,0 +1,94 @@
...
+    // They are used for [ConditionalFact] and [ConditionalTheory]
+    // tests.  They are called at most once, and the return value cached.
+    internal static class ConditionalTestDetectors
+    {
+        public static bool IsRootCertificateInstalled()
+        {
+            try
+            {
+                ServiceUtilHelper.EnsureRootCertificateInstalled();
+                return true;
```
zhenlan: I know we called `EnsureRootCertificateInstalled()` here, but I would like us to still keep the detection logic as you had in your previous PR. Let the detection tell if the installation is successful or not instead of making assumptions. In the future, in case we want to factor out installation logic at other places, it will be very easy too.

zhenlan:
Feels like all the tests that are disabled by `[ActiveIssue(1123, PlatformID.AnyUnix)]` should be disabled by stronger condition instead.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
roncain: Agree, but I would like @iamjasonp  to recommend a name for the condition and a proposed implementation for the detector.  Presumably it failed to write to the certificate store because it has not been configured as per https://github.com/dotnet/wcf/blob/master/Documentation/manualtest-guide.md .

If this is correct, I could make a condition "Certificate_Store_Configured" or something like that.  But how would the detector work?  And why wouldn't we add that to every single test requiring a root cert.  Thoughts welcome.

/cc @iamjasonp

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
roncain: Good question.  The initial population of [ConditionalFact] across tests was based on which tests used endpoints that had an embedded EnsureRootCertificateInstalled().  Meaning that if the tests said they needed it via their endpoint's attempt to install the root cert, then we needed a matching [ConditionalFact].  It could mean the original endpoint was over-constrained.  I'll experiment with removing the root-cert condition.
/cc @iamjasonp

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ConditionalTestDetectors.cs
roncain: I'm not sure I totally agree but let's discuss.  If EnsureRootCertificateInstalled() said it succeeded, you have to trust it -- that's its job.  Or add code to it to double check (though I'm not sure what additional work it would do.  I don't think the detectors should second guess when EnsureRootCertificateInstalled said it succeeded.  The only place where it might make sense is in the exception path.  Maybe EnsureRootCertificateInstalled failed to write but there is a usable cert already there.  But even in that case, I believe that job belongs to EnsureRootCertificateInstalled, not the detector.

PR #1104, opened 2016-04-30
Server setup script fix

zhenlan:
Fix error handling in server setup script
- Fixed `:Failure` that is completely messed up when `%_cmd%` contains quotes or parenthesis
- Redirect both standard output and standard error to log so error information can be displayed properly in failure case

Rename WcfTestServerSetup.cmd to SetupWcfIISHostedService.cmd so it follows the same "DoWhat" naming convention as the rest of scripts.


roncain:
LGTM.
Non-blocking nit observation: I'd prefer a name like SetupWcfTestServer.cmd, because it is intended to be used only for that purpose.  The new name would be perfect for the script I would like to see -- one that configures IIS on my personal machine using my current repo for the vdir.

zhenlan:
Yep, it's forwarding looking renaming 😄

PR #1107, opened 2016-05-02
Add port80 to firewall because crl service need it

hongdai:
* This is because there is only one server certificate created per machine.
To run both selfhost and IIS hosted server on the same machine, we need to
use the same port.


hongdai:

roncain:
LGTM

zhenlan:
It's confusing to see `_WCF Test Server PortHttp80` and `_WCF Test Server PortHttp` at the same time 😄. Hope we can consolidate ports for self-hosted services (with IIS hosted services) in the default case one day.

Issue #1108, opened 2016-05-02
NetTcp TransportSecurity tests fail cross-machine


roncain:
The NetTcp tests using TransportSecurity pass when self-hosted and run locally but fail in cross-machine use, including CI.   This is a manual Windows-to-Windows Self-hosted cross machine test involving domain-joined machines.  It does not occur yet in CI because the ConditionalFact detectors skip these tests.  By setting environment variables, we force them to run.

Repro steps:

1. Start the self-hosted WCF services on machine A (domain-joined)

2. On machine B running Windows and domain-joined, do these steps
 - Set these environment variables:
```
    set Server_Accepts_Certificates=true
    set Domain_Joined=true
    set NTLM_Available=true
    set Basic_Authentication_Available=true
    set Digest_Authentication_Available=true
    set Windows_Authentication_Available=true
    set ExplicitUserName=valid-domain-user-name
    set ExplicitPassword=valid-domain-user-password
```
 - Run the OuterLoops on machine B

The failing tests are:
```
Tcp_ClientCredentialTypeTests.SameBinding_DefaultSettings_EchoString
Tcp_ClientCredentialTypeTests.SameBinding_SecurityModeTransport_EchoString
StreamingTests.NetTcp_TransportSecurity_StreamedRequest_RoundTrips_String
StreamingTests.NetTcp_TransportSecurity_Streamed_RoundTrips_String_WithSingleThreadedSyncContext
StreamingTests.NetTcp_TransportSecurity_Streamed_RoundTrips_String
StreamingTests.NetTcp_TransportSecurity_Streamed_Async_RoundTrips_String
StreamingTests.NetTcp_TransportSecurity_StreamedRequest_Async_RoundTrips_String
StreamingTests.NetTcp_TransportSecurity_StreamedResponse_Async_RoundTrips_String
StreamingTests.NetTcp_TransportSecurity_StreamedResponse_RoundTrips_String
StreamingTests.NetTcp_TransportSecurity_Streamed_TimeOut_Long_Running_Operation
StreamingTests.NetTcp_TransportSecurity_Streamed_Async_RoundTrips_String_WithSingleThreadedSyncContext
```

Failure usually looks like this on the CI machines:
```
     Tcp_ClientCredentialTypeTests.SameBinding_DefaultSettings_EchoString [FAIL]
        System.ServiceModel.Security.SecurityNegotiationException : Authentication failed on the remote side (the stream might still be available for additional authentication attempts).
[...]

PR #1109, opened 2016-05-02
Disable NetTcp tests requiring TransportSecurity

roncain:
After moving to the new test infrastructure, tests for NetTcp with
TransportSecurity failed. We initially suppressed them with a
[ConditionalFact] that prevented them from running cross-machine.
But in attempting to make the [ConditionalFact] actually detect
the presence of certificates, we discovered these tests fail when
run in CI or self-hosted cross-machine

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeTests.cs
@@ -17,6 +17,7 @@ public class Tcp_ClientCredentialTypeTests : ConditionalWcfTest
     //                         - ClientCredentialType = Windows
     [ConditionalFact(nameof(Root_Certificate_Installed))]
 #if !FEATURE_NETNATIVE
+    [ActiveIssue(1108)] // Does not work cross-machine
     [ActiveIssue(592, PlatformID.AnyUnix)] // NegotiateStream works on Windows but is not yet supported on Unix
 #else
     [ActiveIssue(832)] // Windows Stream Security is not supported in NET Native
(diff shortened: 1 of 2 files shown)


roncain:

zhenlan:
Thanks @roncain for putting up a PR for this.

I can see all relevant tests are already disabled on all other platforms but Windows, so technically we should scope the new ActiveIssue to Windows only. However, I do expect these tests to pass on Windows, so it's either a client/server setup issue or current conditions used in ConditonalFact are not sufficient. I understand this is to unblock PR #1096, but #1096 will not buy us much if these tests continue to fail. I feel it's probably the time for us to try to enable these tests or at least to understand better what is going on here before disabling them.

Just to add another data point, some of these tests pass on self-host CI job. We will lose coverage by disabling all tests like what in this PR.

roncain:
@Zhenlan -- I agree we should investigate asap rather than disable.  Issue #1108 has repro steps that will force the certificate detectors to return 'true', meaning we can ask these tests to run even before PR #1096 is merged.  I also agree #1096 can wait.  It's importance is to light-up the tests requiring certs so we can find issues -- and I apparently it did that.

If we do disable them, however, I'd prefer not to remove the *nix-only [ActiveIssue] because that truly is a different issue, and we'd need to put it back as soon as we fixed the issue under discussion.

zhenlan:
Sounds good. Let's focus on resolving #1108 first.

> If we do disable them, however, I'd prefer not to remove the *nix-only [ActiveIssue] because that truly is a different issue

Right, we shouldn't remove #nix-only [ActiveIssue]. If we disable these tests, we should add new ActiveIssue that is limited to Windows only.

roncain:
Please do not merge now.  Once #1108 has been resolved, this PR can be discarded.

roncain:
I am going to discard this PR.  PR #1096 added [ConditionalFact] to these tests using a condition currently satisfied only on selfhost/localhost.  They are skipped otherwise.  If we find they fail when we allow their [ConditionalFact] to be true on cross-machine tests, we can open a new issue.

PR #1110, opened 2016-05-02
Update package publishing for azure

StephenBonikowsky:
These changes support a few different scenarios, the primary scenario I am trying to validate right now with help from @chcosta is building and running our tests against the WCF product "packages" just created in a local build.


StephenBonikowsky:
I'd like to be able to merge this PR even while I am working on using the hooks added by this PR. Reason being that I have additional changes waiting in the wings to PR after this is in and don't want to block.

BuildToolsVersion.txt
```diff
@@ -1 +1 @@
-1.0.25-prerelease-00330-01
```
roncain: CodeFlow shows there is some extra whitespace (newline) in our version.  I recommend copying from CoreFx directly.  Not blocking.

roncain:
LGTM.  Tiny white-space nit comment not required to merge.  It would be good to test build in TFS, but currently the TFS is not building for me.  If you merge @StephenBonikowsky I recommend watching the rolling TFS builds.

BuildToolsVersion.txt
StephenBonikowsky: Looks like I shouldn't even be editing this file, already up-to-date.

PR #1116, opened 2016-05-03
Reduce the number of times the Timer is modified

### LINKED ISSUE #1074: Streamline the request timeout handling
Profiles show up to 54% of CPU time spent in the call to TimerQueueTimer.Change(). This is thanks to WCF changing the timer timeout on each async call. Having just 1 timeout for the end of the entire operation and potentially even one timeout for multiple channel calls should greatly help with the time spent managing the timers.

mconnew:
We are seeing high contention on the call to Timer.Change which takes a lock to modify a global priority queue of timers. By reducing the number of calls to Timer.Change, we increase throughput and won't interfere with other code running in the process which also uses Timers.
Also explicitly dispose Timer to remove from finalization queue.
Fixes #1074

### DIFF
--- a/src/System.Private.ServiceModel/src/Internals/System/Runtime/IOTimer.cs
+++ b/src/System.Private.ServiceModel/src/Internals/System/Runtime/IOTimer.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading;
+
+namespace System.Runtime
+{
+    public class IOTimer<TState> : IDisposable
+    {
+        private readonly Action<TState> _callback;
+        private readonly TState _state;
+        private Timer _timer;
+        private bool _enabled;
+        private DateTime _dueDateTime;
+        private readonly double MaxSkew = 100.0; // Milliseconds
+
+        public IOTimer(Action<TState> callback, TState state)
+        {
+            _callback = callback;
+            _state = state;
+            _enabled = false;
+            _dueDateTime = DateTime.MinValue;
+            // Creating a closure of the instance method OnTimer. This is to avoid a cast which
+            // includes checking if the cast is compatible when the timer fires.
+            _timer = new Timer(OnTimer, null, Timeout.Infinite, Timeout.Infinite);
+        }
+
+        public void Cancel()
+        {
+            // As this is an IOTimer, it is likely that the timer will either be reused or rescheduled
+            // before the timer would actually fire. By disabling the callback from running, the timer
+            // is effectively cancelled without the cost of changing the timer schedule in the timer manager.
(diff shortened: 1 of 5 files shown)


mconnew:

src/System.Private.ServiceModel/src/Internals/System/Runtime/IOTimer.cs
```diff
@@ -0,0 +1,67 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading;
+
+namespace System.Runtime
+{
+    public class IOTimer<TState> : IDisposable
```
roncain: Should we make this internal?  It's in the internals folder.

src/System.Private.ServiceModel/src/Internals/System/Runtime/IOTimer.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        private readonly Action<TState> _callback;
+        private readonly TState _state;
+        private Timer _timer;
+        private bool _enabled;
+        private DateTime _dueDateTime;
+        private readonly double MaxSkew = 100.0; // Milliseconds
```
roncain: Nit: Should this be cased like the other private fields?   Like _maxSkewMs?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CoreClrSocketConnection.cs
```diff
@@ -585,7 +592,7 @@ protected override int ReadCore(byte[] buffer, int offset, int size, TimeSpan ti
             catch (ObjectDisposedException objectDisposedException)
             {
                 Exception exceptionToThrow = ConvertObjectDisposedException(objectDisposedException, TransferOperation.Read);
-                if (object.ReferenceEquals(exceptionToThrow, objectDisposedException))
```
roncain: Nit: For future consideration, since we decided to do an all-up CodeFormatter pass, I'd prefer not to include Resharper cleanup in the PRs.  It contributes to the noise to review.

roncain:
LGTM.
I made some comments for consideration but none blocking.

src/System.Private.ServiceModel/src/Internals/System/Runtime/IOTimer.cs
mconnew: It was supposed to be a constant which would make the name correct. I'll fix that.

src/System.Private.ServiceModel/src/Internals/System/Runtime/IOTimer.cs
mconnew: agreed

mconnew:

PR #1118, opened 2016-05-04
Don't need ERRORLEVEL to be set/used by setup/cleanup scripts.

StephenBonikowsky:
* Since we no longer execute setup and cleanup scripts from build.cmd we no longer need !ERRORLEVEL! after the call to msbuild, doesn't hurt to keep it but its out of sync with corefx and I think it makes more sense to change it in WCF since we don't technically need it anymore.


build.cmd
```diff
@@ -116,7 +116,7 @@ goto :AfterBuild
 
 :build
 %_buildprefix% msbuild "%_buildproj%" /nologo /maxcpucount /v:minimal /clp:Summary /nodeReuse:false /flp:v=normal;LogFile="%_buildlog%";Append "/l:BinClashLogger,%_binclashLoggerDll%;LogFile=%_binclashlog%" !unprocessedBuildArgs! %_buildpostfix% %OfficialBuildIdArg%
-set BUILDERRORLEVEL=!ERRORLEVEL!
```
roncain: LGTM
Question only -- this is just sync'ing with CoreFx, right?

StephenBonikowsky: Correct.

PR #1119, opened 2016-05-04
Syncing .sh scripts with corefx

StephenBonikowsky:
* Just bringing them up-to-date.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -55,32 +55,6 @@ check_native_prereqs()
 
 prepare_managed_build()
 {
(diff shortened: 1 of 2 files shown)


StephenBonikowsky:
run-test.sh is also pretty out-of-sync but last I heard we are not using it, is that still correct?

roncain:
LGTM

roncain:
@iamjasonp may have more knowledge about using run-test.sh.  In practice I never use it.

PR #1125, opened 2016-05-06
Clean target

roncain:
Enables the clean target to close the SelfHost WCF service if it is running.

### DIFF
--- a/build.override.targets
+++ b/build.override.targets
@@ -10,16 +10,20 @@
   </PropertyGroup>
   
   <PropertyGroup>
+    <TraversalCleanDependsOn>
+      WcfClean;
+      $(TraversalCleanDependsOn);
+    </TraversalCleanDependsOn>
+  </PropertyGroup>
+  
+  <PropertyGroup>
     <WcfSelfHostSetupLog>$(MSBuildThisFileDirectory)\SelfHostedWcfServiceSetup.log</WcfSelfHostSetupLog>
     <WcfSelfHostCleanupLog>$(MSBuildThisFileDirectory)\SelfHostedWcfServiceCleanup.log</WcfSelfHostCleanupLog>
   </PropertyGroup>
   
   <!--if ServiceUri is set, the service has been already started on another machine-->
   <Target Name="WcfSetup" Condition="($(WithCategories.Contains('OuterLoop')) or '$(OuterLoop)' == 'true') and ('$(ServiceUri)' == '') and $(OSEnvironment) == 'Windows_NT'" >
-  
-    <Delete Files="$(WcfSelfHostSetupLog)" Condition="Exists('$(WcfSelfHostSetupLog)')" />
-    <Delete Files="$(WcfSelfHostCleanupLog)" Condition="Exists('$(WcfSelfHostCleanupLog)')" />
-  
+    <Delete Files="$(WcfSelfHostSetupLog)" Condition="Exists('$(WcfSelfHostSetupLog)')" /> 
     <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\StartWCFSelfHostedSvc.cmd" ContinueOnError="true">
       <Output TaskParameter="ExitCode" PropertyName="WcfSetupErrorCode"/>
     </Exec>
@@ -33,12 +37,22 @@
   </Target>
 
   <Target Name="WcfCleanUp" Condition="$(WcfSetupErrorCode)!='-1' and $(WcfSetupErrorCode)!=''" >
-    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\CleanUpWCFSelfHostedSvc.cmd" ContinueOnError="true">
-    </Exec>
-
     <Message Text="-------------- WCF self-host outerloop cleanup start --------------" Importance="High" />
+    <Delete Files="$(WcfSelfHostCleanupLog)" Condition="Exists('$(WcfSelfHostCleanupLog)')" />
+    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\StopWcfSelfHostedSvc.cmd" ContinueOnError="true" />
     <Message Text="$([System.IO.File]::ReadAllText($(WcfSelfHostCleanupLog)))" Condition="Exists('$(WcfSelfHostCleanupLog)')" Importance="High" />
     <Message Text="-------------- WCF self-host outerloop cleanup end --------------" Importance="High" />
-    
+  </Target>
+  
+  <!-- The WcfClean target executes via clean.cmd or build /t:Clean.
+       Currently it runs only on Windows in order to ensure the SelfHost
+       WCF service has been closed to ensure the bin folder can be deleted.
+  -->
+  <Target Name="WcfClean" Condition="$(OSEnvironment) == 'Windows_NT'">
+    <Message Text="-------------- WCF clean start --------------" Importance="High" />
+    <Delete Files="$(WcfSelfHostCleanupLog)" Condition="Exists('$(WcfSelfHostCleanupLog)')" />
+    <Exec Command="$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\scripts\StopWcfSelfHostedSvc.cmd" ContinueOnError="true" />
+    <Message Text="$([System.IO.File]::ReadAllText($(WcfSelfHostCleanupLog)))" Condition="Exists('$(WcfSelfHostCleanupLog)')" Importance="High" />
+    <Message Text="-------------- WCF clean end --------------" Importance="High" />
   </Target>
 </Project>
--- a/build.proj
+++ b/build.proj
@@ -138,8 +138,8 @@
   <!-- Override RestorePackages from dir.traversal.targets and do a batch restore -->
   <Target Name="RestorePackages" DependsOnTargets="BatchRestorePackages" />
 
-  <!-- Override clean from dir.traversal.targets and just remove the full BinDir -->
-  <Target Name="Clean">
+  <!-- Override CleanAllProjects from dir.traversal.targets and just remove the full BinDir -->
+  <Target Name="CleanAllProjects">
     <RemoveDir Directories="$(BinDir)" />
   </Target>
 
--- a/src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvcDoWork.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvcDoWork.cmd
@@ -4,8 +4,11 @@ setlocal
 REM This script should always be called from StartWCFSelfHostedSvc.cmd
 
 set _setuplog=%~dp0..\..\..\..\SelfHostedWcfServiceSetup.log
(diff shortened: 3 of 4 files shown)


src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvcDoWork.cmd
```diff
@@ -4,9 +4,12 @@ setlocal
 REM This script should always be called from StartWCFSelfHostedSvc.cmd
 
 set _setuplog=%~dp0..\..\..\..\SelfHostedWcfServiceSetup.log
+set _setupMonitorFile=%~dp0..\..\..\..\SelfHostedWcfServiceMonitor.log
```
roncain: Reviews: I'm open to other ideas, but I used a file as a monitor to solve the following problem:  our CMD to start the SelfHost service launches the EXE *and blocks*.  After the EXE exits, the CMD resumes and cleans up.  This means that when hooking up the clean to kill the SelfHost exe, it also allows the startup CMD to resume and finish the cleanup.  This means log files and the bin folder are still in use *after* the SelfHost .exe has been killed.

In this PR, I introduced a short-lived monitor file that is created when the service is started by this script, and deleted when this script has finished cleanup.  That allows the new script to block the clean until this monitor is cleared.

I believe the right fix is probably making an endpoint on the util service to cleanup and shutdown so that we don't have these chained back-to-back CMD files exchanging control.  It makes it much simpler to ask for a clean shutdown in exactly one way.

StephenBonikowsky: For what you are needing to do in the environment we currently have it LGTM, however...

We really should not be chaining scripts together, we have already received feedback from corefx along these lines as well. We should have build.override.targets call each script separately, the scripts can set msbuild properties as needed for additional logic points as we move through the setup and cleanup flow.

src/System.Private.ServiceModel/tools/scripts/StopWcfSelfHostedSvcIfRunning.cmd
```diff
@@ -0,0 +1,67 @@
...
+	echo Warning: An error occurred while killing the SelfHostedWCFService.exe.
+	exit /b 1
+	)
+	
+REM Terminating the Self host WCF process started with StartSelfHostWcfSvc.cmd
+REM results in that script gaining control to finish cleanup.  If we are in that
+REM state, we need to wait for that cleanup to finish, or the bin folder will still
+REM be in use.
+
+for /l %%x in (1, 1, 60) do (
```
zhenlan: Should 60 be a const?

src/System.Private.ServiceModel/tools/scripts/StopWcfSelfHostedSvcIfRunning.cmd
```diff
@@ -0,0 +1,67 @@
...
+  REM The community preferred approach to wait is this.
+  ping 127.0.0.1 -n 6 > nul
+
+echo %%x
+
+)
+
+echo Failed to finish cleanup for Self hosted WCF service
+if EXIST %_setupMonitorFile% del %_setupMonitorFile%
+exit /b -1
```
zhenlan: What does -1 mean? Who is going to use it? Should it be documented somewhere if -1 is necessary?

src/System.Private.ServiceModel/tools/scripts/StopWcfSelfHostedSvcIfRunning.cmd
```diff
@@ -0,0 +1,67 @@
...
+
+set _setupMonitorFile=%~dp0..\..\..\..\SelfHostedWcfServiceMonitor.log
+
+REM Detect if the process is running locally
+tasklist /FI "imagename eq SelfHostedWCFService.exe" | findstr /i /C:"SelfHostedWCFService.exe" 1>nul
+if [%ERRORLEVEL%] == [0] goto stopService
+ 
+REM If it's not running, exit with 0 to indicate nothing was done  
+echo Detected Self hosted WCF Test Service is not running.
+exit /b 0
```
zhenlan: Recommend to follow the common coding style as all existing scripts - all code path has a common exit at the end of script.

zhenlan:
With this PR we will end up with 4 scripts on the same subject

`StopWcfSelfHostedSvc.cmd`
`StopWcfSelfHostedProcess.cmd`
`StopWcfSelfHostedSvcIfRunning.cmd`
`CleanUpWCFSelfHostedSvc.cmd`

I am not sure I will be able to remember what they are for in a week. It probably suggests we should try harder and come up a better solution.

roncain:
@Zhenlan -- are your comments meant to block the PR or suggestions for the future?  It would help to clarify.  I agree there are too many scripts.  In fact, as Wes, @StephenBonikowsky and I have said in common -- the setup and teardown logic should not be in scripts like they are today.  They should be in .exe's or MSBuild targets.  We also have cert cleanup in multiple places.  I would like to return to a common utility .exe as we did for the Bridge that allowed launch, shutdown, cleanup, etc. and handled MSBuild properties. I opened #1131 for this conversion.

roncain:
@Zhenlan and other reviewers -- I deleted the CMD file I added and moved all its logic verbatim into the existing CMD to stop the WCF self-host service.  I also responded to the review comments.  My statement to move away from orchestrated CMD files remains, but at least our count of CMD files remains the same for now.

roncain:

build.proj
```diff
@@ -138,8 +138,8 @@
   <!-- Override RestorePackages from dir.traversal.targets and do a batch restore -->
   <Target Name="RestorePackages" DependsOnTargets="BatchRestorePackages" />
 
-  <!-- Override clean from dir.traversal.targets and just remove the full BinDir -->
-  <Target Name="Clean">
+  <!-- Override CleanAllProjects from dir.traversal.targets and just remove the full BinDir -->
```
roncain: Reviewers: this change was merged into CoreFx first and then copied here verbatim.

roncain:

build.override.targets
```diff
@@ -9,6 +9,13 @@
     </TraversalBuildDependsOn>
   </PropertyGroup>
   
+    <PropertyGroup>
```
zhenlan: NIT: indentation is not aligned

build.override.targets
```diff
@@ -9,6 +9,13 @@
...
+    <TraversalCleanDependsOn>
+      WcfClean;
```
zhenlan: In line 8, we have `WcfCleanup` and we have `WcfClean` here, do we really need 2 different targets for "clean"? Can we have a better solution?

roncain: We do need 2 different targets, because the conditions under which they execute are different (one is triggered only if WcfSetup discovered it started the service, the other is called unconditionally on a Clean).  But I'll make only one of them do the work, and make the other one depend on the worker one when the right conditions are met.

src/System.Private.ServiceModel/tools/scripts/StartWCFSelfHostedSvcDoWork.cmd
```diff
@@ -4,9 +4,12 @@ setlocal
...
 
 echo Preparing to launch the WCF self hosted service
 
+if EXIST %_setupMonitorFile% del %_setupMonitorFile%
```
zhenlan: NIT NIT: naming... "monitor" is a little confusing when I first saw it. This is really a "semaphore" file.

src/System.Private.ServiceModel/tools/scripts/StopWcfSelfHostedSvc.cmd
```diff
@@ -25,10 +38,38 @@ if '%_runelevated%' == '' (
 )
 
 if NOT [%ERRORLEVEL%]==[0] (
-	echo Warning: An error occurred while killing the SelfHostedWCFService.exe. >>%_cleanuplog%
-	exit /b 1
+  _EXITCODE=%ERRORLEVEL%
```
zhenlan: NIT: indentation

src/System.Private.ServiceModel/tools/scripts/StopWcfSelfHostedSvc.cmd
```diff
@@ -25,10 +38,38 @@ if '%_runelevated%' == '' (
...
+    echo Self hosted WCF service cleanup has completed. >>%_cleanuplog%
+    goto done
+  )
+  
+  REM The "proper" way to wait is 'timeout' but that unconditionally
+  REM exits the process with "Input redirection is not supported".
+  REM The community preferred approach to wait is this.
+  ping 127.0.0.1 -n 6 > nul
+
+echo %%x
```
zhenlan: NIT: indentation. btw, it will be nicer to show some message than just a pure number.

src/System.Private.ServiceModel/tools/scripts/StopWcfSelfHostedSvc.cmd
```diff
@@ -25,10 +38,38 @@ if '%_runelevated%' == '' (
...
+
+)
+
+REM Timed out waiting for cleanup script to finish
+echo Failed to finish cleanup for Self hosted WCF service
+set _EXITCODE=1
+
+:done
 
+if EXIST %_setupMonitorFile% del %_setupMonitorFile%
```
zhenlan: If this file still exists at this point, it means something went wrong. This file is not created by this script. Should this script delete it?

src/System.Private.ServiceModel/tools/scripts/StopWcfSelfHostedSvc.cmd
```diff
@@ -25,10 +38,38 @@ if '%_runelevated%' == '' (
 )
```
iamjasonp: Remark only: Wow, this is confusing as I'm seeing StopWcfSelfHostedProcess.cmd vs StopWcfSelfHostedSvc,cmd; both very similarly named. Any way for us to merge that command here?

roncain: I understand, but those are from a different PR and I would prefer to lump those into the discussion about moving to an EXE.  In a nutshell, runElevated.vbs was able to run only CMD files and did not behave correctly attempting to do a TASKKILL alone.  Thus, StopWcfSelfHostedProcess was created to that it could be invoked from StopWcfSelfHostedSvc.  I can rename here if that helps, but having umpteen scripts to try to orchestrate what is conceptually simpler as an EXE seems like throw-away work.

src/System.Private.ServiceModel/tools/scripts/StopWcfSelfHostedSvc.cmd
roncain: Agreed -- will remove.

build.override.targets
roncain: I can't find a way to make this work as a single target.  I would like to keep them separate targets for now.  Details: having one target depend on another makes that target get skipped if the other target ran first.  Hence, 'build /t:Clean;BuildAndTest' executs the WcfClean once only.

StephenBonikowsky:
LGTM
Open issues for unresolved feedback and merge this.

roncain:

iamjasonp:
LGTM

roncain:

roncain:
Merging based on successful outerloop selfhosted

PR #1144, opened 2016-05-11
Move and rename some scenario tests.

StephenBonikowsky:
* Moving a test that have nothing to do with ClientCredentialType out of that test class into appropriate locations.
* Moving some very basic NetTcp tests to a new Tcp project folder under Bindings.
* Renaming tests.
* Remove unnecessary package reference in json file.
* Sort and remove unneeded using statements.


roncain:
LGTM, subject to passing OuterLoop CI.

StephenBonikowsky:
Removed tests from Scenarios\Security\TransportSecurity\Tcp\ClientCredentialTypeTests.cs
   The custom binding didn't actually use ClientCredentialType as the require client credential property on SslStreamSecurityBindingElement is false by default. I moved this test to Binding\Custom\CustomBindingTests.cs
   Three other tests using NetTcpBinding use ClientCredentialType = Windows by default as part of the standard binding, moving them to a new Tcp project under Binding\Tcp. It is debateable where they belong but it seems to me when using the standard binding and not changing any default settings it should go here.

PR #1145, opened 2016-05-11
Updating folders and content for Scenario TransportSecurity tests.

StephenBonikowsky:
* Solution folders were missing and/or did not contain all the files.


StephenBonikowsky:

roncain:
Sorry -- I missed seeing something in #1144 that I should have caught, and it's in this one too.  The SLN file now contains links to .cs files, and it shouldn't.  I'm not sure what caused that (were they added to the SLN manually?), but the solution folders under tests\scenarios should contain only the .csproj files.  The .csproj's are the ones that supply the list of .cs files.  I think it is probably a matter of using VS to open the SLN and removing them (not deleting them) from the SLN folders.

StephenBonikowsky:
I'll try that.

StephenBonikowsky:
Amended with Ron's feedback and rebased to recent merges.

roncain:
LGTM

PR #1147, opened 2016-05-12
Improve debugging of certificate installation

roncain:
Adds Console output for the root and client certificate
installers triggered by ConditionalFact.  The changes now
show what exceptions occur, whether the cert was found in
the store by thumbprint or whether it had to fallback to the
name only.  It also shows information about the cert used.
including valid dates, issuer, etc.

On Linux, these changes allow us to see the reasons cert
installation fail.  And the exceptions contain suggestions
why it may have failed.

On cert-based tests that fail for cert validity issues, it
will likely help understand why.


roncain:

roncain:
@dotnet-bot test outerloop Ubuntu14.01

StephenBonikowsky:
LGTM

PR #1155, opened 2016-05-13
Prevent spin-waiting for async Task to complete by using Task.RunSynchronously

mconnew:
When calling Task.GetAwaiter().GetResult(), the wait which results when the task goes async is a SpinWait. This change causes the Task to be started synchronously on the current thread and avoid the spinwait. To avoid the spinwait, there needs to be no timeout on the wait, the cancellation token for the task needs to be not cancellable and there needs to be an ambient task scheduler present in the task. This third item isn't the case as when the wait was being started, no continuations have been asynchronously called so there wasn't an ambient task schedule. On a 24 core machine, we were spin waiting for 25% of total CPU time.

### DIFF
--- a/src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
+++ b/src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
@@ -179,16 +179,121 @@ namespace System.Runtime
             }
         }
 
-        // Task.GetAwaiter().GetResult() calls an internal variant of Wait() which doesn't wrap exceptions in
-        // an AggregateException.
-        public static void WaitForCompletion(this Task task)
+        public static void RunSyncWithParams<T1>(this Func<T1, Task> function, T1 param1)
         {
-            task.GetAwaiter().GetResult();
+            var state = Tuple.Create(function, param1);
+            var syncTask = new Task<Task>(obj => {
+                var tuple = obj as Tuple<Func<T1, Task>, T1>;
+                var func = tuple.Item1;
+                var p1 = tuple.Item2;
+                return func(p1);
+            }, state, CancellationToken.None);
+            syncTask.RunSynchronously(TaskScheduler.Default);
+            syncTask.GetAwaiter().GetResult().GetAwaiter().GetResult();
         }
 
-        public static TResult WaitForCompletion<TResult>(this Task<TResult> task)
+        public static void RunSyncWithParams<T1, T2>(this Func<T1, T2, Task> function, T1 param1, T2 param2)
         {
-            return task.GetAwaiter().GetResult();
+            var state = Tuple.Create(function, param1, param2);
+            var syncTask = new Task<Task>(obj => {
+                var tuple = obj as Tuple<Func<T1, T2, Task>, T1, T2>;
+                var func = tuple.Item1;
+                var p1 = tuple.Item2;
+                var p2 = tuple.Item3;
+                return func(p1, p2);
+            }, state, CancellationToken.None);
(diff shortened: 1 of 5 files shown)


mconnew:

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
```diff
@@ -179,16 +179,121 @@ public static void ToApmEnd(this IAsyncResult iar)
             }
         }
 
-        // Task.GetAwaiter().GetResult() calls an internal variant of Wait() which doesn't wrap exceptions in
-        // an AggregateException.
-        public static void WaitForCompletion(this Task task)
+        public static void RunSyncWithParams<T1>(this Func<T1, Task> function, T1 param1)
```
roncain: Nit: not required, but it would be great to have a comment on the first one of these that explains what is going on just like your PR description.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
```diff
@@ -179,16 +179,121 @@ public static void ToApmEnd(this IAsyncResult iar)
...
         {
-            task.GetAwaiter().GetResult();
+            var state = Tuple.Create(function, param1);
+            var syncTask = new Task<Task>(obj => {
+                var tuple = obj as Tuple<Func<T1, Task>, T1>;
+                var func = tuple.Item1;
+                var p1 = tuple.Item2;
+                return func(p1);
+            }, state, CancellationToken.None);
```
roncain: Question only: I don't have a good sense of what it means to have CancellationToken.None and still use a timeout parameter for most of these.  Does this defeat timeouts?

roncain:
LGTM subject to these questions:
 - Could you help explain the impact of CancellationToken.None and no longer using a timed wait?  Did this defeat timeouts in any way?
 - Did private binaries of this resolve the issue on the 24 core machine?

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
mconnew: There are two timeouts at play here. The function being passed in handles any timeouts or cancellations. For example RequestChannel.RequestAsyncInternal has the following method signature:
```csharp
private async Task<Message> RequestAsyncInternal(Message message, TimeSpan timeout)
```
This has two parameters, message and timeout, which will map to param1 and param2. Inside the delegate, the function us passed these two parameters. If one of them was a cancellation token, it would get passed in too. It's up to the function to handle the cancellation token or timeout that is passed in. In this particular overload, a new task object is instantiated beginning on line 185. It is this task which has a CancellationToken.None attached to it and not the inner function being called. The significance of this is that waiting on this outer Task means there is no concern about needing to stop waiting for it to complete before the actual execution has completed. E.g. if the inner function took 10 seconds to complete and the outer Task was constructed with a CancellationToken which cancels after 5 seconds, then the code couldn't be executed on the same thread as the method that is waiting on the Task would need to be returned to once the CT has cancelled which would happen after 5 seconds. In this case, the execution actually happens on a different thread so that the current thread is available in the case of a potential cancellation.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
mconnew: I'll add some documentation

mconnew:
I'm closing this pull request as I have a cleaner approach which works a little better and improves perf in more scenarios.

Issue #1129, opened 2016-05-10
Tests that are disabled by ActiveIssue 1123 on *nix should be disabled by ConditionalFact


zhenlan:
Please find more details of the discussion [here](https://github.com/dotnet/wcf/pull/1096/files#r62285935).

Issue #1123, opened 2016-05-05
Https tests fail on *nix from CI


roncain:
The following Https tests fail when run from CI against *nix platforms:

```
CustomBindingTests.DefaultSettings_Https_Text_Echo_RoundTrips_String
HttpsTests.SameBinding_Soap12_EchoString
HttpsTests.CrossBinding_Soap11_EchoString
HttpsTests.SameBinding_Soap11_EchoString
```
Failure in all cases looks the same and is probably related to the fact we need to do manual setup on the *nix machines before they work properly.

```
     CustomBindingTests.DefaultSettings_Https_Text_Echo_RoundTrips_String [FAIL]
Test case FAILED with errors : warning : Unexpected exception was caught while doing the basic echo test for variation... [/mnt/j/workspace/dotnet_wcf/master/outerloop_ubuntu14.04_debug_prtest/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/Binding.Custom.Tests.csproj]
        'Client:: CustomBinding/DefaultValues
        Server:: CustomBinding/DefaultValues'
        Exception: System.ServiceModel.CommunicationException: An error occurred while sending the request. ---> System.Net.Http.HttpRequestException: An error occurred while sending the request. ---> System.Net.Http.CurlException: SSL peer certificate or SSH remote key was not OK

```

zhenlan:
If the root cause is cert install on *nix, this is a dup of #700.

roncain:
I wouldn't mark it a dupe just yet.  This issue captures the fact the tests fail, and how.  #700 is the work item to automate setup on *nix machines.  I think it is better to say #700 represents the infrastructure work necessary to fix this issue.  This issue is the "why" for the need for #700.  If you still want to mark it a dupe, let's propagate the information contained in this issue back to #700.

zhenlan:
related to #1129

roncain:
Closing for this reason: #1305 replaced the ActiveIssue(1123) with a new ConditionalFact -- SSL_Available.  It appears all these errors were related to OpenSSL not being correctly installed on the machines where the tests ran.  Any issues discovered with these tests after we get OpenSSL setup properly will open new issues.

PR #1160, opened 2016-05-13
Add NetHttpsBinding test coverage.

### LINKED ISSUE #507: Need scenario test coverage for Client Certificates on http.
Current tests for Client Certificate are using NetTcp, need to add the http pivot.

StephenBonikowsky:
* Adding tests for NetHttpsBinding.
* Fixes #507

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
@@ -1,0 +1,111 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
+using System;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+public static class Binding_Http_NetHttpsBindingTests
+{
+    [Fact]
+    [OuterLoop]
+    public static void DefaultCtor_NetHttps_Echo_RoundTrips_String()
+    {
+        string testString = "Hello";
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
+
+        try
+        {
+            // *** SETUP *** \\
+            NetHttpsBinding netHttpsBinding = new NetHttpsBinding();
+            factory = new ChannelFactory<IWcfService>(netHttpsBinding, new EndpointAddress(Endpoints.HttpBaseAddress_NetHttps));
+            serviceProxy = factory.CreateChannel();
+
+            // *** EXECUTE *** \\
+            string result = serviceProxy.Echo(testString);
+
+            // *** VALIDATE *** \\
+            Assert.True(String.Equals(testString, result), String.Format("Expected result was {0}. Actual was {1}", testString, result));
+
+            // *** CLEANUP *** \\
+            factory.Close();
+            ((ICommunicationObject)serviceProxy).Close();
+        }
+        finally
+        {
+            // *** ENSURE CLEANUP *** \\
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+        }
+    }
+
+    [Fact]
+    [OuterLoop]
+    public static void TransportWithMessageCredential_NotSupported_NetHttps()
+    {
+        string testString = "Hello";
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
+
+        // BasicHttpsSecurityMode.TransportWithMessageCredential is accessible but not supported.
+        // Verify the correct exception and message is thrown.
+        // When/if Message Security is supported this test will fail and will serve as a reminder to add test coverage.
+        Assert.Throws<PlatformNotSupportedException>(() =>
+        {
+            try
+            {
+                // *** SETUP *** \\
+                NetHttpsBinding netHttpsBinding = new NetHttpsBinding(BasicHttpsSecurityMode.TransportWithMessageCredential);
+                factory = new ChannelFactory<IWcfService>(netHttpsBinding, new EndpointAddress(Endpoints.HttpBaseAddress_NetHttps));
+                serviceProxy = factory.CreateChannel();
+
+                // *** EXECUTE *** \\
+                string result = serviceProxy.Echo(testString);
+            }
+            finally
+            {
+                // *** ENSURE CLEANUP *** \\
+                ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+            }
+        });
+    }
+
+    [Fact]
+    [OuterLoop]
+    public static void NonDefaultCtor_NetHttps_Echo_RoundTrips_String()
+    {
+        string testString = "Hello";
+        ChannelFactory<IWcfService> factory = null;
+        IWcfService serviceProxy = null;
+
+        try
+        {
+            // *** SETUP *** \\
+            NetHttpsBinding netHttpsBinding = new NetHttpsBinding(BasicHttpsSecurityMode.Transport);
+            factory = new ChannelFactory<IWcfService>(netHttpsBinding, new EndpointAddress(Endpoints.HttpBaseAddress_NetHttps));
+            serviceProxy = factory.CreateChannel();
+
+            // *** EXECUTE *** \\
+            string result = serviceProxy.Echo(testString);
+
+            // *** VALIDATE *** \\
+            Assert.NotNull(result);
+            Assert.Equal(testString, result);
+
+            // *** CLEANUP *** \\
+            factory.Close();
+            ((ICommunicationObject)serviceProxy).Close();
+        }
+        finally
+        {
+            // *** ENSURE CLEANUP *** \\
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


StephenBonikowsky:
Clearly something isn't happy 😀

src/System.Private.ServiceModel/tests/Scenarios/Client/ChannelLayer/DuplexChannelShapeTests.4.0.0.cs
```diff
@@ -12,7 +12,7 @@
 using System.Xml;
 using Xunit;
 
-public static class DuplexChannelShapeTests
+public static partial class DuplexChannelShapeTests
```
roncain: Need to add " : ConditionalWcfTest" to this line to make it derive from the base class containing the ConditionalFact conditions.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+public static class Binding_Http_NetHttpsBindingTests
+{
+    [Fact]
+    [OuterLoop]
+    public static void StandardUsage_NetHttps_Echo_RoundTrips_String()
```
roncain: Nit: do we need "StandardUsage" in the name?  What does it mean?  Default ctor?

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        finally
+        {
+            // *** ENSURE CLEANUP *** \\
+            ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+        }
+    }
+
+    [Fact]
+    [OuterLoop]
+    public static void TransportWithMessageCredential_NotSupported_NetHttps()
```
roncain: Yeah, I think until we have an attribute or something like that to mark these kinds of tests, just having the convention of having _NotSupported_ in the name seems like the best approach -- like you have here.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+        });
+
+        // *** ADDITIONAL VALIDATION *** \\
+        Assert.Equal(expectedExceptionMessage, exception.Message);
+    }
+
+    [Fact]
+    [OuterLoop]
+    public static void NonStandardUsage_NetHttps_Echo_RoundTrips_String()
```
roncain: Nit: similar question -- what does "NonStandardUsage" mean.  I think the intent here is something like "NetHttps-TransportSecurity-Echo-RoundTrips-String"

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            // *** SETUP *** \\
+            NetHttpsBinding netHttpsBinding = new NetHttpsBinding();
+            factory = new ChannelFactory<IWcfService>(netHttpsBinding, new EndpointAddress(Endpoints.HttpBaseAddress_NetHttps));
+            serviceProxy = factory.CreateChannel();
+
+            // *** EXECUTE *** \\
+            string result = serviceProxy.Echo(testString);
+
+            // *** VALIDATE *** \\
+            Assert.NotNull(result);
```
roncain: Nit: I'd prefer to see Assert.True used in both cases so you can provide a descriptive error message.  These could be combined into Assert.True(String.Equals(testString, result), String.Format("Expected result was '{0}' Actual was '{1}'", testString, result));

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            }
+            finally
+            {
+                // *** ENSURE CLEANUP *** \\
+                ScenarioTestHelpers.CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
+            }
+        });
+
+        // *** ADDITIONAL VALIDATION *** \\
+        Assert.Equal(expectedExceptionMessage, exception.Message);
```
roncain: This will probably fail in NET Native (retail) because the string resources are stripped.  I think we can probably live without the extra string validation for these kind of NotSupported tests.

src/System.Private.ServiceModel/tests/Scenarios/Client/ChannelLayer/DuplexChannelShapeTests.cs
```diff
@@ -0,0 +1,133 @@
...
+            requestMessage.Headers.MessageId = new UniqueId(Guid.NewGuid());
+
+            // *** EXECUTE *** \\
+            // Send the Message and receive the Response.
+            channel.Send(requestMessage);
+            replyMessage = channel.Receive(TimeSpan.FromSeconds(5));
+
+            // *** VALIDATE *** \\
+            // If the incoming Message did not contain the same UniqueId used for the MessageId of the outgoing Message we would have received a Fault from the Service
+            Assert.Equal(requestMessage.Headers.MessageId.ToString(), replyMessage.Headers.RelatesTo.ToString());
```
roncain: An Assert.True with a descriptive message would be very useful here

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
StephenBonikowsky: One of the other tests was using the binding to create the factory, so I was thinking of this particular style as being the standard way of creating a binding, factory and channel. But I guess that is pretty subjective. I'll update the name.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
StephenBonikowsky: Only difference between this test and the first one in this file is not using the default constructor, I'll rename the test.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
StephenBonikowsky: 🆗

src/System.Private.ServiceModel/tests/Scenarios/Client/ChannelLayer/DuplexChannelShapeTests.cs
StephenBonikowsky: 🆗

StephenBonikowsky:
Updated

StephenBonikowsky:

roncain:
Note: looking at the Console output for the Ubuntu failures, we apparently succeeded in installing the certs:

```
Using root cert in store by thumbprint '41571D814ABFE7669F757C9253F2695D1E024FF9' = '[Subject]
    CN=DO_NOT_TRUST_WcfBridgeRootCA, O=DO_NOT_TRUST, OU=Created by https://github.com/dotnet/wcf
  
  [Issuer]
    CN=DO_NOT_TRUST_WcfBridgeRootCA, O=DO_NOT_TRUST, OU=Created by https://github.com/dotnet/wcf
  
  [Serial Number]
    0082DD601AF83DA77F
  
  [Not Before]
    5/2/16 4:58:11 PM
  
  [Not After]
    6/1/16 5:58:11 PM
  
  [Thumbprint]
    41571D814ABFE7669F757C9253F2695D1E024FF9
  '
  Using client cert in store by thumbprint 'C14A4D88D4F596A2ABFE3A416A79EDF5C54C99E7' = '[Subject]
    CN=WCF Client Certificate, O=DO_NOT_TRUST, OU=Created by https://github.com/dotnet/wcf
  
  [Issuer]
    CN=DO_NOT_TRUST_WcfBridgeRootCA, O=DO_NOT_TRUST, OU=Created by https://github.com/dotnet/wcf
  
  [Serial Number]
    18DBB351782B6DA7
  
  [Not Before]
    5/2/16 4:58:11 PM
  
  [Not After]
    6/1/16 5:58:11 PM
  
  [Thumbprint]
    C14A4D88D4F596A2ABFE3A416A79EDF5C54C99E7

```
But then the tests fail with:

```
     Binding_Http_NetHttpsBindingTests.NonDefaultCtor_NetHttps_Echo_RoundTrips_String [FAIL]
        System.ServiceModel.CommunicationException : An error occurred while sending the request.
        ---- System.Net.Http.HttpRequestException : An error occurred while sending the request.
        -------- System.Net.Http.CurlException : SSL peer certificate or SSH remote key was not OK
```
These look like the same reason for failure as listed in #1123.  We may want to ActiveIssue these tests with #1123 as we did those other tests.

StephenBonikowsky:
I think you are right, will update with that ActiveIssue.

roncain:
Be aware #1129 has asked to *not* use ActiveIssue(1123) for those other tests but condition with ConditionalFact.  But because we haven't decided what that ConditionalFact would test for, I don't see any option but to ActiveIssue these tests with 1123 and know we need to come back and ConditionalFact them later.

StephenBonikowsky:
After discussing this further going to just add the conditional fact and merge, even if they will fail on Linux.

roncain:
LGTM
As we discussed, I'll see if an improved ConditionalFact allows these to be skipped.

PR #1163, opened 2016-05-14
Prevent spin-waiting by using WaitHandle from Task

mconnew:
Created NoSpin variations of WaitForCompletion which immediately uses a WaitHandle. The original variations are still there as they will be more performant when a Task is expected to complete immediately. Switching to NoSpin variations should only be done when it's known an operation will take more than a fraction of a second to complete or when a profile shows it would be advantageous.

### DIFF
--- a/src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
+++ b/src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
@@ -180,9 +180,25 @@ namespace System.Runtime
         }
 
         // Task.GetAwaiter().GetResult() calls an internal variant of Wait() which doesn't wrap exceptions in
-        // an AggregateException.
+        // an AggregateException. It does spinwait so if it's expected that the Task isn't about to complete,
+        // then use the NoSpin variant.
         public static void WaitForCompletion(this Task task)
         {
+            task.GetAwaiter().GetResult();
+        }
+
+        // If the task is about to complete, this method will be more expensive than the regular method as it
+        // always causes a WaitHandle to be allocated. If it is expected that the task will take longer than
+        // the time of a spin wait, then a WaitHandle will be allocated anyway and this method avoids the CPU
+        // cost of the spin wait.
+        public static void WaitForCompletionNoSpin(this Task task)
+        {
+            if (!task.IsCompleted)
+            {
+                ((IAsyncResult) task).AsyncWaitHandle.WaitOne();
+            }
+
+            // Call GetResult() to get any exceptions that were thrown
             task.GetAwaiter().GetResult();
         }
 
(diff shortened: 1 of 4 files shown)


mconnew:

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
```diff
@@ -180,26 +180,64 @@ public static void ToApmEnd(this IAsyncResult iar)
...
 
+        // If the task is about to complete, this method will be more expensive than the regular method as it
+        // always causes a WaitHandle to be allocated. If it is expected that the task will take longer than
+        // the time of a spin wait, then a WaitHandle will be allocated anyway and this method avoids the CPU
+        // cost of the spin wait.
+        public static void WaitForCompletionNoSpin(this Task task)
+        {
+            if (!task.IsCompleted)
+            {
+                ((IAsyncResult) task).AsyncWaitHandle.WaitOne();
```
roncain: Just checking -- this is logically an infinite wait  (which of course, so it GetResult()).  Will this wait handle be set by if the task is cancelled?

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
```diff
@@ -180,26 +180,64 @@ public static void ToApmEnd(this IAsyncResult iar)
...
-            return task.Wait(timeout);
+            bool completed = true;
+            if (!task.IsCompleted)
+            {
+                completed = ((IAsyncResult)task).AsyncWaitHandle.WaitOne(timeout);
+            }
+
+            if (completed)
+            {
+                // Throw any exceptions if there are any
```
roncain: Nit: Is this comment accurate?  How is it throwing away exceptions by using GetResult()?  And question: if completed==false, why does it bypass the GetAwaiter().GetResult()?  What other mechanism will trigger a faulted task or exception when completed==false?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportDuplexSessionChannel.cs
```diff
@@ -536,27 +536,9 @@ private void AsyncReceiveCancelled()
...
-
-                Message message = GetPendingMessage();
-
-                if (message != null)
-                {
-                    StartNextReceiveAsync();
-                }
-
-                return message;
+                return ReceiveAsync(timeout).WaitForCompletionNoSpin();
```
roncain: Question only: are we able to address the TODO now because of the change, or is this just something you noticed we needed to clean-up?  Just curious.

And why does it call WaitForCompletionNoSpin() without the timeout rather than the one that takes a timeout?  Is that because the ReceiveAsync's timeout will guarantee this flavor of WaitForCompletionNoSpin will complete if a timeout occurs?

roncain:
LGTM, subject only to a few questions (not blocking for merge).
BTW, I believe you need to rebase to get CI jobs to succeed again.  I'm investigating the reason but it worked for other PR's.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
mconnew: The WaitHandle is waiting on task completion. If a task is cancelled, or throws an exception or finishes without error, the task completes. The difference with using AsyncWaitHandle is an actual native WaitHandle is allocated whereas `GetResult()` uses a ManualResetEventSlim which does a SpinWait first.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
mconnew: It doesn't throw it away, it throws the exception. `GetAwaiter().GetResult()` will throw an exception if the task faulted. AsyncWaitHandle doesn't, it simply returns whether the WaitHandle was signaled (task completed, returns true) or it timed out (returns false). If the task timed out, calling task.GetAwaiter().GetResult() would cause an infinite wait which isn't what we want. We want to know that it didn't complete within timeout. If the task completed, we want to call task.GetAwaiter().GetResult() so that if the task faulted or was cancelled, the exception will get thrown.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WebSocketTransportDuplexSessionChannel.cs
mconnew: With WebSockets, we maintain a pending receive because it's a duplex transport channel. This means the majority of the time we won't have any data waiting for us so we don't want to spin wait. If there is data waiting, then it won't use the WaitHandle at all as the Task will complete immediately.

src/System.Private.ServiceModel/src/Internals/System/Runtime/TaskHelpers.cs
roncain: Sounds good -- sorry I misread the comment...

roncain:
LGTM.  When you have rebased and are ready, try an OuterLoop test run.  We expect 3 failures in NetHttps tests in *nix machines that can be ignored.  If that's all that fail, I'd go ahead and merge.

mconnew:

mconnew:

PR #1164, opened 2016-05-14
Prevent HttpClient from setting a timer on internal cancellation token

mconnew:
As HttpClient has a timeout set, it creates a linked cancellation token and calls CancelAfter on that linked token. By setting the default timeout to Infinite, HttpClient skips the call to CancelAfter. We were getting about 5% CPU usage updating the global TimerQueue because of this.
Also fixed an issue where we weren't unregistering a callback on our cancelation token which would have caused an unnecessary no-op callback.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
@@ -298,6 +298,11 @@ namespace System.ServiceModel.Channels
 
                     httpClient = new HttpClient(clientHandler);
 
+                    // We provide our own CancellationToken for each request. Setting HttpClient.Timeout to -1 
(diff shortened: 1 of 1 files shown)


mconnew:

roncain:
I think a rebase is required to make the PR tests pass, though I can't explain why yet.
Question only: can you double check that we have a test that verifies a normal Http request times out properly?
LGTM, subject to getting a CI pass by rebasing and verifying we have test coverage of the timeout.

mconnew:

roncain:
@dotnet-bot test outerloop Ubuntu14.04 Release

mconnew:
ExpectedExceptionTests.SendTimeout_For_Long_Running_Operation_Throws_TimeoutException tests the timeout.  
The Ubuntu failures are for NetHttp and certificates so aren't related.

roncain:
LGTM

PR #1165, opened 2016-05-15
Add additional runtimes

ellismg:
- Add "debian.8-x64". This is the correct rid for Debian. I want to
  remove "debian.8.2-x64" before RTM.

- Add a generic "linux-x64" rid which we can use in cases where we want
  to build for a distro where we don't yet have a set of runtime
  packages. This will let us restore the managed assets (since they are
  specific to "unix" or "linux" and then folks can do a native build to
  get the shims and runtime and copy them over (this is a more
  reasonable approach than our current strategy of just using
  ubuntu.14.04-x64)

- Sort runtime sections

### DIFF
--- a/src/Common/test-runtime/project.json
+++ b/src/Common/test-runtime/project.json
@@ -17,7 +17,10 @@
   },
   "frameworks": {
     "dnxcore50": {
-      "imports": [ "netstandard1.6", "portable-net45+win8" ],
+      "imports": [
+        "netstandard1.6",
+        "portable-net45+win8"
+      ],
       "dependencies": {
         "System.Net.Http.WinHttpHandler": "4.0.0-rc3-24109-00",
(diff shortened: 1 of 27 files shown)


ellismg:
I'm making these sorts of changes across CoreCLR, CoreFX and WCF.  I want to remove the uses of the "debian.8.2-x64" RID so we can remove it before RTM.

ellismg:
I generated these with a tool, which looked at all runtime sections that had and ubuntu.14.04 and updated it to my set of RIDs. This is why the imports section was changed as well in some places.

zhenlan:
LGTM. Thanks @ellismg.

roncain:

roncain:
LGTM -- passed all OuterLoop tests for all OSes our CI supports.

ellismg:
Thanks All!

PR #1178, opened 2016-05-18
Remove certificate matching by name

### LINKED ISSUE #1172: ConditionalFact detector/installer for certificates should not match on name only
The ConditionalFact detector/installer for both root and client certificates has a fallback that matches based only on name (IssuedBy and Subject Name) if it cannot install or locate a cert that matches the thumbprint given by the server.  I believe this is a mistake and should be removed.  I think we are matching certs we can not use and end up debugging why, when it would be better to see the test skip for "certificate not installed".

This issue can be used for discussion on whether name-only matches can be made to work.

roncain:
The ConditionalFact detector for certificates was using a fallback
strategy to locate certificates based on IssuedBy and SubjectName
when it could not locate by thumbprint.  But this allowed matches
on left-over certificates that would not work.

With this change, we match only on thumbprint or report that there
is no certificate installed and let the ConditionalFact skip the
certificate tests.

Fixes #1172

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ServiceUtilHelper.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ServiceUtilHelper.cs
@@ -35,6 +35,7 @@ public static class ServiceUtilHelper
             if (!s_rootCertAvailabilityChecked)
             {
                 X509Certificate2 rootCertificate = null;
+                string thumbprint = null;
 
                 try
                 {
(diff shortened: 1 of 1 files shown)


roncain:

roncain:
Reviewers: the OuterLoop failures are the known NetHttps test failures we consciously chose to allow for a short time.  Console output reveals they did, in fact, succeed in installing the root and client certs.

iamjasonp:
Wondering if we're actually better off trying to ensure that the leftovers are cleaned up properly... I'm trying to recall why we chose to match by name previously instead of matching by thumbprint (and i think it was for historical reasons where we didn't know in advance what the thumbprint for a cert was going to be)

Changes LGTM

roncain:
*nix OuterLoop failures are the known NetHttps tests we are consciously allowing to fail in CI for now.

PR #1182, opened 2016-05-19
Updating tests with ActiveIssue 1123.

StephenBonikowsky:
* Some new tests added recently fail for known Issue 1123 on Linux platforms.
* The correct fix is to create a conditional fact that skips the test based on an invalid environment setup.
* The setup that is causing the failure is currently still under investigation.


roncain:
LGTM.

Issue #1137, opened 2016-05-10
PlatformNotSupportedException running async duplex streaming over NetHttp


KKhurin:
This looks like a stress bug that popped up in the perf runs.
This is on the private for #1074...

~~~
0:100> !clrstack
OS Thread Id: 0xae4 (100)
        Child SP               IP Call Site
000000ab77d7b968 00007ffbfa4c8a5c [HelperMethodFrame: 000000ab77d7b968] 
000000ab77d7ba60 00007ffb8c3d5d46 System.ServiceModel.Channels.ServiceChannel.PrepareCall(System.ServiceModel.Dispatcher.ProxyOperationRuntime, Boolean, System.ServiceModel.Dispatcher.ProxyRpc ByRef) [D:\git\wcf\src\System.Private.ServiceModel\src\System\ServiceModel\Channels\ServiceChannel.cs @ 565]
000000ab77d7bae0 00007ffb8c3d5318 System.ServiceModel.Channels.ServiceChannel+SendAsyncResult.Begin() [D:\git\wcf\src\System.Private.ServiceModel\src\System\ServiceModel\Channels\ServiceChannel.cs @ 1672]
000000ab77d7bb10 00007ffb8c3d4671 System.ServiceModel.Channels.ServiceChannel.BeginCall(System.String, Boolean, System.ServiceModel.Dispatcher.ProxyOperationRuntime, System.Object[], System.TimeSpan, System.AsyncCallback, System.Object) [D:\git\wcf\src\System.Private.ServiceModel\src\System\ServiceModel\Channels\ServiceChannel.cs @ 713]
000000ab77d7bc60 00007ffb8c3d3b07 System.ServiceModel.Channels.ServiceChannelProxy+TaskCreator.CreateGenericTask(System.ServiceModel.Channels.ServiceChannel, System.ServiceModel.Dispatcher.ProxyOperationRuntime, System.Object[]) [D:\git\wcf\src\System.Private.ServiceModel\src\System\ServiceModel\Channels\ServiceChannelProxy.cs @ 194]
000000ab77d7bcf0 00007ffb8c3d2eaf System.ServiceModel.Channels.ServiceChannelProxy.Invoke(System.Reflection.MethodInfo, System.Object[]) [D:\git\wcf\src\System.Private.ServiceModel\src\System\ServiceModel\Channels\ServiceChannelProxy.cs @ 136]
[...]

PR #1185, opened 2016-05-19
Reset original operation context before causing task continuation to run

mconnew:
Addresses the problem with OperationContext.Current being set to the wrong thing in #1137. This doesn't fix the incorrect exception that is thrown.


roncain:
>                     OperationContext.Current = asyncResult.AsyncState as OperationContext;

In the scenario I am trying to duplicate for the stress issue, this always evaluates to null.  I'm not sure if that is expected or not.  Therefore when it arrives a PrepareCall, OperationContext.Current is null.

---
Refers to: src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannelProxy.cs:218 in 7ced7b1. [](commit_id = 7ced7b17deb0d1e870bf42916a70b29d89d12664, deletion_comment = False)

roncain:
Reviewers: I have taken these change plus another suggestion into #1191.  If that PR solves the issue, we will close this PR and merge #1191.

PR #1188, opened 2016-05-20
Add comments for recently made public classes do not match contract

hongdai:
* Add comments to explain why the classes are made public, but not on contract


zhenlan:
LGTM

roncain:
LGTM

Issue #486, opened 2015-10-29
Verify web sockets are being used in non-duplex scenarios.


StephenBonikowsky:
When WebSocketTransportUsage is set to 'Always' but the scenario is not using a duplex contract how do we know at the ServiceModel layer that the scenario is actually using web sockets?

Current test using this scenario...
Extensibility.WebSockets.WebSocketTests.WebSocketHttpRequestReplyBinaryStreamed
Extensibility.WebSockets.WebSocketTests.WebSocketHttpRequestReplyTextStreamed

Maybe we should have a simpler test (not streaming) to verify this issue.

StephenBonikowsky:
Test had already been added for this scenario some time ago.
Test case: Extensibility.WebSockets.WebSocketTests.WebSocket_Http_VerifyWebSocketsUsed()

PR #1195, opened 2016-05-20
Verify WebSockets are actually used when not using a callback.

StephenBonikowsky:
* Verify WebSockets are actually being used when WebSocketTransportUsage is set to Always.
* Wcf automatically uses WebSockets when a callback is used, but this test verifies that it still uses
  WebSockets when not using a callback.
* Fixes Issue #486

### DIFF
--- a/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/VerifyWebSockets.cs
+++ b/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/VerifyWebSockets.cs
@@ -1,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
+using System;
+using System.ServiceModel;
+
+namespace WcfService
+{
+    [ServiceBehavior(InstanceContextMode = InstanceContextMode.PerSession, AddressFilterMode = AddressFilterMode.Any)]
+    public class VerifyWebSockets : IVerifyWebSockets
+    {
+        private bool successMessage;
+
+        // This operation cannot return anything so in order to validate that it was called set the InstanceContextMode
+        //  to PerSession and set a variable that we can get with a call to the ValidateWebSocketsUsed operation.
+        public void ForceWebSocketsUse()
+        {
+            successMessage = true;
+        }
+
+        public bool ValidateWebSocketsUsed()
+        {
+            return successMessage;
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


StephenBonikowsky:

StephenBonikowsky:
WebSockets don't work on Unix platforms, since the test is attributed to be skipped on AnyUnix there is no point to run OuterLoop tests on those platforms.

StephenBonikowsky:
Jenkins appears to be busted...
 

> Identified problems 
> No identified problem
> No problems were identified. If you know why this problem occurred, please add a suitable Cause for it.

iamjasonp:
Code LGTM.
@dotnet-bot test outerloop Windows_NT please

Looks like there were some transient issues on Windows_NT outerloops, so let's test them again.

iamjasonp:
@dotnet-bot test outerloop Windows_NT please

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/VerifyWebSockets.cs
```diff
@@ -0,0 +1,28 @@
...
+    [ServiceBehavior(InstanceContextMode = InstanceContextMode.PerSession, AddressFilterMode = AddressFilterMode.Any)]
+    public class VerifyWebSockets : IVerifyWebSockets
+    {
+        private String successMessage = "";
+
+        // This operation cannot return anything so in order to validate that it was called set the InstanceContextMode
+        //  to PerSession and set a variable that we can get with a call to the ValidateWebSocketsUsed operation.
+        public void ForceWebSocketsUse()
+        {
+            successMessage = "WebSocketsWasUsed";
```
roncain: Nit: not blocking, but I generally like to decouple the service from the test as much as possible, including magic strings the server side returns.  I'd rather do something like pass a String into ForceWebSocketsUse, store it as you do here, and return it in ValidateWebSocketsUsed.  That way the magic string "WebSocketsWasUses" will exist only in the test.

roncain: Nit: not required to fix -- you could have just set a 'bool' field if this was called and had ValidateWebSocketsUsed could return the bool.  That would eliminate the need for a magic string duplicated in both server and test code.

roncain:
LGTM.
One nit comment about using bool instead of string in the new service interface.

StephenBonikowsky:

StephenBonikowsky:
Updated with the NIT comments and added the missing code to make the test work web hosted.

PR #1212, opened 2016-05-25
Add conditional fact for the two new tests need root cert

hongdai:
* The tests are failing in the projectN lab.
We need the contitional fact to ensure root certificate is installed
to validate server certificate.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
@@ -5,15 +5,13 @@
 
 using System;
 using System.ServiceModel;
-using System.ServiceModel.Channels;
-using System.Text;
-using System.Threading;
-using System.Threading.Tasks;
+
+using Infrastructure.Common;
 using Xunit;
 
-public static class Binding_Http_NetHttpsBindingTests
+public class Binding_Http_NetHttpsBindingTests: ConditionalWcfTest
 {
-    [Fact]
+    [ConditionalFact(nameof(Root_Certificate_Installed))]
     [OuterLoop]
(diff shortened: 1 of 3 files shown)


hongdai:
Not ready for review. I'll add a fix to another test having the same issue.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/NetHttpsBindingTests.cs
```diff
@@ -5,15 +5,13 @@
...
-using System.Threading.Tasks;
+
+using Infrastructure.Common;
 using Xunit;
 
-public static class Binding_Http_NetHttpsBindingTests
+public class Binding_Http_NetHttpsBindingTests: ConditionalWcfTest
 {
-    [Fact]
+    [ConditionalFact(nameof(Root_Certificate_Installed))]
```
roncain: ConditionalFact is not recognized by ToF, which means adding this attribute will cause this test to be skipped silently.

An alternative is to keep the [Fact] attribute and add a direct call like Assert.True(Root_Certificate_Installed(), "Root cert is not installed");

We can discuss how to do this more generally separately.  I'm just observing this change will effectively disable the tests in ToF for both CoreCLR and NET Native.

hongdai: I will keep these tests consistent with other tests that need Cert installed. We will have a general way to solve the issue that they will not run for NET Native.

hongdai:

hongdai:
It's ready for review now.

StephenBonikowsky:
LGTM
I will be testing a way to get all these tests using ConditionalFact to work in NetNative as a separate task.

roncain:
LGTM

iamjasonp:
LGTM

Issue #945, opened 2016-03-30
Automate kinit step and kdestroy step for ambient credential NegotiateStream/security tests


iamjasonp:
The following tests fail when Kerberos is not initialized, but passes when `kinit` is run and server is running as LOCAL SYSTEM 

```
   StreamingTests.NetTcp_TransportSecurity_StreamedRequest_RoundTrips_String [FAIL]
   StreamingTests.NetTcp_TransportSecurity_Streamed_RoundTrips_String_WithSingleThreadedSyncContext [FAIL]
   StreamingTests.NetTcp_TransportSecurity_Streamed_RoundTrips_String [FAIL]
   StreamingTests.NetTcp_TransportSecurity_Streamed_Async_RoundTrips_String [FAIL]
   StreamingTests.NetTcp_TransportSecurity_StreamedRequest_Async_RoundTrips_String [FAIL]
   StreamingTests.NetTcp_TransportSecurity_StreamedResponse_Async_RoundTrips_String [FAIL]
   StreamingTests.NetTcp_TransportSecurity_StreamedResponse_RoundTrips_String [FAIL]
   StreamingTests.NetTcp_TransportSecurity_Streamed_TimeOut_Long_Running_Operation [FAIL]
   StreamingTests.NetTcp_TransportSecurity_Streamed_Async_RoundTrips_String_WithSingleThreadedSyncContext [FAIL]
   Tcp_ClientCredentialTypeTests.SameBinding_DefaultSettings_EchoString [FAIL]
   Tcp_ClientCredentialTypeTests.SameBinding_SecurityModeTransport_EchoString [FAIL]
   NegotiateStream_Http_Tests.NegotiateStream_Http_AmbientCredentials [FAIL]
```

We need to be able to: 
- Run the Bridge/test host as LOCAL SYSTEM or another credential with access to the HOST/hostname SPN 
- Automate the `kinit user@dc.domain.com` step to specify the ambient Kerberos credential 
- Automate the `kdestroy -A` step to destroy the Kerberos credential

iamjasonp:
For the moment, use manual test guide: https://github.com/dotnet/wcf/blob/master/Documentation/manualtest-guide.md

zhenlan:
@iamjasonp let's consider a solution with the new test infrastructure, for example, host relevant service separately ran with LOCAL SYSTEM account.

Lxiamail:
Since we have the manual coverage, the infrastructure work is not low cost. closing it for now.

PR #1218, opened 2016-05-25
Update tests marked with ActiveIssue #592 to #945

iamjasonp:
Update the comment in the test file itself, as #592 is no longer an accurate
description for why this test must be skipped. #945 better tracks the issues around
why this test is currently skipped.

skip ci please


StephenBonikowsky:
LGTM
Didn't know we had another magical phrase to skip the CI tests 🎊

Issue #1168, opened 2016-05-16
PR service and CI PR synching appears mismatched for tests vs WCF services


roncain:
PR service and the Jenkins PR syncing scripts for Outerloops appear to use different approaches, sometimes resulting in a situation where the tests that run are not in sync with the WCF services they execute against.

For example, after PR #1103 was merged, no PR submitted after that point would succeed until it was rebased onto master to incorporate changes in #1103, even PR's that had absolutely nothing to do with changes from #1103.

This does not affect self-hosted tests -- only those where there server and test client are on separate machines and synced independently.

zhenlan:
@iamjasonp while we can workaround this issue for now (by rebasing other PRs), it will be great if we can fix this sooner.

iamjasonp:
Next time someone changes the service is going to be validation for this :smile:

iamjasonp:
@roncain and I think that this is sufficiently tested after PR #1346 - we find that the PeerTrust tests are running successfully with both a server side and a client side change. 

Closing this issue - please reopen if there are additional issues that are found

PR #1223, opened 2016-05-26
Create security subfolders in IIS host server

hongdai:
* This will enable most security tests run in Domain machines.
* This will enable client certificate tests in IIS hosted lab runs.
* Fix tests with incorrect conditional fact.
* Fix service side test issue as it's the first time we enable some
of the security tests.

PR #1225, opened 2016-05-27
Exclude scenario tests if not building for OuterLoop

### LINKED ISSUE #1224: Suppress ConditionalFact detectors when not running OuterLoops
The ConditionalFact detectors are attempting to run when only InnerLoops are run.  And they attempt to contact ServiceUri, which usually fails.  They do this because the Fact discoverer comes before the Trait discoverer, and xunit doesn't know to eliminate the OuterLoop-marked tests until they have already executed their [ConditionalFact].

Task: find a way to return false from all ConditionalFact detectors without contacting the server when we are running only InnerLoops.

roncain:
The ConditionalFact detectors were being executed during
innerloop-only test runs. This is due to the fact we build
all test projects unconditionally and because xunit's trait
and Fact discoverers are executed before deciding whether
there are any tests to run. But unless the Self-host WCF
service was running (which happens only during OuterLoop
runs), this generated many exceptions in the console log.

With this change, all projects under tests\Scenarios are
excluded from the build unless OuterLoop has been requested.
This means there are no are no binaries containing
ConditionalFact when a normal non-OuterLoop build is done.

Fixes #1224

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/ServiceModel.Scenario.Tests.builds
+++ b/src/System.Private.ServiceModel/tests/Scenarios/ServiceModel.Scenario.Tests.builds
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
-  <ItemGroup>
+  <ItemGroup Condition="($(WithCategories.Contains('OuterLoop'))) or ('$(OuterLoop)' == 'true')">
     <Project Include="**\*.csproj" />
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.traversal.targets))\dir.traversal.targets" />


roncain:

roncain:
Reviewers: compare the console logs of the CI innerloop jobs with those of other current PR's to understand what is being fixed here.  Search for "root cert".  This PR also resolves the issue where it appeared OuterLoops were always being run on our dev boxes.  They weren't, but the ConditionalFact's were.

roncain:
The OuterLoop failure is due to the merge of #1223 happening concurrently with the CI runs.  It reiterates the need to fix #1168.  I will need to rebase before they will pass again.

roncain:

zhenlan:
LGTM

With this, people need to build OuterLoop explicitly to make sure outerloop tests are not broken.

roncain:
Reviewers: I decided the initial approach of not building scenario tests unless Outerloop was specified had some issues (ex: hiding scenario test build errors during normal 'build.cmd', still showing errors if the innerloops were run after running outerloops, etc.).  The current PR uses a completely different approach.  It introduces a 'Services_Available' TestProperty that is set to 'true' if the build has asked for OuterLoop tests.  The ConditionalFact detectors that require access to the WCF services check this during xunit Fact discovery and invoke the detection logic only when OuterLoops are being run.

I believe this is more consistent with how we are already using TestProperties to capture build-time properties into the test binaries (ex: ServiceUri).

@zhenlan  marked LGTM on the prior approach, but I would appreciate a fresh review of this new approach.

roncain:

roncain:

roncain:

roncain:
Thanks @zhenlan .  Yes, build errors in OuterLoop tests will not be caught until explicitly ask to build /p:WithCategories=OuterLoop.  I experimented with a couple other approaches, but in the end this one was the simplest.  And it makes innerloop builds and runs much faster locally and in CI.

PR #1228, opened 2016-05-27
Enable client certificates and server certificate validation using HttpClientHandler

mconnew:


roncain:
I think this PR needs to be based on release/1.0.0-rc2 rather than release/1.0.0.  Either that or I got my PR wrong and need to fix mine :grin:

@Zhenlan -- could you help clarify?

Issue #1211, opened 2016-05-24
System.ObjectDisposedException: Message is closed.


KKhurin:
OS: Ubuntu 14.04 x64
Transport: Http
Test: async Streaming echo
We get the following exception:
~~~
System.IO.IOException: An exception has been thrown when reading the stream. ---> System.IO.IOException: The read operation failed, see inner exception. ---> System.ObjectDisposedException: Message is closed.
   at System.ServiceModel.Channels.BodyWriterMessage.get_Version()
   at System.ServiceModel.Channels.Message.WriteMessagePostamble(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.Message.<OnWriteMessageAsync>d__79.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.ServiceModel.Channels.Message.<WriteMessageAsync>d__78.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.ServiceModel.Channels.TextMessageEncoderFactory.TextMessageEncoder.<WriteMessageAsync>d__39.MoveNext()
   --- End of inner exception stack trace ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.ServiceModel.Channels.MaxMessageSizeStream.<ReadAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.ServiceModel.Channels.ReadAheadWrappingStream.<EnsureBufferedAsyncCore>d__15.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
[...]

PR #1234, opened 2016-05-31
Wait until message has been sent before completing HttpClient.SendAsync

mconnew:
Fixes the problem where HttpClientHandler completes the task returned from SendAsync
once the response headers have be returned when running on Unix.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
@@ -12,6 +12,7 @@ using System.Net.Http.Headers;
 using System.Runtime;
 using System.Threading.Tasks;
 using System.Linq;
+using System.Threading;
 
 namespace System.ServiceModel.Channels
 {
@@ -21,23 +22,21 @@ namespace System.ServiceModel.Channels
         protected MessageEncoder _messageEncoder;
         protected Stream _stream = null;
         private bool _disposed;
+        protected TaskCompletionSource<bool> _writeCompletedTcs;
 
         public MessageContent(Message message, MessageEncoder messageEncoder)
         {
             _message = message;
             _messageEncoder = messageEncoder;
+            _writeCompletedTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
 
             SetContentType(_messageEncoder.ContentType);
             PrepareContentHeaders();
         }
 
-        public Message Message
-        {
-            get
-            {
-                return _message;
-            }
-        }
+        public Message Message { get { return _message; } }
+
+        internal  Task WriteCompletionTask { get { return _writeCompletedTcs.Task; } }
 
         private void PrepareContentHeaders()
         {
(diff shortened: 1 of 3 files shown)


mconnew:

roncain:
I assume this is a fix for #1211, right? Could you add a "Fixes" comment?

roncain:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
```diff
@@ -21,23 +22,21 @@ internal abstract class MessageContent : HttpContent
...
-        public Message Message
-        {
-            get
-            {
-                return _message;
-            }
-        }
+        public Message Message { get { return _message; } }
+
+        internal  Task WriteCompletionTask { get { return _writeCompletedTcs?.Task; } }
```
roncain: Nit: none of the other dereferences of _writeCompletedTcs check for null.  Is this one strictly needed, or should the others check too?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceModelHttpMessageHandler.NetNative.cs
```diff
@@ -79,11 +79,6 @@ public bool SupportsCertificateValidationCallback
         {
             get { return false; }
         }
-
-        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
-        {
-            return base.SendAsync(request, cancellationToken);
-        }
```
roncain: Question only: I can see why this SendAsync is being removed, because it was only calling the base.  But in the non-NET Native version of this class, you are adding a new override of SendAsync to await the WriteCompletedTask.  Could you explain in this PR thread why the NET Native version is not doing the same override?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
```diff
@@ -180,15 +179,34 @@ protected override Task<Stream> CreateContentReadStreamAsync()
             // WriteMessageAsync might run synchronously and try to write to the stream. ProducerConsumerStream
             // will block on the write until the stream is being read from. The WriteMessageAsync method needs
             // to run on a different thread to prevent a deadlock.
-            _stream = new ProducerConsumerStream();
-            var bufferedStream = new BufferedWriteStream(_stream, WriteBufferSize);
-            Task.Run(() => _messageEncoder.WriteMessageAsync(_message, bufferedStream));
-            return Task.FromResult(_stream);
+            var resultStream = new ProducerConsumerStream();
```
roncain: Just a couple questions to help me understand:
Before the change, _stream held the ProducerConsumerStream, and it looks like the BufferedWriteStream wrapped around it was just allowed to fall out of scope and be GC'ed after the write completed.

After the change, _stream holds the BufferedWriteStream, not the ProducerConsumerStream.

Questions are:
  1. Is this change intended to control the lifetime of the BufferedWriteStream and not let it be GC'ed till we're done?
  2. Are there any ramifications of _stream now holding the "outer" BufferedWriteStream instead of the "inner" ProducerConsumerStream?

Don't block on this question -- I just wanted to understand what's going on here.

roncain:
LGTM
I think we should add the "Fixes" comment to associate this fix with an issue, and there are a couple (non-blocking) questions and nits.  Have we had a chance to run these changes under stress for both Windows and Linux?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: This is a remnant from an earlier version of the fix where I was lazily creating the TCS. I realized there was too much possibility for race conditions so abandoned that approach. I'll remove the null check.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/MessageContent.cs
mconnew: Disposing a BufferedWriteStream also disposes the wrapped stream. But it also does one other step when Dispose is called, it flushes out any unwritten buffered data to the inner stream. Our message encoders will already do this, but if an external MessageEncoder didn't call Flush, then we might not send the final bytes of a request. This was a stress bug waiting to be reported :)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceModelHttpMessageHandler.NetNative.cs
mconnew: These are partial classes and not derived types. So the SendAsync method in NetNative was a duplicate method with the one in the base type. The implementation in NetNative was a harmless leftover from when we had a separate wrapper around WinHttpHandler for windows coreclr. Now it's no longer harmless and was causing a problem.

mconnew:
Stress test was run on Linux with this fix.

PR #1238, opened 2016-06-01
Add dummy events to generate correct eventlog channel id's

mconnew:


zhenlan:
cc: @vancem

src/System.Private.ServiceModel/src/Internals/WcfEventSource.cs
```diff
@@ -2926,6 +2962,9 @@ public class Keywords
             public const EventKeywords WFActivities = (EventKeywords)0x2000000;
             public const EventKeywords WFServices = (EventKeywords)0x4000000;
             public const EventKeywords WFInstanceStore = (EventKeywords)0x8000000;
+            // This DummyKeyword is used to emit events that aren't being listened for in
+            // order to initialize EventSource with the correct values for the eventlog channels
+            public const EventKeywords DummyKeyword = (EventKeywords) 0x080000000;
```
vancem: I am surprised you need this new dummy keyword.  I would expect just leaving the Keywords property off the 4 events above would work just fine.

src/System.Private.ServiceModel/src/Internals/WcfEventSource.cs
```diff
@@ -2629,6 +2661,10 @@ public class EventIds
             public const int ThrowingEtwExceptionVerbose = 57409;
             public const int ThrowingEtwException = 57410;
             public const int HttpHandlerPickedForUrl = 62326;
+            public const int DummyAdminEventId = 65530;
+            public const int DummyOperationalEventId = 65531;
+            public const int DummyAnalyticEventId = 65532;
+            public const int DummyDebugEventId = 65533;
```
vancem: Note that EventSource works by making a big array that is index by event ID.  This array is as big as the biggest event number used.  Thus it is best if you DON'T make big numbers.   Now the other numbers are what they are and for compat we have to live with that, but we  don't need to make it worse here.   From what I could tell there were plenty of IDs not being used less than 57K, so you should use something that does not make the range bigger.

src/System.Private.ServiceModel/src/Internals/WcfEventSource.cs
```diff
@@ -12,6 +12,38 @@ internal sealed class WcfEventSource : EventSource
...
+        }
+
+        [Event(EventIds.DummyAnalyticEventId, Level = EventLevel.Critical, Channel = EventChannel.Analytic, Opcode = EventOpcode.Info, Task = EventTask.None,
+    Keywords = Keywords.DummyKeyword)]
+        private void ACDummyAnalyticChannelEvent()
+        {
+            WriteEvent(EventIds.DummyAnalyticEventId);
+        }
+
+        [Event(EventIds.DummyDebugEventId, Level = EventLevel.Critical, Channel = EventChannel.Debug, Opcode = EventOpcode.Info, Task = EventTask.None,
```
vancem: In addition to leaving out the keywords property  (see comment below), I would also leave out the Opcode and level since the defaults are fine it is is less code.   For what it is worth...

src/System.Private.ServiceModel/src/Internals/WcfEventSource.cs
```diff
@@ -12,6 +12,38 @@ internal sealed class WcfEventSource : EventSource
     {
         public static WcfEventSource Instance = new WcfEventSource();
 
+        // The following four dummy events are to set the correct values for the eventlog channel keyword values. EventSource allocates
+        // the channel id's in the order that they are first used in the event methods lexically. So the first method attributed with [Event]
+		// which specifies an EventChannel will have the first channel keyword assigned to that Channel. We need EventSource to assign
+		// keyword bit 63 to Admin, bit 62 to Operational, bit 61 to Analytic and bit 60 to Debug.
+        [Event(EventIds.DummyAdminEventId, Level = EventLevel.Critical, Channel = EventChannel.Admin, Opcode = EventOpcode.Info, Task = EventTask.None,
+    Keywords = Keywords.DummyKeyword)]
```
roncain: Nit: (non-blocking) -- the indenting of the new [Event]s doesn't match the others

roncain:

roncain:

roncain:
LGTM, subject to Vance's comments.
Just to confirm -- have you tried running this on NET Native?

mconnew:
This approach didn't work.

Issue #1245, opened 2016-06-03
ProjectK RTM VS experience needs to explicitly include "System.Private.DataContractSerialization" in project.json


hongdai:
"System.ServiceModel.Http" depends on  "System.Runtime.Serialization.Xml", it should have included 
System.Private.DataContractSerialization automatically.

The package restore shows it has included System.Private.DataContractSerialization, but the runtime fails to find it.

zhenlan:
This is the exception
```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'System.ServiceModel.Channels.MessageVersion' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'System.ServiceModel.EnvelopeVersion' threw an exception.
 ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime.Serialization.Xml, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
   at System.ServiceModel.EnvelopeVersion..cctor()
   --- End of inner exception stack trace ---
   at System.ServiceModel.EnvelopeVersion.get_None()
   at System.ServiceModel.Channels.MessageVersion..cctor()
   --- End of inner exception stack trace ---
   at System.ServiceModel.Channels.TextMessageEncodingBindingElement..ctor()
   at System.ServiceModel.HttpBindingBase..ctor()
   at System.ServiceModel.BasicHttpBinding..ctor(BasicHttpSecurityMode securityMode)
   at ConsoleApp3.Program.Main(String[] args)
```
After including `System.Runtime.Serialization.Xml` package in project.json the exception goes away although the app does not use anything directly from S.R.S.Xml. This is updated project.json:
```
{
  "version": "1.0.0-*",
  "buildOptions": {
    "emitEntryPoint": true
  },

  "dependencies": {
    "Microsoft.NETCore.App": {
      "type": "platform",
      "version": "1.0.0-rc3-004338"
    }
  },

  "frameworks": {
    "netcoreapp1.0": {
      "imports": "dnxcore50",
      "dependencies": {
        "System.ServiceModel.Http": "4.1.0-rc3-24128-00",
        "System.Runtime.Serialization.Xml": "4.1.1-rc3-24128-00"
      }
    }
   }
}
```
[...]

PR #1247, opened 2016-06-03
Fix package baseline

ericstj:
Package baselining has been broken since the first package update after
I added it.  The problem was that the package that contains the corefx
baseline was missing from the update list so the update script updated
it to be out of sync with the CoreFxExpectedPrerelease property. I've
added the baseline package to the list of packages coming from CoreFx.

This was silently being dropped because we don't import the target if it
doesn't exist.  I've updated the import condition to be less permissive.
We just need it to not import before the restore so constraining it to
pkgproj will do that, and fail the build if this issue should occur in
the future.

Fixes https://github.com/dotnet/wcf/issues/1245

/cc @hongdai @zhenlan @weshaggard @dagood

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -82,7 +82,7 @@
     <ValidatePackageVersions>true</ValidatePackageVersions>
     <ProhibitFloatingDependencies>true</ProhibitFloatingDependencies>
     <CoreFxExpectedPrerelease>rc3-24117-00</CoreFxExpectedPrerelease>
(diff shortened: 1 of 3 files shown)


dagood:
LGTM.

Looking at the versions repo, this package doesn't seem to be publishing in the master branch (from CoreFX build pipeline or TFS) anymore. Is that a problem? Or are we already planning on removing the package post-1.0.0?

zhenlan:

zhenlan:
Thanks @ericstj for the quick fix. Will you please port this to the 1.0.0 release branch through ASK mode too?

ericstj:
Yeah, I'll port to release.

zhenlan:
@ericstj - Great. Please go ahead and merge or do you need me to click the merge button?

ericstj:
Was waiting for outerloop to pass.  Looks like it's backed up or machines are offline.  I'll merge.

zhenlan:
Given that outerloop passed on CentOS and OSX (in addition to Windows) already, it's good to me.

Issue #1220, opened 2016-05-25
Remove ActiveIssue #833 from tests and verify fixed.


StephenBonikowsky:
The following tests are attributed with Issue #833 

- Binding\Custom\CustomBindingTests.cs\DefaultSettings_Tcp_Binary_Echo_RoundTrips_String
- Client\ExpectedExceptions\ExpectedExceptionTests.cs\TCP_ServiceCertExpired_Throw_MessageSecurityException
- Client\ExpectedExceptions\ExpectedExceptionTests.cs\TCP_ServiceCertRevoked_Throw_MessageSecurityException
- Security\TransportSecurity\Tcp\IdentityTests.cs\VerifyServiceIdentityMatchDnsEndpointIdentity
- Security\TransportSecurity\Tcp\IdentityTests.cs\ServiceIdentityNotMatch_Throw_MessageSecurityException

This is a NETNative issue, so need to validate that these tests run and pass in that configuration.

PR #1261, opened 2016-06-07
Fixing issues related to full xunit not being supported in ToF.

### LINKED ISSUE #1221: Implement workaround in NETNative for ConditionalFact not being understood.
All tests being attributed with a ConditionalFact will not get run in NETNative as ToF uses a "toy" version of xunit that doesn't understand ConditionalFact, therefore since it doesn't recognize ConditionalFact its as if the test doesn't even have the 'Fact' attribute and ToF doesn't even think it is a test.

Test using an IF/DEF based on if(ToF) to add the 'Fact' attribute. Then the first step in the test case would do an Assert(<call the conditional fact method>).

If this works we would add this logic to all tests we "expect" to work in NETNative/ToF environment if certain setup/environment conditions are met.

The difference between these and other runs would be that whereas in other runs the tests would show as skipped if they failed the condition, these tests would show as failures, albeit with a good error message to know why.

StephenBonikowsky:
* Added a new constant variable only in ToF to if/def against in order to get around ConditionalFact not being supported as well as conditional ActiveIssue attribute.
* Fixes #1221

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.4.0.0.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.4.0.0.cs
@@ -47,11 +47,25 @@ public partial class CustomBindingTests : ConditionalWcfTest
     }
 
     // Https: Client and Server bindings setup exactly the same using default settings.
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+#else
     [ConditionalFact(nameof(Root_Certificate_Installed))]
-    [OuterLoop]
     [ActiveIssue(1123, PlatformID.AnyUnix)]
+#endif
+    [OuterLoop]
     public static void DefaultSettings_Https_Text_Echo_RoundTrips_String()
     {
+#if FULLXUNIT_NOTSUPPORTED
+        bool root_Certificate_Installed = Root_Certificate_Installed();
+        if (!root_Certificate_Installed)
+        {
+            Console.WriteLine("---- Test SKIPPED --------------");
+            Console.WriteLine("Attempting to run the test in ToF, a ConditionalFact evaluated as FALSE.");
+            Console.WriteLine("Root_Certificate_Installed evaluated as {0}", root_Certificate_Installed);
+            return;
+        }
+#endif
         string testString = "Hello";
         CustomBinding binding = null;
         ChannelFactory<IWcfService> factory = null;
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Tcp/NetTcpBindingTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Tcp/NetTcpBindingTests.cs
@@ -12,15 +12,25 @@ public partial class Binding_Tcp_NetTcpBindingTests : ConditionalWcfTest
     // Default settings are:
     //                         - SecurityMode = Transport
     //                         - ClientCredentialType = Windows
-    [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Windows_Authentication_Available))]
-#if !FEATURE_NETNATIVE
-    [ActiveIssue(945, PlatformID.AnyUnix)] // NegotiateStream works on Windows and Linux, but we have not yet automated ambient credential configuration in Linux
-#else
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
     [ActiveIssue(832)] // Windows Stream Security is not supported in NET Native
+#else
+    [ConditionalFact(nameof(Windows_Authentication_Available))]
 #endif
     [OuterLoop]
     public static void DefaultSettings_Echo_RoundTrips_String()
     {
+#if FULLXUNIT_NOTSUPPORTED
+        bool windows_Authentication_Available = Windows_Authentication_Available();
+        if (!windows_Authentication_Available)
+        {
+            Console.WriteLine("---- Test SKIPPED --------------");
+            Console.WriteLine("Attempting to run the test in ToF, a ConditionalFact evaluated as FALSE.");
+            Console.WriteLine("Windows_Authentication_Available evaluated as {0}", windows_Authentication_Available);
+            return;
+        }
+#endif
         string testString = "Hello";
         ChannelFactory<IWcfService> factory = null;
         IWcfService serviceProxy = null;
@@ -52,15 +62,25 @@ public partial class Binding_Tcp_NetTcpBindingTests : ConditionalWcfTest
     // Simple echo of a string using NetTcpBinding on both client and server with SecurityMode=Transport
     // By default ClientCredentialType will be 'Windows'
     // SecurityMode is Transport by default with NetTcpBinding, this test explicitly sets it.
-    [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Windows_Authentication_Available))]
-#if !FEATURE_NETNATIVE
-    [ActiveIssue(945, PlatformID.AnyUnix)] // NegotiateStream works on Windows and Linux, but we have not yet automated ambient credential configuration in Linux
-#else
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
     [ActiveIssue(832)] // Windows Stream Security is not supported in NET Native
+#else
+    [ConditionalFact(nameof(Windows_Authentication_Available))]
 #endif
     [OuterLoop]
     public static void SecurityModeTransport_Echo_RoundTrips_String()
     {
+#if FULLXUNIT_NOTSUPPORTED
+        bool windows_Authentication_Available = Windows_Authentication_Available();
+        if (!windows_Authentication_Available)
+        {
+            Console.WriteLine("---- Test SKIPPED --------------");
+            Console.WriteLine("Attempting to run the test in ToF, a ConditionalFact evaluated as FALSE.");
+            Console.WriteLine("Windows_Authentication_Available evaluated as {0}", windows_Authentication_Available);
+            return;
+        }
+#endif
         string testString = "Hello";
         ChannelFactory<IWcfService> factory = null;
         IWcfService serviceProxy = null;
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
@@ -11,10 +11,24 @@ using Infrastructure.Common;
 
 public class DuplexClientBaseTests : ConditionalWcfTest
 {
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+#else
     [ConditionalFact(nameof(Root_Certificate_Installed))]
+#endif
     [OuterLoop]
     public static void DuplexClientBaseOfT_OverHttp_Call_Throws_InvalidOperation()
     {
+#if FULLXUNIT_NOTSUPPORTED
+        bool root_Certificate_Installed = Root_Certificate_Installed();
+        if (!root_Certificate_Installed)
+        {
+            Console.WriteLine("---- Test SKIPPED --------------");
+            Console.WriteLine("Attempting to run the test in ToF, a ConditionalFact evaluated as FALSE.");
+            Console.WriteLine("Root_Certificate_Installed evaluated as {0}", root_Certificate_Installed);
+            return;
+        }
+#endif
         DuplexClientBase<IWcfDuplexService> duplexService = null;
         IWcfDuplexService proxy = null;
 
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
@@ -278,7 +278,7 @@ public partial class ExpectedExceptionTests : ConditionalWcfTest
     }
 
     [Fact]
-#if FEATURE_NETNATIVE
+#if FULLXUNIT_NOTSUPPORTED
     [ActiveIssue(833)] // Not supported in NET Native
 #endif
     [OuterLoop]
@@ -316,11 +316,28 @@ public partial class ExpectedExceptionTests : ConditionalWcfTest
         }
     }
 
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+    [ActiveIssue(983)]
+#else
     [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed))]
+#endif
     [OuterLoop]
     // Confirm that the Validate method of the custom X509CertificateValidator is called and that an exception thrown there is handled correctly.
     public static void TCP_ServiceCertFailedCustomValidate_Throw_Exception()
     {
+#if FULLXUNIT_NOTSUPPORTED
+        bool root_Certificate_Installed = Root_Certificate_Installed();
+        bool client_Certificate_Installed = Client_Certificate_Installed();
+        if (!root_Certificate_Installed || !client_Certificate_Installed)
+        {
+            Console.WriteLine("---- Test SKIPPED --------------");
+            Console.WriteLine("Attempting to run the test in ToF, a ConditionalFact evaluated as FALSE.");
+            Console.WriteLine("Root_Certificate_Installed evaluated as {0}", root_Certificate_Installed);
+            Console.WriteLine("Client_Certificate_Installed evaluated as {0}", client_Certificate_Installed);
+            return;
+        }
+#endif
         string testString = "Hello";
         NetTcpBinding binding = null;
         EndpointAddress endpointAddress = null;
@@ -360,8 +377,8 @@ public partial class ExpectedExceptionTests : ConditionalWcfTest
         Assert.Equal(MyCertificateValidator.exceptionMsg, exception.Message);
     }
 
+#if FULLXUNIT_NOTSUPPORTED
     [Fact]
-#if FEATURE_NETNATIVE
     [ActiveIssue(833)] // Not supported in NET Native
 #endif
     [OuterLoop]
--- a/src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
@@ -16,7 +16,10 @@ public class WebSocketTests : ConditionalWcfTest
 {
     [Fact]
     [OuterLoop]
+#if FULLXUNIT_NOTSUPPORTED
+#else
     [ActiveIssue(420, PlatformID.AnyUnix)]
+#endif
     public static void WebSocket_Http_RequestReply_BinaryStreamed()
     {
         NetHttpBinding binding = null;
@@ -78,10 +81,10 @@ public class WebSocketTests : ConditionalWcfTest
 
     [Fact]
     [OuterLoop]
-#if !FEATURE_NETNATIVE
-    [ActiveIssue(420, PlatformID.AnyUnix)]
-#else
+#if FULLXUNIT_NOTSUPPORTED
     [ActiveIssue(526)]
+#else
+    [ActiveIssue(420, PlatformID.AnyUnix)]
 #endif
     public static void WebSocket_Http_Duplex_BinaryStreamed()
     {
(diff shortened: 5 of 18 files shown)


StephenBonikowsky:

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.cs
```diff
@@ -93,11 +108,33 @@ public static void NegotiateStream_Http_AmbientCredentials()
         }
     }
 
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+#else
     [ConditionalFact(nameof(Windows_Authentication_Available), nameof(Root_Certificate_Installed))]
-    [ActiveIssue(851)]
```
roncain: These NegotiateStream tests are now failing in OuterLoops because the removal of ActiveIssue(851) allows them to run.  But they cannot succeed without manual setup.  I agree #851 was not a good issue to use, so I created new issue 1265 asking to create a new ConditionalFact to detect whether the manual setup has been done.  I recommend putting back the line [ActiveIssue(1265)] so these don't get run by CI until we can make them conditional.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/ClientCredentialTypeCertificateCanonicalNameTests.cs
```diff
@@ -24,12 +24,213 @@ public class Tcp_ClientCredentialTypeCertificateCanonicalNameTests : Conditional
...
     // to pass a variation or fail a variation. 
 
-    [ConditionalTheory(nameof(Root_Certificate_Installed))]
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+#else
+    [ConditionalFact(nameof(Root_Certificate_Installed))]
+#endif
     [OuterLoop]
-    [MemberData("TcpClientCredentialType_Certificate_With_Only_CanonicalName_MemberData")]
```
roncain: We had a separate discussion on converting the [Theory] tests to be individual tests in this PR.  Stephen and I disagree about the use of [Theory] tests (there are good arguments on both sides), but I don't disagree strongly enough to alter this PR.  I mention it here only because others may not have heard that conversation.

 I still believe we should have left the original test alone and added new tests #if'ed with FULLXUNIT_NOTSUPPORTED so that moving to Helix would have run the tests as they were originally written.  But okay to continue the PR with these tests refactored like this.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/StreamingTests.cs
```diff
@@ -17,15 +17,30 @@
 
 public class StreamingTests : ConditionalWcfTest
 {
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+    [ActiveIssue(832)] // Windows Stream Security is not supported in NET Native
```
roncain: Non-blocking observation only: as pointed out in a separate pre-PR code review, this is a good example of an ActiveIssue that needs to be retained when we ultimately move to Helix.  It is a true product issue on NET Native and unrelated to ToF or Xunit.  We agreed that temporarily FULLXUNIT_NOTSUPPORTED was an acceptable compromise but need to be aware to manually migrate these #if'ed out ActiveIssues ourside the #if or they will fail in Helix.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/ClientCredentialTypeTests.cs
```diff
@@ -111,13 +157,35 @@ public static void BasicAuthenticationEmptyUser_throw_ArgumentException()
...
 
-    [ConditionalFact(nameof(Domain_Joined),
-                     nameof(Root_Certificate_Installed),
-                     nameof(Digest_Authentication_Available), 
-                     nameof(Explicit_Credentials_Available))]
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+#else
+    [ConditionalFact(nameof(Server_Domain_Joined), nameof(Root_Certificate_Installed), nameof(Digest_Authentication_Available), nameof(Explicit_Credentials_Available))]
+#endif
```
roncain: Non-blocking nit: I prefer the original indented style because the single line is too long to read.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.4.0.0.cs
```diff
@@ -47,11 +47,25 @@ public static void DefaultSettings_Http_Text_Echo_RoundTrips_String()
...
+#endif
+    [OuterLoop]
     public static void DefaultSettings_Https_Text_Echo_RoundTrips_String()
     {
+#if FULLXUNIT_NOTSUPPORTED
+        bool root_Certificate_Installed = Root_Certificate_Installed();
+        if (!root_Certificate_Installed)
+        {
+            Console.WriteLine("---- Test SKIPPED --------------");
+            Console.WriteLine("Attempting to run the test in ToF, a ConditionalFact evaluated as FALSE.");
```
roncain: Non-blocking nit from pre-PR review -- I'd prefer to combine these into a single line, but agree it's okay to merge as is.

roncain:
LGTM subject to a couple things:
1. Do we have buy-off to merge this while in Escrow (@Zhenlan)?  It seems unlikely we will need to create new tests to merge back to the RTM branch at this point, but I just wanted to confirm.
2. Self-host OuterLoops are failing due to the removal of ActiveIssue(851).  I recommend we replace that with [ActiveIssue(1265)] until we can make a ConditionalFact for NegotiateStream tests that require manual setup.

iamjasonp:
NIT: A lot of this seems super boilerplate, I'm wondering if you could consolidate it into a method somehow - for example:

`bool ConditionalFact_Root_Certificate_Installed()`. 

Or perhaps you could be even more novel and have something like `EvaluateConditionalFacts (params Func<bool> evaluationFunctions)`.

Not blocking for merge.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Tcp/NetTcpBindingTests.cs
```diff
@@ -12,15 +12,28 @@ public partial class Binding_Tcp_NetTcpBindingTests : ConditionalWcfTest
     // Default settings are:
     //                         - SecurityMode = Transport
     //                         - ClientCredentialType = Windows
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+    [ActiveIssue(832)] // Windows Stream Security is not supported in NET Native
+#else
     [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Windows_Authentication_Available))]
-#if !FEATURE_NETNATIVE
     [ActiveIssue(945, PlatformID.AnyUnix)] // NegotiateStream works on Windows and Linux, but we have not yet automated ambient credential configuration in Linux
```
zhenlan: We should remove this AcitveIssue 945 and the Root_Certificate_Installed condition from this test. The Windows_Authentication_Available condition is supposed to be sufficient to skip the test when requirement is not met.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Tcp/NetTcpBindingTests.cs
```diff
@@ -12,15 +12,28 @@ public partial class Binding_Tcp_NetTcpBindingTests : ConditionalWcfTest
...
     public static void DefaultSettings_Echo_RoundTrips_String()
     {
+#if FULLXUNIT_NOTSUPPORTED
+        bool root_Certificate_Installed = Root_Certificate_Installed();
+        bool windows_Authentication_Available = Windows_Authentication_Available();
+        if (!root_Certificate_Installed || !windows_Authentication_Available)
+        {
+            Console.WriteLine("---- Test SKIPPED --------------");
+            Console.WriteLine("Attempting to run the test in ToF, a ConditionalFact evaluated as FALSE.");
+            Console.WriteLine("Root_Certificate_Installed evaluated as {0}", root_Certificate_Installed);
```
zhenlan: This test should not need root certificate installed. Do you see otherwise?

src/System.Private.ServiceModel/tests/Scenarios/Binding/Tcp/NetTcpBindingTests.cs
```diff
@@ -52,15 +65,28 @@ public static void DefaultSettings_Echo_RoundTrips_String()
...
     // SecurityMode is Transport by default with NetTcpBinding, this test explicitly sets it.
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+    [ActiveIssue(832)] // Windows Stream Security is not supported in NET Native
+#else
     [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Windows_Authentication_Available))]
-#if !FEATURE_NETNATIVE
     [ActiveIssue(945, PlatformID.AnyUnix)] // NegotiateStream works on Windows and Linux, but we have not yet automated ambient credential configuration in Linux
-#else
-    [ActiveIssue(832)] // Windows Stream Security is not supported in NET Native
```
zhenlan: Same comment as I had for `DefaultSettings_Echo_RoundTrips_String` test.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
```diff
@@ -11,10 +11,24 @@
 
 public class DuplexClientBaseTests : ConditionalWcfTest
 {
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+#else
     [ConditionalFact(nameof(Root_Certificate_Installed))]
+#endif
     [OuterLoop]
     public static void DuplexClientBaseOfT_OverHttp_Call_Throws_InvalidOperation()
```
zhenlan: Interestingly we don't have a positive test case in this scenario. If you think this is _not_ intentional, please open an issue to add a positive test case.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
```diff
@@ -278,7 +278,7 @@ public static void DuplexCallback_Throws_FaultException_ReturnsFaultedTask()
     }
 
     [Fact]
-#if FEATURE_NETNATIVE
+#if FULLXUNIT_NOTSUPPORTED
     [ActiveIssue(833)] // Not supported in NET Native
```
zhenlan: I know you have #1220 to track this... so does this test pass (can 833 be removed)? It's okay if you plan to remove 833 in another PR... just to confirm.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -374,12 +387,11 @@ public static void WebSocket_Https_Duplex_TextStreamed()
 
     [Fact]
     [OuterLoop]
-#if !FEATURE_NETNATIVE
-    [ActiveIssue(4429)] // In #CoreFX
+#if FULLXUNIT_NOTSUPPORTED
+    [ActiveIssue(526)]
+#else
     [ActiveIssue(470)]
```
zhenlan: 470 should not apply to this test as 470 only applies to wss (https) while this test is http.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -472,7 +484,10 @@ public static void WebSocket_Http_Duplex_TextStreamed()
 
     [Fact]
     [OuterLoop]
+#if FULLXUNIT_NOTSUPPORTED
+#else
     [ActiveIssue(420, PlatformID.AnyUnix)]
+#endif
```
zhenlan: Why we have #if here since we only want to disable this test on anyunix?

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -534,7 +549,10 @@ public static void WebSocket_Http_RequestReply_TextStreamed()
...
```
zhenlan: Why add #if here?

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -574,7 +592,10 @@ public static void WebSocket_Http_WSTransportUsageDefault_DuplexCallback_GuidRou
...
```
zhenlan: Why #if here?

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
```diff
@@ -614,7 +635,10 @@ public static void WebSocket_Http_WSTransportUsageAlways_DuplexCallback_GuidRoun
...
```
zhenlan: Why #if here?

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
```diff
@@ -166,15 +222,28 @@ public static void SameBinding_Soap12_EchoString()
...
     }
 
-    [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed))]
-#if FEATURE_NETNATIVE
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
     [ActiveIssue(959)] // Server certificate validation not supported in NET Native
 #else
+    [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed))]
     [ActiveIssue(959, PlatformID.AnyUnix)] // Server certificate validation not supported on Linux and OS X
```
zhenlan: According to 959, this should have been fixed for *nix already. 959 should be removed in #else.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
```diff
@@ -215,11 +284,30 @@ public static void ServerCertificateValidation_EchoString()
         }
     }
 
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
+    [ActiveIssue(1248)]
+#else
     [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed), nameof(Server_Accepts_Certificates))]
     [ActiveIssue(960, PlatformID.AnyUnix)]
```
zhenlan: 960 has been closed. It should be removed from this test.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
```diff
@@ -12,14 +12,28 @@
 
 public class IdentityTests : ConditionalWcfTest
 {
-    [ConditionalFact(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed))]
-#if FEATURE_NETNATIVE
+#if FULLXUNIT_NOTSUPPORTED
+    [Fact]
     [ActiveIssue(833)] // Not supported in NET Native
```
zhenlan: Removing 833 is a separate PR?

zhenlan:
Since we touched 4.0.0 tests, can we please run latest toolchain+4.0.0 package tests?

src/System.Private.ServiceModel/tests/Scenarios/Binding/Tcp/NetTcpBindingTests.cs
StephenBonikowsky: Done

src/System.Private.ServiceModel/tests/Scenarios/Binding/Tcp/NetTcpBindingTests.cs
StephenBonikowsky: I wasn't evaluating if the existing conditions that were being used were correct or not, I removed it.

src/System.Private.ServiceModel/tests/Scenarios/Client/ClientBase/DuplexClientBaseTests.cs
StephenBonikowsky: Because it throws on "CreateChannel" it looks to me like this is the expected behavior and there is no positive scenario for using duplex client base with http?

@iamjasonp since this looks like his test.

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.cs
StephenBonikowsky: Yeah, I was going to do that separately, as soon as I merge this PR I will validate those tests.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: removing

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: ToF does not understand conditional ActiveIssues so it just treats this as an unconditional ActiveIssue and skips it.

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: ditto

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: ditto

src/System.Private.ServiceModel/tests/Scenarios/Extensibility/WebSockets/WebSocketTests.cs
StephenBonikowsky: ditto

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
StephenBonikowsky: Ok, it would be great if whoever fixes or verifies an issue has been fixed would then update the affected tests. This PR was not intended to evaluate the veracity of every condition or active issue, however; good catch.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Tcp/IdentityTests.cs
StephenBonikowsky: yes

StephenBonikowsky:

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
StephenBonikowsky: Now fails with Issue #1295

StephenBonikowsky:

StephenBonikowsky:
@zhenlan I ran latest toolchain+4.0.0 tests. I got two failures, one was a test that should not have been added to the 4.0.0 tests, I opened Issue #1298 and the other failure is more interesting because the test should be for 4.0.0 but ConditionalFact logic that we have added is making an API call that is not supported in 4.0.0.

In my opinion neither of these issues blocks this PR and can be handled separately.

StephenBonikowsky:
All CI tests passed. Amended one last time with just the additional [ActiveIssue(1297, PlatformID.AnyUnix)] on test case HttpsTests.ClientCertificate_EchoString

PR #1277, opened 2016-06-09
Sync dir.props to core fx

StephenBonikowsky:
First commit just brings in sync with the version in corefx.
Second commit is needed to unblock our move from TFS to VSO. (for build purposes)

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -85,7 +85,9 @@
     <CoreFxExpectedPrerelease>rc3-24206-00</CoreFxExpectedPrerelease>
     <CoreClrExpectedPrerelease>rc3-24206-00</CoreClrExpectedPrerelease>
     <ExternalExpectedPrerelease>rc3-24206-00</ExternalExpectedPrerelease>
+    <WCFExpectedPrerelease>rc3-24206-00</WCFExpectedPrerelease>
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:

dir.props
```diff
@@ -82,17 +83,31 @@
     <ValidatePackageVersions>true</ValidatePackageVersions>
     <ProhibitFloatingDependencies>true</ProhibitFloatingDependencies>
     <CoreFxExpectedPrerelease>rc3-24206-00</CoreFxExpectedPrerelease>
-    <CoreFxVersionsIdentityRegex>^(?i)((System\..*)|(NETStandard\.Library)|(Microsoft\.CSharp)|(Microsoft\.Private\.PackageBaseline)|(Microsoft\.NETCore.*)|(Microsoft\.TargetingPack\.Private\.(CoreCLR|NETNative))|(Microsoft\.Win32\..*)|(Microsoft\.VisualBasic))(?&lt;!TestData)(?&lt;!System\.Data\.SqlClient)(?&lt;!System\.IO\.Compression)$</CoreFxVersionsIdentityRegex>
+    <CoreClrExpectedPrerelease>rc3-24206-00</CoreClrExpectedPrerelease>
+    <ExternalExpectedPrerelease>rc3-24206-00</ExternalExpectedPrerelease>
+    <WCFExpectedPrerelease>rc3-24206-00</WCFExpectedPrerelease>
```
iamjasonp: Question only: Do we strictly need another ExpectedPrerelease version here, or can we piggypack onto one of the existing ones? The advantage here is that we decouple ourselves, but the disadvantage is another thing to update and/or screw up :laughing:

iamjasonp:
LGTM

dir.props
StephenBonikowsky: The only way I know in which it matters is for package validation when our packages have a different version than corefx, but the only circumstance in which I have seen this be needed was just yesterday and it was because Wcf packages are still building out of TFS which hopefully won't be the case soon.

PR #1282, opened 2016-06-12
Clean packages on official and CI builds

ericstj:
This is needed because we'll be producing packages with constant stable
version and content updates.

/cc @weshaggard @jhendrixMSFT

We'll also need to update the VSO build definitions so pass BuildType=Official.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -73,7 +73,15 @@
 
   <UsingTask TaskName="GatherDirectoriesToRestore" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll" />
 
-  <Target Name="BatchRestorePackages">
+  <!-- by default in the release branch we will re-download packages every time -->
+  <PropertyGroup Condition="'$(SkipCleanPackages)' == ''">
+    <BatchRestorePackagesDepensOn>
+      CleanPackages;
+      CleanPackagesCache;
+      $(BatchRestorePackagesDepensOn);
+    </BatchRestorePackagesDepensOn>
+  </PropertyGroup>
+  <Target Name="BatchRestorePackages" DependsOnTargets="$(BatchRestorePackagesDepensOn)">
     <MakeDir Directories="$(PackagesDir)" Condition="!Exists('$(PackagesDir)')" /> 
 
     <Message Importance="High" Text="[$([System.DateTime]::Now.ToString('HH:mm:ss.ff'))] Restoring all packages..." />


weshaggard:
Don't we need this for more then official/CI builds? Would it be needed for local devs also?

zhenlan:
Thanks @ericstj. Any more information so I can understand what's the difference when passing BuildType with CI, Official or not passing it?

cc: @StephenBonikowsky - sounds like this will be needed for VSO build too

netci.groovy
```diff
@@ -205,14 +205,14 @@ branchList.each { branchName ->
             if (osGroupMap[os] == 'Windows_NT') {
                 newJob.with {
                     steps {
-                        batchFile("build.cmd /p:ConfigurationGroup=${configurationGroup} /p:OSGroup=${osGroupMap[os]} /p:WithCategories=OuterLoop /p:ServiceUri=%WcfServiceUri%")
+                        batchFile("build.cmd /p:ConfigurationGroup=${configurationGroup} /p:OSGroup=${osGroupMap[os]} /p:WithCategories=OuterLoop /p:ServiceUri=%WcfServiceUri% /p:BuildType:CI")
```
zhenlan: Maybe it's the same but just to confirm... should it be `p:BuildType:CI` or `p:BuildType=CI`? The latter format is used for all other parameters.

zhenlan:
>                         batchFile("C:\\Packer\\Packer.exe .\\bin\\build.pack .\\bin")

Should `/p:BuildType` be added here too?

---
Refers to: netci.groovy:281 in 501a02c. [](commit_id = 501a02cb552423959d181de48b315d425c75e53c, deletion_comment = False)

netci.groovy
```diff
@@ -285,7 +285,7 @@ branchList.each { branchName ->
             else {
```
zhenlan: >batchFile("call \"C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\vcvarsall.bat\" x86 && build.cmd /p:ConfigurationGroup=${configurationGroup} /p:OSGroup=${osGroupMap[os]}") 

I cannot add a comment there... should `/p:BuildType` be added to line 280 too?

ericstj:
> Don't we need this for more then official/CI builds? Would it be needed for local devs also?
@weshaggard I don't want local dev's to be forced to build clean every time.

I guess for now I'll just add an opt-out for devs rather than trying to get all the workflows updated.  That's simpler and less risky given the timeline.

weshaggard:
LGTM

Issue #1262, opened 2016-06-07
All TCP ExplicitUserNameAndPassword tests are failing


hongdai:
They are failing because of  https://github.com/dotnet/corefx/issues/9160.

hongdai:
Tests failed:
NegotiateStream_Tcp_Tests.NegotiateStream_Tcp_With_ExplicitUserNameAndPassword failed.
NegotiateStream_Tcp_Tests.NegotiateStream_Tcp_With_ExplicitUserNameAndPassword_With_Spn failed.
NegotiateStream_Tcp_Tests.NegotiateStream_Tcp_With_ExplicitUserNameAndPassword_With_Upn failed.

The client will throw below exception from NCL:
System.Net.Security.dll!System.Net.Security.NegoState.ThrowCredentialException(long error) Line 842 C#
 System.Net.Security.dll!System.Net.Security.NegoState.ProcessReceivedBlob(byte[] message, System.Net.LazyAsyncResult lazyResult) Line 608 C#
 System.Net.Security.dll!System.Net.Security.NegoState.ReadCallback(System.IAsyncResult transportResult) Line 758 C#
 System.Net.Security.dll!System.Net.LazyAsyncResult.Complete(System.IntPtr userToken) Line 481 C#
 System.Net.Security.dll!System.Net.LazyAsyncResult.ProtectedInvokeCallback(object result, System.IntPtr userToken) Line 431 C#
 System.Net.Security.dll!System.Net.StreamFramer.ReadFrameComplete(System.IAsyncResult transportResult) Line 292 C#
 System.Net.Security.dll!System.Net.StreamFramer.ReadFrameCallback(System.IAsyncResult transportResult) Line 171 C#
 System.Net.Security.dll!System.Threading.Tasks.TaskToApm.InvokeCallbackWhenTaskCompletes.AnonymousMethod__0() Line 137 C#
 System.Private.CoreLib.ni.dll!System.Runtime.CompilerServices.TaskAwaiter.OutputWaitEtwEvents.AnonymousMethod__0() Unknown
 System.Private.CoreLib.ni.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state) Unknown
 System.Private.CoreLib.ni.dll!System.Threading.Tasks.AwaitTaskContinuation.RunCallback(System.Threading.ContextCallback callback, object state, ref System.Threading.Tasks.Task currentTask) Unknown
[...]

PR #1301, opened 2016-06-15
Disable TCP Negotiate tests using explicit credentials

roncain:
Currently the TCP NegotiateStream tests that use explicit
credentials fail.  Issue #1262 tracks the underlying
issue https://github.com/dotnet/corefx/issues/9160

These tests are currently run only manually because they
require valid user credentials.


iamjasonp:
LGTM

zhenlan:
#1262 listed 3 tests. Looks like `NegotiateStream_Tcp_With_ExplicitUserNameAndPassword_With_Upn` is disabled with wrong ActiveIssue currently.

roncain:
Good catch @zhenlan -- I missed one test that had the wrong ActiveIssue.  Fixed now.  These tests can only be run manually.

zhenlan:
LGTM

> These tests can only be run manually.
It's my understanding even if we run those tests manually, they will fail. That's why we use ActiveIssue here.

Issue #1102, opened 2016-04-29
Update/rename BridgeCertificateInstaller for cross-platform installation of certificates on clients


roncain:
The existing BridgeCertificateInstaller is a cross-platform exe that is currently run from the build.sh on platforms other than Windows.  It needs to be renamed (no "Bridge") and updated to match what is happening in the new test infrastructure CertificateGenerator.exe.  It should also be relocated under the tools folder to match where CertificateGenerator lives.

Success is determined by all OuterLoop tests with [ConditionalFact] for certificates both executing and passing on Linux when run via ./build.sh.  Tests skipped for other reasons (ex: Kerberos_Available) are still expected to be skipped.

zhenlan:
Thanks @roncain for opening the issue. A couple questions
 
1. Is this tool currently broken if we don't change anything? This helps to understand the urgency.
2. Is this tool only used for Linux? If yes, should we consider to make it x-plat so the same tool is used on Windows, Mac etc. too?

roncain:
Prior to the new test infrastructure, running the OuterLoops on any *nix machine required the manual execution of BridgeCertificateInstaller.sh.  The script invoked an x-plat exe to ask the Bridge for the CA cert, and then the script installed the cert on the machine.  Without the Bridge, the exe will fail, and the .sh file will not install the cert.  Without this cert, all the Https will fail.  I have only run this script on Linux, but I believe its intent was that it could run on any *nix.  The intent of opening this issue is to regain the ability to install certs on *nix machines by asking the new test infrastructure cert endpoint for the cert.

/cc @iamjasonp @zhenlan

zhenlan:
Thanks @roncain for the clarification.

> Without this cert, all the Https will fail.

I understand client cert related tests will fail, but Https should pass...

Looks like this [line](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/tools/test/BridgeCertificateInstaller/CertificateInstaller.cs#L76) in CertificateInstaller.cs needs to be updated with the new URL that provides cert download. @hongdai, could you please provide the new URL?

PR #1313, opened 2016-06-21
Automate certificate install for OSX, other *nixes

iamjasonp:
Adds support to BridgeCertificateInstaller.sh to install root certificates on non-Ubuntu OSes

### DIFF
--- a/src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateInstaller.sh
+++ b/src/System.Private.ServiceModel/tools/setupfiles/BridgeCertificateInstaller.sh
@@ -34,28 +34,64 @@ run_installer()
 install_root_cert()
 {
     __cafile=$1
-
     chmod a+r "$__cafile"
-    cp -f "$__cafile" /usr/local/share/ca-certificates
 
-    if [ $? -ne 0 ]; then 
-        return $? 
-    fi 
-
+    # Find the command for updating the OS certificate bundle
     echo "Updating root certificates with cert from '$__cafile'"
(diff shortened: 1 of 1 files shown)


zhenlan:
Nice! @iamjasonp, can you please move the script to `tools/scripts` folder and clean up the `Bridge` in naming etc.?

roncain:
LGTM, subject to @Zhenlan 's comments to move to scripts and eliminate Bridge naming in this file. Nice to see these changes!
Question only: what steps are required before this script can be run? I typically needed to do a full outerloop run to get everything built before I could run the prior one. Not blocking for this PR.

iamjasonp:
Yes, due to the changes in build.sh (cmd), we have to do a full outerloop build before this command can run. I can put those instructions into the script usage info if it makes sense

iamjasonp:
As discussed f2f with @zhenlan, we will: 

- merge this change without the "bridge" name cleanup, and then, 
- take the cleanup as a separate PR, to prevent muddling of the changes

There are a good number of references to "Bridge" which I will clean up at the same time - this will include things like the projects, names, and the like which will require a bit more effort. 

I want to push this through to unblock anyone else who might be waiting :smile:

iamjasonp:
#1102 tracks renaming the Bridge to... whatever 😅 . I suspect a naming war will ensue in the next PR

roncain:
LGTM to push this through as-is and deal with #1102 separately.

iamjasonp:
Merging - this has been tested with Ubuntu, OS X, CentOS

Issue #1205, opened 2016-05-23
Server side code should not log to the console.


StephenBonikowsky:
Some server side code here...
...\wcf\src\System.Private.ServiceModel\tools\IISHostedWcfService\App_code\WcfChannelExtensibilityService.cs
Is logging to the Console, this should be removed.

roncain:
@StephenBonikowsky -- we need to merge PR #1325 before tackling this issue, because it alters the MessageInspector tests to not use a flawed workaround.  At that point, I would like to broaden this current issue to include cleaning up the MessageInspector test, probably like this:
- Remove the console logging on the server side
- Rather than write to console, make the server side return the information it would have logged (not the silly messages -- just the wind speeds that made it through the intercepting channel)
- Make the MessageInterceptor test verify the data returned from the server matched what it expected.

Optional: I might also consider making the custom MessageInterceptor used in this test set some flags or something like that to indicates its methods were called, and the test can verify that those methods were called when the message was sent.

Feel free to reassign to me.

StephenBonikowsky:
Fixed by https://github.com/dotnet/wcf/pull/1394

PR #1325, opened 2016-06-23
Fix issue where OnOpen and OnClose were not called on custom channels and factories

### LINKED ISSUE #1241: CustomBinding_Message_Interceptor test shows we can modify communication object while it's in Opened state
This test is passing in .netcore. However, it fails in Desktop because the test is modifying the communication object when it's already "Opened". 

In Desktop, it fails with below exception:
The communication object, System.ServiceModel.Channels.HttpChannelFactory`1[System.ServiceModel.Channels.IRequestChannel], cannot be modified while it is in the Opened state.

at System.ServiceModel.Channels.CommunicationObject.ThrowIfDisposedOrImmutable()
   at System.ServiceModel.Channels.CommunicationObject.Open(TimeSpan timeout)
   at InterceptingChannelFactory`1.OnOpen(TimeSpan timeout) in E:\test\GenerateTests\GenerateTests\scenarios\Extensibility\MessageInterceptor\InterceptingChannelFactory.cs:line 48
   at System.ServiceModel.Channels.CommunicationObject.Open(TimeSpan timeout)
   at System.ServiceModel.Channels.ServiceChannelFactory.TypedServiceChannelFactory`1.OnOpen(TimeSpan timeout)
   at System.ServiceModel.Channels.CommunicationObject.Open(TimeSpan timeout)
   at System.ServiceModel.ChannelFactory.OnOpen(TimeSpan timeout)
   at System.ServiceModel.Channels.CommunicationObject.Open(TimeSpan timeout)
   at System.ServiceModel.ChannelFactory.EnsureOpened()
   at System.ServiceModel.ChannelFactory`1.CreateChannel(EndpointAddress address, Uri via)
   at System.ServiceModel.ChannelFactory`1.CreateChannel()
   at MessageInterceptorTests.CustomBinding_Message_Interceptor() in E:\test\GenerateTests\GenerateTests\scenarios\Extensibility\MessageInterceptor\MessageInterceptorTests.cs:line 49
   at ConsoleApplication1.Program.RunTests() in E:\test\GenerateTests\GenerateTests\Program.cs:line 37

roncain:
The MessageInterceptor scenario test exposed a product issue where we were not correctly invoking OnOpen or OnClose on custom channels or channel factories.  The test also used a workaround to get past the product issue, and this workaround failed when run on the full framework.

This PR fixes the product issue and removes the workarounds from the MessageInterceptor test.

Fixes #1241

### DIFF
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
@@ -67,8 +67,6 @@ namespace System.IdentityModel.Selectors
 
             public ChainTrustValidator(bool useMachineContext, X509ChainPolicy chainPolicy, uint chainPolicyOID)
             {
-                Contract.Assert(useMachineContext == false, "CoreCLR does not have ctor allowing useMachineContext = true");
-
                 _useMachineContext = useMachineContext;
                 _chainPolicy = chainPolicy;
                 _chainPolicyOID = chainPolicyOID;
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
@@ -758,14 +758,22 @@ namespace System.ServiceModel.Channels
 
         internal protected virtual Task OnCloseAsync(TimeSpan timeout)
         {
-            Contract.Requires(false, "OnCloseAsync needs to be implemented on derived classes");
-            return TaskHelpers.CompletedTask();
+            // Derived types aware of IAsyncCommunicationObject should override OnCloseAsync.
+            // However, because IAsyncCommunicationObject is internal, external implementations
+            // such as custom channels and factories cannot override it, yet still expect their OnClose
+            // logic to be called.  Moreover, we cannot know BeginClose is implemented, so run the
+            // synchronous Close asynchronously.
+            return Task.Run(() => OnClose(timeout));
         }
 
         internal protected virtual Task OnOpenAsync(TimeSpan timeout)
         {
-            Contract.Requires(false, "OnOpenAsync needs to be implemented on derived classes");
-            return TaskHelpers.CompletedTask();
+            // Derived types aware of IAsyncCommunicationObject should override OnOpenAsync.
+            // However, because IAsyncCommunicationObject is internal, external implementations
+            // such as custom channels and factories cannot override it, yet still expect their OnOpen
+            // logic to be called.  Moreover, we cannot know BeginOpen is implemented, so run the
+            // synchronous Open asynchronously.
+            return Task.Run(() => OnOpen(timeout));
         }
     }
 
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
@@ -5,6 +5,7 @@
 
 using System.Diagnostics;
 using System.Diagnostics.Contracts;
+using System.Reflection;
 using System.Runtime;
 using System.Threading.Tasks;
 using System.ServiceModel.Diagnostics;
@@ -26,6 +27,8 @@ namespace System.ServiceModel.Channels
         private bool _traceOpenAndClose;
         private object _eventSender;
         private CommunicationState _state;
+        private bool _onOpenAsyncCalled;
+        private bool _onCloseAsyncCalled;
 
         protected CommunicationObject()
             : this(new object())
@@ -758,22 +761,46 @@ namespace System.ServiceModel.Channels
 
         internal protected virtual Task OnCloseAsync(TimeSpan timeout)
         {
-            // Derived types aware of IAsyncCommunicationObject should override OnCloseAsync.
-            // However, because IAsyncCommunicationObject is internal, external implementations
-            // such as custom channels and factories cannot override it, yet still expect their OnClose
-            // logic to be called.  Moreover, we cannot know BeginClose is implemented, so run the
-            // synchronous Close asynchronously.
-            return Task.Run(() => OnClose(timeout));
+#if !FEATURE_NETNATIVE  // Avoid Reflection in NET Native and rely on CoreCLR to catch this problem.
+            // All product types are required to override this method
+            Contract.Assert(String.IsNullOrEmpty(GetType().Namespace) || 
+                            !GetType().Namespace.StartsWith("System.ServiceModel"),
+                            String.Format("Type '{0}' is required to override OnCloseAsync", GetType()));
+#endif
+
+            // If we have already executed this code path, just return a completed Task.
+            // It means this type did not override OnCloseAsync or OnBeginClose.
+            if (_onCloseAsyncCalled)
+            {
+                return TaskHelpers.CompletedTask();
+            }
+
+            // External types cannot override this method because it is not public.
+            // Redirect this call to the APM path.
+            _onCloseAsyncCalled = true;
+            return Task.Factory.FromAsync(OnBeginClose, OnEndClose, timeout, TaskCreationOptions.RunContinuationsAsynchronously);
         }
 
         internal protected virtual Task OnOpenAsync(TimeSpan timeout)
         {
-            // Derived types aware of IAsyncCommunicationObject should override OnOpenAsync.
-            // However, because IAsyncCommunicationObject is internal, external implementations
-            // such as custom channels and factories cannot override it, yet still expect their OnOpen
-            // logic to be called.  Moreover, we cannot know BeginOpen is implemented, so run the
-            // synchronous Open asynchronously.
-            return Task.Run(() => OnOpen(timeout));
+#if !FEATURE_NETNATIVE  // Avoid Reflection in NET Native and rely on CoreCLR to catch this problem.
+            // All product types are required to override this method
+            Contract.Assert(String.IsNullOrEmpty(GetType().Namespace) || 
+                            !GetType().Namespace.StartsWith("System.ServiceModel"),
+                            String.Format("Type '{0}' is required to override OnOpenAsync", GetType()));
+#endif
+
+            // If we have already executed this code path, just return a completed Task.
+            // It means this type did not override OnOpenAsync or OnBeginOpen.
+            if (_onOpenAsyncCalled)
+            {
+                return TaskHelpers.CompletedTask();
+            }
+
+            // External types cannot override this method because it is not public.
+            // Redirect this call to the APM path.
+            _onOpenAsyncCalled = true;
+            return Task.Factory.FromAsync(OnBeginOpen, OnEndOpen, timeout, TaskCreationOptions.RunContinuationsAsynchronously);
         }
     }
 
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
@@ -226,14 +226,20 @@ namespace System.ServiceModel.Channels
             CleanupServerCertificate();
         }
 
+        protected internal override Task OnCloseAsync(TimeSpan timeout)
+        {
+            OnClose(timeout);
+            return TaskHelpers.CompletedTask();
+        }
+
         protected override IAsyncResult OnBeginClose(TimeSpan timeout, AsyncCallback callback, object state)
         {
-            throw ExceptionHelper.PlatformNotSupported("SslStreamSecurityUpgradeProvider async path");
+            return OnCloseAsync(timeout).ToApm(callback, state);
         }
 
         protected override void OnEndClose(IAsyncResult result)
         {
-            throw ExceptionHelper.PlatformNotSupported("SslStreamSecurityUpgradeProvider async path");
+            result.ToApmEnd();
         }
 
         private void SetupServerCertificate(SecurityToken token)
@@ -275,14 +281,20 @@ namespace System.ServiceModel.Channels
             }
         }
 
+        protected internal override Task OnOpenAsync(TimeSpan timeout)
+        {
+            OnOpen(timeout);
+            return TaskHelpers.CompletedTask();
+        }
+
         protected override IAsyncResult OnBeginOpen(TimeSpan timeout, AsyncCallback callback, object state)
         {
-            throw ExceptionHelper.PlatformNotSupported("SslStreamSecurityUpgradeProvider async path");
+            return OnOpenAsync(timeout).ToApm(callback, state);
         }
 
         protected override void OnEndOpen(IAsyncResult result)
         {
-            throw ExceptionHelper.PlatformNotSupported("SslStreamSecurityUpgradeProvider async path");
+            result.ToApmEnd();
         }
     }
 
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/StreamedFramingRequestChannel.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/StreamedFramingRequestChannel.cs
@@ -39,14 +39,19 @@ namespace System.ServiceModel.Channels
             get { return _startBytes; }
         }
 
+        protected internal override Task OnOpenAsync(TimeSpan timeout)
+        {
+            return TaskHelpers.CompletedTask();
+        }
+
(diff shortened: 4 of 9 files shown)


roncain:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -758,14 +758,22 @@ internal void ThrowPending()
 
         internal protected virtual Task OnCloseAsync(TimeSpan timeout)
         {
-            Contract.Requires(false, "OnCloseAsync needs to be implemented on derived classes");
-            return TaskHelpers.CompletedTask();
+            // Derived types aware of IAsyncCommunicationObject should override OnCloseAsync.
+            // However, because IAsyncCommunicationObject is internal, external implementations
+            // such as custom channels and factories cannot override it, yet still expect their OnClose
+            // logic to be called.  Moreover, we cannot know BeginClose is implemented, so run the
```
roncain: Reviewers: FYI, I experimentally made this logic call OnBeginOpen and OnBeginClose via Task.Factory.FromAsync, which logically makes more sense.  However we have some communication object's that throw PNSE from their async paths (e.g. SslStreamSecurityUpgradeProvider).  If we believe calling OnBeginOpen and OnBeginClose is the better fix, I will need to augment this PR to fix our own PNSE's.  If there is a chance customer's custom channels also throw from their async paths, the fix here to call the sync OnOpen and OnClose might be the most conservative fix.  Thoughts welcome.

roncain:

roncain:
Reviewers: the MessageInterceptor test still needs to be rewritten because it really isn't doing any validation other than writing to the console.  Issue #1205 is meant to cleanup the console logging, and I would prefer to broaden that issue to rewrite the test to do proper validation.  I'd prefer to keep test cleanup out of this PR because it is really a product fix.

roncain:

roncain:
Note: The Ubuntu failures look like failures in build tools restoration and are not due to this PR.

roncain:

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
```diff
@@ -67,8 +67,6 @@ public ChainTrustValidator()
 
             public ChainTrustValidator(bool useMachineContext, X509ChainPolicy chainPolicy, uint chainPolicyOID)
             {
-                Contract.Assert(useMachineContext == false, "CoreCLR does not have ctor allowing useMachineContext = true");
```
roncain: Reviewers: @iamjasonp confirmed this Assert was no longer needed, and in fact must be removed because it overconstrains and fires.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -758,14 +767,48 @@ internal void ThrowPending()
 
         internal protected virtual Task OnCloseAsync(TimeSpan timeout)
         {
-            Contract.Requires(false, "OnCloseAsync needs to be implemented on derived classes");
```
roncain: Reviewers: for the record, this Contract.Requires was originally intended to alert us wherever we failed to override OnXXXAsync.  But apparently this does not fire, even in Debug builds.  A Contract.Assert likely would have fired.  Because it really is a serious issue if we don't override, I converted this to a hard faulted Task.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -226,6 +226,20 @@ protected override void OnClose(TimeSpan timeout)
             CleanupServerCertificate();
         }
 
+        protected internal override Task OnCloseAsync(TimeSpan timeout)
```
roncain: Reviewers: this newly added code is to address the issue that some of our types were not overriding these methods and hit the CommunicationObject base class -- which is an error for us.  Please review the code for how I have implemented these methods.  Unfortunately, I am using Task.Run to call synchronous code that has no async equivalent.  Creating async equivalents is a bigger redesign I would prefer not to do here.

roncain:
Reviewers: OS X failed because the machine went offline.  Ubuntu release failed for the known flakey streamed test.  I don't believe this PR is responsible.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -37,6 +39,13 @@ protected CommunicationObject(object mutex)
             _mutex = mutex;
             _eventSender = this;
             _state = CommunicationState.Created;
+
+            // All WCF product types fully implement IAsyncCommunicationObject.
+            // But because the interface is internal, custom CommunicationObjects cannot override
+            // any of its methods.  The heuristic to detect this is based on the type's namespace
+            // belonging to System.ServiceModel.
+            string ns = this.GetType().Namespace;
+            _canUseIAsyncCommunicationObject = !String.IsNullOrEmpty(ns) && ns.StartsWith("System.ServiceModel");
```
roncain: Reviewers: after discussion with @mconnew, we decided on this approach to detect "external" CommunicationObjects.  Unfortunately, IAsyncCommunicationObject is internal and external customers cannot override its methods.  But because our base classes implement it, we can't distinguish simply by testing whether the type implements IAsyncCommunicationObject -- because that would be true even for external types.  And we wanted to avoid a deeper probe into MethodInfo's or we would need to include metadata in NET Native.

I can add deeper Debug-only-on-CoreCLR testing if we want to verify this logic is enough.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
iamjasonp: Thanks Ron

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -758,14 +767,48 @@ internal void ThrowPending()
 
         internal protected virtual Task OnCloseAsync(TimeSpan timeout)
         {
-            Contract.Requires(false, "OnCloseAsync needs to be implemented on derived classes");
-            return TaskHelpers.CompletedTask();
+            // Derived classes are required to override this method because we
+            // use these Task-based API's exclusively with the product.
+            return Task.FromException(new InvalidOperationException(
```
mconnew: You could just throw. The difference would be functionally equivalent to the difference between an exception being thrown before or after an await which completed asynchronously. I don't think it matters much as we shouldn't be hitting this code anyway and the exception message will tell you exactly where it came from.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -758,14 +767,48 @@ internal void ThrowPending()
...
+        internal Task CallOnCloseAsync(TimeSpan timeout)
+        {
+            if (_canUseIAsyncCommunicationObject)
+            {
+                return OnCloseAsync(timeout);
+            }
+
+            // This type must be a custom CommunicationObjecct, such as
+            // a custom channel or channel factory.  Because it cannot
+            // have implemented OnCloseAsync, redirect to the synchronous version.
```
mconnew: Nit:..redirect to the APM version

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -808,7 +851,7 @@ public static void OnClose(CommunicationObject communicationObject, TimeSpan tim
 
         public static IAsyncResult OnBeginClose(CommunicationObject communicationObject, TimeSpan timeout, AsyncCallback callback, object state)
         {
-            return communicationObject.OnCloseAsync(timeout).ToApm(callback, state);
+            return communicationObject.CallOnCloseAsync(timeout).ToApm(callback, state);
```
mconnew: Could end up in an infinite loop if the caller doesn't implement the APM methods either. Probably want to only call CallOnCloseAsync if _canUseIAsyncCommunicationObject otherwise have the call terminating with Task.Completed.ToApm(callback, state); The same on the other calls.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -226,6 +226,20 @@ protected override void OnClose(TimeSpan timeout)
...
+        }
+
+        private async Task OnCloseAsyncInternal(TimeSpan timeout)
+        {
+            if (_clientCertificateAuthenticator != null)
+            {
+                await Task.Run(() => SecurityUtils.CloseTokenAuthenticatorIfRequired(_clientCertificateAuthenticator, timeout));
+            }
+            CleanupServerCertificate();
+        }
```
mconnew: There's no need to do anything async artificially. Just do everything synchronously and return Task.CompletedTask

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -275,6 +289,32 @@ protected override void OnOpen(TimeSpan timeout)
...
 
+        protected internal override Task OnOpenAsync(TimeSpan timeout)
+        {
+            return OnOpenAsyncInternal(timeout);
+        }
+
+        private async Task OnOpenAsyncInternal(TimeSpan timeout)
+        {
+            TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
+            await Task.Run(() =>
```
mconnew: No need to use a Task, just run the code.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -275,6 +289,32 @@ protected override void OnOpen(TimeSpan timeout)
...
+                SecurityUtils.OpenTokenAuthenticatorIfRequired(ClientCertificateAuthenticator, timeoutHelper.RemainingTime()));
+
+            if (_serverTokenProvider != null)
+            {
+                await Task.Run(() =>
+                    SecurityUtils.OpenTokenProviderIfRequired(_serverTokenProvider, timeoutHelper.RemainingTime()));
```
mconnew: Same as above

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SslStreamSecurityUpgradeProvider.cs
```diff
@@ -275,6 +289,32 @@ protected override void OnOpen(TimeSpan timeout)
...
+                using (CancellationTokenSource cts = new CancellationTokenSource(timeout))
+                {
+                    SecurityToken token = await _serverTokenProvider.GetTokenAsync(cts.Token);
+                    SetupServerCertificate(token);
+                    await Task.Run(() => 
+                        SecurityUtils.CloseTokenProviderIfRequired(_serverTokenProvider, timeoutHelper.RemainingTime()));
```
mconnew: Same as above

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/StreamedFramingRequestChannel.cs
```diff
@@ -39,14 +39,19 @@ private byte[] Preamble
...
 
+        protected internal override Task OnOpenAsync(TimeSpan timeout)
+        {
+            return TaskHelpers.CompletedTask();
+        }
+
         protected override IAsyncResult OnBeginOpen(TimeSpan timeout, AsyncCallback callback, object state)
         {
-            return new CompletedAsyncResult(callback, state);
+            throw ExceptionHelper.PlatformNotSupported("StreamedFramingRequestChannel async path");
```
mconnew: What happens if a derived class implements OnBeginOpen and calls base.OnBeginOpen()?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/WindowsStreamSecurityUpgradeProvider.cs
```diff
@@ -149,15 +154,30 @@ protected override void OnOpen(TimeSpan timeout)
...
+
+        private async Task OnOpenAsyncInternal(TimeSpan timeout)
+        {
+            if (!_isClient)
+            {
+                SecurityTokenRequirement sspiTokenRequirement = TransportSecurityHelpers.CreateSspiTokenRequirement(Scheme, _listenUri);
+                _serverCredential = await Task.Run(() => 
+                    TransportSecurityHelpers.GetSspiCredential(_securityTokenManager, sspiTokenRequirement, timeout,
+                    out _extractGroupsForWindowsAccounts));
+            }
```
mconnew: Just return Task.Completed

roncain:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -758,14 +761,46 @@ internal void ThrowPending()
 
         internal protected virtual Task OnCloseAsync(TimeSpan timeout)
         {
-            Contract.Requires(false, "OnCloseAsync needs to be implemented on derived classes");
-            return TaskHelpers.CompletedTask();
+#if !FEATURE_NETNATIVE  // Avoid Reflection in NET Native and rely on CoreCLR to catch this problem.
```
roncain: Reviewers: after experimenting with the original approach (namely: attempt to predict which types are external and cannot override the async methods), I found it too convoluted and fragile.  Plus it dragged in more metadata in NET Native.  Therefore I have changed the approach to (1) assert in CoreCLR builds that all our product code has overridden the methods, and (2) for types that cannot override them, redirect to the APM equivalents.

I think this is simpler and more robust.  Feedback welcome.

roncain:

src/System.Private.ServiceModel/src/Resources/Strings.resx
```diff
@@ -6756,8 +6756,7 @@
   <data name="SSLProtocolNegotiationFailed" xml:space="preserve">
     <value>SSL protocol negotiation failed. Requested '{0}' but was offered '{1}'.</value>
   </data>
-  <!-- Copied from System.Net.Primitives net_ssl_io_cert_validation as we're doing the Ssl certificate validation ourselves for UWP apps-->
   <data name="ssl_io_cert_validation" xml:space="preserve">
-    <value>The remote certificate is invalid according to the validation procedure.</value> 
+    <value>The remote certificate is invalid according to the validation procedure.</value>
```
roncain: Rebase mistake -- will revert.

roncain:
Reviewers: this has been tested in NET Native too now

src/System.Private.ServiceModel/src/Resources/Strings.resx
iamjasonp: What happened here? Is all we're looking at the whitespace?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/StreamedFramingRequestChannel.cs
```diff
@@ -175,14 +180,19 @@ protected override void OnClosed()
...
+
         protected override IAsyncResult OnBeginClose(TimeSpan timeout, AsyncCallback callback, object state)
         {
-            return base.BeginWaitForPendingRequests(timeout, callback, state);
+            return OnCloseAsync(timeout).ToApm(callback, state);
         }
 
         protected override void OnEndClose(IAsyncResult result)
         {
-            base.EndWaitForPendingRequests(result);
```
mconnew: As you are removing the call to the base APM method and instead calling in to the Async version, the base APM method(s) should be removed.

mconnew:
LGTM, although I would prefer cleaning up the now-defunct code in {Begin|End}WaitForPendingRequests

roncain:

roncain:

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/StreamedFramingRequestChannel.cs
roncain: I removed the unused APM methods from RequestChannel

PR #1333, opened 2016-06-24
Add condition to proj ref of S.P.SM

StephenBonikowsky:
* This project reference is causing strange build behavior in the VSO build pipeline. Since in that environment it is supposed to build test using the actual packages anyway these project references should not be used.


StephenBonikowsky:
Adding @ericstj  and @chcosta

StephenBonikowsky:

StephenBonikowsky:
I tested this in a private branch on VSO and it worked.

iamjasonp:
LGTM, given that you've tested this in VSO.

ericstj:
I think you should have a look at the way we're doing test > product references in CoreFx now using pkgprojs.  Also if you already have a targets file defining the path to the S.P.SM project, why not avoid a ton of duplication and put the item in the targets file as well, conditioned of course.  Not needed for this PR, but something to condsider.

PR #1341, opened 2016-06-28
Updated all wcf packages references to Beta-24228-04

StephenBonikowsky:
* Also had to update the minor version for each wcf package.


roncain:
LGTM.
The 2 Ubuntu failures are for a test that already has an ActiveIssue on AnyUnix, so they should not have run.  I suspect these would to away if you rebased, but you don't need to do that to merge.

StephenBonikowsky:
Added unconditional ActiveIssue on that test.

PR #1346, opened 2016-07-01
Peertrust

### LINKED ISSUE #1183: Enable X509CertificateValidationMode PeerTrust
Currently X509CertificateValidationMode enums PeerTrust and PeerOrChainTrust throw PlatformNotSupported from System.ServiceModel.Security.X509ServiceCertificateAuthentication.TryGetCertificateValidator

There is no known reason why it shouldn't be supported.

roncain:
Enable X509CertificateValidationMode PeerTrust

Add back to the product the X509CertificateValidator support for PeerTrust.
This code had been removed on the initial port from the full framework.

This also requires new test support to fetch the peer trust certificate
and install it into the client's TrustedPeople store so that the PeerTrust validation
logic can find it.

This also adds a new ConditionalFact to indicate that the peer trust certificate
has been successfully installed into the TrustedPeople store.

Fixes #1183


src/System.Private.ServiceModel/src/Resources/Strings.resx
```diff
@@ -6756,8 +6756,16 @@
   <data name="SSLProtocolNegotiationFailed" xml:space="preserve">
     <value>SSL protocol negotiation failed. Requested '{0}' but was offered '{1}'.</value>
   </data>
-  <!-- Copied from System.Net.Primitives net_ssl_io_cert_validation as we're doing the Ssl certificate validation ourselves for UWP apps-->
   <data name="ssl_io_cert_validation" xml:space="preserve">
-    <value>The remote certificate is invalid according to the validation procedure.</value> 
+    <value>The remote certificate is invalid according to the validation procedure.</value>
```
roncain: Reviewers: this strings were copied verbatim from IdentityModel in the full framework.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
```diff
@@ -13,7 +13,9 @@ namespace System.IdentityModel.Selectors
 {
     public abstract class X509CertificateValidator
     {
+        private static X509CertificateValidator s_peerTrust;
```
roncain: Reviewers -- this code was copied verbatim from the full framework with only these changes:
- rename of fields to be consistent with new conventions
- use of X509Store rather than IdentityModel's X509CertificateStore.
- SR.Format

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/CertificateManager.cs
```diff
@@ -73,7 +73,7 @@ private static StoreLocation PlatformSpecificRootStoreLocation
                 using (X509Store store = new X509Store(storeName, storeLocation))
                 {
                     store.Open(OpenFlags.ReadOnly);
-                    resultCert = CertificateFromThumbprint(store, certificate.Thumbprint);
+                    resultCert = CertificateFromThumbprint(store, certificate.Thumbprint, validOnly: false);
```
roncain: Reviewers: I altered this to allow invalid certs to be retrieved.  We are seeing issues on Linux where the certs are considered invalid, and it caused the auto-detection to disable all cert tests.  It also meant the CI and lab machines would not be able to force the certs to be loaded by manually setting the TestProperties to enable the cert tests. I felt it was better to always allow the cert to be loaded if possible, and let tests pass or fail based on cert validity.

The real test that the certs are valid now happens in the Infrastructure scenario tests for certs.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/CertificateManager.cs
```diff
@@ -106,82 +106,45 @@ private static StoreLocation PlatformSpecificRootStoreLocation
...
-
-        // Retrieves a client certificate matching the given issuer name and subject name from the certificate store
-        public static X509Certificate2 ClientCertificateFromName(string issuerName, string subjectName)
+        public static X509Certificate2 ClientCertificateFromThumprint(string thumbprint, bool validOnly)
         {
-            return CertificateFromName(StoreName.My, StoreLocation.CurrentUser, issuerName, subjectName);
+            return CertificateFromThumbprint(StoreName.My, StoreLocation.CurrentUser, thumbprint, validOnly);
         }
 
-        private static X509Certificate2 CertificateFromName(StoreName storeName,
```
roncain: Reviewers: the find by name logic was dead code, so I eliminated it.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/CertificateManager.cs
```diff
@@ -204,5 +167,18 @@ public static X509Certificate2 InstallCertificateToMyStore(X509Certificate2 cert
 
             return certificate;
         }
+
+        // Install the certificate into the TrustedPeople store.
+        // It will not install the certificate if it is already present in the store.
+        // It returns the thumbprint of the certificate, regardless whether it was added or found.
+        public static X509Certificate2 InstallCertificateToTrustedPeopleStore(X509Certificate2 certificate)
```
roncain: Reviewers: for PeerTrust to work, it is necessary that the server's certificate be installed on the client into the TrustedPeople store for CurrentUser.  This is the code that gets called by the ConditionalFact for Server_Certificate_Installed.  It is also validated by new scenario tests.

src/System.Private.ServiceModel/tests/Scenarios/Infrastructure/SetupValidationTests.cs
```diff
@@ -5,21 +5,37 @@
 
 using Infrastructure.Common;
 using System;
+using System.Security.Cryptography.X509Certificates;
 using Xunit;
 
 public class SetupValidationTests : ConditionalWcfTest
 {
+#if FULLXUNIT_NOTSUPPORTED
```
roncain: Reviewers: I had forgotten to make these run on NET Native so am adding that support now.

src/System.Private.ServiceModel/tools/CertificateGenerator/Program.cs
```diff
@@ -56,6 +56,7 @@ private static void UninstallAllCerts()
             RemoveCertificatesFromStore(StoreName.Root, StoreLocation.CurrentUser);
 
             RemoveCertificatesFromStore(StoreName.TrustedPeople, StoreLocation.LocalMachine);
+            RemoveCertificatesFromStore(StoreName.TrustedPeople, StoreLocation.CurrentUser);
```
roncain: Reviewers: nit level, but our cert generator was not cleaning up where we put the cert for PeerTrust.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/Util.cs
```diff
@@ -103,7 +119,7 @@ public static X509Certificate2 CertificateFromSubject(StoreName name, StoreLocat
             }
         }
 
-        public static X509Certificate2 CertificateFromFridendlyName(StoreName name, StoreLocation location, string fridendlyName)
+        public static X509Certificate2 CertificateFromFriendlyName(StoreName name, StoreLocation location, string friendlyName)
```
roncain: Reviewers: nit level and not required for this product feature, but these misspellings were constantly fooling my searches for "friendlyName", so I fixed them.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
```diff
@@ -108,6 +208,44 @@ private static string GetChainStatusInformation(X509ChainStatus[] chainStatus)
                 return String.Empty;
             }
         }
+
+
```
iamjasonp: NIT: extra space

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
```diff
@@ -108,6 +208,44 @@ private static string GetChainStatusInformation(X509ChainStatus[] chainStatus)
...
+        class PeerOrChainTrustValidator : X509CertificateValidator
+        {
+            X509CertificateValidator chain;
+            PeerTrustValidator peer;
```
iamjasonp: NIT: style, should be `_chain` and `_peer`. Also remove `this.` from blocks below

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
```diff
@@ -54,6 +83,77 @@ public override void Validate(X509Certificate2 certificate)
...
+
+                // The following code could be written as:
+                // DateTime now = DateTime.UtcNow;
+                // if (now > certificate.NotAfter.ToUniversalTime() || now < certificate.NotBefore.ToUniversalTime())
+                //
+                // this is because X509Certificate2.xxx doesn't return UT.  However this would be a SMALL perf hit.
+                // I put a DebugAssert so that this will ensure that the we are compatible with the CLR we shipped with
+
+                DateTime now = DateTime.Now;
+                DiagnosticUtility.DebugAssert(now.Kind == certificate.NotAfter.Kind && now.Kind == certificate.NotBefore.Kind, "");
```
iamjasonp: Contract.Assert?

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
```diff
@@ -54,6 +83,77 @@ public override void Validate(X509Certificate2 certificate)
...
+            internal bool TryValidate(X509Certificate2 certificate, out Exception exception)
+            {
+                // Checklist
+                // 1) time validity of cert
+                // 2) in trusted people store
+                // 3) not in disallowed store
+
+                // The following code could be written as:
+                // DateTime now = DateTime.UtcNow;
+                // if (now > certificate.NotAfter.ToUniversalTime() || now < certificate.NotBefore.ToUniversalTime())
```
iamjasonp: I wonder if we should take this suggestion here... I don't like dealing with local times as it causes odd issues that are hard to diagnose.

I don't recommend us port over to desktop, but we might as well make the change here IMO. 

@mconnew, any thoughts on what perf hit we might take?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/CertificateManager.cs
iamjasonp: Validity could mean a lot of things... I tend to think that we should trust what the OS/underlying layers believe is valid or not as it helps us have a layer of security in addition to whatever checks we do here. 

I tend to take the more conservative approach here. If we leave this change in, could we add an issue to track flipping this back to `validOnly = true` ?

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ServiceUtilHelper.cs
```diff
@@ -199,6 +216,80 @@ public static void EnsureClientCertificateInstalled()
...
     }
 
+    // Tries to ensure that the server machine certificate is installed into
+    // the TrustedPeople store.  It obtains the server certificate from the service
+    // utility endpoint and either installs it or verifies that a matching
+    // one is already installed.  InvalidOperationException will be thrown
+    // if an error occurred attempting to install the certificate.  This
+    // method may be called multiple times but will attempt the installation
+    // once only.
+    public static void EnsureServerCertificateInstalled()
```
iamjasonp: I almost wonder if we should be using an entirely different certificate here rather than rely on the existing server cert. It's not going to be immediately apparent whether this works because the certificate is chain building properly, or whether the certs are in the trusted people store. 

Ideally, shouldn't the server present a certificate that _can't_ chain build to our Root CA cert? Then it would be obvious and provable that we used an otherwise-non-valid cert to authenticate to a server that would otherwise not be trusted

iamjasonp:
There are three effective changes that are packaged into this PR - not a problem, just make sure that when (if?) you do squash, make sure to not squash the non-peertrust related changes to the same commits as the peertrust+peertrust tests commit

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ServiceUtilHelper.cs
roncain: This sounds like a good discussion item.  What are your thoughts about taking it to design discussions and making a different PR for the code that generates a different cert (if that's what we decide)?  It will have an affect on the cert gen tools and could affect machine setup automation.  We might also consider what kinds of additional positive or negative tests to add.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
roncain: @iamjasonp -- what are the advantages of doing what the comment says if the cert times are local? I wonder if we risk subtle bugs or behavioral changes by diverging from desktop.

Incidentally, if the comments about certs using local time are correct, then our CertificateGenerator may be subtly  doing the wrong thing [here](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/tools/CertificateGenerator/CertificateGenerator.cs#L91) and for expired certs [here](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/tools/CertificateGenerator/Program.cs#L111).

src/System.Private.ServiceModel/tests/Common/Unit/ServiceInterfaces.cs
```diff
@@ -175,15 +175,3 @@ public interface IWcfDuplexTaskReturnCallback
     Task<Guid> ServicePingFaultCallback(Guid guid);
 }
 
-[ServiceContract]
```
roncain: Reviewers: this code was both dead and redundant with the same definition elsewhere.  Removed for cleanup.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/CertificateManager.cs
roncain: It is worth observing that Unix considers the root cert invalid when retrieving it by thumbprint, and we still don't understand why.  Issue #1347 tracks this.  But with the changes in this PR, this still allows all the cert tests to attempt to run if an "invalid" root cert is in the store.  Only one infrastructure test is disabled currently -- on Unix, the one that tries to verify that the root cert is valid is disabled by #1347.

src/System.Private.ServiceModel/tests/Common/Infrastructure/src/ServiceUtilHelper.cs
roncain: @iamjasonp -- regarding passing a cert that can't chain build to our root CA cert.  I agree in principle, but looking at code coverage, I see the PeerTrust validator "wins" and we don't get to the code path that checks chain building.  In fact, to get coverage of that, we would need to pass in a cert that *fails* peer trust but *passes* chain trust. The only way I know how to do that is to pass in a cert not in the TrustedPeople store. Just FYI.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/X509CertificateValidator.cs
iamjasonp: I poked around the Unix PAL and Windows PAL and we are using local time. In fact, the Unix PAL specifically normalizes times to local (Windows returns it as local anyway). 

https://github.com/dotnet/corefx/blob/c2c725978f726e2058c61232d6d66628ab600366/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Pal.Unix/OpenSslCertificateFinder.cs#L91 shows that they're normalizing time for reporting. 

Creation of certificates should not be impacted here - if we see a DateTime.Kind != UTC, i believe we do the right thing and convert it to UTC. Ultimately, date/time stored in a certificate looks to be UTC anyway, so what we're doing is a "convenience" function, kind of - probably a relic of how Windows APIs reported time to begin with

Based on the code I'm inspecting - I think as long as we're consistent between how we're using time at each step, we should be okay. So I think we should just leave the code as is, even though Utc would probably be the better technological solution.

roncain:
Anyone willing to sign off on this PR?
/cc @iamjasonp @mconnew @StephenBonikowsky @hongdai @Zhenlan

roncain:

StephenBonikowsky:
LGTM

iamjasonp:
LGTM

src/System.Private.ServiceModel/System.Private.ServiceModel.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
```
roncain: Reviewers: the update the SLN was necessary to accommodate the prior PR that moved the Infrastructure source project.  As a consequence of that update, VS touched several .csproj's to fix spacing and GUID's.  I prefer to take those touches so we don't see them again.

roncain:
Reverting .csproj changes VS made automatically, as they cause build breaks.

PR #1348, opened 2016-07-01
Fix Digest authentication test that should not run on NET Native

roncain:
Adds some NET Native versions of ConditionalFact checking to prevent
this test from running under NET Native.


iamjasonp:
LGTM

zhenlan:
LGTM

PR #1349, opened 2016-07-05
Compile tests net standard1.3v2

StephenBonikowsky:
Copying over the comments from the primary commit.

* Various 'shared files' have been updated with changes necessary to support this new way of determining what tests projects compile against and what TFM it runs on.
* All test project.json files have been updated with the needed changes to support the new system.
* All WCF 'Scenario' test projects have NOT been updated to reference the WCF Contract package projects. They continue to list a pre-release package of each needed WCF Contract in the their respective project.json files, then after the project is compiled the test .csproj file overwrites to the output directory the locally built System.Private.ServiceModel assembly which is what is then used for the test runtime.
* All WCF 'Unit' test projects were updated to reference the WCF Contract package projects they needed, they also needed a reference to the System.Private.ServiceModel.pkgproj. In addition for 3 of the 5 test projects references to additional CoreFx package dependencies were explicitly added to the project.json files since transitive dependencies are not being pulled in via the package project references in the test projects.
* That transitive dependencies are not being pulled in via the package references is why the Scenario test projects were not updated to reference package projects, it would then have required explicitly adding every required dependency to all the test projects which is counter to the whole point of using package references in the first place.
* I will create a new Issue for this transitive dependencies problem as soon as I know the correct repo to do it in.


BuildToolsVersion.txt
```diff
@@ -1 +1 @@
-1.0.25-prerelease-00512-01
+1.0.26-prerelease-00513-02
```
roncain: It looks like CoreFx is up to 530-02.  Should we move too?

dir.props
```diff
@@ -119,8 +119,12 @@
...
     </ValidationPattern>
     <ValidationPattern Include="xUnitExtensionsVersions">
-      <IdentityRegex>^(?i)xunit\.netcore\.extensions$</IdentityRegex>
-      <ExpectedVersion>1.0.0-prerelease-00312-04</ExpectedVersion>
+      <IdentityRegex>^(?i)Microsoft\.xunit\.netcore\.extensions$</IdentityRegex>
+      <ExpectedVersion>1.0.0-prerelease-00508-01</ExpectedVersion>
+    </ValidationPattern>
+    <ValidationPattern Include="buildToolsTestSuiteVersions">
+      <IdentityRegex>^(?i)Microsoft\.DotNet\.BuildTools\.TestSuite$</IdentityRegex>
+      <ExpectedVersion>1.0.0-prerelease-00508-01</ExpectedVersion>
```
roncain: CoreFx is currently on 520 here.  Do you know if this needs to be kept in sync with BuildToolsVersion.txt?  I sure wish CoreFx had a better way to keep these expected versions in a separate file.

roncain:
LGTM, subject to getting the CI build issues fixed.  Something is wrong with Duplex tests on some platforms:
```
error MSB4018: The "GenerateAssemblyList" task failed unexpectedly. [/mnt/j/workspace/dotnet_wcf/master/ubuntu14.04_debug_prtest/src/System.ServiceModel.Duplex/tests/System.ServiceModel.Duplex.Tests.csproj]

```

StephenBonikowsky:
Added a commit with changes to sh files from corefx, hoping these changes fix the *nix innerloop failures.

StephenBonikowsky:
@karajas @ericstj 
All the inner loops on *nix are failing because they cannot find...
Could not find file '/Users/dotnet-bot/j/workspace/dotnet_wcf/master/osx_release_prtest/bin/TestDependencies/System.ServiceModel.Duplex.Tests-OSX.AnyCPU.Release.netcoreapp1.0.dependencylist.txt'

When I run locally targeting Linux my local bin contains "System.ServiceModel.Duplex.Tests-Windows_NT.AnyCPU.Debug.netcoreapp1.0.dependencylist.txt", it looks like these files are not getting named correctly on *nix platforms.

Do you guys know where this is coming from?

src/System.ServiceModel.Duplex/tests/project.json
```diff
@@ -2,37 +2,22 @@
...
-    "linux-x64": {},
-    "opensuse.13.2-x64": {},
-    "osx.10.10-x64": {},
-    "rhel.7-x64": {},
-    "ubuntu.14.04-x64": {},
-    "ubuntu.16.04-x64": {},
-    "win7-x64": {},
-    "win7-x86": {}
+  "supports": {
+    // "coreFx.Test.netcore50": { }, // System.Net.Security has no run-time assembly compatible with win10-x86 / win10-x86-aot / win10-x64 / win10-x64-aot
```
karajas: Looks like this isn't valid json. 
eg . "_comment": "comment text goes here...",

StephenBonikowsky: Thanks Karthik!

StephenBonikowsky:
Doesn't look like the failures have anything to do with the project.json files. If they did they would be failing for me locally.
I come back to the fact that when I set /p:FilterToOSGroup=Linux locally I still get the dependencylist.txt files named as Windows. I know CoreFx builds tests on the actual Linux machines, I'm wondering if the naming of this file makes any assumptions based on where it is being built?

roncain:
Recommend: move to the latest buildtoolsversion.txt.

History: buildtools issue [821](https://github.com/dotnet/buildtools/issues/821) reported by Eric was for this exact same problem.

@karajas removed a workaround in PR [824](https://github.com/dotnet/buildtools/pull/824) specifically to fix this issue.  Our version of buildtools does not have that fix.

If I run the same script CI did on my local Ubuntu box *without* the workaround removed, I get the exact same failure we see in CI.  In fact, the TestDependencies folder exists, but it is empty.  I verified the WriteLinesToFile were blank, and hence no file was created, and GenerateAssemblyList fails.

When I manually removed the workaround in my Ubuntu's Tools/tests.targets (to match just that part of a later version of buildtools) the **failure went away** and TestDependencies had all the necessary dependencylist.txt files.  The fix is removing lines 185 to 197 in tests.targets

Naturally, I don't know what else will break moving to the latest buildtools (or at least latest enough to pick up the fix @karajas made), but I think it's our only option.  I will experiment further.

/cc @karajas @StephenBonikowsky @iamjasonp @ericstj

StephenBonikowsky:
Working on pulling in the changes you mentioned and then will try to resolve the bin-clasher errors.

StephenBonikowsky:
The three test failures on Ubuntu are the streaming time-out issue that is being investigated.

roncain:
LGTM

PR #1351, opened 2016-07-08
Fix issue with ProducerConsumerStream/MessageContent not completing after entire request has been sent

### LINKED ISSUE #1307: ServiceContractTests.BasicHttp_DefaultSettings_Echo_RoundTrips_String_StreamedRequest is flaky in ToF
This test is timing out in ToF and causing 45 other tests in the same project to fall over and not report results. 

I have disabled this test only in ToF to prevent this flakiness from impeding our ability to get good test results from Contract.Service.Tests.csproj

mconnew:
Fixes #1307

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
@@ -32,7 +32,9 @@ namespace System.ServiceModel.Channels
 
         public override void Flush()
         {
-            _dataAvail.TrySetResult(0);
+            int bytesWrittenToBuffer = _currentBuffer.BytesWritten;
+            _currentBuffer = WriteBufferWrapper.EmptyContainer;
+            _dataAvail.TrySetResult(bytesWrittenToBuffer);
         }
 
         public override int Read(byte[] buffer, int offset, int count)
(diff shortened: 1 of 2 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
```diff
@@ -20,7 +20,6 @@ internal class ProducerConsumerStream : Stream
         public ProducerConsumerStream()
         {
             _buffer = new TaskCompletionSource<WriteBufferWrapper>();
-            _dataAvail = new TaskCompletionSource<int>();
```
roncain: I'm a little concerned this field is no longer initialized here.  It is used in Flush without a null check, and because Flush is public, its possible it could be called before ReadAsync.  Unless there's a good reason, it would be better to guarantee initialization here or to null check before using it in Flush.

roncain:
LGTM.
I'd like to see a null check or Contract.Assert for _dataAvail in Flush(), but don't block on it.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ProducerConsumerStream.cs
mconnew: This would require Dispose() being called before any calls to WriteAsync for there to be an issue. This could happen if a MessageEncoder throws an exception before it ever gets to call WriteAsync so this shouldn't be a problem with our own encoders, but might be for external encoders. Although I don't like the extra allocation, I prefer it over an added null check in Flush() so I'll put that line back.

PR #1352, opened 2016-07-08
Change test infrastructure to use REST endpoints

iamjasonp:
After moving to the "new" infrastructure and ripping out all the REST endpoints, it turns out we do need them because certificate acquisition where we rely on a WCF Core client in test is incredibly finicky and prone to failure - requiring a build and subsequent run of the WCF Core application. 

We will move these endpoints back to being REST endpoints like the Bridge days, so we can now use cURL or a simple web browser to acquire certificates rather than needing a WCF client. 

This change will also remove some of the extra cruft that is no longer needed, and also rename the Util class to TestHost. 

One will acquire a certificate by using an endpoint like `http://locahost/TestHost.svc/RootCert`


src/System.Private.ServiceModel/tools/CertificateGenerator/App.config
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <appSettings>
-    <add key="testserverbase" value=""/>
-    <add key="CertExpirationInDay" value="30"/>
+    <add key="testserverbase" value="jasonpa-edge.redmond.corp.microsoft.com"/>
+    <add key="CertExpirationInDay" value="60"/>
```
iamjasonp: Oops, this change needs to be reverted

iamjasonp:
#1353 is preventing testing this properly at the moment. Working on getting to the bottom of #1353.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/TestHost.cs
```diff
@@ -16,74 +16,117 @@
...
+
+            return new MemoryStream(response);
         }
 
-        public byte[] GetRootCert(bool exportAsPem)
+        public Stream Crl(string serialNum)
+        {
+            // The test.crl is generated by the cert util tool and will not expire until the root cert expires.
+            // We should investigate if we can generate it in the run time. This is not a blocking issue.
+            string downloadFilePath = @"c:\\WCFTest\\test.crl";
```
roncain: Nit: can we bring literals like this to the top of the file?  I know you inherited this literal, and it is not part of this PR per se, but as long as we're touching this file it would be great to pull it out into a const at the top of the file.  I'm finding these buried literals all over the place.  Not blocking -- we can do this in separate cleanup PR.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/TestHost.cs
```diff
@@ -16,74 +16,117 @@
...
+            WebOperationContext.Current.OutgoingResponse.ContentType = "application/x-pkcs7-crl";
+            WebOperationContext.Current.OutgoingResponse.Headers["Content-Disposition"] = "attachment; filename=\"wcf-crl.crl\"";
+
+            return File.OpenRead(downloadFilePath);
+        }
+
+        public Stream RootCert(bool asPem)
         {
             X509Certificate2 rootCert = CertificateFromSubject(StoreName.Root, StoreLocation.LocalMachine, "DO_NOT_TRUST_WcfBridgeRootCA");
-            if (exportAsPem)
```
roncain: Aside: here is another (inherited) buried literal I want made more visible.  I'll stop mentioning them.  Let's just plan on a cleanup PR later that allows us to share these literals rather than copy&pasting them everywhere.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/TestHost.cs
```diff
@@ -93,9 +136,14 @@ public byte[] GetPeerCert(bool exportAsPem)
             }
         }
 
-        public string GetFQDN()
+        public Stream Ping()
```
roncain: Can this just be a bool?  What is the advantage of using a message as a stream?  Is it easier or harder to consume this response from curl or a browser?

src/System.Private.ServiceModel/tools/SelfHostedWcfService/Program.cs
```diff
@@ -303,10 +298,10 @@ private static void Main()
             webSocketHttpVerifyWebSocketsUsedTestServiceHostServiceHost.Open();
 
             //Start the crlUrl service last as the client use it to ensure all services have been started
-            Uri crlUrl = new Uri(string.Format("http://localhost/CrlService.svc", s_httpPort));
-            WebServiceHost host = new WebServiceHost(typeof(CrlService), crlUrl);
+            Uri utilUrl = new Uri(string.Format("http://localhost/TestHost.svc", s_httpPort));
```
roncain: Nit: utilUrl could be testHostUrl.  Not required.

src/System.Private.ServiceModel/tools/TestRootCertificateInstaller/CertificateInstaller.cs
```diff
@@ -1,137 +0,0 @@
-// Licensed to the .NET Foundation under one or more agreements.
```
roncain: Wow -- nice!  We drop the entire need for this projects now that we can directly curl the REST endpoint!

roncain:
LGTM, subject only to the question about why Ping returns a stream.  Not blocking for checkin however.

src/System.Private.ServiceModel/tests/Common/Infrastructure/ServiceUtilHelper.cs
```diff
@@ -280,49 +274,17 @@ public static void EnsurePeerCertificateInstalled()
...
-            peerCertificate = new X509Certificate2(certdata, "test", X509KeyStorageFlags.PersistKeySet);
-            peerCertificate = CertificateManager.InstallCertificateToTrustedPeopleStore(peerCertificate);
-        }
-        finally
-        {
-            CloseCommunicationObjects((ICommunicationObject)serviceProxy, factory);
-        }
-
-        return peerCertificate;
+        X509Certificate2 peerCertificate = new X509Certificate2(GetResourceFromServiceAsByteArray(PeerCertificateResource), "test", X509KeyStorageFlags.PersistKeySet);
```
roncain: When I get the dependencies working better, we fail at this line:

WindowsCryptographicException
Cannot find the requested object
   at Internal.Cryptography.Pal.CertificatePal.FromBlobOrFile(Byte[] rawData, String fileName, String password, X509KeyStorageFlags keyStorageFlags)
   at System.Security.Cryptography.X509Certificates.X509Certificate..ctor(Byte[] rawData, String password, X509KeyStorageFlags keyStorageFlags)
   at ServiceUtilHelper.InstallPeerCertificateFromServer() in c:\OSS\wcf\src\System.Private.ServiceModel\tests\Common\Infrastructure\ServiceUtilHelper.cs:line 286
   at ServiceUtilHelper.EnsurePeerCertificateInstalled() in c:\OSS\wcf\src\System.Private.ServiceModel\tests\Common\Infrastructure\ServiceUtilHelper.cs:line 236

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/TestHost.cs
```diff
@@ -93,9 +136,14 @@ public byte[] GetPeerCert(bool exportAsPem)
...
         {
-            return Dns.GetHostEntry("127.0.0.1").HostName;
+            return new MemoryStream(Encoding.UTF8.GetBytes("Service has started"));
+        }
+
+        public Stream State()
+        {
+            return new MemoryStream();
```
iamjasonp: This is just scaffolded in so that we can provide server state at a future date - such as NTLM available - as a JSON response.

We can either have a JSON response with all the required information (ping once) or individual endpoints like NtlmSupported, KerberosSupported, IsWindows to determine service state. I prefer the "ping once" approach where we obtain all relevant server state once.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/TestHost.cs
iamjasonp: We will likely not be keying off the return value, but rather off the HTTP response (i.e., we're likely to see a 50x if the server is misconfigured, a 200 if the server is set up right. 

I return a stream here because in all likelihood, the consumer of Ping is either a script (in which case, you care about the HTTP response code), or someone typing into a browser manually (in which case, a response where you can see "we work!" is useful)

An alternative is that we use `GetFqdn` as 'Ping`. However, having `Ping` is not super expensive and we can see very clearly the service's alive/not status. Thoughts, @roncain?

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/TestHost.cs
iamjasonp: Ugh :) we use pretty much exactly once (aside from the "DO_NOT_TRUST_WcfBridgeRootCA" literal) 

... which I also want to change to get rid of the "Bridge" reference

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/TestHost.cs
roncain: I'm fine with a streamed message we could echo.  You're right, the 200 is probably what we would key off anyway. Is there any useful info we could add to the message, such as the machine it is running on?  No changes required for this PR -- just questions.

roncain:
@iamjasonp -- I think you can now rebase and get past the #1353 issue.  All that remained to sign-off for me was getting Peer_Certificate_Installed to return true on Windows.  If you can solve that, this LGTM.

iamjasonp:
@dotnet-bot test outerloop ubuntu14.04 please

iamjasonp:
Merging - the issues in Ubuntu are due to build tools initialization issues unrelated to this change.

PR #1355, opened 2016-07-08
Update main SLN for new infrastructure test project location

roncain:
The Infrastructure test project was removed, and the source project
was moved in a prior PR.  This PR contains only the fix ups needed
to the main SLN file so the project can be used in VS again.


iamjasonp:
LGTM

PR #1362, opened 2016-07-11
Latest CoreFx changes to various shared scripts.

StephenBonikowsky:
* Covers numerous small changes.

### DIFF
--- a/run-test.sh
+++ b/run-test.sh
@@ -15,9 +15,10 @@ wait_on_pids()
 
 usage()
 {
-	# *** start WCF Content ***
+    # *** start WCF Content ***
     echo "Runs .NET tests on FreeBSD, Linux or OSX"
 	# *** start WCF Content ***
+    echo "Runs .NET CoreFX tests on FreeBSD, Linux, NetBSD or OSX"
     echo "usage: run-test [options]"
     echo
     echo "Input sources:"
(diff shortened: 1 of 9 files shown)


run-test.sh
```diff
@@ -336,21 +354,9 @@ do
...
-        --bridge-host)
-        BridgeHost=$2
-        export BridgeHost
+        --corefx-packages)
+        CoreFxPackages=$2
         ;;
-        --xunit-args)
-        XunitArgs=$2
-        ;;
-		# *** end WCF Content ***
```
roncain: I think we should retain this part, just in case we need to use it again.  But because we don't know where CoreFx is taking this script, it's hard to say.  Not blocking, but if you remove this part, we should probably also remove the echos at the top of the file that describe these parameters.

roncain:
LGTM.  My only comment was optional.  

@iamjasonp or @mconnew -- do either of you actively use run-tests.sh any more?  I know it was needed for private testing but it's becoming increasingly expensive to maintain as CoreFx continues to evolve it.  I don't use it myself.  Personally I think if private testing becomes necessary, it would be easier to modify run-tests.sh at that point than to keep maintaining it.

StephenBonikowsky:
Updated with minor feedback from @roncain 
Merging.

PR #1364, opened 2016-07-11
Install root certificates when running outerloops

iamjasonp:
WCF certificate root installer should run under *nix platforms and when the ServiceUri variable is specified. We also have to ensure that we run under superuser priviliges but we cannot fail the build if the command fails - we should output a warning if this is so

### DIFF
--- a/src/System.Private.ServiceModel/tools/scripts/InstallRootCertificate.sh
+++ b/src/System.Private.ServiceModel/tools/scripts/InstallRootCertificate.sh
@@ -22,8 +22,8 @@ acquire_certificate()
    
     # Need to make a call as the original user as we need to write to the cert store for the current 
     # user, not as root
-    echo "Making a call to '${__service_host}/TestHost.svc/GetRootCertificate' as user '$SUDO_USER'"
-    sudo -E -u $SUDO_USER $__curl_exe -o $__cafile "http://${__service_host}/TestHost.svc/GetRootCert?asPem=true" > /dev/null 2> /dev/null
+    echo "Making a call to '${__service_host}/TestHost.svc/RootCert' as user '$SUDO_USER'"
+    sudo -E -u $SUDO_USER $__curl_exe -o $__cafile "http://${__service_host}/TestHost.svc/RootCert?asPem=true" 
     
     return $?
 }


iamjasonp:
Reviewers: feel free to comment, but I'm hunting down the "certificate not trusted" issue that is happening on CentOS/Ubuntu.

roncain:
@iamjasonp -- thanks, this looks like a good approach for build-time certificate installation, meaning it will address the dev experience (run locally on Ubuntu) and any CI experience where building and running tests happens on the same machine.  But as I understand Helix, the building happens on a Windows machine (targeting a particular OS), and the test binaries are shipped off to a machine with that OS for separate running.  Therefore I believe this build target would not execute on the machine that ran the tests.  For that, I believe we will need something like the Diagnostics.Process API's to execute at test runtime -- probably in the same place where our Infrastructure ConditionalFact logic handles the root cert today.  Those API's would probably do exactly what you are doing here, but via Process rather then MSBuild <Exec />.

If we chose that install-at-test-runtime approach, do we also need an install-at-build-time approach?

Thoughts?

iamjasonp:
@roncain - You know - I was thinking just that over my morning coffee. You're absolutely right that there's no ability for Helix to run this build target and hence install-at-runtime is probably the correct thing to do given that we don't want to bifurcate the experience more than we need to

That said, the command to run is going to be the correct one to run - the question also then becomes where the script is going to come from if we're going to be shipping off binaries only to Helix. Hopefully we can prove that script as an extra asset in addition to our binaries. Anyway, that's for later. 

As for this change, I don't think we need to merge this PR (except for the script changes, which I'll do separately now) but I'll leave this open for the time being so I can test on this. I'm still getting unexpected certificate errors.

iamjasonp:
Closing this - we are going with the install-at-runtime approach, @hongdai is pursuing this

iamjasonp:
For future reference - the command line that was run was 

```
sudo -n $(MSBuildThisFileDirectory)/src/System.Private.ServiceModel/tools/scripts/InstallRootCertificate.sh --service-host $(ServiceUri) --cert-file /tmp/wcf-ca.crt > $(WcfRootCertificateInstallLog) 2> $(WcfRootCertificateInstallLog
```

PR #1372, opened 2016-07-15
Fixup InstallRotCertificate.sh hitting old endpoint

iamjasonp:
Three changes: 

1. We were masking errors by redirecting the result of cURL to /dev/null. I have removed the redirection
2. The endpoint GetRootCertificate was changed to RootCert
3. InstallRootCertificate.sh was edited in Windows, and so it needs to be chmodded to be executable

skip ci please
Replaces #1364


roncain:
LGTM. Love the commit name 😁 (no need to change it)

iamjasonp:
Balls :) I will go and change the commit name to be correct

roncain:
LGTM

PR #1377, opened 2016-07-19
Update the Facade test projects to use .pkgproj references.

StephenBonikowsky:
* Facade test.csproj files now use .pkgproj references.
* Facade project.json files have been cleaned-up.
* UnitTests.Common.csproj and project.json files have been cleaned-up.
* Wherever dependent packages need to be referenced use the last stable released version.

* ToDo: Add a run in VSO to run tests against latest-latest (latest dependencies/latest WCF).
* ToDo: Next PR to make same changes to Scenario test projects and related common code.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/Infrastructure.Common.csproj
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/Infrastructure.Common.csproj
@@ -24,21 +24,6 @@
     <None Include="testproperties.props" />
     <None Include="testproperties.targets" />
   </ItemGroup>
-  <Choose>
-    <When Condition="'$(HELIXBRANCH)' == ''" >
-      <ItemGroup>
-        <!-- Compile tests against the contract, but copy our local-built implementation for testing -->
-        <ProjectReference Include="$(WcfSourceProj)">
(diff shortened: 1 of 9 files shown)


iamjasonp:
LGTM

src/System.ServiceModel.Duplex/tests/System.ServiceModel.Duplex.Tests.csproj
```diff
@@ -23,9 +23,6 @@
     <None Include="project.json" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="$(WcfSourcePkgProj)">
-      <Name>System.Private.ServiceModel</Name>
-    </ProjectReference>
```
zhenlan: Based on my understanding of what @ericstj recommended, as a workaround, we need to keep the project reference to [$(WcfSourcePkgProj)](https://github.com/dotnet/wcf/blob/master/src/test.props#L4) so the tests can run against the system.private.servicemodel.dll that was just built.

StephenBonikowsky: Updated.
@ericstj could you confirm, not just the façade pkgproj but also the S.P.SM.pkgproj?

zhenlan:
LGTM

PR #1382, opened 2016-07-19
Improve PeerTrust scenario tests

roncain:
This adds a new certificate that can be used for PeerTrust,
new test endpoints that use it, and new scenario tests that
test PeerTrust, PeerOrChainTrust and ChainTrust.


src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
```diff
@@ -21,7 +21,11 @@ public partial class HttpsTests : ConditionalWcfTest
                      nameof(SSL_Available))]
 #endif
     [OuterLoop]
-    public static void Https_SecModeTrans_ClientCredTypeNone_ServerCertValModePeerTrust_EchoString()
+    [ActiveIssue(1381)] // requires additional SSL configuration of new port to work
```
roncain: Reviewers: I'm adding this test but also disabling it with an ActiveIssue pending further discussion.  As far as I can tell, to get a different cert working end to end with https could require using a different port and therefore changes to our setup scripts, etc.  If we see sufficient value in this test, we can pursue it, otherwise I'll remove it.  The TCP equivalent of these tests has an end-to-end peer cert positive test that covers that part of the product.

iamjasonp: :+1: I'm ok with keeping this test disabled for now

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
```diff
@@ -78,31 +84,38 @@ public static void Https_SecModeTrans_ClientCredTypeNone_ServerCertValModePeerTr
...
+                     nameof(Peer_Certificate_Installed),
                      nameof(SSL_Available))]
 #endif
     [OuterLoop]
-    public static void Https_SecModeTrans_ClientCredTypeNone_ServerCertValModePeerOrChainTrust_EchoString()
+    // Asking for PeerTrust alone should throw SecurityNegotiationException
+    // if the certificate is not in the TrustedPeople store.  For this test
+    // we use a valid chain-trusted certificate that we know is not in the
+    // TrustedPeople store.
+    public static void Https_SecModeTrans_CertValMode_PeerTrust_Fails_Not_In_TrustedPeople()
```
iamjasonp: Is this ActiveIssue 1381 as well?

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
```diff
@@ -111,13 +124,91 @@ public static void Https_SecModeTrans_ClientCredTypeNone_ServerCertValModePeerOr
...
+        CommunicationException communicationException = null;
 
         try
         {
             // *** SETUP *** \\
             BasicHttpsBinding binding = new BasicHttpsBinding(BasicHttpsSecurityMode.Transport);
             binding.Security.Transport.ClientCredentialType = HttpClientCredentialType.None;
-            endpointAddress = new EndpointAddress(new Uri(Endpoints.Https_SecModeTrans_ClientCredTypeNone_ServerCertValModePeerTrust_Address));
+            endpointAddress = new EndpointAddress(new Uri(
+                                Endpoints.Https_SecModeTrans_ClientCredTypeNone_ServerCertValModePeerOrChainTrust_Address));
```
iamjasonp: Is this endpoint correct? 
The test is named Https_SecModeTrans_CertValMode_PeerTrust_Fails_Not_In_TrustedPeople
but you're using a cert that is capable of doing PeerOrChainTrust?

src/System.Private.ServiceModel/tools/CertificateGenerator/CertificateManager.cs
```diff
@@ -186,6 +199,16 @@ public static X509Certificate2 CreateAndInstallLocalMachineCertificates(Certific
...
+                // Create the PeerTrust cert
+                certificateCreationSettings = new CertificateCreationSettings()
+                {
+                    FriendlyName = "WCF Bridge - UserPeerTrustCertificateResource",
+                    Subject = fqdn,
+                    SubjectAlternativeNames = new string[] { fqdn, hostname, "localhost" }
+                };
+                var peerCert = certificateGenerator.CreateMachineCertificate(certificateCreationSettings).Certificate;
+                InstallCertificateToTrustedPeopleStore(peerCert, certificateCreationSettings.ValidityType == CertificateValidityType.Valid);
             }
```
iamjasonp: I don't think this helps us with the case where we have peer-only trust, right? As in, this cert will still be chain trusted because it will build up to the root cert we install to the system

iamjasonp: We may need to just create a new self-signed cert that doesn't chain up to the CA cert in any way. Will think on how to achieve this

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
roncain: No -- this is simply a negative test to demonstrate that asking for PeerTrust only will fail if the cert is not in TrustedPeople.  From the standpoint of these tests, this exercises the code path that rejects such certificates in the X509CertificateValidator.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
roncain: The name of the endpoint caused confusion, so I am renaming.  That endpoint uses a cert guaranteed to be chain trusted and also not in TrustedPeople.  I'll remove the "PeerOr" from the name of the endpoint and its matching service host.

---
In reply to: [71626389](https://github.com/dotnet/wcf/pull/1382#discussion_r71626389) [](ancestors = 71626389)

src/System.Private.ServiceModel/tools/CertificateGenerator/CertificateManager.cs
roncain: Well here is the problem -- let's brainstorm what we want: The intent is that we want at least one cert in TrustedPeople so we can exercise the PeerTrust-only path.  I chose to make it chain-trusted as well so that we could test the ChainTrust path. Therefore the cert passes validation for PeerTrust, PeerOrChainTrust, and ChainTrust (and gives us coverage of all the positive code paths in the validator check). But once we get past the X509CertificateValidator checks, the cert still has to work end to end.  I could not find a way to make a cert work all the way through TCP with any less than what is here.  When I used something closer to the client cert, I was blocked by EndpointIdentity checks (and we've disabled many of the alternate EndpointIdentity code paths).  So the issue is this -- we need a cert that exercises X509CertificateValidation checks *and* is trusted by all the other layers in the stack.

How do you feel about letting this PR merge (after addressing other concerns) and then come back with a 2nd PR for just the cert if we can find a way to make that happen?  I've had to rebase this 3 times already.

---
In reply to: [71627124](https://github.com/dotnet/wcf/pull/1382#discussion_r71627124) [](ancestors = 71627124)

roncain: If we go there (which we can discuss), you're right we will need a new cert, a new REST method to get it, a new ConditionalFact to detect and obtain it, a new TestProperty, etc.   And if we go there for Https, we'll need to update build scripts, etc.  I guess the question to ask (though hard to answer) is "What new WCF product code paths would be tested by this kind of cert?"

---
In reply to: [71627543](https://github.com/dotnet/wcf/pull/1382#discussion_r71627543) [](ancestors = 71627543)

iamjasonp: > How do you feel about letting this PR merge (after addressing other concerns) and then come back with a 2nd PR for just the cert if we can find a way to make that happen? I've had to rebase this 3 times already.

Yeah, I'm ok taking this one later

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.cs
iamjasonp: Ah ok. Got it.

src/System.Private.ServiceModel/tools/CertificateGenerator/CertificateManager.cs
roncain: Thanks @iamjasonp -- are you good with the updates made in the 2nd commit?

iamjasonp: And yeah, in reply to [71698857](https://github.com/dotnet/wcf/pull/1382#discussion_r71698857) - that I didn't think of, how to get this past all of the _rest_ of the checks that we have.

iamjasonp:
LGTM

Issue #1034, opened 2016-04-18
[ConditionalFact] or [ActiveIssue]


zhenlan:
Both ConditionalFact and ActiveIssue can be used to skip tests. There are some pros, cons and expectations of each approach. Start this issue to discuss what we should use going forward.

roncain:
Good idea.  To open the discussion, it is important to understand some concepts:
- [ActiveIssue] is based on xunit trait framework
- [ConditionalFact] is based on the xunit [Fact] base class

Traits are logically part of the "discovery" phase, and [Fact] is logically part of the "execute" phase.

When you use "-notrait" to exclude a group of tests, the xunit discovery phase just throws them away.  You don't see any record they even existed or were skipped.

When you use [ConditionalFact], xunit's [Fact] framework gives it a chance to say "now is your chance to run this test and tell me what result you want, including an indication of whether it failed or was skipped".  If we want a test run to log which tests were skipped and why, some kind of [Fact]-based approach is probably the best choice.

One option we should consider is writing our own form of [ConditionalFact] that allows us to associate a test with a specific issue (or set of issues).  We could also allow control over whether we wanted the test skipped, executed as is, or failed for a different reason, such as "test X failed because attempting to open the firewall for port XYZ failed".  Ideas welcome -- those are just some of the background thoughts.

zhenlan:
This is what was discussed:
- It is preferred that the number of tests that are skipped can be shown as part of test result, which is what [ConditonalFact] can give.
- However, unlike [ActiveIssue], [ConditionalFact] cannot be applied to a test multiple times to indicate different issues on different platforms etc.
- We basically need a new attribute that has the pros of both, but we believe it's hard to change xunit engine.
- **Plan**: we will prototype a custom solution something like below and add it to buildtools repo if it works.
```
  [CondtionalFact("NoIssuesApplicable")]
  [Issue(123, PlatformID.Windows)]
  [Issue(456, PlatformID.OSX)]
  public static void SomeOfMyTest()
  {
	...
  } 
```

roncain:
After prototyping this and experimenting, I've altered the approach slightly.  

The design approach is:

- Replace use of ConditionalFact with a new Fact attribute that can reflect over other WCF-specific attributes to determine whether the test should be run or skipped
- Create new "skip" attribute types to allow skipping based on (1) an active issue, (2) some condition (like ConditionalFact), or (3) the current platform. 

Reasons for the approach:

[...]

PR #1388, opened 2016-07-22
Add new [Fact] and [Theory] attributes to generalize ConditionalFact

roncain:
This PR adds WcfFact and WcfTheory attributes and discoverers to take the place of
ConditionalFact and ConditionalTheory.  The new discoverers have
been generalized to reflect onto the test method for any of the new
"skip" attributes (IssueAttribute and ConditionAttribute).

For design discussion on this see issue #1034

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/ConditionalTestDetectors.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/ConditionalTestDetectors.cs
@@ -4,7 +4,7 @@
 
 
 using System;
-using System.Runtime.InteropServices;
+using Xunit;
 
 namespace Infrastructure.Common
 {
@@ -117,7 +117,7 @@ namespace Infrastructure.Common
         // Returns 'true' if the client is running on a Windows OS
         public static bool IsWindows()
         {
-            return RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            return OSID.AnyWindows.MatchesCurrent();
         }
 
         // Returns 'true' if the server is running as localhost.
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/FrameworkID.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/FrameworkID.cs
@@ -1,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Infrastructure.Common
+{
+    // Enum to indicate a Framework.
+    // There must be a 1:1 match between all the elements in
+    // this enum and RuntimeInformation.FrameworkDescription.
+    // It uses [Flags] because it is a bitmask that allows bitwise
+    // combinations in scenarios like the [Issue] attribute.
+    [Flags]
+    public enum FrameworkID
+    {
+        None =         0x00000000,
+        NetFramework = 0x00000001,   // Net 4.5 || Win8
+        NetCore =      0x00000002,   // netcore50 || wpa81 || other
+        NetNative =    0x00000004,   // netcore50aot
+
+        // 'Any' explicitly names only known flags so "G" formatting
+        // can be used to show a comma separated list of the bitmask.
+        Any = NetFramework | NetCore | NetNative
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/OsID.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/OsID.cs
@@ -1,0 +1,59 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Infrastructure.Common
+{
+    // Enum to indicate a known operating system.
+    // The elements in this enum represent all known operating systems
+    // against which the product can be run.  Please update this enum
+    // as new operating systems bceome available for testing.
+    // It uses [Flags] because it is a bitmask that allows bitwise
+    // combinations in scenarios like the [Issue] attribute.
+
+    [Flags]
+    public enum OSID
+    {
+        None =          0x00000000,
+
+        Windows_7 =     0x00000001,
+        Windows_8_1 =   0x00000002,
+        Windows_10 =    0x00000004,
+        WindowsPhone =  0x00000008,
+        Windows_Nano =  0x00000010,
+
+        CentOS_7 =      0x00000020,
+        CentOS_7_1 =    0x00000040,
+        AnyCentOS = CentOS_7 | CentOS_7_1,
+
+        Debian_8 =      0x00000080,
+        Debian_8_2 =    0x00000100, 
+        AnyDebian = Debian_8 | Debian_8_2,
+
+        Fedora_23 =     0x00000200,
+        AnyFedora = Fedora_23,
+
+        OpenSUSE_13_2 = 0x00000400,
+        AnyOpenSUSE = OpenSUSE_13_2,
+
+        OSX_10_10 =     0x00000800,
+        OSX_10_11 =     0x00001000,
+        AnyOSX = OSX_10_10 | OSX_10_11,
+
+        RHEL_7 =        0x00002000,
+        AnyRHEL = RHEL_7,
+
+        Ubuntu_14_04 =  0x00004000,
+        Ubuntu_16_04 =  0x00008000,
+        AnyUbuntu = Ubuntu_14_04 | Ubuntu_16_04,
+
+        AnyUnix = AnyCentOS | AnyDebian | AnyFedora | AnyOpenSUSE | AnyOSX | AnyRHEL | AnyUbuntu,
+        AnyWindows = Windows_7 | Windows_8_1 | Windows_10 | WindowsPhone | Windows_Nano,
+
+        // 'Any' explicitly names only known flags so "G" formatting
+        // can be used to show a comma separated list of the bitmask.
+        Any = AnyUnix | AnyWindows
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
@@ -1,0 +1,123 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Xunit.Abstractions;
+
+namespace Infrastructure.Common
+{
+    // The [Condition] attribute can be applied to any test method that
+    // is marked with [WcfFact] or [WcfTheory].  It provides zero or more
+    // member names to be evaluated at runtime to determine whether the
+    // test should be run or skipped.
+    // Examples: 
+    //  [Condition(nameof(Root_Certificate_Installed))]
+    //  [Condition(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed))]
+
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+    public class ConditionAttribute : WcfSkippableAttribute
+    {
+        public string[] Conditions { get; private set; }
+
+        public ConditionAttribute(params string[] conditions)
+        {
+            Conditions = conditions;
+        }
+
+        public override string GetSkipReason(ITestMethod testMethod)
+        {
+            // A null or empty list of conditions is treated as "no conditions",
+            // and the test cases will not be skipped.
+            // Example: [Condition()] or [Condition((string[]) null)]
+            int conditionCount = Conditions == null ? 0 : Conditions.Length;
+            if (conditionCount == 0)
+            {
+                return null;
+            }
+
+            MethodInfo testMethodInfo = testMethod.Method.ToRuntimeMethod();
+            Type testMethodDeclaringType = testMethodInfo.DeclaringType;
+
+            List<string> falseConditions = new List<string>(conditionCount);
+
+            foreach (string entry in Conditions)
+            {
+                string conditionMemberName = entry;
+
+                // Null condition member names are silently tolerated
+                if (string.IsNullOrWhiteSpace(conditionMemberName))
+                {
+                    continue;
+                }
+
+                string[] symbols = conditionMemberName.Split('.');
+                Type declaringType = testMethodDeclaringType;
+
+                if (symbols.Length == 2)
+                {
+                    conditionMemberName = symbols[1];
+                    ITypeInfo type = testMethod.TestClass.Class.Assembly.GetTypes(false).Where(t => t.Name.Contains(symbols[0])).SingleOrDefault();
+                    if (type != null)
+                    {
+                        declaringType = type.ToRuntimeType();
+                    }
+                }
+
+                MethodInfo conditionMethodInfo;
+                if ((conditionMethodInfo = LookupConditionalMethod(declaringType, conditionMemberName)) == null)
+                {
+                    falseConditions.Add(String.Format("Condition \"{0}\" not found in type \"{1}\".",
+                                                      conditionMemberName, testMethodDeclaringType.FullName));
+                    continue;
+                }
+
+                // In the case of multiple conditions, collect the results of all
+                // of them to produce a summary skip reason.
+                try
+                {
+                    if (!(bool)conditionMethodInfo.Invoke(null, null))
+                    {
+                        falseConditions.Add(conditionMemberName);
+                    }
+                }
+                catch (Exception exc)
+                {
+                    falseConditions.Add(String.Format("Condition \"{0}\" threw exception {1}: \"{2}\".",
+                                                      conditionMemberName, 
+                                                      exc.GetType().Name,
+                                                      exc.Message));
+                }
+            }
+
+            if (falseConditions.Count == 0)
+            {
+                return null;
+            }
+
+            return String.Format("{0}", String.Join(", ", falseConditions));
+        }
+
+
+        internal static MethodInfo LookupConditionalMethod(Type t, string name)
+        {
+            if (t == null || name == null)
+                return null;
+
+            TypeInfo ti = t.GetTypeInfo();
+
+            MethodInfo mi = ti.GetDeclaredMethod(name);
+            if (mi != null && mi.IsStatic && mi.GetParameters().Length == 0 && mi.ReturnType == typeof(bool))
+                return mi;
+
+            PropertyInfo pi = ti.GetDeclaredProperty(name);
+            if (pi != null && pi.PropertyType == typeof(bool) && pi.GetMethod != null && pi.GetMethod.IsStatic && pi.GetMethod.GetParameters().Length == 0)
+                return pi.GetMethod;
+
+            return LookupConditionalMethod(ti.BaseType, name);
+        }
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
@@ -1,0 +1,77 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using Xunit.Abstractions;
+
+namespace Infrastructure.Common
+{
+    // The [Issue] attribute can be applied to any test method that
+    // is marked with [WcfFact] or [WcfTheory].  It indicates the open
+    // issue number and optionally the framework(s), or
+    // OS(s) to which it applies.
+    // Examples: 
+    //  [Issue(999)]
+    //      means this issue applies to all OSes, all Frameworks
+    //  [Issue(999, Framework=FrameworkID.NetNative)]
+    //      means this issue only applies to NET Native
+    //  [Issue(999, OS=OSID.Ubuntu_14_04)]
+    //      means this issue only applies to Ubuntu 14.04
+
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+    public class IssueAttribute : WcfSkippableAttribute
+    {
+        public int Issue { get; private set; }
+
+        public FrameworkID Framework {get; set; }
+
+        public OSID OS { get; set; }
+
+        // Short name to repository where the issue exists.
+        // It is not required to be a URL, and it won't be used to create a URL.
+        public string Repository { get; set; }
+
+        public IssueAttribute(int issue)
+        {
+            Issue = issue;
+            Repository = "dotnet/wcf";
+
+            // Default framework and OS to "none" so that they must
+            // be explicitly specified to be more specific.
+            Framework = FrameworkID.None;
+            OS = OSID.None;
+        }
+
+        public override string GetSkipReason(ITestMethod testMethod)
+        {
+            string repositoryAndIssue = String.Format("{0} #{1}", Repository, Issue);
+
+            if (Framework.MatchesCurrent())
+            {
+                return String.Format("{0} on framework \"{1}\" (filter is \"{2}\")",
+                                    repositoryAndIssue,
+                                    FrameworkHelper.Current.Name(),
+                                    Framework.Name());
+            }
+
+            if (OS.MatchesCurrent())
+            {
+                return String.Format("{0} on OS \"{1}\" (filter is \"{2}\")",
+                                    repositoryAndIssue,
+                                    OSHelper.Current.Name(),
+                                    OS.Name());
+            }
+
+            // If no specific OS or Framework filters, it applies to all
+            if (OS == OSID.None && Framework == FrameworkID.None)
+            {
+                return String.Format("{0}",
+                    repositoryAndIssue);
+            }
+
+            // Null means "don't skip"
+            return null;
+        }
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfFactDiscoverer.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfFactDiscoverer.cs
@@ -1,0 +1,28 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Infrastructure.Common
+{
+    public class WcfFactDiscoverer : FactDiscoverer
+    {
+        private readonly IMessageSink _diagnosticMessageSink;
+
+        public WcfFactDiscoverer(IMessageSink diagnosticMessageSink) : base(diagnosticMessageSink)
+        {
+            _diagnosticMessageSink = diagnosticMessageSink;
+        }
+
+        public override IEnumerable<IXunitTestCase> Discover(
+            ITestFrameworkDiscoveryOptions discoveryOptions, ITestMethod testMethod, IAttributeInfo factAttribute)
+        {
+            IEnumerable<IXunitTestCase> testCases = base.Discover(discoveryOptions, testMethod, factAttribute);
+            return WcfTestDiscoverer.Discover(discoveryOptions, _diagnosticMessageSink, testMethod, testCases);
+        }
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestDiscoverer.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestDiscoverer.cs
@@ -1,0 +1,81 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Infrastructure.Common
+{
+    // Internal helper class for code common to conditional test discovery through
+    // [WcfFact] and [WcfTheory].  It reflects back onto the test method's attributes
+    // to determine whether the test should be skipped.
+    internal class WcfTestDiscoverer
+    {
+        internal static IEnumerable<IXunitTestCase> Discover(
+                                                        ITestFrameworkDiscoveryOptions discoveryOptions,
+                                                        IMessageSink diagnosticMessageSink,
+                                                        ITestMethod testMethod,
+                                                        IEnumerable<IXunitTestCase> testCases)
+        {
+            MethodInfo testMethodInfo = testMethod.Method.ToRuntimeMethod();
+ 
+            // Evaluate any [Issue] attributes on this method.
+            // We do this first so that tests we know won't be run will avoid calling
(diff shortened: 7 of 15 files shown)


roncain:
Reviewers: this PR is the result of my prototype to combine ConditionalFact and ActiveIssue.  My main goal of this PR is to solicit feedback about the approach.  See #1034 for deeper discussion comments.  If approved (subject to feedback), it can be checked in without affecting existing code.  And then a 2nd PR can make a sweep to change existing tests to use the new attributes.

src/System.Private.ServiceModel/tests/Scenarios/Infrastructure/WcfFactTests.cs
```diff
@@ -0,0 +1,96 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
roncain: Reviewers: this test case was included mostly to illustrate usage and to allow me to test.  I will delete it from the PR before merging.

roncain:
@hongdai -- I don't know whether these proposed changes overlap with your prototype for ETW tracing.  Please take a look and we can discuss if we need to merge our efforts.

src/System.Private.ServiceModel/tests/Common/Infrastructure/ConditionalTestDetectors.cs
```diff
@@ -117,7 +117,7 @@ public static bool IsServerDomainJoined()
         // Returns 'true' if the client is running on a Windows OS
         public static bool IsWindows()
         {
-            return RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            return Platform.MatchesCurrentPlatform(PlatformID.Windows);
```
mconnew: I'm curious, what scenario required this change?

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+    // The [Condition] attribute can be applied to any test method that
+    // is marked with [WcfFact] or [WcfTheory].  It provides zero or more
+    // member names to be evaluated at runtime to determine whether the
+    // test should be run or skipped
+    // Examples: 
+    //  [Condition(nameof(Root_Certificate_Installed))]
+    //  [Condition(nameof(Root_Certificate_Installed), nameof(Client_Certificate_Installed))]
+
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+    public sealed class ConditionAttribute : WcfSkippableAttribute
```
mconnew: >sealed class ConditionAttribute [](start = 11, length = 31)

I noticed that ConditionalFactAttribute is sealed too. Is there some architectural reason why this needs to be the case?

src/System.Private.ServiceModel/tests/Common/Infrastructure/ConditionalTestDetectors.cs
roncain: Now that I've centralized other Platform detection into the Platform class, I wanted to be consistent in using it.  Nothing required this change other than code consistency, and I can revert if you prefer.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
roncain: They don't need to be sealed -- they are just test infrastructure.  So I'll remove that.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            Conditions = conditions;
+        }
+
+        public override string GetSkipReason(ITestMethod testMethod)
+        {
+            // A null or empty list of conditions is treated as "no conditions",
+            // and the test cases will not be skipped.
+            // Example: [Condition()] or [Condition((string[]) null)]
+            int conditionCount = Conditions == null ? 0 : Conditions.Length;
```
mconnew: >Conditions == null [](start = 33, length = 18)

Is there a case where Conditions could be null? The only constructor has a params as the only argument. Doesn't passing no parameters still cause a zero length array to be passed? I know this is nit-picking, so I'm not expecting you to change anything, but could this be converted to a Contract.Assert  to protect against someone adding a different constructor in the future?

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+                    continue;
+                }
+
+                string[] symbols = conditionMemberName.Split('.');
+                Type declaringType = testMethodDeclaringType;
+
+                if (symbols.Length == 2)
+                {
+                    conditionMemberName = symbols[1];
+                    ITypeInfo type = testMethod.TestClass.Class.Assembly.GetTypes(false).Where(t => t.Name.Contains(symbols[0])).FirstOrDefault();
```
mconnew: >FirstOrDefault [](start = 129, length = 14)

SingleOrDefault might be better in case there are multiple matches. I think we would prefer a failure if there's ambiguity rather than it being non-deterministic.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+                // Null condition member names are silently tolerated
+                if (string.IsNullOrWhiteSpace(conditionMemberName))
+                {
+                    continue;
+                }
+
+                string[] symbols = conditionMemberName.Split('.');
+                Type declaringType = testMethodDeclaringType;
+
+                if (symbols.Length == 2)
```
mconnew: This looks like we don't expect a type to have it's namespace. I think it would be worth adding logic so that if symbols.Length > 2, then presume the full namespace qualified type has been specified and instead of doing a search for any type with the same name, try and get the namespace qualified type exactly.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+                    {
+                        declaringType = type.ToRuntimeType();
+                    }
+                }
+
+                MethodInfo conditionMethodInfo;
+                if ((conditionMethodInfo = LookupConditionalMethod(declaringType, conditionMemberName)) == null)
+                {
+                    falseConditions.Add(String.Format("Condition \"{0}\" not found in type \"{1}\".",
+                                                      conditionMemberName, testMethodDeclaringType.Name));
```
mconnew: >testMethodDeclaringType [](start = 75, length = 23)

Use declaringType.FullName instead so that we know what type the method couldn't be found in.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+                {
+                    if (!(bool)conditionMethodInfo.Invoke(null, null))
+                    {
+                        falseConditions.Add(conditionMemberName);
+                    }
+                }
+                catch (Exception exc)
+                {
+                    falseConditions.Add(String.Format("Condition \"{0}\" threw exception \"{1}\".",
+                                                      conditionMemberName, exc.ToString()));
```
mconnew: >exc.ToString() [](start = 75, length = 14)

This is going to output the entire stack trace as part of the string (which will include new lines). The string text suggests you might have intended to just output the exception type and/or it's message. Just wanted to check the behavior matched the intention.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
```diff
@@ -0,0 +1,121 @@
...
+                return null;
+
+            TypeInfo ti = t.GetTypeInfo();
+
+            MethodInfo mi = ti.GetDeclaredMethod(name);
+            if (mi != null && mi.IsStatic && mi.GetParameters().Length == 0 && mi.ReturnType == typeof(bool))
+                return mi;
+
+            PropertyInfo pi = ti.GetDeclaredProperty(name);
+            if (pi != null && pi.PropertyType == typeof(bool) && pi.GetMethod != null && pi.GetMethod.IsStatic && pi.GetMethod.GetParameters().Length == 0)
```
mconnew: >pi.GetMethod.GetParameters() [](start = 114, length = 28)

Is it possible for a PropertyInfo to return a method from GetMethod which takes any parameters? Also, PropertyInfo.GetMethod will return non-public methods too. Is this what we want? This might cause us problems once .Net native is using this.

src/System.Private.ServiceModel/tests/Common/Infrastructure/Platform.cs
```diff
@@ -0,0 +1,71 @@
...
+            get
+            {
+                return GetPlatformName(Current);
+            }
+        }
+
+        public static string GetPlatformName(PlatformID id)
+        {
+            return Enum.Format(typeof(PlatformID), id, "G");
+        }
```
mconnew: Could this be an extension method for PlatformID? That way you could get rid of the CurrentName property and instead do the following.
```csharp
    Platform.Current.GetName();
```

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+    //  [Issue(999, PlatformID.OSX)]
+
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
+    public sealed class IssueAttribute : WcfSkippableAttribute
+    {
+        public int Issue { get; private set; }
+
+        public PlatformID PlatformID { get; private set; }
+
+        public IssueAttribute(int issue, PlatformID platformID)
```
mconnew: Is there any benefit to having a repo name here? Or can we assume that we will always have a tracking bug in our repo so all Issue id's will be for dotnet/wcf?

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/PlatformAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using Xunit;
+using Xunit.Sdk;
+using Xunit.Abstractions;
+
+namespace Infrastructure.Common
+{
+    // The [Platform] attribute can be applied to any test method that
+    // is marked with [WcfFact] or [WcfTheory].  It indicates the which
```
mconnew: >the [](start = 62, length = 3)

Nit: errant 'the' in comment

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/PlatformAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+    // platforms on which the test should be run.  It also optionally permits
+    // an exclusion list to describe on which platforms it should not be run.
+    // If this attribute is absent, it means the test can run on all platforms.
+    // Examples: 
+    //  [Platform(PlatformID.Windows)]
+    //  [Platform(PlatformID.AnyUnix, Exclude=PlatformID.OSX)]
```
mconnew: How would we use this to specify only exclude OSX? It looks like it would be this:
```csharp
    [Platform(PlatformID.Any, Exclude=PlatformID.OSX)]
```
Can we add a default constructor which sets PlatformID to PlatformID.Any? That way you can specify only the exclude on the attribute which I think is going to be a common use case.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/SkippedTestCase.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Infrastructure.Common
+{
+    /// <summary>Wraps another test case that should be skipped.</summary>
```
mconnew: Maybe add comment that this is a copy of [https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/SkippedTestCase.cs](https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/SkippedTestCase.cs)

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfFactDiscoverer.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Linq;
+using Xunit;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Infrastructure.Common
+{
+    public class WcfFactDiscoverer : FactDiscoverer
+    {
+        private readonly IMessageSink _diagnosticMessageSink;
```
mconnew: You've only added this because the base class has its field holding this private. As this class isn't sealed, should you make this a protected property with private set?

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfFactDiscoverer.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Infrastructure.Common
+{
+    public class WcfFactDiscoverer : FactDiscoverer
```
mconnew: Maybe add comment referencing [https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Discoverers/ConditionalFactDiscoverer.cs](https://github.com/dotnet/buildtools/blob/master/src/xunit.netcore.extensions/Discoverers/ConditionalFactDiscoverer.cs) as inspiration for this in case things get refactored, it will be easier to know where to look to see what changes we need to make.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestDiscoverer.cs
```diff
@@ -0,0 +1,100 @@
...
+using Xunit.Sdk;
+
+namespace Infrastructure.Common
+{
+    // Internal helper class for code common to conditional test discovery through
+    // [WcfFact] and [WcfTheory].  It reflects back onto the test method's attributes
+    // to determine whether the test should be skipped.
+    internal class WcfTestDiscoverer
+    {
+
```
mconnew: Nit: blank line

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestDiscoverer.cs
```diff
@@ -0,0 +1,100 @@
...
+                    string skipReason = issue.GetSkipReason(testMethod);
+                    if (skipReason != null)
+                    {
+                        issueSkipList.Add(skipReason);
+                    }
+                }
+
+                if (issueSkipList.Count > 0)
+                {
+                    string skippedReason = string.Format("Active issue(s): {0}", string.Join(",", issueSkipList));
```
mconnew: The output from IssueAttribute.GetSkipReason clashes a little with this. The output would look something like this:  
Active issue(s): issue 123 for "AnyUnix" skipped on "Linux", issue 125 for "Any" skipped on "Linux"  
There's redundancy as every issue will be reporting the current platform. You are already special casing the attribute here and you have a reference to a strongly typed IssueAttribute object. Maybe have GetSkipReason output:
```csharp
"{0} for \"{1}\", Issue, Platform.GetPlatformName(PlatformID)
```
And have WcfTestDiscoverer output:
```csharp
"Active issue(s): {0} skipped on {1}", string.Join(", ", issueSkipList), Platform.CurrentName
```

src/System.Private.ServiceModel/tests/Common/Infrastructure/ConditionalTestDetectors.cs
mconnew: I'm fine with code consistency being a reason to change this. I was wondering if there was some subtlety to comparing things differently that caused a slightly different behavior.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/SkippedTestCase.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public object[] TestMethodArguments { get { return _testCase.TestMethodArguments; } }
+
+        public Dictionary<string, List<string>> Traits { get { return _testCase.Traits; } }
+
+        public string UniqueID { get { return _testCase.UniqueID; } }
+
+        public void Deserialize(IXunitSerializationInfo info) { _testCase.Deserialize(info); }
+
+        public Task<RunSummary> RunAsync(
```
hongdai: Could we make this class WCFTestCase for WCF specific actions? I'll need to add ETW information in the RunAsync. I believe this class can be used for non skipped tests as well, if we don't set SkipReason.

roncain: I'd prefer not to deviate from using "Fact" or "Theory" in the name, because those are fundamental xunit concepts.  I'm open to renaming the prefix part (example: TestFact and TestTheory or WcfTestFact and WcfTheoryFact).  Thoughts?

src/System.Private.ServiceModel/tests/Common/Infrastructure/Platform.cs
roncain: Good idea -- will do.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
roncain: When I first updated ConditionalFact, Stephen Toub observed it was legal for the attribute declaration to pass in a null explicitly, so asked for the check.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/ConditionAttribute.cs
roncain: Let me investigate this apart from the main PR (just to get it checked in first).  I copied this from BuildTools.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
roncain: I like the idea -- will add it.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/PlatformAttribute.cs
roncain: The next commit will have [Framework] and [OS] to allow this granularity.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/SkippedTestCase.cs
roncain: I was going to but @hongdai and I decided that this class will be renamed and diverge from BuildTools to be more extensible.  Actually, I think SkippedTestCase in BuildTools was copied from an xunit sample anyway.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfFactDiscoverer.cs
roncain: I think we've diverged past recognition at this point, and even ConditionAttribute is diverging from ConditionalFact's logic.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfFactDiscoverer.cs
roncain: This was copied from buildtools, but let me mull over refactoring slightly when we see merge in changes from @hongdai

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestDiscoverer.cs
roncain: Good point.  I've restructured a bit so the skip reasons only include the details and let the outer Format add the part that you observed was redundant.  We can tune the message if we find it confusing in practice.

roncain:
Reviewers: I have incorporated feedback and added FrameworkAttribute, OSAttribute, and their corresponding enum and support types.  I have also added the concept of Framework and OS to the IssueAttribute.  I may need to tune the enum matching for OSAttribute as I see the console output from the CI or VSO runs, but I think that can be handled in a separate PR.  Remember, these changes are only additive and will not affect current tests.  A separate PR will happen after this one to convert all [Fact] to [WcfFact] before it affects existing tests.

I am interested in feedback on the new changes.  If you have comments, please indicate whether they must be fixed before merging or whether they can be fixed post-merge.  I would like to merge soon to enable @hongdai to modify WcfFact for ETW tracing.

/cc: @mconnew @StephenBonikowsky @iamjasonp @hongdai

roncain:
Reviewers: I've added finer-grained values for OSID to allow distinguishing at a better granularity.  But I've found trying to extract this from RuntimeInformation alone is fragile and would require non-deterministic string matching.  Please review PR #1397 at the same time, because I will need to check that in first. It will permit me to determine the OSID from the runtime package used for tests.

/cc: @mconnew @iamjasonp @StephenBonikowsky @hongdai

src/System.Private.ServiceModel/tests/Common/Infrastructure/OS.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        // Extension method for OSID to provide name.
+        // Format will expand to enum's name or collection if multiple.
+        public static string Name(this OSID id)
+        {
+            return Enum.Format(typeof(OSID), id, "G");
+        }
+
+        private static OSID DetectCurrentOS()
```
roncain: Reviewers: it is my intent to replace this string matching against the OS description with a lookup table based on $(TestNugetRuntimeId) in TestProperties.  I will need to merge #1397 before I can do that.  I prefer that we merge this current PR asap to unblock @hongdai and let me switch to using TestProperties post-merge.

src/System.Private.ServiceModel/tests/Common/Infrastructure/Framework.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                return FrameworkID.NetNative;
+            }
+
+            if (frameworkName.IndexOf(NetFrameworkFrameworkName, StringComparison.Ordinal) >= 0)
+            {
+                return FrameworkID.NetFramework;
+            }
+
+            if (frameworkName.IndexOf(NetCoreFrameworkName, StringComparison.Ordinal) >= 0)
```
iamjasonp: NIT: Ordinal or OrdinalIgnoreCase?

roncain:
Reviewers: please wait until I update this PR before reviewing.  I am reducing its scope a bit.  I will send out a new comment when it is ready for review again.

/cc: @Zhenlan @mconnew @iamjasonp @hongdai @StephenBonikowsky

src/System.Private.ServiceModel/tests/Common/Infrastructure/Framework.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public static FrameworkID Current
+        {
+            get
+            {
+                if (_currentFrameworkID == 0)
+                {
+                    _currentFrameworkID = DetectCurrentFramework();
+                    Console.WriteLine(String.Format("Detected current framework as \"{0}\" from description \"{1}\"",
+                                     _currentFrameworkID.Name(), CurrentFrameworkDescription));
```
mconnew: Debugging output should be removed

src/System.Private.ServiceModel/tests/Common/Infrastructure/Framework.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                return FrameworkID.NetCore;
+            }
+
+            // We adopt the ID "unknown" whenever we cannot detect the framework to avoid
+            // throwing exceptions during test discovery when new frameworks become available.
+            // The "unknown" in the skip reasons will indicate investigation is required.
+            Console.WriteLine(String.Format(
+                                "FrameworkID cannot be detected from RuntimeInformation.FrameworkDescription \"{0}\".", 
+                                frameworkName));
```
mconnew: Debugging output should be removed

src/System.Private.ServiceModel/tests/Common/Infrastructure/FrameworkID.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Infrastructure.Common
+{
+    [Flags]
```
mconnew: Why is this a flag? A current framework can't be multiple frameworks.

src/System.Private.ServiceModel/tests/Common/Infrastructure/OS.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        public static OSID Current
+        {
+            get
+            {
+                if (_currentOSID == 0)
+                {
+                    _currentOSID = DetectCurrentOS();
+                    Console.WriteLine(String.Format("Detected current OS as \"{0}\" from description \"{1}\"",
```
mconnew: Are these outputs intentional? I Don't think we should have them.

src/System.Private.ServiceModel/tests/Common/Infrastructure/OsID.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Infrastructure.Common
+{
+    [Flags]
+    public enum OSID
+    {
+        None = 0,
+
+        Windows_7 = 1,
+        Windows_8_1 = 2,
```
mconnew: Can we use hex values?

roncain:
Reviewers: this is ready to be reviewed now.  It has been reduced in scope.  As always, please indicate which comments are blocking and which can be addressed post-merge.  Thanks.

/cc: @Zhenlan @mconnew @iamjasonp @hongdai @StephenBonikowsky

src/System.Private.ServiceModel/tests/Common/Infrastructure/OsID.cs
```diff
@@ -0,0 +1,49 @@
...
+        AnyOSX = OSX_10_11,
+
+        RHEL_7_2 = 1024,
+        AnyRHEL = RHEL_7_2,
+
+        Ubuntu_14_04 = 2048,
+        Ubuntu_16_04 = 4096,
+        AnyUbuntu = Ubuntu_14_04 | Ubuntu_16_04,
+
+        Unknown = 8192,     // were unable to determine
```
mconnew: >Unknown = 8192 [](start = 8, length = 14)

Unknown shouldn't be the next number in sequence. It's not possible to have no OS, so maybe Unknown should have a value of 0 so that Unknown is the opposite of Any.

src/System.Private.ServiceModel/tests/Common/Infrastructure/FrameworkID.cs
roncain: Both FrameworkID and OSID are flags because they are bit masks.  They are bit masks to allow combinations like Any.  And you can combine them in attribute declarations to include or exclude certain frameworks like this:

```
    [WcfFact]
    [Issue(999, Framework = FrameworkID.Any ^ FrameworkID.NetFramework)]
    public static void Skip_Issue_All_Frameworks_Except_Full_Framework()
    {
    }

    [WcfFact]
    [Issue(999, Framework = FrameworkID.NetCore | FrameworkID.NetFramework)]
    public static void Skip_Issue_On_NetCore_And_Full_Framework()
    {
    }
```

src/System.Private.ServiceModel/tests/Common/Infrastructure/OsID.cs
roncain: ok

src/System.Private.ServiceModel/tests/Common/Infrastructure/Framework.cs
roncain: These names are supposed to be an exact match against literals buried in RuntimeInformation. I think I want detection to fail and tests to report the problem if they change.

src/System.Private.ServiceModel/tests/Common/Infrastructure/Framework.cs
roncain: Perhaps in a later checkin after we see some VSO runs.  Right now, look at the Console logs on the CI runs and you can see which OS and Framework have been detected.  This would be a great debugging aid if you find tests being run unexpectedly on certain OSes or Frameworks.

src/System.Private.ServiceModel/tests/Common/Infrastructure/Framework.cs
roncain: Same comment as above -- leaving in at least for now because it provides vital information in the console logs across all the platforms.

src/System.Private.ServiceModel/tests/Common/Infrastructure/OS.cs
roncain: Yes, these outputs are intentional, in the same way the Console writes are in other areas like certificate loading.  It captures information in the console logs on lab runs that help debug.  I'd like to leave these in at least for awhile and maybe permanently.

src/System.Private.ServiceModel/tests/Common/Infrastructure/OsID.cs
roncain: I thought about this some more and decided it was better to remove 'Unknown' entirely and just return 'None' if we can't detect.  Then I added infrastructure tests that will fail if they ever encounter a None for OSID or FrameworkID.  This means that if we start running on some new framework or OS other than the ones we know here we will see a test failure, and we'll know we need to update the detection logic.

roncain:
Reviewers: all feedback incorporated and awaiting final sign-off from someone.
I will squash all spurious commits before merging.

(Note: look at the console output for any OuterLoop and search for "detected current" to verify that we really are detecting the OS and Framework on which we are running.  Waiting to see more in VSO runs. We can't yet distinguish between Win7, Win81 and Win10 -- I'll make a new PR when I learn how.)

/cc: @Zhenlan @mconnew @iamjasonp @hongdai @StephenBonikowsky

src/System.Private.ServiceModel/tests/Common/Infrastructure/OsID.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        RHEL_7 =        0x00002000,
+        AnyRHEL = RHEL_7,
+
+        Ubuntu_14_04 =  0x00004000,
+        Ubuntu_16_04 =  0x00008000,
+        AnyUbuntu = Ubuntu_14_04 | Ubuntu_16_04,
+
+        AnyUnix = AnyCentOS | AnyDebian | AnyFedora | AnyOpenSUSE | AnyOSX | AnyRHEL | AnyUbuntu,
+        AnyWindows = Windows_7 | Windows_8_1 | Windows_10 | WindowsPhone | Windows_Nano,
```
iamjasonp: >AnyWindows = Windows_7 | Windows_8_1 | Windows_10 | WindowsPhone | Windows_Nano, [](start = 8, length = 80)

NIT: this probably belongs near the "Windows" entries on Line 26

iamjasonp:
We should definitely look deeper into how to distinguish the different Windowses (sp? :smile:) 

LGTM

roncain:
@mconnew had an interesting suggestion how to refactor slightly to make IssueAttribute and ConditionAttribute both simpler and more powerful.  I am going to merge this PR as-is and open another one with his proposed changes.

PR #1397, opened 2016-07-26
Add TestNugetRuntimeId to TestProperties

roncain:
This change adds a new entry to TestProperties to capture the
MsBuild property $(TestNugetRuntimeId).  This property is set
by Helix runs to indicate the package name for the runtime to
use when running the tests.  The build tools also already set
this property, so it is correct during the dev experience too.

This provides the extra necessary information for PR #1388 to be
able to provide more accurate values for the OS enum. For example,
we will be able to distintuish Ubuntu 14.04 from 16.04 at test
runtime based on this new TestProperty.

The existing RuntimeInformation support does not provide this
granularity explicitly and would require fragile string matching to
try to extract from the RuntimeInformation's OS description alone.


iamjasonp:
LGTM

PR #1399, opened 2016-07-26
Use one single master json file for all scenario test projects.

StephenBonikowsky:
* Delete all the project.json files under Scenarios.
* One 'master' json file added that will be copied to each scenario test dir at the beginning of the build.
* Running the clean script will remove these json files and they have been added to the gitignore file.
* The target that does the work has a condition based on a property's bool value, this is to support the next PR where we will have another target that will copy a different json file when that property is set otherwise.


build.override.targets
```diff
@@ -3,37 +3,39 @@
...
       WcfClean;
       $(TraversalCleanDependsOn);
     </TraversalCleanDependsOn>
   </PropertyGroup>
-  
+
   <PropertyGroup>
     <WcfSelfHostSetupLog>$(MSBuildThisFileDirectory)\SelfHostedWcfServiceSetup.log</WcfSelfHostSetupLog>
     <WcfSelfHostCleanupLog>$(MSBuildThisFileDirectory)\SelfHostedWcfServiceCleanup.log</WcfSelfHostCleanupLog>
+    <ScenarioLatestDependencyJson>$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\test\JsonFiles\stable.dependencies.json</ScenarioLatestDependencyJson>
```
zhenlan: Is there some naming mismatch here... the attribute says Scenario**Latest**DependencyJson while the file it points to is **stable**.dependencies.json?

build.override.targets
```diff
@@ -3,37 +3,39 @@
...
+    <UseStableDependencies Condition="'$(UseStableDependencies)'==''">true</UseStableDependencies>
```
zhenlan: I am leaning towards flipping the bit here. While we are progressing the development, we will have more and more non-stable dependencies so `UseStableDependencies` will really become `UseSpaghettiDependencies` 😄, which can be confusing to people. If we call the knob `UseLatestDependencies`, it will stay true (and we set it here default to false).

build.override.targets
```diff
@@ -55,4 +57,16 @@
     <Message Text="$([System.IO.File]::ReadAllText($(WcfSelfHostCleanupLog)))" Condition="Exists('$(WcfSelfHostCleanupLog)')" Importance="High" />
     <Message Text="-------------- WCF clean end --------------" Importance="High" />
   </Target>
+
+  <!-- This target will copy a master version of project.json that uses last released version of non-WCF dependencies to the scenario test dirs. -->
+  <!-- These non-WCF dependencies "may" explicitly reference a pre-release version but should only do so to pull in needed fixes. -->
+  <Target Name="WcfCopyProjectJson" Condition="'$(UseStableDependencies)'=='true'">
```
zhenlan: It's my understanding this target should be unconditioned. Inside the target, however, we will copy different project.json files depending on this condition. Is this the idea?

.gitignore
```diff
@@ -266,4 +266,8 @@ Session.vim
 
 # Private test configuration and binaries.
 config.ps1
-**/IISApplications
\ No newline at end of file
+**/IISApplications
+
+# WCF copied test project.json files
+src/System.Private.ServiceModel/tests/Scenarios/
+project.json
```
iamjasonp: warning - it seems like there's an extra line break here or something. I believe as written _all_ project.json are ignored. I think the correct thing might be 

```
src/System.Private.ServiceModel/tests/Scenarios/**/project.json
```

build.override.targets
roncain: I agree we should rename to $(UseLatestDependencies) and default to false.  It makes it clearer that you opt-in to latest dependencies (whereas the current name suggests you are only opting out of stable and not clear what that implies).  It would also set a precedent if we ever introduce something like $(UseLastKnownGoodDependencies) in the future.

roncain: I think the way I would prefer to see this done would be have a property for the location of the preferred json file based on some condition.  
And the final pick of stable would occur if nothing else had set it.
This allows us future ways to add other kinds of dependencies (like $(UseLKGDependencies)) and also allows us to name a .json file to use from the command line.

```
    <PropertyGroup>
      <UseLatestDependencies Condition="'$(UseLatestDependencies)'==''">false</UseLatestDependencies>
    </PropertyGroup>
    
    <PropertyGroup Condition = "'$(UseLatestDependencies)' == 'true' >
        <ScenarioDependencyJson>$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\test\JsonFiles\latest.dependencies.json</ScenarioDependencyJson> 
    </PropertyGroup>
        
    <PropertyGroup Condition = "'$(ScenarioDependencyJson)' == '' >
        <ScenarioDependencyJson>$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\test\JsonFiles\stable.dependencies.json</ScenarioDependencyJson> 
    </PropertyGroup>

```

build.override.targets
roncain: +1 to remove the condition.  And the copy inside will (as you do already) unconditionally copy from $(ScenariosDependencyJson) which will have been set outside this target.  There should not be any decisions inside this target about where to copy from or whether it is using stable or latest.

roncain:
We should double check what happens when we build in TFS.  I don't think we will have any issues, but when WcfOpen builds it will create project.json's inside a TFS enlistment.  I think it will just work, and the next 'tfpt scorch' should just remove them gracefully.

StephenBonikowsky:
Updated with feedback.

build.override.targets
```diff
@@ -3,37 +3,49 @@
...
+
   <PropertyGroup>
     <WcfSelfHostSetupLog>$(MSBuildThisFileDirectory)\SelfHostedWcfServiceSetup.log</WcfSelfHostSetupLog>
     <WcfSelfHostCleanupLog>$(MSBuildThisFileDirectory)\SelfHostedWcfServiceCleanup.log</WcfSelfHostCleanupLog>
   </PropertyGroup>
-  
+
+  <PropertyGroup>
+    <UseLatestDependencies Condition="'$(UseLatestDependencies)'==''">false</UseLatestDependencies>
+  </PropertyGroup>
```
iamjasonp: Question only: we don't ever actually depend on the value being "false" do we?

build.override.targets
```diff
@@ -3,37 +3,49 @@
...
+
+  <PropertyGroup Condition = "'$(UseLatestDependencies)' == 'true'" >
+    <ScenarioDependencyJson>$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\test\JsonFiles\latest.dependencies.json</ScenarioDependencyJson>
+  </PropertyGroup>
+
+  <PropertyGroup Condition = "'$(ScenarioDependencyJson)' == ''" >
```
iamjasonp: Should this be 

\<PropertyGroup Condition = "'$(**UseLatestDependencies**)' != 'true'" \>

iamjasonp: Clarified with @StephenBonikowsky - that there might be other cases that `ScenarioDependencyJson` is set to something else, and this is a default value. 

Can you please add a comment to that effect?

build.override.targets
iamjasonp: Clarified - we're relying on the default value below in line 34 to set if no pre-conditions such as "UseLatestDependencies" are true. There may be other conditions we will add later

build.override.targets
StephenBonikowsky: Adding the following...

iamjasonp:
Small nit per https://github.com/dotnet/wcf/pull/1399#discussion_r72494090 - LGTM

build.override.targets
iamjasonp: Sounds good

build.override.targets
```diff
@@ -3,37 +3,50 @@
...
+  <PropertyGroup>
+    <UseLatestDependencies Condition="'$(UseLatestDependencies)'==''">false</UseLatestDependencies>
+  </PropertyGroup>
+
+  <PropertyGroup Condition = "'$(UseLatestDependencies)' == 'true'" >
+    <ScenarioDependencyJson>$(MSBuildThisFileDirectory)\src\System.Private.ServiceModel\tools\test\JsonFiles\latest.dependencies.json</ScenarioDependencyJson>
+  </PropertyGroup>
+
+  <!-- If $(UseLatestDependencies) is 'false' and $(ScenarioDependencyJson) has not been set in command line use 'stable.dependencies.json' as the default value. -->
+  <PropertyGroup Condition = "'$(ScenarioDependencyJson)' == ''" >
```
mconnew:
```xml
<PropertyGroup Condition = "('$(ScenarioDependencyJson)' == '') or ('$(ScenarioDependencyJson)' == 'false')" >
```

build.override.targets
roncain: @iamjasonp raises a good point.  The normal MsBuild convention for boolean properties is that blank means false and that you should only really every check "== 'true'".  I think there is value having the entry here showing this property exists and its default value is false.  But we should not rely on matching the string 'false' anywhere.

---
In reply to: [72494232](https://github.com/dotnet/wcf/pull/1399#discussion_r72494232) [](ancestors = 72494232)

build.override.targets
roncain: This property is supposed to be the file name of the json file, so I wouldn't add a 'false' test

---
In reply to: [72516176](https://github.com/dotnet/wcf/pull/1399#discussion_r72516176) [](ancestors = 72516176)

roncain:
LGTM

PR #1404, opened 2016-07-28
Temporarily put back project.json files to unblock VSO

roncain:
This PR temporarily puts back the project.json files we removed when moving to a single project.json.  It appears the Build-Test job in VSO does not trigger our logic to copy the project.jsons, and therefore fails to build.  This PR is only temporary to unblock VSO while we decide on the correct fix.


StephenBonikowsky:
LGTM merging

PR #1405, opened 2016-07-29
Change to use Content copy installRootCertificate

hongdai:
* Move the file copy to certtest.props
* Change from taget to Content Copy for Helix to work


roncain:
LGTM

hongdai:
I have also verified the copy was successful in Helix.

PR #1429, opened 2016-08-01
Replace Fact, ConditionalFact and ActiveIssue

roncain:
Replace all FactAttribute usages with WcfFactAttribute
Replace all ConditionalFactAttribute with ConditionAttribute
Replace all ActiveIssueAttribute with IssueAttribute

The exception to this is in the conditional code that compiles
only under ToF, because the pseudo xunit does not support the
xunit types they require.  This will eventually be removed.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.4.0.0.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.4.0.0.cs
@@ -2,17 +2,20 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-
-using Infrastructure.Common;
 using System;
 using System.ServiceModel;
 using System.ServiceModel.Channels;
+using Infrastructure.Common;
 using Xunit;
 
 public partial class CustomBindingTests : ConditionalWcfTest
 {
     // Http: Client and Server bindings setup exactly the same using default settings.
+#if FULLXUNIT_NOTSUPPORTED
     [Fact]
+#endif
+
+    [WcfFact]
     [OuterLoop]
     public static void DefaultSettings_Http_Text_Echo_RoundTrips_String()
     {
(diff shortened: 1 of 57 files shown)


roncain:
Reviewers: sorry there are so many files to review.  Most changes were mechanical and just copy/paste.  The conditional compilation is needed to get ToF to run.  I will be doing some analysis of CI test runs as well as N-on-K and K-on-K test runs to ensure we get what we expect.  I would like to merge this fairly soon so that we start seeing the Issue's count as skipped in VSO runs.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Custom/CustomBindingTests.4.0.0.cs
```diff
@@ -2,17 +2,20 @@
...
 using Xunit;
 
 public partial class CustomBindingTests : ConditionalWcfTest
 {
     // Http: Client and Server bindings setup exactly the same using default settings.
+#if FULLXUNIT_NOTSUPPORTED
     [Fact]
+#endif
+
+    [WcfFact]
```
roncain: Reviewers: to help understand the pattern, WcfFact, Condition, and Issue attributes are all defined in ToF (TFS only) so that they compile (but are non-functional).  This permits everything outside the #if to be correct for both GitHub and ToF.  The goal is to be able to remove all the 'if FULLXUNIT_NOTSUPPORTED' blocks when we are on Helix, and the rest should not need to be touched.

src/System.Private.ServiceModel/tests/Scenarios/Client/ChannelLayer/DuplexChannelShapeTests.4.1.0.cs
```diff
@@ -17,11 +16,11 @@ public partial class DuplexChannelShapeTests : ConditionalWcfTest
     // The tests in this file use the IDuplexChannel shape.
 #if FULLXUNIT_NOTSUPPORTED
     [Fact]
-#else
-    [ConditionalFact(nameof(Root_Certificate_Installed))]
 #endif
+    [WcfFact]
+    [Condition(nameof(Root_Certificate_Installed))]
     [OuterLoop]
-    [ActiveIssue(1157)]
```
roncain: >[ActiveIssue(1157)] [](start = 4, length = 19)

I am in the process of adding back all the deleted unconditional ActiveIssues into the ToF conditional blocks, otherwise ToF tries to run them.  So assume that will be done across the board before the merge.

iamjasonp:
LGTM, effectively all boilerplate changes

PR #1430, opened 2016-08-01
Emit WCF ETW events whenever a test fail

hongdai:
* This is to help analyze test failures, especially on non window
platforms

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/TestEventListener.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/TestEventListener.cs
@@ -1,0 +1,92 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+
+namespace Infrastructure.Common
+{
+    /// <summary>Simple event listener</summary>
+    /// This class is based on https://github.com/dotnet/corefx/blob/master/src/Common/tests/System/Diagnostics/Tracing/TestEventListener.cs
+    public sealed class TestEventListener : EventListener
+    {
+        private Dictionary<string, bool> _targetSourceName = new Dictionary<string, bool>();
+        private readonly EventLevel _level;
+
+        private Action<EventWrittenEventArgs> _eventWritten;
+        private List<EventSource> _tmpEventSourceList = new List<EventSource>();
+
+        public TestEventListener(List<string> targetSourceNames, EventLevel level)
+        {
+            // Store the arguments
+            foreach (var targetSourceName in targetSourceNames)
+            {
+                _targetSourceName.Add(targetSourceName, true);
+            }
+            _level = level;
+
+            LoadSourceList();
+        }
+
+        public Action<EventWrittenEventArgs> EventWritten
+        {
+            get { return _eventWritten; }
+            set { _eventWritten = value; }
+        }
+
+        private void LoadSourceList()
+        {
+            // The base constructor, which is called before this constructor,
+            // will invoke the virtual OnEventSourceCreated method for each
+            // existing EventSource, which means OnEventSourceCreated will be
+            // called before _targetSourceName have been set.  As such,
+            // we store a temporary list that just exists from the moment this instance
+            // is created (instance field initializers run before the base constructor)
+            // and until we finish construction... in that window, OnEventSourceCreated
+            // will store the sources into the list rather than try to enable them directly,
+            // and then here we can enumerate that list, then clear it out.
+            List<EventSource> sources;
+            lock (_tmpEventSourceList)
+            {
+                sources = _tmpEventSourceList;
+                _tmpEventSourceList = null;
+            }
+            foreach (EventSource source in sources)
+            {
+                EnableSourceIfMatch(source);
+            }
+        }
+
+        protected override void OnEventSourceCreated(EventSource eventSource)
+        {
+            List<EventSource> tmp = _tmpEventSourceList;
+            if (tmp != null)
+            {
+                lock (tmp)
+                {
+                    if (_tmpEventSourceList != null)
+                    {
+                        _tmpEventSourceList.Add(eventSource);
+                        return;
+                    }
+                }
+            }
+
+            EnableSourceIfMatch(eventSource);
+        }
+
+        private void EnableSourceIfMatch(EventSource source)
+        {
+            if (_targetSourceName.ContainsKey(source.Name))
+            {
+                EnableEvents(source, _level);
+            }
+        }
+        protected override void OnEventWritten(EventWrittenEventArgs eventData)
+        {
+            _eventWritten?.Invoke(eventData);
+        }
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
@@ -2,7 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit.Abstractions;
@@ -17,6 +21,8 @@ namespace Infrastructure.Common
         private readonly string _skippedReason;
         private readonly bool _isTheory;
         private readonly IMessageSink _diagnosticMessageSink;
+
+        static TestEventListener s_testListener = new TestEventListener(new List<string>() { "Microsoft-Windows-Application Server-Applications" }, EventLevel.Verbose);
 
         internal WcfTestCase(IXunitTestCase testCase, string skippedReason = null, bool isTheory = false, IMessageSink diagnosticMessageSink = null)
         {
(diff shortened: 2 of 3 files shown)


src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -40,11 +43,41 @@ internal WcfTestCase(IXunitTestCase testCase, string skippedReason)
...
-        public Task<RunSummary> RunAsync(
+        public async Task<RunSummary> RunAsync(
             IMessageSink diagnosticMessageSink, IMessageBus messageBus, object[] constructorArguments,
             ExceptionAggregator aggregator, CancellationTokenSource cancellationTokenSource)
         {
-            return new XunitTestCaseRunner(this, DisplayName, _skippedReason, constructorArguments, TestMethodArguments, messageBus, aggregator, cancellationTokenSource).RunAsync();
+            ConcurrentQueue<EventWrittenEventArgs> events = new ConcurrentQueue<EventWrittenEventArgs>();
+	          s_testListener._eventWritten = events.Enqueue;
+
+            var runsummary = await (new XunitTestCaseRunner(this, DisplayName, _testCase.SkipReason, constructorArguments, TestMethodArguments, messageBus, aggregator, cancellationTokenSource)).RunAsync();
```
roncain: @hongdai -- I discovered we had a bug in WcfTestCase that I have fixed in PR #1433.  We cannot use XunitTestCaseRunner for tests marked with WcfTheory or it will fail.  It's not a difficult change, so we could potentially merge mine first and apply the fix to this PR.  But if you want to go first, I will update my PR with your changes here.  In either case, we can't move to WcfTheory until my fixes are in one of these PR's.

roncain: Nit: I prefer to avoid use of 'var' unless it is clear from context what the type is.  I can't tell from this code what type 'runsummary' is.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -40,11 +43,41 @@ internal WcfTestCase(IXunitTestCase testCase, string skippedReason)
...
+
+            if (runsummary.Failed > 0)
```
roncain: Be aware that Theory based tests can run multiple tests under this 'await' call (which is why Failed is an int).  It may not matter, but this new ETW logic will fire if any of the multiple tests fail, even if some of them succeed.  I point that out so you know, but we probably can ignore that for now.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -40,11 +43,41 @@ internal WcfTestCase(IXunitTestCase testCase, string skippedReason)
...
+            {
+                s_testListener._eventWritten = null;
+                Console.WriteLine(string.Format("---ETW Trace for Test {0} Begins---", DisplayName));
+                foreach (var item in events)
+                {
+                    try
+                    {
+                        Console.WriteLine(string.Format(item.Message, item.Payload));
+                    }
+                    catch (Exception)
```
roncain: Can you explain why the "try" can fail?  And why does the "catch" enumerate over the Payload items?  If 'try' can fail, isn't it likely the code in 'catch' can also fail?

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -40,11 +43,41 @@ internal WcfTestCase(IXunitTestCase testCase, string skippedReason)
...
+                    {
+                        Console.WriteLine(string.Format("ETW message: {0}, payload below was received", item.Message));
+                        foreach (var payloadPara in item.Payload)
```
roncain: Nit: another place where I would prefer to see the type name instead of 'var'.  I can't tell from context what a "payloadPara" is.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -2,7 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics.Tracing;
```
roncain: Does this compile on N?  The fact you added new dependencies may mean we need to update the ToF wcfopen.settings.targets or make this code conditional in N.  If we make any conditional code in these files, we will need to think about how the appropriate $(DefineConstant) gets set.   I don't think FEATURE_NETNATIVE is set when building tests except in ToF.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
hongdai: I will specify the type explicitly.

Your PR #1433 should go first, I'll merge it.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
hongdai: Good to know. I'll ignore it for now since we don't have a better solution.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
hongdai: Good point! I will test it on projectN.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
hongdai: The exception is caused by mismatch format arguments somehow. The catch won't fail. I will add a comment.

roncain: Do we understand why there is a mismatch?  It sounds like we could be ignoring another bug somewhere else in the stack that should be fixed.  If we don't fix it, it could cause other consumers of this information to fail.  Not strictly required to merge, but we should understand what is happening.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
hongdai: The payload isn't big from my testing result, we don't usually include very large message in the functional test. The risk is also mitigated by we only output ETW for failing cases.

Agree to collect them to a StringBuilder and do a single WriteLine.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
hongdai: We don't understand why there's a mismatch as they actually match when I inspect them under debug. Agree we should investigate the root cause to ensure there isn't a bug. I will open a issue to track it.

iamjasonp:
From what I read, we might have an issue when doing the multiple Console.WriteLines due to our test runners being able to run multiple tests at the same time (for example, look at our CI runs and then look at the output from InstallRootCertificate.sh - it's often interleaved between different tests)

Can we put some kind of mutex on the output so it stays in one block? Or at the very least it might be possible to use a StringBuilder to aggregate the output so there's less chance of output being interleaved with other stuff

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -40,11 +43,41 @@ internal WcfTestCase(IXunitTestCase testCase, string skippedReason)
...
+                        Console.WriteLine(string.Format(item.Message, item.Payload));
+                    }
+                    catch (Exception)
+                    {
+                        Console.WriteLine(string.Format("ETW message: {0}, payload below was received", item.Message));
+                        foreach (var payloadPara in item.Payload)
+                        {
+                            if (payloadPara != null)
+                            {
+                                Console.WriteLine(payloadPara);
```
iamjasonp: Consider: prefix *payloadPara* with "    [DisplayName]"

hongdai: There is only one test running at one time. There is nothing else we can do in the test to ensure the console sequence. Good idea to prefix payloadPara with test name. I'll make the change.

iamjasonp: Yep, agree there's only one test at a time :) prefixing payloadPara will help in case the output log ends up getting split somehow. thanks!

hongdai:
Tested on ProjectN and send out a codeflow to add the new system.diag.tracing dependency.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -44,13 +50,48 @@ internal WcfTestCase(IXunitTestCase testCase, string skippedReason = null, bool
...
             ExceptionAggregator aggregator, CancellationTokenSource cancellationTokenSource)
         {
-            return _isTheory
-                ? new XunitTheoryTestCaseRunner(this, DisplayName, _skippedReason, constructorArguments, _diagnosticMessageSink, messageBus, aggregator, cancellationTokenSource).RunAsync()
-                : new XunitTestCaseRunner(this, DisplayName, _skippedReason, constructorArguments, TestMethodArguments, messageBus, aggregator, cancellationTokenSource).RunAsync();
+            ConcurrentQueue<EventWrittenEventArgs> events = new ConcurrentQueue<EventWrittenEventArgs>();
+	          s_testListener._eventWritten = events.Enqueue;
+
+            RunSummary runsummary = await (_isTheory ? new XunitTheoryTestCaseRunner(this, DisplayName, _skippedReason, constructorArguments, _diagnosticMessageSink, messageBus, aggregator, cancellationTokenSource).RunAsync()
+                : new XunitTestCaseRunner(this, DisplayName, _skippedReason, constructorArguments, TestMethodArguments, messageBus, aggregator, cancellationTokenSource).RunAsync());
```
roncain: Nit: the formatting of this statement would be better if it were broken onto separate lines.  Not blocking but consider reformatting if you touch this file again

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -44,13 +50,48 @@ internal WcfTestCase(IXunitTestCase testCase, string skippedReason = null, bool
...
+            if (runsummary.Failed > 0)
+            {
+                s_testListener._eventWritten = null;
+                StringBuilder etwOutput = new StringBuilder();
+                etwOutput.AppendLine(string.Format("---ETW Trace for Test {0} Begins---", DisplayName));
+                foreach (var item in events)
+                {
+                    try
+                    {
+                        etwOutput.AppendLine(string.Format(DisplayName + ":" + item.Message, item.Payload));
```
roncain: >string.Format(DisplayName + ":" + item.Message, item.Payload [](start = 45, length = 60)

Is this String.Format the way you intended?  I can't tell what it will do. It is passing DisplayName and item.Message as the format template -- which could easily blow up if they contain formatting characters like {1}. Was it intended to be something like String.Format("{0}:{1} {2}", DisplayName, item.Message, item.Payload)?

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -44,13 +50,48 @@ internal WcfTestCase(IXunitTestCase testCase, string skippedReason = null, bool
...
+                    }
+                    //The mumber of parameters in Payload does not match the number of arguments in the item.Message and thus cause a 
+                    // FormatException occationally, In this case, we catch and output all items in the payload and the Message without formatting the message.
+                    // https://github.com/dotnet/wcf/issues/1440 is opened to investigate the root cause of the mismatch exception.
+                    catch (Exception)
+                    {
+                        etwOutput.AppendLine(string.Format("ETW message: {0}, payload below was received", item.Message));
```
roncain: >etwOutput.AppendLine(string.Format("ETW message: {0}, payload below was received", item.Message)); [](start = 24, length = 98)

Since we are trying to understand issue #1440, I recommend we write to the StringBuilder something to help us debug, something like this:
etwOutput.AppendLine(String.Format("ETW message encountered exception '{0}' using DisplayName '{1}', format '{2}' and {3} payload items", 
                                            exception.Message, DisplayName, item.Message, item.Payload.Count()));

src/System.Private.ServiceModel/tests/Common/Infrastructure/TestEventListener.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace Infrastructure.Common
+{
+    /// <summary>Simple event listener</summary>
+    /// This class is based on https://github.com/dotnet/corefx/blob/master/src/Common/tests/System/Diagnostics/Tracing/TestEventListener.cs
+    public sealed class TestEventListener : EventListener
+    {
+        private Dictionary<string, bool> _targetSourceName = new Dictionary<string, bool>();
+        private readonly EventLevel _level;
+
+        public Action<EventWrittenEventArgs> _eventWritten;
```
roncain: > public Action<EventWrittenEventArgs> _eventWritten [](start = 7, length = 51)

It would be better to make this a property, not a public field.  The WcfTestCase class is modifying this member directly

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
```diff
@@ -44,13 +50,48 @@ internal WcfTestCase(IXunitTestCase testCase, string skippedReason = null, bool
...
-                : new XunitTestCaseRunner(this, DisplayName, _skippedReason, constructorArguments, TestMethodArguments, messageBus, aggregator, cancellationTokenSource).RunAsync();
+            ConcurrentQueue<EventWrittenEventArgs> events = new ConcurrentQueue<EventWrittenEventArgs>();
+	          s_testListener._eventWritten = events.Enqueue;
+
+            RunSummary runsummary = await (_isTheory ? new XunitTheoryTestCaseRunner(this, DisplayName, _skippedReason, constructorArguments, _diagnosticMessageSink, messageBus, aggregator, cancellationTokenSource).RunAsync()
+                : new XunitTestCaseRunner(this, DisplayName, _skippedReason, constructorArguments, TestMethodArguments, messageBus, aggregator, cancellationTokenSource).RunAsync());
+
+            if (runsummary.Failed > 0)
+            {
+                s_testListener._eventWritten = null;
```
roncain: >s_testListener._eventWritten = null [](start = 16, length = 35)

Unless I misunderstand the code, I think this code to set _eventWritten to null should be outside this if block.  We don't want to leave a static member initialized to a non-null Action after the test is complete, regardless if we pass or fail.  As mentioned elsewhere, I think _eventWritten should be a property, not a field.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
hongdai: Ok.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
hongdai: Agreed.

src/System.Private.ServiceModel/tests/Common/Infrastructure/TestEventListener.cs
hongdai: Agreed.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/WcfTestCase.cs
hongdai: Agreed.

roncain:
LGTM.
Thanks for merging to my changes and the feedback update @hongdai .

PR #1431, opened 2016-08-01
Add execution permission to the cert installation file

hongdai:
* I'm getting "Command Not Found" error without this fix.
* After this fix, cert tests are passing in Helix for Ubuntu

### DIFF
--- a/src/certtest.props
+++ b/src/certtest.props
@@ -5,6 +5,7 @@
     </Content>
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


roncain:
LGTM

iamjasonp:
This is actually mildly worrying - the repo's copy should already be chmodded correctly

iamjasonp:
Also, there should be no need to sudo to chmod. #1453 changes this.

PR #1433, opened 2016-08-02
Fix WcfTheory discoverer to use theory data

roncain:
The WcfTheoryAttribute test discover was creating an xunit
test runner that worked only for Fact, but not for Theory.

The solution was to refactor slightly and to create a
theory-based xunit test runner for non-skipped tests.


hongdai:
LGTM.

roncain:
Thanks @hongdai -- merged now, so you can go ahead and combine it into your PR.  I'll review as soon as you push up a new commit

PR #1435, opened 2016-08-02
Change unit tests to use WcfFact

roncain:
Modifies all the unit tests to use WcfFact instead of Fact
and WcfTheory instead of Theory.

We retain some conditional code temporarily to allow running
in ToF as well as normally in GitHub.

### DIFF
--- a/src/System.ServiceModel.Duplex/tests/ServiceModel/CallbackBehaviorAttributeTest.cs
+++ b/src/System.ServiceModel.Duplex/tests/ServiceModel/CallbackBehaviorAttributeTest.cs
@@ -3,11 +3,15 @@
 // See the LICENSE file in the project root for more information.
 
 using System.ServiceModel;
+using Infrastructure.Common;
 using Xunit;
 
 public static class CallbackBehaviorAttributeTest
 {
+#if FULLXUNIT_NOTSUPPORTED
     [Fact]
+#endif
+    [WcfFact]
     public static void Default_Ctor_Initializes_Correctly()
     {
         CallbackBehaviorAttribute cba = new CallbackBehaviorAttribute();
@@ -16,7 +20,10 @@ public static class CallbackBehaviorAttributeTest
         Assert.True(cba.UseSynchronizationContext, "UseSynchronizationContext should have been true");
     }
 
+#if FULLXUNIT_NOTSUPPORTED
     [Theory]
+#endif
+    [WcfTheory]
     [InlineData(false)]
     [InlineData(true)]
     public static void AutomaticSessionShutdown_Property_Is_Settable(bool value)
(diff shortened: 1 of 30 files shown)


roncain:
Reviewers: this PR contains a bug fix copied directly from PR #1433, otherwise all the [WcfTheory] tests throw exceptions from xunit.  Ideally #1433 will be merged first, and those files removed from this PR.  But I need to work with @hongdai before merging #1433 because she is changing the same files in #1430.

Would you please review and sign-off strictly the touches to the unit tests themselves?  They are 99.99% mechanical boiler-plat changes. I will not merge until we resolve the issue with #1433 and #1430.

zhenlan:
I looked through all the unit tests. None of them uses conditional issue or conditions. Should we leave unit tests alone? Our WcfFactAttribute etc. infrastructure pieces may go through more iterations of updates in the future, so we can reduce unnecessary churns. It's not a problem for these unit tests to use original xunit attributes if they are sufficient for the needs.

roncain:
@Zhenlan -- you are correct no unit tests use ConditionAttribute, but I feel strongly unit tests should move forward to WcfFact and WcfTheory -- otherwise we cannot use IssueAttribute.  Remember, Issue is only an attribute discovered by WcfFact and WcfTheory. I think we want common handling for issues across all test projects.  Incidentally, this PR uses Issue on a unit test that formerly was doing a forced skip.

zhenlan:
These tests are messier after the change and we almost know for sure we will go through another round to remove those `#if FULLXUNIT_NOTSUPPORTED` blocks. I agree we should handle active issues with IssueAttribute consistently, but none of these tests need IssueAttribute. Can we wait till a need arises? We did not touch unit tests for quite some time.

roncain:
@zhenlan -- let's discuss.  It was my understanding after talking with @StephenBonikowsky that we wanted to prepare the unit tests to use the new syntax ([WcfFact], [Issue], etc.) so we can continue with the verification of UWP in Helix.  Yes, we have not touched unit tests in a long time (our bad), but they are the first tests being verified against UWP.  I believe verifying Helix UWP against only the old [Fact] syntax would not verify what we need to verify.

And it is not appropriate to make them unrunnable in ToF until we have fully moved to UWP and are getting 100% passing results. The "mess" you describe is exactly the same we have in scenarios as a temporary measure to run in all worlds (ToF, GitHub, Helix UWP).  And yes, once we decide to move away from ToF, there will be a one-time pass to remove all that conditional code across all tests.  That one-time pass will be very mechanical and will take less time than what it took to prepare this PR.

roncain:
Reviewers: we are going to wait on merging this PR until @StephenBonikowsky has verified the existing unit tests pass on Helix UWP for both the existing FactAttibute as well as trying a single WcfFactAttribute and WcfTheoryAttribute.  If those look good, we will merge, and then clean out the old ToF conditional logic once we are fully moved to Helix.

/cc: @zhenlan @iamjasonp @StephenBonikowsky

iamjasonp:
LGTM

PR #1437, opened 2016-08-02
Add RunTestsWithIssues TestProperty

roncain:
Setting this property to 'true' means that tests marked
with IssuesAttribute will not be skipped.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/testproperties.props
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/testproperties.props
@@ -150,6 +150,16 @@
     <PropertyGroup Condition="'$(SSL_Available)' == ''">
       <SSL_Available></SSL_Available>
     </PropertyGroup>
+
+    <!-- The default value of $(IncludeTestsWithIssues) is blank, which is equivalent fo 'false'.
+         The check of $(WithCategories) preserves the convention used by other repo's where the
+         keyword 'failing' inside $(WithCategories) asks that tests with [ActiveIssue] be run.
+         Optionally, $(IncludeTestsWithIssues) can be a semicolon-separated list of issue numbers.
+         Tests marked with an [Issue] whose issue number is in this list will be run rather than skipped.
+    -->
+    <PropertyGroup Condition="'$(IncludeTestsWithIssues)' == '' and $(WithCategories.ToLower().Contains('failing'))">
+      <IncludeTestsWithIssues>true</IncludeTestsWithIssues>
+    </PropertyGroup>
   
   <!--
      GeneratedTestPropertiesFileName:
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Collections.Generic;
 using Xunit.Abstractions;
 
 namespace Infrastructure.Common
@@ -22,6 +23,8 @@ namespace Infrastructure.Common
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
     public class IssueAttribute : WcfSkippableAttribute
     {
+        private static HashSet<int> s_includeIssuesHash = DetectIssuesToInclude();
+
         public int Issue { get; private set; }
 
         public FrameworkID Framework {get; set; }
@@ -45,25 +48,34 @@ namespace Infrastructure.Common
 
         public override string GetSkipReason(ITestMethod testMethod)
         {
+            // If we are deliberately including this issue in the test run,
+            // return "don't skip" response unconditionally.
+            if (ShouldIncludeThisIssue(Issue))
+            {
+                return null;
+            }
+
             string repositoryAndIssue = String.Format("{0} #{1}", Repository, Issue);
 
             if (Framework.MatchesCurrent())
             {
+                // Don't skip if we are running tests with issues.
+                // But match the Framework first so we don't run tests 
                 return String.Format("{0} on framework \"{1}\" (filter is \"{2}\")",
-                                    repositoryAndIssue,
-                                    FrameworkHelper.Current.Name(),
-                                    Framework.Name());
+                                     repositoryAndIssue,
+                                     FrameworkHelper.Current.Name(),
+                                     Framework.Name());
             }
 
             if (OS.MatchesCurrent())
             {
                 return String.Format("{0} on OS \"{1}\" (filter is \"{2}\")",
-                                    repositoryAndIssue,
-                                    OSHelper.Current.Name(),
-                                    OS.Name());
+                                     repositoryAndIssue,
+                                     OSHelper.Current.Name(),
+                                     OS.Name());
             }
 
-            // If no specific OS or Framework filters, it applies to all
+            // If no specific OS or Framework filters, it applies to all.
             if (OS == OSID.None && Framework == FrameworkID.None)
             {
                 return String.Format("{0}",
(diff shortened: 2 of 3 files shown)


roncain:
Reviewers: the approach used in this PR should just work with any existing CI lab scripts that build with /p:WithCategories="OuterLoop;failing".  But for complete control, it is recommended that future scripts use either /p:RunTestsWithIssues=true in the build or set the environment variable RunTestsWithIssues=true.

src/System.Private.ServiceModel/tests/Common/Infrastructure/testproperties.props
```diff
@@ -150,6 +150,10 @@
     <PropertyGroup Condition="'$(SSL_Available)' == ''">
       <SSL_Available></SSL_Available>
     </PropertyGroup>
+
+  <PropertyGroup Condition="$(WithCategories.Contains('failing'))">
```
iamjasonp: NIT: is there any way to make this a case-insensitive compare?

iamjasonp:
LGTM, just a minor nit on case-sensitivity of "failing"

src/System.Private.ServiceModel/tests/Common/Infrastructure/testproperties.props
zhenlan: Just trying to understand... when `$(WithCategories.Contains('failing'))` will be true?

zhenlan:
There is a little bit ambiguity here - does `RunTestsWithIssues` mean we will only run tests that are decorated with `IssueAttribute` and does it matter if the Issue condition is actually applicable or not? How do you feel `IncludeTestsWithIssues`?

What will be the scenario we want to set this property to true? Do we care tests skipped by ConditionAttribute?

src/System.Private.ServiceModel/tests/Common/Infrastructure/testproperties.props
roncain: @iamjasonp -- I made the Contains case-insensitive using the pattern used by other targets files.
@zhenlan -- regarding when the condition will be true -- the syntax used by other repo's (and by one flavor of our former nightly runs) is to say /p:WithCategories="OuterLoop;failing" when you want to include tests marked with [ActiveIssue].  These semantics are part of build tools.  Now that we have moved to [Issue], I thought it would be reasonable to allow the build-tools syntax to continue to be used.

roncain:
I have pushed up a 2nd commit that incorporates feedback and also changes the semantics a little.  When experimenting with it, I was not satisfied with the "all or nothing" switch.  It forced me to run tests I didn't want run, including those with long timeouts or hangs. I think there are 3 valid scenarios:

1. Skip all tests marked with IssueAttribute (default VSO and CI)
2. Do not skip any tests marked with IssueAttribute (special VSO run to force all tests to run, including those with issues)
3. Do not skip tests marked with specific issue(s) (dev experience)

Option (3) is one I expect to use routinely to see if a specific bug has been fixed.  The only option today is to comment out IssueAttribute, potentially in many tests across many files.  But I want to be able to say "I think issue 470 is fixed now", so please don't skip any tests marked with 470."  I can do this without touching code by setting IncludeTestsWithIssues=470.

The latest PR allows IncludeTestsWithIssues to be one of these:

- Unspecified (default and means skip all tests marked with Issue)
- "all" (means don't skip any tests based on having IssueAttribute)
- "n1;n2,n3" (means don't skip tests with IssueAttribute containing n1, n2, or n3

To answer the question @zhenlan asked -- this logic has no effect on the ConditionAttribute.  It only controls whether the IssueAttribute alone gets to say "skip this test".  If IssueAttribute decides not to skip the test, ConditionAttribute is still evaluated and might choose to skip for its one reasons.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
```diff
@@ -73,5 +86,71 @@ public override string GetSkipReason(ITestMethod testMethod)
...
+            {
+                return true;
+            }
+
+            // Anything else is interpreted as a semicolon-separate list of
+            // issue numbers to include (i.e. not skip).
+            string[] issues = includeTestsWithIssues.Split(';');
+            foreach (string issue in issues)
+            {
+                if (String.IsNullOrEmpty(issue))
```
zhenlan: NIT: `String.Split` has the option to `RemoveEmptyEntries`

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
```diff
@@ -22,6 +23,9 @@ namespace Infrastructure.Common
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
     public class IssueAttribute : WcfSkippableAttribute
     {
+        private static bool s_includeAllIssues = DetectIssuesToInclude();
+        private static HashSet<int> s_includeIssuesHash;
```
zhenlan: I think the code is correct. It's just these skip test, skip issue, include test, include issue logics which together made my head hurt 😄.

I am wondering if it will be easier to understand if we keep only one variable:
`private static HashSet<int> s_IgnoredIssues = DetectIgnoredIssues();`

and an issue will be ignored based on logic below:
-  `s_IgnoredIssues==null`: No issues will be ignored. This is default behavior.
-  `s_IgnoredIssues.Count==0`: All issues will be ignored. The is the behavior of IncludeTestsWithIssues=all.
-  `s_IgnoredIssues.Count>0` : Only specific issues will be ignored.

An issue is ignored means the inclusion of a test will not be affected by the issue in any way.

src/System.Private.ServiceModel/tests/Common/Infrastructure/testproperties.props
```diff
@@ -150,6 +150,10 @@
     <PropertyGroup Condition="'$(SSL_Available)' == ''">
       <SSL_Available></SSL_Available>
     </PropertyGroup>
+
+  <PropertyGroup Condition="'$(IncludeTestsWithIssues)' == '' and $(WithCategories.ToLower().Contains('failing'))">
+    <IncludeTestsWithIssues>all</IncludeTestsWithIssues>
```
zhenlan: NIT: I personally feel `true` is more intuitive than `all` (if I just guess what value I can set).

zhenlan:
LGTM. I like this desgin and I think option 3 will be useful.
I have some recommendation for implementation but not blocking.

iamjasonp:
> Unspecified (default and means skip all tests marked with Issue)
"all" (means don't skip any tests based on having IssueAttribute)
"n1;n2,n3" (means don't skip tests with IssueAttribute containing n1, n2, or n3

Question on this: so if I wanted to run all the tests that are ActiveIssue'd out for OSX, then I would just specify the activeIssue number, right? 

Is there a provision for running all the tests skipped by a particular OS? Not a blocking issue, just a question (use case: I'm trying to test out some cases that are blocked by filter "AnyOSX" with more than one number)

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
iamjasonp: +1. I was reading through this code and it took a while to figure out what the state of s_includeIssuesHash was and how it impacted skipping/inclusion logic

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
```diff
@@ -73,5 +86,71 @@ public override string GetSkipReason(ITestMethod testMethod)
...
+            string[] issues = includeTestsWithIssues.Split(';');
+            foreach (string issue in issues)
+            {
+                if (String.IsNullOrEmpty(issue))
+                {
+                    continue;
+                }
+
+                int issueAsInt = 0;
+                if (!int.TryParse(issue, out issueAsInt))
```
iamjasonp: NIT: this check will allow -ve numbers through, but i don't think it does any harm even if negative numbers are passed

iamjasonp:
LGTM, some minor non blocking nits

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
roncain: Will do.

src/System.Private.ServiceModel/tests/Common/Infrastructure/xunit/IssueAttribute.cs
roncain: I hear you both. I simplified to a single static hashset with the semantics similar to what @zhenlan proposed.  But to minimize confusion, I avoid all use of "skip" or "ignore" in member names and only use "include".

src/System.Private.ServiceModel/tests/Common/Infrastructure/testproperties.props
roncain: Agreed. The name sounds Boolean.  And 'false' is also acceptable and treated the same as the default (i.e. don't include any tests with Issue)

roncain:
@iamjasonp -- I like the idea about being able to set OS (and maybe framework?) values to include but need to think about it some more for a separate PR.

PR #1439, opened 2016-08-02
Use IssueAttribute to disable WebSockets tests on Win7

roncain:
The Windows platforms we currently test against include Win7,
but the WebSockets protocol is not supported on that OS.
As a result, VSO runs report WebSockets failures on the Win7
runs, forcing us to investigate each time.

With this change, those tests will be skipped on Win7, and any
VSO failures reported will be legitimate issues to investigate.
This new IssueAttribute will not disable these tests anywhere else.


roncain:
@dotnet-bot test all outerloop

iamjasonp:
LGTM

zhenlan:
LGTM. Thanks!

PR #1452, opened 2016-08-04
Remove Issue(1398) from tests requiring cert+OSX

iamjasonp:
OSX certificate install should now work, so we can now allow these tests to run. 

~~There may still be tests failing because we're using SecureTransport instead of OpenSSL, so this PR may be further touched so that such tests point to a new issue instead~~

EDIT: guess not, CI all passed, so this can go ahead


hongdai:
LGTM.

roncain:
LGTM

PR #1453, opened 2016-08-04
Remove call to sudo when chmodding in certtest.props

iamjasonp:
There should be no need to call sudo when attmepting to chmod the InstallRootCertificates.sh
script. The file is owned by the text execution user, not root


hongdai:
LGTM.

PR #1455, opened 2016-08-05
Testing xunit for uwp

StephenBonikowsky:
The unit tests won't run against netcore50 by default. To run them go to the test project and run...
msbuild  /p:TestTFM=netcore50 /p:TestNugetRuntimeId=win10-x64-aot /t:RebuildAndTest /p:UseDotNetNativeToolchain=true /p:TestWithCore=true

### DIFF
--- a/src/System.ServiceModel.Duplex/tests/ServiceModel/CallbackBehaviorAttributeTest.cs
+++ b/src/System.ServiceModel.Duplex/tests/ServiceModel/CallbackBehaviorAttributeTest.cs
@@ -2,13 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-
+using Infrastructure.Common;
 using System.ServiceModel;
 using Xunit;
 
 public static class CallbackBehaviorAttributeTest
 {
-    [Fact]
+    [ConditionalFact]
     public static void Default_Ctor_Initializes_Correctly()
     {
         CallbackBehaviorAttribute cba = new CallbackBehaviorAttribute();
@@ -17,7 +17,7 @@ public static class CallbackBehaviorAttributeTest
         Assert.True(cba.UseSynchronizationContext, "UseSynchronizationContext should have been true");
     }
 
-    [Theory]
+    [WcfTheory]
     [InlineData(false)]
     [InlineData(true)]
     public static void AutomaticSessionShutdown_Property_Is_Settable(bool value)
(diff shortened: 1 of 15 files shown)


src/System.ServiceModel.Http/tests/ServiceModel/BasicHttpBindingTest.cs
```diff
@@ -169,6 +169,7 @@ public static void Name_Property_Sets(string value)
     [Theory]
     [InlineData(null)]
     [InlineData("")]
+    [ActiveIssue(1449)]
```
iamjasonp: Can we do something like 

```
      [Theory]
      [InlineData("")]
      [InlineData(null)]
#if FEATURE_NETNATIVE
      [ActiveIssue(1449)]
#endif
```

or conversely 

```
      [Theory]
      [InlineData("")]
#if !FEATURE_NETNATIVE
      [InlineData(null)]
#endif
```

So that we don't end up indiscriminately skipping all our Theory tests will null in UWP?

iamjasonp:
Nonblocking question about whether or not we can maintain _some_ of the [Theory] coverage while excluding the _null_s in UWP.

Otherwise, LGTM

src/System.ServiceModel.Duplex/tests/ServiceModel/CallbackBehaviorAttributeTest.cs
```diff
@@ -2,13 +2,12 @@
...
 // See the LICENSE file in the project root for more information.
 
-using Infrastructure.Common;
 using System.ServiceModel;
 using Xunit;
 
 public static class CallbackBehaviorAttributeTest
 {
-    [ConditionalFact]
+    [Fact]
```
iamjasonp: Should this be WcfFact?

iamjasonp: Wait, nevermind, this is an intermediate commit only

src/System.ServiceModel.Http/tests/ServiceModel/BasicHttpBindingTest.cs
StephenBonikowsky: We could, my thinking was that I can't remember the last time a unit test failed for a product issue, and Ron's PR is going to pave right over this adding the new Wcf conditional fact that will do what you are requesting in a much more elegant way. So we would only lose coverage for a few days at most.

iamjasonp: >  Ron's PR is going to pave right over this adding the new Wcf conditional fact that will do what you are requesting in a much more elegant way.

Let's leave it as-is then, and have the lost coverage for a few days; no point in doing extra work

iamjasonp:
LGTM; merging on @StephenBonikowsky's behalf per request

src/System.ServiceModel.Http/tests/ServiceModel/BasicHttpBindingTest.cs
roncain: When I merge in my big unit test sweep for WcfFact, I'll convert the ActiveIssue to the new Issue form and condition them on FrameworkID.NETNative

PR #1457, opened 2016-08-05
Remove the custom rd.xml's initially required for UWP

roncain:
We added custom rd.xml files for some of our test libraries
for UWP testing because their types were not Reflectable.
The issue was that build tools considered them framework
projects because their names did not end with ".Tests.csproj".
As a consequence, they were signed as framework assemblies,
and the ILC prevented reflection into framework assemblies.

The fix is to set $(IsTestProject) to true inside the .csproj of
each of our test libraries.  And to remove the custom rd.xml's.


StephenBonikowsky:

StephenBonikowsky:
LGTM

Issue #1449, opened 2016-08-04
Unit tests using null values are failing in UWP.


StephenBonikowsky:
Unit tests that use 'null' as an "InlineData" value are failing. No other data is available right now as there are no logs produced. Theory tests do not run in ToF so this is the first time these tests have been run in a long time (in NetNative).

System.ServiceModel.Primitives\tests\ServiceModel: MessageContractTest.WrapperName_Property_Sets_Throws_ArgumentNull
System.ServiceModel.Primitives\tests\ServiceModel: MessageContractTest.WrapperNamespace_Property_Sets
System.ServiceModel.Primitives\tests\Channels: CustomBindingTest.CustomBinding_Name_Property_Set_Throws
System.ServiceModel.Http\tests\ServiceModel: BasicHttpBindingTest.Name_Property_Set_Invalid_Value_Throws
System.ServiceModel.Http\tests\ServiceModel: BasicHttpBindingTest.Namespace_Property_Set_Invalid_Value_Throws

roncain:
Stack trace from bin\tests\...\testResults.xml:
```
      </test>
      <test name="MessageContractTest.WrapperNamespace_Property_Sets" type="MessageContractTest" method="WrapperNamespace_Property_Sets" time="0" result="Fail">
        <failure exception-type="$BlockedFromReflection_1_d9deefc4">
          <message><![CDATA[$BlockedFromReflection_1_d9deefc4 : Reflection_InsufficientMetadata_NoHelpAvailable, <unavailable>. For more information, visit http://go.microsoft.com/fwlink/?LinkId=623485]]></message>
          <stack-trace><![CDATA[   at System.Reflection.Runtime.General.ReflectionCoreCallbacksImplementation.GetTypeInfo(Type type) in f:\dd\ndp\fxcore\CoreRT\src\System.Private.Reflection.Core\src\System\Reflection\Runtime\General\ReflectionCoreCallbacksImplementation.cs:line 38
   at Internal.Reflection.Core.Execution.Binder.DefaultBinder.BindToMethod($MethodBase[] match, Object[]& args) in f:\dd\ndp\fxcore\src\Common\Internal\Reflection\Execution\Binder\DefaultBinder.cs:line 172
[...]

PR #1458, opened 2016-08-08
Reactivate unit tests disabled by issue #1449

roncain:
Some unit tests were disabled when run against UWP due to
a known issue in the toolchain with InlineDataAttribute(null).
This PR applies a workaround to that toolchain issue and reactivates
the tests.


roncain:
Reviewers: this lights up several of the disabled unit tests in UWP.  I left breadcrumbs in the code so we can come back and remove the workaround when the issue is resolved in the toolchain.

mconnew:
LGTM

iamjasonp:
LGTM

PR #1460, opened 2016-08-08
Enhance OSID detection when running under NET Native

### LINKED ISSUE #1459: Infrastructure cannot determine OS version when run under UWP.
The OSID chosen for the currently running machine is "AnyWindows" when running under UWP.  This incorrectly matches any IssueAttributes marked with OSID.Windows_7, causing them to be skipped.  The cause is that the OSPlatform returns only the description "Microsoft Windows" without any version information, and we use the version information to match specific Windows versions.

The fix will be to either find the correct way to manifest the test runner so that it shows the real version of Windows it is running on or to use a better heuristic when version is not available.

The following tests were skipped in UWP due to this issue:

WebSocketTests.WebSocket_Http_WSTransportUsageAlways_DuplexCallback_GuidRoundtrip
WebSocketTests.WebSocket_Http_RequestReply_TextStreamed
WebSocketTests.WebSocket_Http_RequestReply_BinaryStreamed
WebSocketTests.WebSocket_WSScheme_WSTransportUsageAlways_DuplexCallback_GuidRoundtrip
WebSocketTests.WebSocket_Http_WSTransportUsageDefault_DuplexCallback_GuidRoundtrip

roncain:
RuntimeInformation.OSDescription is hard-coded to return a generic
"Microsoft Windows" description when compiled for NET Native. This
masks from our OSID detection which version of Windows is running.

Prior to this fix, we returned "AnyWindows" to indicate we did not
know, but this caused some IssueAttributes to skip Win7 tests.

The fix is to refine the OSID detection slightly when the hard-coded
"Microsoft Windows" description is returned.

Fixes #1459

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/OSHelper.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/OSHelper.cs
@@ -12,6 +12,12 @@ namespace Infrastructure.Common
     // RuntimeInformation.OSDescription
     public static class OSHelper
     {
+        // Hard-coded string literals used by RuntimeInformation.OSDescription
+        // for conditionally compiled code for specific OS versions.  Other
+        // versions of Windows obtain their description at runtime.
+        private const string MicrosoftWindowsName = "Microsoft Windows";
+        private const string MicrosoftWindowsPhoneName = "Microsoft Windows Phone";
+
         private static bool _detectedOSID = false;
         private static OSID _currentOSID = 0;
         private static string _currentOSDescription;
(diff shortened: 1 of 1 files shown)


roncain:
Reviewers: this change lights up 3 of the WebSockets Outerloop tests in UWP.  The other WS tests are skipped for issues that are NET Native specific.

src/System.Private.ServiceModel/tests/Common/Infrastructure/OSHelper.cs
```diff
@@ -169,7 +169,39 @@ private static OSID OSIDfromOSDescription()
...
                 {
-                    return pair.Item2;
+                    OSID detectedID =  pair.Item2;
+
+                    // A match of "AnyWindows" means we know it was Windows but don't know which version.
+                    if (detectedID == OSID.AnyWindows)
+                    {
+                        // "Microsoft Windows" is hard-coded at compilation time by RuntimeInformation.OSDescription
+                        // for win8 and netcore50 builds.  See if we had an exact match (not just prefix-match).
+                        if (string.Equals("Microsoft.Windows", CurrentOSDescription, StringComparison.OrdinalIgnoreCase))
```
mconnew: Did you intend to have a period between Microsoft and Windows?

iamjasonp:

iamjasonp:
LGTM (but please look at @mconnew's https://github.com/dotnet/wcf/pull/1460#discussion_r73973726)

src/System.Private.ServiceModel/tests/Common/Infrastructure/OSHelper.cs
roncain: Doh!  Good catch.  No, the period is wrong.  I fixed it by pulling these hard-coded literals up into private const's to avoid this kind of type-o in the future.

PR #1462, opened 2016-08-09
Fix UWP issue #1450 and re-enable tests disabled for it.

### LINKED ISSUE #1450: Unit tests for valid and invalid Encodings fail in UWP.
Theory test are now being run for the first time in a long time and several tests that use the 'ValidEncodings' and 'InvalidEncodings' properties from System.ServiceModel.Tests.Common.TestData are failing.

No call stack is currently available as no test case logging yet exists for UWP runs from git.

Test cases are...
System.ServiceModel.Primitives\tests\Channels: TextMessageEncodingBindingElementTest.WriteEncoding_Property_Sets
System.ServiceModel.Primitives\tests\Channels: TextMessageEncodingBindingElementTest.WriteEncoding_Property_Set_Throws_For_Invalid_Encodings
System.ServiceModel.Http\tests\ServiceModel: BasicHttpBindingTest.TextEncoding_Property_Sets
System.ServiceModel.Http\tests\ServiceModel: BasicHttpBindingTest.TextEncoding_Property_Set_Invalid_Value_Throws

roncain:
Issue #1450 was caused in UWP because the MemberData attribute
was hiding specific Encoding types from the reducer, resulting in
MissingMetadataException when xunit was doing test discovery.

The fix is to add back a custom rd.xml file to the project containing
the TestData class to grant Reflection permissions for types affected by this.

Fixes #1450

### DIFF
--- a/src/System.ServiceModel.Http/tests/ServiceModel/BasicHttpBindingTest.cs
+++ b/src/System.ServiceModel.Http/tests/ServiceModel/BasicHttpBindingTest.cs
@@ -239,40 +239,6 @@ public static class BasicHttpBindingTest
         Assert.Throws<ArgumentNullException>(() => binding.Namespace = value);
     }
 
-    public static MemberDataSet<Encoding> ValidEncodings
-    {
-        get
-        {
-            return new MemberDataSet<Encoding>
-                {
-                    { Encoding.BigEndianUnicode },
-                    { Encoding.Unicode },
-                    { Encoding.UTF8 },
-                };
-        }
-    }
-
-    public static MemberDataSet<Encoding> InvalidEncodings
-    {
-        get
-        {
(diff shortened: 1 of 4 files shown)


src/System.ServiceModel.Http/tests/ServiceModel/BasicHttpBindingTest.cs
```diff
@@ -239,40 +239,6 @@ public static void Namespace_Property_Set_Invalid_Value_Throws(string value)
         Assert.Throws<ArgumentNullException>(() => binding.Namespace = value);
     }
 
-    public static MemberDataSet<Encoding> ValidEncodings
```
roncain: Reviewers: this section was removed because they were already in TestData.cs.  I wanted them to exist in only one place, and in the one place that had an rd.xml for its types.

hongdai:
LGTM.

PR #1466, opened 2016-08-09
Omit empty ETW "Begin" and "End" messages if no events to report

### LINKED ISSUE #1465: Don't log ETW begin and end messages if no ETW events to trace
Running in NET Native, all failing tests have this kind of extraneous output in the test result:
```
  ---ETW Trace for Test IdentityTests.ServiceIdentityNotMatch_Throw_MessageSecurityException Begins---
  ---ETW Trace for Test IdentityTests.ServiceIdentityNotMatch_Throw_MessageSecurityException Ends---
```

The new logic to write ETW traces to the Console on test failure should not write out the "Begins" string and the "Ends" string unless there is a non-zero number of events in the payload.

roncain:
The new logic to Console.WriteLine ETW events for tests that fail
is writing out extraneous empty "Begin" and "End" blocks when run
on UWP.  This is because UWP doesn't currently support ETW, and it
has no events to report.

The fix is to skip the entire Console.WriteLine block of code if
there are no ETW events to report.

Fixes #1465


hongdai:
LGTM.

Issue #1467, opened 2016-08-09
Several tests Assert about missing UWP Certificate when run in UWP.


roncain:
SslStreamSecurityUpgradeProvider.OnInitiateUpgradeAsync is triggering this Assert in several tests when run under UWP:
```
Contract.Assert(clientCertificate != null, "Missing UWP Certificate as an attachment to X509Certificate2");
```
Failing tests include:

CustomBindingTests.DefaultSettings_Tcp_Binary_Echo_RoundTrips_String
ExpectedExceptionTests.TCP_ServiceCertExpired_Throw_MessageSecurityException
ExpectedExceptionTests.TCP_ServiceCertFailedCustomValidate_Throw_Exception
ExpectedExceptionTests.TCP_ServiceCertRevoked_Throw_SecurityNegotiationException
Tcp_ClientCredentialTypeCertificateCanonicalNameTests.Certificate_With_CanonicalName_Fqdn_Address_EchoString
Tcp_ClientCredentialTypeCertificateCanonicalNameTests.Certificate_With_CanonicalName_Localhost_Address_EchoString
Tcp_ClientCredentialTypeCertificateCanonicalNameTests.Certificate_With_CanonicalName_DomainName_Address_EchoString
Tcp_ClientCredentialTypeTests.NetTcp_SecModeTrans_CertValMode_PeerTrust_Fails_Not_In_TrustedPeople
Tcp_ClientCredentialTypeTests.TcpClientCredentialType_Certificate_With_ServerAltName_EchoString
Tcp_ClientCredentialTypeTests.NetTcp_SecModeTrans_CertValMode_PeerOrChainTrust_Succeeds_ChainTrusted
Tcp_ClientCredentialTypeTests.NetTcp_SecModeTrans_CertValMode_ChainTrust_Succeeds_ChainTrusted
Tcp_ClientCredentialTypeTests.NetTcp_SecModeTrans_CertValMode_PeerTrust_Succeeds_In_TrustedPeople
IdentityTests.ServiceIdentityNotMatch_Throw_MessageSecurityException
IdentityTests.VerifyServiceIdentityMatchDnsEndpointIdentity

Typical stacktrace in testresults.xml:

```
[...]

PR #1468, opened 2016-08-09
Disable tests triggering Assert in UWP

roncain:
Several tests are now triggering a Contract.Assert when
run in UWP.  See issue #1467.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.4.1.0.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.4.1.0.cs
@@ -21,7 +21,10 @@ using Infrastructure.Common;
 
 public partial class ExpectedExceptionTests : ConditionalWcfTest
 {
+#if FULLXUNIT_NOTSUPPORTED
     [Fact]
+#endif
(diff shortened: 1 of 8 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.4.1.0.cs
```diff
@@ -21,7 +21,10 @@
 
 public partial class ExpectedExceptionTests : ConditionalWcfTest
 {
+#if FULLXUNIT_NOTSUPPORTED
     [Fact]
+#endif
+    [WcfFact]
```
roncain: Reviewers: a couple of these touches were cleanup when I noticed this file had been missed in updating from FactAttribute.  All the places that added the new IssueAttribute needed to move to WcfFactAttribute to work properly, so I updated the rest just as cleanup.

roncain:
Please do not merge yet.  We would prefer to fix the issue instead. but this PR is ready if we decide we need to disable them for UWP temporarily.

mconnew:
I've found the problem, it was a simple bug. The assert needed to move up inside an if block. I'll have a PR out very soon for this.

mconnew:
Issue is addressed in #1471

PR #1470, opened 2016-08-10
Fix project solution files

mconnew:
Renamed S.P.SM.sln to S.P.SM.All.sln and made it not auto build test projects
Added sln files for each test project
Create S.P.SM.sln file which only includes the S.P.SM project


mconnew:
I got fed up with VS taking so long to load our solution, so I fixed it and made our development process easier by creating a solution for each test project which includes the test project itself, plus two referenced projects and S.P.SM. This allows a cleaner VS environment and faster VS load times. The big solution still takes a while to load the first time as it builds the .vs folder, but subsequent loads are quite quick.

hongdai:
LGTM.

mconnew:
@dotnet-bot test outerloop Ubuntu14.01

mconnew:

Issue #1235, opened 2016-06-01
NegotiateStream in Tcp is not supported for UWP.


StephenBonikowsky:
I found two issues in corefx that might be what fixes this issue, @iamjasonp @mconnew could one of you confirm?

dotnet/corefx#2928
dotnet/corefx#4721

Using this issue as the ActiveIssue for the tests in...
\Scenarios\Security\TransportSecurity\Negotiate\NegotiateStream_Tcp_Tests.cs

roncain:
Tests that fail for me run on NET Native are:
* NegotiateStream_Tcp_AmbientCredentials
* NegotiateStream_Tcp_With_ExplicitSpn
* NegotiateStream_Tcp_With_SPN
* NetTcp_TransportSecurity_StreamedRequest_RoundTrips_String
* NetTcp_TransportSecurity_Streamed_RoundTrips_String_WithSingleThreadedSyncContext
* NetTcp_TransportSecurity_Streamed_RoundTrips_String
* NetTcp_TransportSecurity_Streamed_Async_RoundTrips_String
* NetTcp_TransportSecurity_StreamedResponse_Async_RoundTrips_String
* NetTcp_TransportSecurity_StreamedResponse_RoundTrips_String
* NetTcp_TransportSecurity_Streamed_TimeOut_Long_Running_Operation
* NetTcp_TransportSecurity_Streamed_Async_RoundTrips_String_WithSingleThreadedSyncContext

iamjasonp:
> NegotiateStream_Tcp_AmbientCredentials
NegotiateStream_Tcp_With_ExplicitSpn
NegotiateStream_Tcp_With_SPN

These I think will require a domain-joined server - is that the case?

zhenlan:
@DavidGoll will networking have NegotiateStream support in UWP/NS2.0 timeframe?

CIPop:
@zhenlan Yes. This is tracked by https://github.com/dotnet/corefx/issues/20033, in particular https://github.com/dotnet/corefx/issues/14560.

zhenlan:
Great, I see dotnet/corefx#14560 is in UWP6.0 milestone. Thanks @CIPop!

cc: @mconnew

mconnew:
Resovled by #2043

PR #1472, opened 2016-08-11
Depend on current version of System.NetSecurity

### LINKED ISSUE #1262: All TCP ExplicitUserNameAndPassword tests are failing

hongdai:
* We need the fix for issue https://github.com/dotnet/corefx/issues/9160
* I have verified manually we can now set Explict username/pwd for TCP.

fixes #1262

### DIFF
--- a/src/Common/test-runtime/project.json
+++ b/src/Common/test-runtime/project.json
@@ -17,7 +17,8 @@
     "Microsoft.DotNet.xunit.performance.runner.cli": "1.0.0-alpha-build0037",
     "Microsoft.DotNet.xunit.performance.runner.Windows": "1.0.0-alpha-build0037",
     "Microsoft.DotNet.BuildTools.TestSuite": "1.0.0-prerelease-00704-04",
-    "xunit.console.netcore": "1.0.3-prerelease-00704-04"
+    "xunit.console.netcore": "1.0.3-prerelease-00704-04",
+    "System.Net.Security": "4.0.1-beta-24322-03"
   },
   "frameworks": {
     "netstandard1.3": { },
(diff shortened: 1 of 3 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/project.json
```diff
@@ -15,10 +15,16 @@
...
+      "imports": [
+        "dotnet5.4"
+      ],
+    "dependencies": {
+        "System.Net.Security": "4.0.1-beta-24322-03",
+     },
+   },
   },
   "supports": {
-    "coreFx.Test.netcore50": {},
```
hongdai: @StephenBonikowsky @roncain I have to remove this line as this line does package validation for projectN and it fails for System.Net.Security as it does not support projectN. I'm going to validate this change both on ProjectN from projectK and ProjectN using Helix. If they both pass, I think the check here is unnecessary.

hongdai:
Hold on the review, it does not work on Linux platform, I need to figure out why.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/project.json
StephenBonikowsky: I remember seeing that before, makes sense.

iamjasonp:
You might need to add the reference to test-runtime/project.json as well :)

hongdai:
@iamjasonp Yeap, the only way to make it work without updating all test Jason file is to add it to test.runtime.

roncain:
LGTM.
Does it build and run in Net Native? I see #1235 will keep it from running the test but it would be good to confirm other security test run okay. It'd be enough to run against UWP locally (no need for TFS)

iamjasonp:
LGTM

hongdai:
I have verified both ProjectN from ProjectK run and .Net native, no regression found.

PR #1473, opened 2016-08-11
Redirect InstallRootCertificate.sh c_rehash output

iamjasonp:
In InstallRootCertificate.sh, redirect the c_rehash stdout to /dev/null, and
continue to allow the stderr to print to console

This will help reduce clutter in output when in our test setup steps,
especially since in our tests we may run this script more than once and c_rehash
output is quite verbose and prints every root cert out


hongdai:
LGTM. This change will make log much cleaner.

roncain:
LGTM

PR #1483, opened 2016-08-18
Disable NetTcp duplex callback test that fails in CI and Helix

roncain:
This test was passing 10 days ago when it was first added,
but over the past 2 days it has been failing in both CI and Helix.

Disabling it with IssueAttribute to return to having clean CI runs.


hongdai:
LGTM.

PR #1484, opened 2016-08-19
Cleanup selfhost

roncain:
Refactors the way service hosts are created and started by the self-hosted program.  This simplifies the code to open a endpoint and makes it less error-prone.  It also centralizes the open step so that we can Console.WriteLine which services are running at which address


roncain:
Reviewers: this PR was not strictly planned, but I got tired of making the same mistakes each time I have tried to add or rename an endpoint.  Our copy/paste/repair approach was unwieldy, so I decided we were overdue for this clean-up. This is largely a mechanical refactor, so I'd recommend not trying to go line-by-line and instead just review the approach taken here.

I will squash the commits before merging.

iamjasonp:
Much cleaner - thanks @roncain 
LGTM

zhenlan:
LGTM. I like less code 😄

PR #1492, opened 2016-08-26
Add script for scheduled self-hosted service restart

iamjasonp:
Test infrastructure requires that the self-hosted WCF Service be restarted periodically
Commit this script so that it's in source control rather than in somebody's file share
internally :) 

Note that this file is designed to live outside the repo - by convention we place this file into C:\wcftest

skip ci please

cc: @roncain


roncain:
LGTM

Issue #1250, opened 2016-06-03
Negative test times-out.


StephenBonikowsky:
ExpectedExceptionTests.SendTimeout_For_Long_Running_Operation_Throws_TimeoutException

This test times-out for around 6 minutes when running manually on NetNative, possibly other platforms as well.

roncain:
This fails for me in K-on-K too.  I believe I have seen it very rarely in GitHub from time to time.

zhenlan:
Will see if this still repos.

StephenBonikowsky:
Looks like it no longer occurs.

PR #1497, opened 2016-08-31
Activate test case, reason for being skipped no longer reproes.

StephenBonikowsky:
* This test was observed to hang back when it was still being run under the ToF infrastructure.
* Failure has not reproduced in current N run environment.
* Fixes Issue #1250


roncain:
LGTM

iamjasonp:
LGTM

PR #1502, opened 2016-09-02
Fix UpnEndpointIdentity on UWP

### LINKED ISSUE #1454: UpnEndpointIdentityTest.Ctor_UpnName throws PlatformNotSupported in UWP.
No call stack is currently available as no test case logging yet exists for UWP runs from git.

Test case...
System.ServiceModel.Security\tests\ServiceModel: UpnEndpointIdentityTest.Ctor_UpnName

mconnew:
Fixes #1454

### DIFF
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimComparer.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimComparer.cs
@@ -21,9 +21,7 @@ namespace System.IdentityModel.Claims
         private static IEqualityComparer<Claim> s_dnsComparer;
         private static IEqualityComparer<Claim> s_rsaComparer;
         private static IEqualityComparer<Claim> s_thumbprintComparer;
-#if SUPPORTS_WINDOWSIDENTITY
         private static IEqualityComparer<Claim> s_upnComparer;
-#endif // SUPPORTS_WINDOWSIDENTITY
         private static IEqualityComparer<Claim> s_x500DistinguishedNameComparer;
         private IEqualityComparer _resourceComparer;
 
@@ -113,7 +111,6 @@ namespace System.IdentityModel.Claims
             }
         }
 
-#if SUPPORTS_WINDOWSIDENTITY
         public static IEqualityComparer<Claim> Upn
         {
             get
@@ -125,7 +122,6 @@ namespace System.IdentityModel.Claims
                 return s_upnComparer;
             }
         }
-#endif // SUPPORTS_WINDOWSIDENTITY
 
         public static IEqualityComparer<Claim> X500DistinguishedName
         {
@@ -160,7 +156,7 @@ namespace System.IdentityModel.Claims
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("claim");
 
             return claim.ClaimType.GetHashCode() ^ claim.Right.GetHashCode()
-                ^ ((claim.Resource == null) ? 0 : _resourceComparer.GetHashCode(claim.Resource));
+                   ^ ((claim.Resource == null) ? 0 : _resourceComparer.GetHashCode(claim.Resource));
         }
 
         private class ObjectComparer : IEqualityComparer
@@ -240,6 +236,7 @@ namespace System.IdentityModel.Claims
         private class X500DistinguishedNameObjectComparer : IEqualityComparer
         {
             private IEqualityComparer _binaryComparer;
+
             public X500DistinguishedNameObjectComparer()
             {
                 _binaryComparer = new BinaryObjectComparer();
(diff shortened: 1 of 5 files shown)


src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimComparer.cs
```diff
@@ -258,12 +255,36 @@ int IEqualityComparer.GetHashCode(object obj)
...
+                SecurityIdentifier sid2;
+                if (!TryLookupSidFromName(upn2, out sid2))
+                    return false;
+
+                return sid1 == sid2;
+#else
+                // If WindowsIdentity isn't supported, then we can't
+                // retrieve the SecurityIdentifier's to compare so
+                // must return false
+                return false;
```
iamjasonp: I guess this is going to be a necessary evil in order for UWP to work... we however need to clearly show this in the release notes.

src/System.Private.ServiceModel/src/System/ServiceModel/UpnEndpointIdentity.cs
```diff
@@ -53,6 +57,110 @@ internal UpnEndpointIdentity(WindowsIdentity windowsIdentity)
...
+                {
+                    string upn = (string)this.IdentityClaim.Resource;
+                    if (!_hasUpnSidBeenComputed)
+                    {
+                        try
+                        {
+                            NTAccount userAccount = new NTAccount(upn);
+                            _upnSid = userAccount.Translate(typeof(SecurityIdentifier)) as SecurityIdentifier;
+                        }
+#pragma warning suppress 56500 // covered by FxCOP
```
iamjasonp: Can we omit this line?

src/System.Private.ServiceModel/src/System/ServiceModel/UpnEndpointIdentity.cs
```diff
@@ -53,6 +57,110 @@ internal UpnEndpointIdentity(WindowsIdentity windowsIdentity)
...
+        string GetUpnFromWindowsIdentity(WindowsIdentity windowsIdentity)
+        {
+            string downlevelName = null;
+            string upnName = null;
+ 
+            try
+            {
+                downlevelName = windowsIdentity.Name;
+ 
+                if (IsMachineJoinedToDomain())
```
iamjasonp: Won't this line cause us to _always_ throw if a call to GetUpnFromWindowsIdentity is called?
Is that what we expect? 

This will mean that our behaviour between desktop and .NET Core will be different, by the way

iamjasonp: I almost think that this entire method should just be PNSE, unless something else uses `IsMachineDomainJoined` or `GetUpnFromDownlevelName`. 

Your call on which is better though

iamjasonp:
LGTM

src/System.Private.ServiceModel/src/System/IdentityModel/Claims/ClaimComparer.cs
```diff
@@ -258,12 +255,36 @@ int IEqualityComparer.GetHashCode(object obj)
             }
         }
 
-#if SUPPORTS_WINDOWSIDENTITY
+//#if SUPPORTS_WINDOWSIDENTITY
```
roncain: Remove this line?

src/System.Private.ServiceModel/src/System/ServiceModel/UpnEndpointIdentity.cs
roncain: I see it will always throw *and catch* PNSE and then fallback to the downlevelName.  I know the number of calls to this could be low, but it's annoying to have unconditional exceptions being thrown.  Can that call to IsMachineJoinedToDomain itself be selectively commented out?  It does feel weird to differ from desktop unconditionally and silently, but maybe we have no other options.  Do we know which real-life scenarios will fail and which might still pass (falling back to downlevelName)?

---
In reply to: [77481060](https://github.com/dotnet/wcf/pull/1502#discussion_r77481060) [](ancestors = 77481060)

src/System.Private.ServiceModel/src/System/ServiceModel/Diagnostics/SecurityTraceRecordHelper.cs
```diff
@@ -40,5 +40,9 @@ internal static void TraceIdentityDeterminationSuccess(EndpointAddress epr, Endp
         internal static void TraceIdentityDeterminationFailure(EndpointAddress epr, Type identityVerifier)
         {
         }
+
+        internal static void TraceSpnToSidMappingFailure(string spn, Exception e)
+        {
+        }
```
roncain: Do we have any intent to implement these (currently) empty method bodies?  If so, we should open an issue.  If not, I wonder if we should just nuke these methods and calls.
P.S. I'm asking partly because I'm encountering them in the WS* stuff too and not sure which is the correct choice.

src/System.Private.ServiceModel/src/System/ServiceModel/UpnEndpointIdentity.cs
mconnew: Good point. I don't even notice those anymore. I think we should have a task to cleanup all of these littered through our code. I'll remove this instance though.

src/System.Private.ServiceModel/src/System/ServiceModel/UpnEndpointIdentity.cs
mconnew: I was trying to keep the code as identical as possible only throwing when we get to some code we can implement. Looking at the code, downlevelName is in the format domain\username. The fully qualified domain name is username@fqdn.tld. I'll give this some more thought. We do need to find a solution to this but none of the code in corefx uses the P/Invoke api call that we use on full framework so we can't expect to find an alternative way to do this using framework api's, at least not yet.

---
In reply to: [77633040](https://github.com/dotnet/wcf/pull/1502#discussion_r77633040) [](ancestors = 77633040,77481060)

roncain:
LGTM.
@mconnew -- I think when this is merged we can remove the Issue(10) we have on a few tests.  It was accidentally unconditional, when it really only applied to UWP.  Don't alter this PR -- just help me remember when you merge, and I'll make the change and verify it.

PR #1512, opened 2016-09-07
Update IIS Server setup scripts to unlock authentication methods

### LINKED ISSUE #1511: Unlock all subsection under Authentication in the SetupWcfIISHostedService.cmd
We need to ensure all Authentication types are unlocked from the root for authentication tests to pass.

hongdai:
* Authentication methods sections need to be unlocked from the root
for individual website to overrid them.

Fixes #1511

### DIFF
--- a/src/System.Private.ServiceModel/tools/scripts/SetupWcfIISHostedService.cmd
+++ b/src/System.Private.ServiceModel/tools/scripts/SetupWcfIISHostedService.cmd
@@ -187,9 +187,21 @@ if ERRORLEVEL 1 goto :Failure
 
 :SkipCertInstall
 
-:: Unlock the configuration of sslFlags
-echo Unlock the IIS config section to allow sslFlags to be overriden
+:: Unlock the configuration of sslFlags and Authentication
+echo Unlock the IIS config section to allow sslFlags and Authentication methods to be overriden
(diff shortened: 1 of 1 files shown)


roncain:
LGTM. Thanks Hong!

src/System.Private.ServiceModel/tools/scripts/SetupWcfIISHostedService.cmd
```diff
@@ -199,6 +202,11 @@ echo.
 echo All operations completed successfully!
 goto :Done
 
+:Unlock
+call :Run %_appcmd% unlock config /section:%1
+if ERRORLEVEL 1 goto :Failure
```
zhenlan: This may cause problem as technically we are still in `Unlock` function when we do this jump in case of error. I would recommend we stick with calling `:Run` function directly. We will have to repeat `if ERRORLEVEL ...` a few times, but we avoid potential weirdness/headache when things go wrong.

hongdai: Agreed. The redundant line is limited in this case, and it's cleaner because we jump less in the script.

zhenlan:
LGTM. Thanks @hongdai.

PR #1513, opened 2016-09-08
Update readme.md badges to show all CI builds

iamjasonp:
Our default CI builds build 
* CentOS
* OS X 
* Ubuntu14.04
* Windows 8.1

Update the badges to show the default builds we have. The other builds that aren't built by default have been left out as CI does not run green all the time and we don't use those yet as an indicator for product health


roncain:
LGTM

(Non-blocking) question: I notice the Windows 8.1 badge shows a test failure regarding the client certificate.  Is that a known issue?  Spurious failure?

iamjasonp:
> (Non-blocking) question: I notice the Windows 8.1 badge shows a test failure regarding the client certificate. Is that a known issue? Spurious failure?

I think it's spurious - the debug build shows no issue so I didn't look into this nay further

Issue #1257, opened 2016-06-06
SecurityBindingElement.CreateUserNameOverTransportBindingElement() not supported.


feinoujc:
possibly related to #9 but I wasn't sure so I'm opening this because I wasn't sure what the roadmap was. 

I get this error when trying to use the `SecurityBindingElement.CreateUserNameOverTransportBindingElement();` binding element as part of a CustomBinding

(running on OSX with .NET core RC2)

Here is my code

``` cs
    public class AcmeServiceChannelFactory : ChannelFactory<AcmeService>
    {
        public AcmeServiceChannelFactory()
            : base(CreateDefaultBinding(), new EndpointAddress("https://acme.com/service.asmx"))
        {
            Credentials.UserName.UserName = "username";
            Credentials.UserName.Password = "password";
        }

        internal static Binding CreateDefaultBinding()
        {
            var binding = new CustomBinding();
            var securityBinding = SecurityBindingElement.CreateUserNameOverTransportBindingElement();
            securityBinding.IncludeTimestamp = false;
            securityBinding.SecurityHeaderLayout = SecurityHeaderLayout.Lax;
            binding.Elements.Add(securityBinding);
            var textbinding = new TextMessageEncodingBindingElement
            {
                MessageVersion = MessageVersion.Soap11,
                ReaderQuotas = new XmlDictionaryReaderQuotas
                {
                    MaxArrayLength = int.MaxValue,
                    MaxDepth = 64,
                    MaxStringContentLength = int.MaxValue
                }
            };

            binding.Elements.Add(textbinding);

            var httpsBinding = new HttpsTransportBindingElement
            {
                MaxReceivedMessageSize = int.MaxValue,
                MaxBufferSize = int.MaxValue
            };

            binding.Elements.Add(httpsBinding);
            binding.OpenTimeout =
            binding.ReceiveTimeout =
            binding.SendTimeout = TimeSpan.FromMinutes(5);
            return binding;
        }
    }

```

error:

```
System.PlatformNotSupportedException: TransportSecurityBindingElement.BuildChannelFactoryCore is not supported.
   at System.ServiceModel.Channels.TransportSecurityBindingElement.BuildChannelFactoryCore[TChannel](BindingContext context)
   at System.ServiceModel.Channels.Binding.BuildChannelFactory[TChannel](BindingParameterCollection parameters)
[...]

Issue #1494, opened 2016-08-29
Enable IAsyncRequestChannel


roncain:
We introduced IAsyncRequestChannel internally to use Task-based API's rather than BeginRequest/EndRequest.  But no caller is yet aware to use this new API, and it is not plumbed through all the RequestChannel implementations.

To complete this task, modify ServiceChannel to support IAsyncRequestChannel, and modify all the IRequestChannel implementations to use it.

roncain:
I recommend this work be done after #1257 is resolved because its implementation introduces several request channels that will require modification.

zhenlan:
This will simplify the internal programming model and reduce the code size. The perf impact is to be determined. This is estimated ~1 week work.

roncain:
Currently in progress in the ws-trust feature branch.  Using that branch because its largest impact is across the security layers.

roncain:
The ws-trust feature branch now has working async request channel tests.  But based on what I've learned, we need to discuss how to proceed.  There is a large body of Begin/End async usage throughout many of the IdentityModel and ServiceModel types related to ws-trust.  Rewriting them to be task-based would take probably 3-4 weeks, minimum, and would then be very different from the full framework.  The alternative is porting the Begin/End AsyncResult code from the full framework.  So far, I have just ported APM for the simplest ws-trust test case and wrapped Tasks around it.

StephenBonikowsky:
@mconnew Could you please review this issue, I think you may have done this already. If possible lets close this.

Issue #31, opened 2015-05-22
Support WS* Bindings to allow interop with federated WCF services


sujitdmello:
Customer has a large deployment of WCF services that use *WsHttpBinding* & *WsFederationHttpBinding*. They are currently developing.NET Core based front-end Web APIs (using ASP.NET 5) that would have to consume these services. Without support for these WS-* bindings, there is no way to call these services securely using federated security. Enabling WS-* on the WCF client would enable these interop scenarios for large enterprises.

roncain:
Thanks for the suggestion @sujitdmello .  Let's use this issue to discuss it and collect other input from the community.  Inviting @iamjasonp and @mconnew for their thoughts.

popcatalin81:
My input:
WS-* is absolutely necessary for any SOAP based stack for interoperability with other WS technologies or in case of existence of an IdP (Identity Provider).

Almost any decent sized WCF based application will make use of WS-* (maybe WS-Trust being the most common). If Core WCF is reduced to a SOAP alternative to REST services with minor additions then there's no point in Core WCF at all. If Core WCF cannot be used interoperably with other WS-* stacks (IE: Full .Net WCF), then there's no point in it either.

From a security point of view Username, Certificate and Windows authentication is simply insufficient especially when considering cloud and hybrid platforms, WS-* extensions are needed for this scenario.

My 2c.

iamjasonp:
Good feedback, @popcatalin81 and @sujitdmello - thanks! 

Our first cut of the APIs in the WCF client for .NET Core was based on what was available on the API surface for Windows Store 8 apps... (maybe this list will look familiar :smiley: [WCF in Metro apps](http://blogs.msdn.com/b/piyushjo/archive/2011/10/19/wcf-for-metro-apps-supported-functionality.aspx)). 

I'm of the opinion too that WS-* protocols should be supported. 

WCF on the desktop supported [these protocols](https://msdn.microsoft.com/en-us/library/ms730294(v=vs.110).aspx)... I've come up with a (likely incomplete) list of things that isn't supported yet in no particular order:

* Messaging - WS-Addressing 2004/08
* Security - WSS SOAP Message Security 1.0
* Security - WSS SOAP Message Security UsernameToken Profile 1.1
* Security - WSS SOAP Message Security X.509 Certificate Token Profile 1.1
* Security - WSS SOAP Message Security Kerberos Token Profile 1.1
* Security - WS-SecureConversation
* **Security - WS-Trust** 
[...]

PR #1515, opened 2016-09-08
Enable SecurityBindingElement.CreateUserNameOverTransportBindingElement

roncain:
Add initial support for the simple-case use of SecurityBindingElement.CreateUserNameOverTransportBindingElement.  This entails bringing in a significant number of files from the full framework and new tests to ensure the simple default case works for synchronous operations.  It is the first step to handle issue #1257, but not enough to consider that issue fixed.

Other PR's will enable the async paths and expand testing of these new layers.

**UPDATE: this PR has been replaced by PR #1530 and will be discarded once that one is merged.**

### DIFF
--- a/src/System.Private.ServiceModel/src/Resources/Strings.resx
+++ b/src/System.Private.ServiceModel/src/Resources/Strings.resx
@@ -6781,6 +6781,6 @@
     <value>There was an error serializing the security key identifier clause. Please see the inner exception for more details.</value>
   </data>
   <data name="CannotReadKeyIdentifierClause" xml:space="preserve">
-    <value>Cannot read KeyIdentifierClause from element '{0}' with namespace '{1}'.  Custom KeyIdentifierClauses require custom SecurityTokenSerializers, please refer to the SDK for examples.</value>
+    <value>Cannot read KeyIdentifierClause from element '{0}' with namespace '{1}'.  Custom KeyIdentifierClauses require custom SecurityTokenSerializers.</value>
   </data>
 </root>
--- a/src/System.Private.ServiceModel/src/Resources/Strings.resx
+++ b/src/System.Private.ServiceModel/src/Resources/Strings.resx
@@ -6771,4 +6771,16 @@
   <data name="Xml_InvalidNodeType" xml:space="preserve">
     <value>'{0}' is an invalid XmlNodeType.</value>
   </data>
+  <data name="ErrorDeserializingKeyIdentifierClause" xml:space="preserve">
+    <value>'There was an error deserializing the security key identifier clause XML. Please see the inner exception for more details.</value>
+  </data>
+  <data name="ErrorSerializingKeyIdentifier" xml:space="preserve">
+    <value>There was an error serializing the security key identifier. Please see the inner exception for more details.</value>
+  </data>
+  <data name="ErrorSerializingKeyIdentifierClause" xml:space="preserve">
+    <value>There was an error serializing the security key identifier clause. Please see the inner exception for more details.</value>
+  </data>
+  <data name="CannotReadKeyIdentifierClause" xml:space="preserve">
+    <value>Cannot read KeyIdentifierClause from element '{0}' with namespace '{1}'.  Custom KeyIdentifierClauses require custom SecurityTokenSerializers, please refer to the SDK for examples.</value>
+  </data>
 </root>
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/SecurityTokenProvider.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/SecurityTokenProvider.cs
@@ -5,6 +5,7 @@
 
 using System.IdentityModel.Tokens;
 using System.Runtime;
+using System.ServiceModel;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -22,6 +23,16 @@ namespace System.IdentityModel.Selectors
         public virtual bool SupportsTokenCancellation
         {
             get { return false; }
+        }
+
+        public SecurityToken GetToken(TimeSpan timeout)
+        {
+            SecurityToken token = this.GetTokenCore(timeout);
+            if (token == null)
+            {
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new SecurityTokenException(SR.Format(SR.TokenProviderUnableToGetToken, this)));
+            }
+            return token;
         }
 
         public async Task<SecurityToken> GetTokenAsync(CancellationToken cancellationToken)
@@ -58,6 +69,8 @@ namespace System.IdentityModel.Selectors
         }
 
         // protected methods
+        protected abstract SecurityToken GetTokenCore(TimeSpan timeout);
+
         protected abstract Task<SecurityToken> GetTokenCoreAsync(CancellationToken cancellationToken);
 
         protected virtual Task<SecurityToken> RenewTokenCoreAsync(CancellationToken cancellationToken, SecurityToken tokenToBeRenewed)
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/SecurityTokenSerializer.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/SecurityTokenSerializer.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.IdentityModel.Tokens;
 using System.ServiceModel;
 using System.Xml;
+using TokenEntry = System.ServiceModel.Security.WSSecurityTokenSerializer.TokenEntry;
 
 namespace System.IdentityModel.Selectors
 {
@@ -206,42 +207,6 @@ namespace System.IdentityModel.Selectors
             public abstract bool SupportsCore(SecurityKeyIdentifier keyIdentifier);
 
             public abstract void WriteKeyIdentifierCore(XmlDictionaryWriter writer, SecurityKeyIdentifier keyIdentifier);
-        }
-
-        internal abstract class TokenEntry
-        {
-            private Type[] _tokenTypes = null;
-
-            protected abstract XmlDictionaryString LocalName { get; }
-            protected abstract XmlDictionaryString NamespaceUri { get; }
-            public Type TokenType { get { return GetTokenTypes()[0]; } }
-            public abstract string TokenTypeUri { get; }
-            protected abstract string ValueTypeUri { get; }
-
-            public bool SupportsCore(Type tokenType)
-            {
-                Type[] tokenTypes = GetTokenTypes();
-                for (int i = 0; i < tokenTypes.Length; ++i)
-                {
-                    if (tokenTypes[i].IsAssignableFrom(tokenType))
-                        return true;
-                }
-                return false;
-            }
-
-            protected abstract Type[] GetTokenTypesCore();
-
-            public Type[] GetTokenTypes()
-            {
-                if (_tokenTypes == null)
-                    _tokenTypes = GetTokenTypesCore();
-                return _tokenTypes;
-            }
-
-            public virtual bool SupportsTokenTypeUri(string tokenTypeUri)
-            {
-                return (this.TokenTypeUri == tokenTypeUri);
-            }
         }
     }
 }
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
@@ -1,4 +1,5 @@
 using System.IdentityModel.Tokens;
+using System.ServiceModel;
 using System.Xml;
 
 namespace System.IdentityModel.Tokens
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
@@ -35,9 +35,6 @@ namespace System.IdentityModel.Tokens
         public override bool Matches(SecurityKeyIdentifierClause keyIdentifierClause)
         {
             GenericXmlSecurityKeyIdentifierClause that = keyIdentifierClause as GenericXmlSecurityKeyIdentifierClause;
-
-            // PreSharp Bug: Parameter 'that' to this public method must be validated: A null-dereference can occur here.
-#pragma warning suppress 56506
             return ReferenceEquals(this, that) || (that != null && that.Matches(this.ReferenceXml));
         }
 
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
@@ -1,0 +1,78 @@
+using System.IdentityModel.Tokens;
+using System.Xml;
+
+namespace System.IdentityModel.Tokens
+{
+    public class GenericXmlSecurityKeyIdentifierClause : SecurityKeyIdentifierClause
+    {
+        XmlElement referenceXml;
+
+        public GenericXmlSecurityKeyIdentifierClause(XmlElement referenceXml)
+            : this(referenceXml, null, 0)
+        {
+        }
+
+        public GenericXmlSecurityKeyIdentifierClause(XmlElement referenceXml, byte[] derivationNonce, int derivationLength)
+            : base(null, derivationNonce, derivationLength)
+        {
+            if (referenceXml == null)
+            {
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("referenceXml");
+            }
+            this.referenceXml = referenceXml;
+        }
+
+        public XmlElement ReferenceXml
+        {
+            get { return this.referenceXml; }
+        }
+
+        public override bool Matches(SecurityKeyIdentifierClause keyIdentifierClause)
+        {
+            GenericXmlSecurityKeyIdentifierClause that = keyIdentifierClause as GenericXmlSecurityKeyIdentifierClause;
+
+            // PreSharp Bug: Parameter 'that' to this public method must be validated: A null-dereference can occur here.
+#pragma warning suppress 56506
+            return ReferenceEquals(this, that) || (that != null && that.Matches(this.ReferenceXml));
+        }
+
+        private bool Matches(XmlElement xmlElement)
+        {
+            if (xmlElement == null)
+                return false;
+
+            return CompareNodes(this.referenceXml, xmlElement);
+        }
+
+        private bool CompareNodes(XmlNode originalNode, XmlNode newNode)
+        {
+            if (originalNode.OuterXml == newNode.OuterXml)
+                return true;
+
+            if (originalNode.LocalName != newNode.LocalName || originalNode.InnerText != newNode.InnerText)
+                return false;
+
+            if (originalNode.InnerXml == newNode.InnerXml)
+                return true;
+
+            if (originalNode.HasChildNodes)
+            {
+                if (!newNode.HasChildNodes || originalNode.ChildNodes.Count != newNode.ChildNodes.Count)
+                    return false;
+
+                bool childrenStatus = true;
+                for (int i = 0; i < originalNode.ChildNodes.Count; i++)
+                {
+                    childrenStatus = childrenStatus & CompareNodes(originalNode.ChildNodes[i], newNode.ChildNodes[i]);
+                }
+
+                return childrenStatus;
+            }
+            else if (newNode.HasChildNodes)
+                return false;
+
+            return true;
+        }
+    }
+}
+
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
@@ -1,3 +1,7 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
 using System.IdentityModel.Tokens;
 using System.ServiceModel;
 using System.Xml;
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.IdentityModel.Tokens;
 using System.ServiceModel;
 using System.Xml;
 
--- a/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
+++ b/src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
@@ -1,31 +1,29 @@
-//------------------------------------------------------------------------------
-//     Copyright (c) Microsoft Corporation.  All rights reserved.
-//------------------------------------------------------------------------------
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.IdentityModel.Selectors;
+using System.Runtime;
+using System.ServiceModel.Security;
+using System.Xml;
+using TokenEntry = System.ServiceModel.Security.WSSecurityTokenSerializer.TokenEntry;
 
 namespace System.IdentityModel.Tokens
 {
-    using System.Collections.Generic;
-    using System.IdentityModel;
-    using System.IdentityModel.Security;
-    using System.IdentityModel.Selectors;
-    using System.Runtime;
-    using System.Xml;
-using System.Collections;
-
     /// <summary>
     /// Abstract class for SecurityKeyIdentifierClause Serializer.
     /// </summary>
     internal class KeyInfoSerializer : SecurityTokenSerializer
     {
-        readonly List<SecurityTokenSerializer.KeyIdentifierEntry> keyIdentifierEntries;
-        readonly List<SecurityTokenSerializer.KeyIdentifierClauseEntry> keyIdentifierClauseEntries;
-        readonly List<SecurityTokenSerializer.SerializerEntries> serializerEntries;
-        readonly List<TokenEntry> tokenEntries;
+        private readonly List<SecurityTokenSerializer.KeyIdentifierEntry> _keyIdentifierEntries;
+        private readonly List<SecurityTokenSerializer.KeyIdentifierClauseEntry> _keyIdentifierClauseEntries;
+        private readonly List<SecurityTokenSerializer.SerializerEntries> _serializerEntries;
+        private readonly List<TokenEntry> _tokenEntries;
 
-
-        DictionaryManager dictionaryManager;
-        bool emitBspRequiredAttributes;
-        SecurityTokenSerializer innerSecurityTokenSerializer;
+        private DictionaryManager _dictionaryManager;
+        private bool _emitBspRequiredAttributes;
+        private SecurityTokenSerializer _innerSecurityTokenSerializer;
 
         /// <summary>
         /// Creates an instance of <see cref="SecurityKeyIdentifierClauseSerializer"/>
@@ -51,54 +49,56 @@ using System.Collections;
             SecurityTokenSerializer innerSecurityTokenSerializer,
             Func<KeyInfoSerializer, IEnumerable<SerializerEntries>> additionalEntries)
         {
-            this.dictionaryManager = dictionaryManager;
-            this.emitBspRequiredAttributes = emitBspRequiredAttributes;
-            this.innerSecurityTokenSerializer = innerSecurityTokenSerializer;
+            _dictionaryManager = dictionaryManager;
+            _emitBspRequiredAttributes = emitBspRequiredAttributes;
+            _innerSecurityTokenSerializer = innerSecurityTokenSerializer;
 
-            this.serializerEntries = new List<SecurityTokenSerializer.SerializerEntries>();
+            _serializerEntries = new List<SecurityTokenSerializer.SerializerEntries>();
 
-            this.serializerEntries.Add(new XmlDsigSep2000(this));
-            this.serializerEntries.Add(new XmlEncApr2001(this));
-            this.serializerEntries.Add(new System.IdentityModel.Security.WSTrust(this, trustDictionary));
-            if ( additionalEntries != null )
+            // Issue #31 in progress
+            //serializerEntries.Add(new XmlDsigSep2000(this));
+            //serializerEntries.Add(new XmlEncApr2001(this));
+            //serializerEntries.Add(new WSTrust(this, trustDictionary));
+
+            if (additionalEntries != null)
             {
-                foreach ( SerializerEntries entries in additionalEntries( this ) )
+                foreach (SerializerEntries entries in additionalEntries(this))
                 {
-                    this.serializerEntries.Add(entries);
+                    _serializerEntries.Add(entries);
                 }
             }
 
             bool wsSecuritySerializerFound = false;
-            foreach ( SerializerEntries entry in this.serializerEntries )
+            foreach (SerializerEntries entry in _serializerEntries)
             {
-                if ( ( entry is WSSecurityXXX2005 ) || ( entry is WSSecurityJan2004 ) )
+                if ((entry is WSSecurityXXX2005) || (entry is WSSecurityJan2004))
                 {
                     wsSecuritySerializerFound = true;
                     break;
                 }
             }
 
-            if ( !wsSecuritySerializerFound )
+            if (!wsSecuritySerializerFound)
             {
-                this.serializerEntries.Add( new WSSecurityXXX2005( this ) );
+                _serializerEntries.Add(new WSSecurityXXX2005((WSSecurityTokenSerializer) innerSecurityTokenSerializer));
             }
 
-            this.tokenEntries = new List<TokenEntry>();
-            this.keyIdentifierEntries = new List<SecurityTokenSerializer.KeyIdentifierEntry>();
-            this.keyIdentifierClauseEntries = new List<SecurityTokenSerializer.KeyIdentifierClauseEntry>();
+            _tokenEntries = new List<TokenEntry>();
+            _keyIdentifierEntries = new List<SecurityTokenSerializer.KeyIdentifierEntry>();
+            _keyIdentifierClauseEntries = new List<SecurityTokenSerializer.KeyIdentifierClauseEntry>();
 
-            for (int i = 0; i < this.serializerEntries.Count; ++i)
+            for (int i = 0; i < _serializerEntries.Count; ++i)
             {
-                SecurityTokenSerializer.SerializerEntries serializerEntry = this.serializerEntries[i];
-                serializerEntry.PopulateTokenEntries(this.tokenEntries);
-                serializerEntry.PopulateKeyIdentifierEntries(this.keyIdentifierEntries);
-                serializerEntry.PopulateKeyIdentifierClauseEntries(this.keyIdentifierClauseEntries);
+                SecurityTokenSerializer.SerializerEntries serializerEntry = _serializerEntries[i];
+                serializerEntry.PopulateTokenEntries(_tokenEntries);
+                serializerEntry.PopulateKeyIdentifierEntries(_keyIdentifierEntries);
+                serializerEntry.PopulateKeyIdentifierClauseEntries(_keyIdentifierClauseEntries);
             }
         }
 
         public DictionaryManager DictionaryManager
         {
-            get { return this.dictionaryManager; }
+            get { return _dictionaryManager; }
         }
 
         /// <summary>
@@ -108,7 +108,7 @@ using System.Collections;
         {
             get
             {
-                return this.emitBspRequiredAttributes;
+                return _emitBspRequiredAttributes;
             }
         }
 
@@ -116,11 +116,11 @@ using System.Collections;
         {
             get
             {
-                return this.innerSecurityTokenSerializer == null ? this : this.innerSecurityTokenSerializer;
+                return _innerSecurityTokenSerializer == null ? this : _innerSecurityTokenSerializer;
             }
            set
             {
-                this.innerSecurityTokenSerializer = value;
+                _innerSecurityTokenSerializer = value;
             }
         }
 
@@ -132,7 +132,7 @@ using System.Collections;
         protected override SecurityToken ReadTokenCore(XmlReader reader, SecurityTokenResolver tokenResolver)
         {
             XmlDictionaryReader localReader = XmlDictionaryReader.CreateDictionaryReader( reader ); 
-            throw DiagnosticUtility.ExceptionUtility.ThrowHelperError( new XmlException( SR.GetString( SR.CannotReadToken, reader.LocalName, reader.NamespaceURI, localReader.GetAttribute( XD.SecurityJan2004Dictionary.ValueType, null ) ) ) );
+            throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperError( new XmlException( SR.Format( SR.CannotReadToken, reader.LocalName, reader.NamespaceURI, localReader.GetAttribute( XD.SecurityJan2004Dictionary.ValueType, null ) ) ) );
         }
 
         protected override bool CanWriteTokenCore(SecurityToken token)
@@ -142,15 +142,15 @@ using System.Collections;
 
         protected override void WriteTokenCore(XmlWriter writer, SecurityToken token)
         {
-            throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.GetString(SR.StandardsManagerCannotWriteObject, token.GetType())));
+            throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.StandardsManagerCannotWriteObject, token.GetType())));
         }
 
         protected override bool CanReadKeyIdentifierCore(XmlReader reader)
         {
             XmlDictionaryReader localReader = XmlDictionaryReader.CreateDictionaryReader(reader);
-            for (int i = 0; i < this.keyIdentifierEntries.Count; i++)
+            for (int i = 0; i < _keyIdentifierEntries.Count; i++)
             {
-                KeyIdentifierEntry keyIdentifierEntry = this.keyIdentifierEntries[i];
+                KeyIdentifierEntry keyIdentifierEntry = _keyIdentifierEntries[i];
                 if (keyIdentifierEntry.CanReadKeyIdentifierCore(localReader))
                     return true;
             }
@@ -164,7 +164,7 @@ using System.Collections;
             SecurityKeyIdentifier keyIdentifier = new SecurityKeyIdentifier();
             while (localReader.IsStartElement())
             {
-                SecurityKeyIdentifierClause clause = this.InnerSecurityTokenSerializer.ReadKeyIdentifierClause(localReader);
+                SecurityKeyIdentifierClause clause = InnerSecurityTokenSerializer.ReadKeyIdentifierClause(localReader);
                 if (clause == null)
                 {
                     localReader.Skip();
@@ -176,7 +176,7 @@ using System.Collections;
             }
             if (keyIdentifier.Count == 0)
             {
-                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.GetString(SR.ErrorDeserializingKeyIdentifierClause)));
+                throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.Format(SR.ErrorDeserializingKeyIdentifierClause)));
             }
             localReader.ReadEndElement();
 
@@ -185,9 +185,9 @@ using System.Collections;
 
         protected override bool CanWriteKeyIdentifierCore(SecurityKeyIdentifier keyIdentifier)
         {
-            for (int i = 0; i < this.keyIdentifierEntries.Count; ++i)
+            for (int i = 0; i < _keyIdentifierEntries.Count; ++i)
             {
-                KeyIdentifierEntry keyIdentifierEntry = this.keyIdentifierEntries[i];
+                KeyIdentifierEntry keyIdentifierEntry = _keyIdentifierEntries[i];
                 if (keyIdentifierEntry.SupportsCore(keyIdentifier))
                     return true;
             }
@@ -198,9 +198,9 @@ using System.Collections;
         {
             bool wroteKeyIdentifier = false;
             XmlDictionaryWriter localWriter = XmlDictionaryWriter.CreateDictionaryWriter(writer);
-            for (int i = 0; i < this.keyIdentifierEntries.Count; ++i)
+            for (int i = 0; i < _keyIdentifierEntries.Count; ++i)
             {
-                KeyIdentifierEntry keyIdentifierEntry = this.keyIdentifierEntries[i];
+                KeyIdentifierEntry keyIdentifierEntry = _keyIdentifierEntries[i];
                 if (keyIdentifierEntry.SupportsCore(keyIdentifier))
                 {
                     try
@@ -218,7 +218,7 @@ using System.Collections;
                             throw;
                         }
 
-                        throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.GetString(SR.ErrorSerializingKeyIdentifier), e));
+                        throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.Format(SR.ErrorSerializingKeyIdentifier), e));
                     }
                     wroteKeyIdentifier = true;
                     break;
@@ -226,7 +226,7 @@ using System.Collections;
             }
 
             if (!wroteKeyIdentifier)
-                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.GetString(SR.StandardsManagerCannotWriteObject, keyIdentifier.GetType())));
+                throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.StandardsManagerCannotWriteObject, keyIdentifier.GetType())));
 
             localWriter.Flush();
         }
@@ -234,9 +234,9 @@ using System.Collections;
         protected override bool CanReadKeyIdentifierClauseCore(XmlReader reader)
         {
             XmlDictionaryReader localReader = XmlDictionaryReader.CreateDictionaryReader(reader);
-            for (int i = 0; i < this.keyIdentifierClauseEntries.Count; i++)
+            for (int i = 0; i < _keyIdentifierClauseEntries.Count; i++)
             {
-                KeyIdentifierClauseEntry keyIdentifierClauseEntry = this.keyIdentifierClauseEntries[i];
+                KeyIdentifierClauseEntry keyIdentifierClauseEntry = _keyIdentifierClauseEntries[i];
                 if (keyIdentifierClauseEntry.CanReadKeyIdentifierClauseCore(localReader))
                     return true;
             }
@@ -246,9 +246,9 @@ using System.Collections;
         protected override SecurityKeyIdentifierClause ReadKeyIdentifierClauseCore(XmlReader reader)
         {
             XmlDictionaryReader localReader = XmlDictionaryReader.CreateDictionaryReader(reader);
-            for (int i = 0; i < this.keyIdentifierClauseEntries.Count; i++)
+            for (int i = 0; i < _keyIdentifierClauseEntries.Count; i++)
             {
-                KeyIdentifierClauseEntry keyIdentifierClauseEntry = this.keyIdentifierClauseEntries[i];
+                KeyIdentifierClauseEntry keyIdentifierClauseEntry = _keyIdentifierClauseEntries[i];
                 if (keyIdentifierClauseEntry.CanReadKeyIdentifierClauseCore(localReader))
                 {
                     try
@@ -265,18 +265,18 @@ using System.Collections;
                         {
                             throw;
                         }
-                        throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.GetString(SR.ErrorDeserializingKeyIdentifierClause), e));
+                        throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.Format(SR.ErrorDeserializingKeyIdentifierClause), e));
                     }
                 }
             }
-            throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.GetString(SR.CannotReadKeyIdentifierClause, reader.LocalName, reader.NamespaceURI)));
+            throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.Format(SR.CannotReadKeyIdentifierClause, reader.LocalName, reader.NamespaceURI)));
         }
 
         protected override bool CanWriteKeyIdentifierClauseCore(SecurityKeyIdentifierClause keyIdentifierClause)
         {
-            for (int i = 0; i < this.keyIdentifierClauseEntries.Count; ++i)
+            for (int i = 0; i < _keyIdentifierClauseEntries.Count; ++i)
             {
-                KeyIdentifierClauseEntry keyIdentifierClauseEntry = this.keyIdentifierClauseEntries[i];
+                KeyIdentifierClauseEntry keyIdentifierClauseEntry = _keyIdentifierClauseEntries[i];
                 if (keyIdentifierClauseEntry.SupportsCore(keyIdentifierClause))
                     return true;
             }
@@ -287,9 +287,9 @@ using System.Collections;
         {
             bool wroteKeyIdentifierClause = false;
             XmlDictionaryWriter localWriter = XmlDictionaryWriter.CreateDictionaryWriter(writer);
-            for (int i = 0; i < this.keyIdentifierClauseEntries.Count; ++i)
+            for (int i = 0; i < _keyIdentifierClauseEntries.Count; ++i)
             {
-                KeyIdentifierClauseEntry keyIdentifierClauseEntry = this.keyIdentifierClauseEntries[i];
+                KeyIdentifierClauseEntry keyIdentifierClauseEntry = _keyIdentifierClauseEntries[i];
                 if (keyIdentifierClauseEntry.SupportsCore(keyIdentifierClause))
                 {
                     try
@@ -307,7 +307,7 @@ using System.Collections;
                             throw;
                         }
                         
-                        throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.GetString(SR.ErrorSerializingKeyIdentifierClause), e));
+                        throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new XmlException(SR.Format(SR.ErrorSerializingKeyIdentifierClause), e));
                     }
                     wroteKeyIdentifierClause = true;
                     break;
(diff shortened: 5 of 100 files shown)


roncain:
Reviewers: this PR is *not* intended to be merged into master, but **I would appreciate review comments now**.  If there are broad comments that apply to multiple files, just comment once and assume I will deal with them across all files.

A few notes about the commits:
 - The first commit is a raw port of the minimum set of file I needed to get the default case to work.  It is not worth your time to review it, because they are verbatim copies from the full framework and not intended to be compiled.  This commit was done only to make diff'ing easier.

- The 2nd commit contains all the edits to both the new and existing files.  This is the one I would like reviewed.  I have deliberately departed from prior convention and left full framework code in place, but commented out.  These areas are marked with issue #31 which is being used to track all the WSTrust work, and I intend to come back and enable them once the API's they need are available. PNSE is thrown from every code path departing from the simplest scenario I tested, and those will be removed as testing broadens.

- The 3rd commit contains the new tests.  The async test is disabled pending resolution of #1494.

Sorry for the large PR.  This WS* area is sufficiently entangled that I could not fragment it.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/SecurityTokenSerializer.cs
```diff
@@ -207,41 +208,5 @@ public virtual bool CanReadKeyIdentifierCore(XmlDictionaryReader reader)
 
             public abstract void WriteKeyIdentifierCore(XmlDictionaryWriter writer, SecurityKeyIdentifier keyIdentifier);
         }
-
-        internal abstract class TokenEntry
```
roncain: > internal abstract class TokenEntry [](start = 7, length = 35)

Reviewers: the duplicate-and-diverge history between IdentityModel and WCF has really made this merge awkward.  I am attempting to eliminate the internal IdentityModel types that WCF has duplicates for and to converge on the WCF versions.  This deletion is an example.  If you have any caveats to share, I'd like to hear them.  I'd actually prefer to refactor more broadly, but to make this PR reviewable, I'm just trying to eliminate duplication and stick with existing factoring.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/LayeredChannel.cs
```diff
@@ -57,12 +58,12 @@ protected override void OnClose(TimeSpan timeout)
 
         protected override IAsyncResult OnBeginClose(TimeSpan timeout, AsyncCallback callback, object state)
         {
-            return _innerChannel.BeginClose(timeout, callback, state);
+            throw ExceptionHelper.PlatformNotSupported();
```
roncain: >throw ExceptionHelper.PlatformNotSupported(); [](start = 12, length = 45)

Reviewers: PNSE's like this are a deliberate attempt to make all our CommunicationObjects enforce the fact we have moved to the OnOpenAsync and OnCloseAsync model across the board.  It is not strictly related to WS* support other than continuing our existing practice in newly added/modified types. It should not be possible for an app to cause these methods to be called, regardless if they use sync or async open.   But they are useful to uncover coding mistakes of our own and to discover dead AsyncResult code.

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/SecurityTokenProvider.cs
```diff
@@ -58,6 +69,8 @@ public virtual bool SupportsTokenCancellation
         }
 
         // protected methods
+        protected abstract SecurityToken GetTokenCore(TimeSpan timeout);
```
iamjasonp: I'm wondering if we should just omit this and utilize GetTokenCoreAsync exclusively through our .NET Core code. Maintaining two codepaths for effectively the same thing isn't ideal 

Furthermore, the things we're porting over right now seem not to use the timeout parameter, though I did see it in S.SM.IssuanceTokenProvider and S.SM.SecuritySessionSecurityTokenProvider for .NET Desktop (but then this seems like something we might want to rewrite using task async anyway)

src/System.Private.ServiceModel/src/Resources/Strings.resx
```diff
@@ -6771,4 +6771,16 @@
...
+    <value>'There was an error deserializing the security key identifier clause XML. Please see the inner exception for more details.</value>
+  </data>
+  <data name="ErrorSerializingKeyIdentifier" xml:space="preserve">
+    <value>There was an error serializing the security key identifier. Please see the inner exception for more details.</value>
+  </data>
+  <data name="ErrorSerializingKeyIdentifierClause" xml:space="preserve">
+    <value>There was an error serializing the security key identifier clause. Please see the inner exception for more details.</value>
+  </data>
+  <data name="CannotReadKeyIdentifierClause" xml:space="preserve">
+    <value>Cannot read KeyIdentifierClause from element '{0}' with namespace '{1}'.  Custom KeyIdentifierClauses require custom SecurityTokenSerializers, please refer to the SDK for examples.</value>
```
iamjasonp: >, please refer to the SDK for examples [](start = 152, length = 38)

Is the SDK comment still relevant for us?

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
```diff
@@ -0,0 +1,79 @@
+using System.IdentityModel.Tokens;
+using System.ServiceModel;
```
iamjasonp: Question only: Is adding `using System.ServiceModel` part of the consolidation of code between System.ServiceModel and System.IdentityModel?

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        protected override bool CanWriteTokenCore(SecurityToken token)
+        {
+            return false;
+        }
+
+        protected override void WriteTokenCore(XmlWriter writer, SecurityToken token)
+        {
+            throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.Format(SR.StandardsManagerCannotWriteObject, token.GetType())));
```
iamjasonp: >ServiceModel [](start = 18, length = 12)

NIT: Consider aliasing `using DiagnosticUtility = System.ServiceModel.DiagnosticUtility` to avoid this weirdness of having to prefix DiagnosticUtility... not a usage pattern we use typically isall

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
```diff
@@ -0,0 +1,372 @@
...
+            {
+                KeyIdentifierEntry keyIdentifierEntry = _keyIdentifierEntries[i];
+                if (keyIdentifierEntry.SupportsCore(keyIdentifier))
+                {
+                    try
+                    {
+                        keyIdentifierEntry.WriteKeyIdentifierCore(localWriter, keyIdentifier);
+                    }
+#pragma warning suppress 56500 // covered by FxCOP
+                    catch (Exception e)
```
iamjasonp: Delete

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
```diff
@@ -0,0 +1,372 @@
...
+            for (int i = 0; i < _keyIdentifierClauseEntries.Count; ++i)
+            {
+                KeyIdentifierClauseEntry keyIdentifierClauseEntry = _keyIdentifierClauseEntries[i];
+                if (keyIdentifierClauseEntry.SupportsCore(keyIdentifierClause))
+                {
+                    try
+                    {
+                        keyIdentifierClauseEntry.WriteKeyIdentifierClauseCore(localWriter, keyIdentifierClause);
+                    }
+#pragma warning suppress 56500 // covered by FxCOP
```
iamjasonp: Delete

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
```diff
@@ -0,0 +1,372 @@
...
+            for (int i = 0; i < _keyIdentifierClauseEntries.Count; i++)
+            {
+                KeyIdentifierClauseEntry keyIdentifierClauseEntry = _keyIdentifierClauseEntries[i];
+                if (keyIdentifierClauseEntry.CanReadKeyIdentifierClauseCore(localReader))
+                {
+                    try
+                    {
+                        return keyIdentifierClauseEntry.ReadKeyIdentifierClauseCore(localReader);
+                    }
+#pragma warning suppress 56500 // covered by FxCOP
```
iamjasonp: Recommend going through the files and removing all the pragma/FxCop stuff as it's not used in the .NET core codebase

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SamlSecurityToken.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        public SamlSecurityToken(SamlAssertion assertion)
+        {
+            Initialize(assertion);
+        }
+
+        protected void Initialize(SamlAssertion assertion)
+        {
+            if (assertion == null)
+                throw ServiceModel.DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("assertion");
```
iamjasonp: >ServiceModel [](start = 22, length = 12)

Consider `using DiagnosticUtility = System.ServiceModel.DiagnosticUtility ... `. Will not further comment on these

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SamlSerializer.cs
```diff
@@ -2,15 +2,249 @@
...
+    using System.Collections.Generic;
+    using System.Collections.ObjectModel;
+    using System.IdentityModel;
+    using System.IdentityModel.Selectors;
+    using System.Runtime.Serialization;
+    using System.Security;
+    using System.Security.Cryptography;
+    using System.Security.Cryptography.X509Certificates;
+    using System.Text;
+    using System.Xml;
```
iamjasonp: NIT: Formatting, usings should be outside namespace

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SecurityToken.cs
```diff
@@ -14,5 +13,42 @@ public abstract class SecurityToken
...
+                return this.SecurityKeys[0];
+
+            return null;
+        }
+
+        bool CanCreateLocalKeyIdentifierClause()
+        {
+            return (this.Id != null);
+        }
     }
```
iamjasonp: NIT: Formatting for `this`.

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SecurityToken.cs
```diff
@@ -14,5 +13,42 @@ public abstract class SecurityToken
...
 }
```
iamjasonp: Have you tried in N if the typeof and GetType( ) calls work properly?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -312,6 +314,18 @@ protected void Fault()
             OnFaulted();
         }
 
+        internal void Fault(Exception exception)
+        {
+            lock (this.ThisLock)
```
iamjasonp: NIT: elide `this.`

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -802,6 +827,48 @@ internal protected virtual Task OnOpenAsync(TimeSpan timeout)
...
+                {
+                    if (this._exceptions.Count > 0)
+                    {
+                        return this._exceptions.Dequeue();
+                    }
+                }
+
+                return null;
+            }
+        }
```
iamjasonp: NIT: `this.` / formatting

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/LayeredChannel.cs
iamjasonp: Sounds like a plan. Wondering if a Contract.Requires would help with static analysis as well? 

---
In reply to: [78079821](https://github.com/dotnet/wcf/pull/1515#discussion_r78079821) [](ancestors = 78079821)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/LayeredChannelFactory.cs
```diff
@@ -65,11 +66,28 @@ protected override void OnClose(TimeSpan timeout)
...
+        protected internal override Task OnCloseAsync(TimeSpan timeout)
+        {
+            return OnCloseAsyncInternal(timeout);
+        }
+
+        private async Task OnCloseAsyncInternal(TimeSpan timeout)
+        {
+            TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
+            await base.OnCloseAsync(timeoutHelper.RemainingTime());
+            await ((IAsyncCommunicationObject)_innerChannelFactory).CloseAsync(timeout);
```
iamjasonp: >timeout [](start = 79, length = 7)

timeoutHelper.RemainingTime()

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannel.cs
```diff
@@ -18,7 +18,7 @@
 namespace System.ServiceModel.Channels
 {
     // This class is sealed because the constructor could call Abort, which is virtual
-    internal sealed class ServiceChannel : CommunicationObject, IChannel, IClientChannel, IDuplexContextChannel, IOutputChannel, IRequestChannel, IServiceChannel
+    internal /*sealed*/ class ServiceChannel : CommunicationObject, IChannel, IClientChannel, IDuplexContextChannel, IOutputChannel, IAsyncRequestChannel, IServiceChannel
```
iamjasonp: >sealed [](start = 15, length = 6)

Question only: why did we unseal this?

iamjasonp:
> //    }

Wow. Why is this file so big... 

---
Refers to: src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReliableChannelBinder.cs:4286 in b765b75. [](commit_id = b765b754c19c8142a6a07fd10fc35369622afccf, deletion_comment = False)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannelFactory.cs
```diff
@@ -471,28 +478,25 @@ protected override void OnClose(TimeSpan timeout)
...
             {
                 return OnCloseAsyncInternal(timeout);
             }
 
-            protected internal override Task OnOpenAsync(TimeSpan timeout)
-            {
-                this.OnOpen(timeout);
-                return TaskHelpers.CompletedTask();
-            }
-
```
iamjasonp: Is this removal deliberate given our shift to the *Async methods?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/CryptoHelper.cs
```diff
@@ -0,0 +1,375 @@
...
+
+            //    AsymmetricAlgorithm asymmetricAlgorithm = algorithmObject as AsymmetricAlgorithm;
+            //    SignatureDescription signatureDescription = algorithmObject as SignatureDescription;
+            //    if (asymmetricAlgorithm != null || signatureDescription != null)
+            //        return CryptoAlgorithmType.Asymmetric;
+
+            //    return CryptoAlgorithmType.Unknown;
+            //}
+
+            //switch (algorithm)
```
iamjasonp: When we finally do bring these in, we should see if there's anything that's been deprecated and prevent that algorithm from being used. A quick glance shows that we should be ok, but we might need to engage with security people to figure out what shouldn't be supported

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/LayeredChannel.cs
roncain: Good idea, but we experienced some issues with Contract.Requires in other places (i.e. didn't work properly), and we don't do static analysis, so I'm a little reluctant to add it.
Note: based on past experience, if static analysis becomes available and we enable it, I think we can expect the list of contract precondition violations to number in the 10,000's

---
In reply to: [78151294](https://github.com/dotnet/wcf/pull/1515#discussion_r78151294) [](ancestors = 78151294,78079821)

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/SecurityTokenProvider.cs
roncain: Good observations.  I think they should be addressed when I get the IAsyncRequestChannel stuff fully Task-based and we look at how and whether the sync paths use the async paths (like CommunicationObject open and close).  I'm trying to avoid a refactor in this PR and just keep the API's as they are in the full framework.

---
In reply to: [78136608](https://github.com/dotnet/wcf/pull/1515#discussion_r78136608) [](ancestors = 78136608)

src/System.Private.ServiceModel/src/Resources/Strings.resx
roncain: Hmmm, good catch.  I'll remove it.  This was verbatim from the full framework, but even in the full framework that's a pretty ambiguous error message.

---
In reply to: [78136772](https://github.com/dotnet/wcf/pull/1515#discussion_r78136772) [](ancestors = 78136772)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
roncain: Yup -- good idea.  Will do across all files.

---
In reply to: [78142791](https://github.com/dotnet/wcf/pull/1515#discussion_r78142791) [](ancestors = 78142791)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
roncain: I'll do this for all the files in the PR but not in the rest of the code base -- it would explode the PR if I did.

---
In reply to: [78142863](https://github.com/dotnet/wcf/pull/1515#discussion_r78142863) [](ancestors = 78142863)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SamlSecurityToken.cs
roncain: Done

---
In reply to: [78143532](https://github.com/dotnet/wcf/pull/1515#discussion_r78143532) [](ancestors = 78143532)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SamlSerializer.cs
roncain: Done

---
In reply to: [78143615](https://github.com/dotnet/wcf/pull/1515#discussion_r78143615) [](ancestors = 78143615)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SecurityToken.cs
roncain: I ran the tests under N and they passed.  I don't know if this code path gets hit.

---
In reply to: [78148023](https://github.com/dotnet/wcf/pull/1515#discussion_r78148023) [](ancestors = 78148023)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SecurityToken.cs
roncain: Done -- will do for other files in this PR as I find them.

---
In reply to: [78147853](https://github.com/dotnet/wcf/pull/1515#discussion_r78147853) [](ancestors = 78147853)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
roncain: Yuck -- this file already has a ton of this. usage.  I'll fix the code I added but want to prevent sweeping code changes outside of the scope of this PR.

---
In reply to: [78150268](https://github.com/dotnet/wcf/pull/1515#discussion_r78150268) [](ancestors = 78150268)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannel.cs
roncain: Great question -- I need to add a new virtual method RequestAsync, but the compiler disallows it when the class is sealed.  Given this is already internal, I think it's fine to remove the sealed.  Note the comment why it was sealed -- FxCop must have warned them.

---
In reply to: [78154483](https://github.com/dotnet/wcf/pull/1515#discussion_r78154483) [](ancestors = 78154483)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannelFactory.cs
roncain: Great question -- I did remove it deliberately.  This base ServiceChannelFactory class does not override OnBeginOpen. It appears the TypedServiceChannelFactory does the overrides and does not call base, and therefore this became dead code.  By not overriding here, if it is ever called it will fall all the way through to CommunicationObject where an Assert will fire to alert us we need this override here too.  That's part of the pattern to move to the new async open/close.  Any type that *needs* to override but does not triggers the Assert, and then we go fix it.  I relied on that a lot in this PR.

---
In reply to: [78156592](https://github.com/dotnet/wcf/pull/1515#discussion_r78156592) [](ancestors = 78156592)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/CryptoHelper.cs
roncain: Absolutely yes.  Not looking forward to lighting this up.  My initial light-up is for TransportSecurity, but many of these helpers are MessageSecurity -- meaning the light-up is nothing less than lighting up MessageSecurity.

---
In reply to: [78158972](https://github.com/dotnet/wcf/pull/1515#discussion_r78158972) [](ancestors = 78158972)

roncain:
> //    }

It should never have been allowed to contain so many classes and needs to be split up.  There are a few others that are stunningly overstuffed.  Not for this PR though.

---
In reply to: [245869524](https://github.com/dotnet/wcf/pull/1515#issuecomment-245869524) [](ancestors = 245869524)

---
Refers to: src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReliableChannelBinder.cs:4286 in b765b75. [](commit_id = b765b754c19c8142a6a07fd10fc35369622afccf, deletion_comment = False)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
```diff
@@ -0,0 +1,843 @@
...
+        }
+
+        async Task CloseProtocolFactoryAsync(TimeSpan timeout)
+        {
+            if (_securityProtocolFactory != null && !SessionMode)
+            {
+                await _securityProtocolFactory.CloseAsync(timeout);
+                _securityProtocolFactory = null;
+            }
+        }
```
iamjasonp: I'm guessing this code was as before, but we probably should snapshot _securityProtocolFactory before .Close-ing - there's a potential null ref here. Also for the sync path. 
Additionally, consider consolidating so that Close() just calls CloseAsync().GetAwaiter().GetResult() so we don't have to maintain two paths

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
```diff
@@ -0,0 +1,843 @@
...
+            //    ends.Add(new OperationEndCallback(securityProtocolFactory.EndClose));
+            //}
+
+            //if (sessionClientSettings != null)
+            //{
+            //    begins.Add(new OperationWithTimeoutBeginCallback(sessionClientSettings.BeginClose));
+            //    ends.Add(new OperationEndCallback(sessionClientSettings.EndClose));
+            //}
+
+            //return OperationWithTimeoutComposer.BeginComposeAsyncOperations(timeout, begins.ToArray(), ends.ToArray(), callback, state);
```
iamjasonp: >             [](start = 0, length = 12)

NIT: given our intent is to nuke the AsyncResult paths, maybe we should just wipe them out right away instead of cluttering the review with //'ed out code? Up to you when to do it, will not comment on this further

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
```diff
@@ -0,0 +1,843 @@
...
+
+        protected internal override Task OnCloseAsync(TimeSpan timeout)
+        {
+            return OnCloseAsyncInternal(timeout);
+        }
+
+        private async Task OnCloseAsyncInternal(TimeSpan timeout)
+        {
+            TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
+            await base.OnCloseAsync(timeout);
```
iamjasonp: >timeout [](start = 36, length = 7)

timeout -> timeoutHelper.RemainingTime()

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
```diff
@@ -0,0 +1,843 @@
...
+                SessionClientSettings.Open(this, InnerChannelFactory, ChannelBuilder, timeoutHelper.RemainingTime());
+            }
+            else
+            {
+                ThrowIfProtocolFactoryNotSet();
+                _securityProtocolFactory.Open(true, timeoutHelper.RemainingTime());
+            }
+        }
+
+        private async Task OnOpenCoreAsync(TimeSpan timeout)
```
iamjasonp: >private [](start = 8, length = 7)

Question only: Do the new Open/CloseAsyncs need to be private? Or is this to prevent us from internally calling these methods by accident at the moment?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
```diff
@@ -0,0 +1,843 @@
...
+
+            public void Send(Message message, TimeSpan timeout)
+            {
+                ThrowIfFaulted();
+                ThrowIfDisposedOrNotOpen(message);
+                TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
+                SecurityProtocol.SecureOutgoingMessage(ref message, timeoutHelper.RemainingTime());
+                InnerChannel.Send(message, timeoutHelper.RemainingTime());
+            }
+        }
```
iamjasonp: I'm assuming that SendAsync is already on your radar when doing Issue31; no further comment needed if you're already aware of the future work

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportSecurityBindingElement.cs
```diff
@@ -49,10 +51,113 @@ internal override bool SupportsRequestReply
             get { return true; }
         }
 
+        internal override SecurityProtocolFactory CreateSecurityProtocolFactory<TChannel>(BindingContext context, SecurityCredentialsManager credentialsManager, bool isForService, BindingContext issuerBindingContext)
+        {
+            if (context == null)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("context");
+            if (credentialsManager == null)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull("credentialsManager");
```
iamjasonp: Are we missing a null check for issuerBindingContext?

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
roncain: Yes, it was to get DiagnosticUtility to resolve

---
In reply to: [78141546](https://github.com/dotnet/wcf/pull/1515#discussion_r78141546) [](ancestors = 78141546)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
roncain: I'm doing this in all the files I touched but don't want to blow up this PR by visiting all the other files.

---
In reply to: [78143348](https://github.com/dotnet/wcf/pull/1515#discussion_r78143348) [](ancestors = 78143348)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
roncain: I added the snapshot, though this code was copied from the full framework.  Regarding consolidating Close and CloseAsync -- the Close is verbatim from the full framework and has an 'aborted' parameter.  I didn't want that part in the async path, but I'm open to suggestions otherwise.

---
In reply to: [78341698](https://github.com/dotnet/wcf/pull/1515#discussion_r78341698) [](ancestors = 78341698)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
roncain: These comments are going to be deleted as we wire up the Task async code, but I'd prefer to leave them as reference code till then. 

---
In reply to: [78341973](https://github.com/dotnet/wcf/pull/1515#discussion_r78341973) [](ancestors = 78341973)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
roncain: I was trying to be explicit about the modifiers as I added code.  OnOpenCore was effectively already private, so I followed that lead.

---
In reply to: [78342565](https://github.com/dotnet/wcf/pull/1515#discussion_r78342565) [](ancestors = 78342565)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
roncain: Yup, it's a blip on the radar.

---
In reply to: [78346337](https://github.com/dotnet/wcf/pull/1515#discussion_r78346337) [](ancestors = 78346337)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportSecurityBindingElement.cs
roncain: Possibly -- but this is copied from the full framework.  I traced all the calls it makes and I believe none assert all the way down till the point is really is dereferenced.  But I'd rather not start cleaning up the code copied from the full framework at the risk of making this PR even larger.

---
In reply to: [78346772](https://github.com/dotnet/wcf/pull/1515#discussion_r78346772) [](ancestors = 78346772)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/ReceiveSecurityHeader.cs
```diff
@@ -31,9 +57,589 @@ public override string Namespace
...
+            {
+                ThrowIfProcessingStarted();
+                _replayDetectionEnabled = value;
+            }
+        }
+
         protected override void OnWriteHeaderContents(XmlDictionaryWriter writer, MessageVersion messageVersion)
         {
             throw ExceptionHelper.PlatformNotSupported();
         }
```
iamjasonp: Does this PNSE have an implementation that we should be putting in (and perhaps commenting out with an Issue31 comment)?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityAppliedMessage.cs
```diff
@@ -0,0 +1,471 @@
...
+using System.Runtime;
+using System.Security.Cryptography;
+using System.ServiceModel.Channels;
+using System.ServiceModel.Security.Tokens;
+using System.Xml;
+using System.IdentityModel.Tokens;
+using System.Collections.Generic;
+using IPrefixGenerator = System.IdentityModel.IPrefixGenerator;
+using ISecurityElement = System.IdentityModel.ISecurityElement;
+// using XmlAttributeHolder = System.IdentityModel.XmlAttributeHolder;
```
iamjasonp: Is this also Issue31?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityAuditHelper.cs
```diff
@@ -14,5 +16,15 @@ public static bool IsSecurityAuditSupported
...
+
+        public static void WriteMessageAuthenticationSuccessEvent(AuditLogLocation auditLogLocation, bool suppressAuditFailure, Message message,
+    Uri serviceUri, string action, string clientIdentity)
+        {
+        }
+
+        public static void WriteMessageAuthenticationFailureEvent(AuditLogLocation auditLogLocation, bool suppressAuditFailure, Message message,
+     Uri serviceUri, string action, string clientIdentity, Exception exception)
+        {
+        }
```
iamjasonp: NIT: weird formatting in newly added code where we do multiline

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityProtocolFactory.cs
```diff
@@ -686,6 +693,31 @@ public virtual void OnClose(TimeSpan timeout)
...
+        private async Task OnCloseAsyncInternal(TimeSpan timeout)
+        {
+            TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
+            if (!_actAsInitiator)
+            {
+                foreach (SupportingTokenAuthenticatorSpecification spec in _channelSupportingTokenAuthenticatorSpecification)
+                {
+                    await SecurityUtils.CloseTokenAuthenticatorIfRequiredAsync(spec.TokenAuthenticator, timeoutHelper.RemainingTime());
+                }
+                foreach (string action in _scopedSupportingTokenAuthenticatorSpecification.Keys)
```
iamjasonp: I don't see the keys being used - might it make more sense to just enumerate through _scopedSupportingTokenAuthenticatorSpecification.Values to get the `ICollection<SupportingTokenAuthenticatorSpecification>` rather than indexing based on action?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/ReceiveSecurityHeader.cs
roncain: Reasonable suggestion.  I'll do it.

---
In reply to: [78404876](https://github.com/dotnet/wcf/pull/1515#discussion_r78404876) [](ancestors = 78404876)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityProtocolFactory.cs
iamjasonp: (though I get that we want to keep the code similar to the sync path...) Feel free to not fix this 

---
In reply to: [78406802](https://github.com/dotnet/wcf/pull/1515#discussion_r78406802) [](ancestors = 78406802)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityAppliedMessage.cs
roncain: Great catch.  I'm removing this line because XmlAttributeHolder in ServiceModel exists and appears to be identical to the IdentityModel one

---
In reply to: [78405348](https://github.com/dotnet/wcf/pull/1515#discussion_r78405348) [](ancestors = 78405348)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityProtocolFactory.cs
roncain: Yeah, I was constantly tempted to refactor, but in the end did a verbatim copy and then Taskified it.  I sort of prefer keeping them looking as similar as possible so we'd recognize when one of them changed.

---
In reply to: [78407020](https://github.com/dotnet/wcf/pull/1515#discussion_r78407020) [](ancestors = 78407020,78406802)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SendSecurityHeader.cs
```diff
@@ -40,9 +130,853 @@ public override string Namespace
...
+        }
+
+        public bool HasId
+        {
+            get { return true; }
+        }
+
+        public string Id
+        {
+            get { return token.Id; }
```
iamjasonp: null ref possible - maybe we want to Contract.Assert at the ctor as a preemptive measure?

iamjasonp: ... Not blocking

---
In reply to: [78423001](https://github.com/dotnet/wcf/pull/1515#discussion_r78423001) [](ancestors = 78423001)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/TransportSecurityProtocol.cs
```diff
@@ -0,0 +1,305 @@
...
+           IList<SecurityToken> signedEndorsingTokens, IList<SecurityToken> signedTokens, Dictionary<SecurityToken, ReadOnlyCollection<IAuthorizationPolicy>> tokenPoliciesMapping)
+        {
+            SecurityMessageProperty security = SecurityMessageProperty.GetOrCreate(message);
+            AddSupportingTokenSpecification(security, basicTokens, endorsingTokens, signedEndorsingTokens, signedTokens, tokenPoliciesMapping);
+            security.ServiceSecurityContext = new ServiceSecurityContext(security.GetInitiatorTokenAuthorizationPolicies());
+        }
+
+        protected virtual void VerifyIncomingMessageCore(ref Message message, TimeSpan timeout)
+        {
+            // $$$ throw ExceptionHelper.PlatformNotSupported();   // #31 in progress
```
iamjasonp: Should this throw be taken out now that it seems the method is implemented?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/TransportSecurityProtocol.cs
```diff
@@ -0,0 +1,305 @@
...
+            securityHeader.RequireMessageProtection = false;
+            securityHeader.ExpectBasicTokens = expectBasicTokens;
+            securityHeader.ExpectSignedTokens = expectSignedTokens;
+            securityHeader.ExpectEndorsingTokens = expectEndorsingTokens;
+            securityHeader.MaxReceivedMessageSize = factory.SecurityBindingElement.MaxReceivedMessageSize;
+            securityHeader.ReaderQuotas = factory.SecurityBindingElement.ReaderQuotas;
+
+            // Due to compatibility, only honor this setting if this app setting is enabled
+            // Issue #31 in progress
+            //if (ServiceModelAppSettings.UseConfiguredTransportSecurityHeaderLayout)
```
iamjasonp: For later - it looks like this is just an AppSetting that was copied over. Do we have a way to do AppSettings in .NET Core? If not, recommend we get rid of this or otherwise come up with some keyword with which we can search such code

src/System.Private.ServiceModel/src/System/ServiceModel/Security/WSKeyInfoSerializer.cs
```diff
@@ -0,0 +1,389 @@
...
+            }
+
+            protected class SecurityContextTokenEntry : TokenEntry
+            {
+                private WSSecureConversation _parent;
+                private Type[] _tokenTypes;
+
+                public SecurityContextTokenEntry( WSSecureConversation parent )
+                {
+                    _parent = parent;
```
iamjasonp: >parent [](start = 30, length = 6)

Future - consider adding a Contract.Assert for _parent != null

src/System.Private.ServiceModel/src/System/ServiceModel/Security/WSSecureConversation.cs
```diff
@@ -0,0 +1,599 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IdentityModel.Selectors;
+using System.IdentityModel.Tokens;
+using System.Runtime;
+using System.ServiceModel;
+using System.ServiceModel.Security.Tokens;
+using System.Xml;
+//using StrEntry = WSSecurityTokenSerializer.StrEntry;
```
iamjasonp: Issue31 temporary removal or another instance where we're removing for consolidation?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/Tokens/SecurityTokenParameters.cs
```diff
@@ -57,11 +75,56 @@ public SecurityTokenParameters Clone()
 
         protected abstract SecurityTokenParameters CloneCore();
 
+        internal protected abstract SecurityKeyIdentifierClause CreateKeyIdentifierClause(SecurityToken token, SecurityTokenReferenceStyle referenceStyle);
+
+        internal protected abstract void InitializeSecurityTokenRequirement(SecurityTokenRequirement requirement);
+
+        internal SecurityKeyIdentifierClause CreateKeyIdentifierClause<TExternalClause, TInternalClause>(SecurityToken token, SecurityTokenReferenceStyle referenceStyle)
+    where TExternalClause : SecurityKeyIdentifierClause
```
iamjasonp: >where [](start = 4, length = 5)

NIT: spacing before "where" clauses

src/System.Private.ServiceModel/src/System/ServiceModel/Security/WSSecureConversation.cs
roncain: Removed it -- it was not needed in this file anyway

---
In reply to: [78424958](https://github.com/dotnet/wcf/pull/1515#discussion_r78424958) [](ancestors = 78424958)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/TransportSecurityProtocol.cs
roncain: I'll investigate what they want the story to be in NET Core

---
In reply to: [78423934](https://github.com/dotnet/wcf/pull/1515#discussion_r78423934) [](ancestors = 78423934)

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/SecurityTokenSerializer.cs
mconnew: Is this the right way around to do things? WCF originally had all these classes internally, and then IdentityModel was created and duplicated there and not all the WCF versions were eliminated. If IdentityModel were to come back, then we would likely want to do another sweep and remove them again.

---
In reply to: [78077394](https://github.com/dotnet/wcf/pull/1515#discussion_r78077394) [](ancestors = 78077394)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            get { return this.referenceXml; }
+        }
+
+        public override bool Matches(SecurityKeyIdentifierClause keyIdentifierClause)
+        {
+            GenericXmlSecurityKeyIdentifierClause that = keyIdentifierClause as GenericXmlSecurityKeyIdentifierClause;
+
+            // PreSharp Bug: Parameter 'that' to this public method must be validated: A null-dereference can occur here.
+#pragma warning suppress 56506
```
mconnew: Is this necessary now?

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
```diff
@@ -0,0 +1,370 @@
...
+            _emitBspRequiredAttributes = emitBspRequiredAttributes;
+            _innerSecurityTokenSerializer = innerSecurityTokenSerializer;
+
+            _serializerEntries = new List<SecurityTokenSerializer.SerializerEntries>();
+
+            // Issue #31 in progress
+            //serializerEntries.Add(new XmlDsigSep2000(this));
+            //serializerEntries.Add(new XmlEncApr2001(this));
+            //serializerEntries.Add(new WSTrust(this, trustDictionary));
+
```
mconnew: Would it be possible to add skeleton classes that throw for these? That way it's less likely this will get missed as missing these items to serialize could be really difficult bug to track down.

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SamlAssertion.cs
```diff
@@ -0,0 +1,671 @@
...
+using System.IdentityModel.Selectors;
+using System.Runtime;
+using System.Runtime.InteropServices;
+using System.Security;
+using System.Security.Cryptography;
+using System.Xml;
+
+namespace System.IdentityModel.Tokens
+{
+    public class SamlAssertion // : ICanonicalWriterEndRootElementCallback  // Issue #31 in progress
```
mconnew: >ICanonicalWriterEndRootElementCallback [](start = 36, length = 38)

FYI, this interface will be used as part of message signing.

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SamlSerializer.cs
```diff
@@ -2,15 +2,250 @@
...
+            get
+            {
+                if (_dictionaryManager == null)
+                    _dictionaryManager = new DictionaryManager();
+
+                return _dictionaryManager;
+            }
+        }
+
+        // Issue #31 in progress
```
mconnew: I'm not expecting you to go back and change everything, this is just an idea, but #if'ing the blocks out might make things easier as you can temporarily define the item to see how much left there is to resolve.

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SecurityToken.cs
mconnew: If you can call the method with the generic type, then it must be able to get the Type object. I'm not seeing any dependence on the RuntimeType so I would expect it to be safe.

---
In reply to: [78184790](https://github.com/dotnet/wcf/pull/1515#discussion_r78184790) [](ancestors = 78184790,78148023)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
```diff
@@ -0,0 +1,845 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime;
+using System.Runtime.InteropServices;
+using System.Security.Authentication.ExtendedProtection;
+using System.ServiceModel;
+//using System.ServiceModel.Diagnostics.Application;
```
mconnew: Nit, remove?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
```diff
@@ -0,0 +1,845 @@
...
+            //return new OperationWithTimeoutAsyncResult(new OperationWithTimeoutCallback(OnOpen), timeout, callback, state);
+        }
+
+        protected override void OnEndOpen(IAsyncResult result)
+        {
+            throw ExceptionHelper.PlatformNotSupported(); // Issue #31 in progress
+            //OperationWithTimeoutAsyncResult.End(result);
+        }
+
+        void OnOpenCore(TimeSpan timeout)
```
mconnew: >void [](start = 8, length = 4)

Isn't current code standard that access modifiers should be explicit?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
```diff
@@ -0,0 +1,845 @@
...
+                    null,
+                    typeof(TChannel) == typeof(IRequestChannel),
+                    timeoutHelper.RemainingTime());
+                OnProtocolCreationComplete(securityProtocol);
+                await SecurityProtocol.OpenAsync(timeoutHelper.RemainingTime());
+                await base.OnOpenAsync(timeoutHelper.RemainingTime());
+            }
+
+            void EnableChannelBindingSupport()
+            {
```
mconnew: We don't have channel binding support wired up for HttpTransport yet. I'm unsure if it's even supported yet.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
```diff
@@ -0,0 +1,845 @@
...
+            public IOutputSession Session
+            {
+                get
+                {
+                    return ((IOutputSessionChannel)InnerChannel).Session;
+                }
+            }
+        }
+
+        class SecurityRequestChannel : ClientSecurityChannel<IRequestChannel>, IRequestChannel
```
mconnew: Should this be implementing IAsyncRequestChannel?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannel.cs
mconnew: Why are the methods virtual? They come from IAsyncRequestChannel so there's no requirement for them to be virtual.

---
In reply to: [78188771](https://github.com/dotnet/wcf/pull/1515#discussion_r78188771) [](ancestors = 78188771,78154483)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannel.cs
```diff
@@ -1221,6 +1222,16 @@ public Message Request(Message message, TimeSpan timeout)
             return (Message)Call(message.Headers.Action, false, operation, new object[] { message }, Array.Empty<object>(), timeout);
         }
 
+        public virtual Task<Message> RequestAsync(Message message)
```
mconnew: There's no requirement for these to be virtual. The reason many classes have Async methods as virtual is to prevent breaking existing code which derives from the class which might not have overridden the Async methods. As this class was sealed, nothing derives from it (and it's internal) so there's no need to make the methods virtual.

src/System.Private.ServiceModel/src/System/ServiceModel/Description/ClientCredentials.cs
```diff
@@ -18,6 +18,7 @@ public class ClientCredentials : SecurityCredentialsManager, IEndpointBehavior
         private WindowsClientCredential _windows;
         private HttpDigestClientCredential _httpDigest;
         private bool _isReadOnly;
+        private bool _useIdentityConfiguration = false;
```
mconnew: We don't have config. I would remove this field, and on the property getter always return false and have the setter throw if true is passed, otherwise ignore.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/CryptoHelper.cs
mconnew: These algorithms are still all in the latest soap spec and some of them are required to be supported by a soap version which we officially support but you wouldn't want to use them for a real scenario. I believe nothing in that list is deprecated although some is discouraged.

---
In reply to: [78192012](https://github.com/dotnet/wcf/pull/1515#discussion_r78192012) [](ancestors = 78192012,78158972)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/EncryptedKey.cs
```diff
@@ -0,0 +1,132 @@
...
+            //{
+            //    this.referenceList = new ReferenceList();
+            //    this.referenceList.ReadFrom(reader);
+            //}
+            //if (reader.IsStartElement(CarriedKeyElementName, EncryptedType.NamespaceUri))
+            //{
+            //    reader.ReadStartElement(CarriedKeyElementName, EncryptedType.NamespaceUri);
+            //    this.carriedKeyName = reader.ReadString();
+            //    reader.ReadEndElement();
+            //}
```
mconnew: I believe everything you need to uncomment this code is available.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityProtocolFactory.cs
```diff
@@ -686,6 +693,31 @@ public virtual void OnClose(TimeSpan timeout)
...
+        }
+
+        private async Task OnCloseAsyncInternal(TimeSpan timeout)
+        {
+            TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
+            if (!_actAsInitiator)
+            {
+                foreach (SupportingTokenAuthenticatorSpecification spec in _channelSupportingTokenAuthenticatorSpecification)
+                {
+                    await SecurityUtils.CloseTokenAuthenticatorIfRequiredAsync(spec.TokenAuthenticator, timeoutHelper.RemainingTime());
```
mconnew: >timeoutHelper.RemainingTime() [](start = 104, length = 29)

Is it worth converting to CancellationToken's? You can instantiate a single one at the beginning and reuse.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/CryptoHelper.cs
```diff
@@ -0,0 +1,364 @@
...
+            //            return new SHA256CryptoServiceProvider();
+            //        else
+            //            return new SHA256Managed();
+            //    default:
+            //        throw DiagnosticUtility.ExceptionUtility.ThrowHelperWarning(new MessageSecurityException(SR.GetString(SR.UnsupportedCryptoAlgorithm, digestMethod)));
+
+            //}
+        }
+
+        [SuppressMessage("Microsoft.Security.Cryptography", "CA5354:DoNotUseSHA1", Justification = "Cannot change. Required as SOAP spec requires supporting SHA1.")]
```
iamjasonp: For later - Remove the FxCop suppression, but it might be useful to keep the comment around in case in future we are questioned about why SHA1 is being used

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityProtocolFactory.cs
iamjasonp: I think it's probably worth converting to it eventually, but I'm fine with leaving this as a work item for later given the state of the temporary implementation

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
roncain: No -- I'll try to find and remove all those from the files already in this PR.  We should cleanup the others in a separate PR to keep this one as "small" as possible.

---
In reply to: [78483477](https://github.com/dotnet/wcf/pull/1515#discussion_r78483477) [](ancestors = 78483477)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
roncain: I was able to bring in the first 2, but will need to go deeper on WSTrust.  It is abstract in ServiceModel and cannot be instantiated like this.  Another duplicate-and-diverge story between IdentityModel and ServiceModel.

---
In reply to: [78483747](https://github.com/dotnet/wcf/pull/1515#discussion_r78483747) [](ancestors = 78483747)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SamlSerializer.cs
roncain: I thought about that approach, but I'm finding the commented out lines give a better experience when searching.  You know the line is commented out even with FindStr. I know -- just the opposite of what we favored in our first pass of WCF, but I'm hoping these commented sections are uncommented soon.

---
In reply to: [78484083](https://github.com/dotnet/wcf/pull/1515#discussion_r78484083) [](ancestors = 78484083)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
roncain: Sigh -- yes.  I'm finding these all over the place.  But I didn't want to CodeFormat the entire solution.  What's your thinking -- fix by hand?  CodeFormat the entire code base after the initial checkin?

---
In reply to: [78484691](https://github.com/dotnet/wcf/pull/1515#discussion_r78484691) [](ancestors = 78484691)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
roncain: Yes, I'll make that change and stub them for now.

---
In reply to: [78485068](https://github.com/dotnet/wcf/pull/1515#discussion_r78485068) [](ancestors = 78485068)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
roncain: What do you think we should do?  I don't want to PNSE here or we defeat the default scenario.  Maybe just stub it all out or remove the call to it?

---
In reply to: [78484932](https://github.com/dotnet/wcf/pull/1515#discussion_r78484932) [](ancestors = 78484932)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannel.cs
roncain: Good point -- will remove virtual.  Don't remember why I did it in the first place.

---
In reply to: [78485372](https://github.com/dotnet/wcf/pull/1515#discussion_r78485372) [](ancestors = 78485372)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/WrapperSecurityCommunicationObject.cs
```diff
@@ -45,14 +46,17 @@ protected override void OnAbort()
             _innerCommunicationObject.OnAbort();
         }
 
+
         protected override IAsyncResult OnBeginClose(TimeSpan timeout, AsyncCallback callback, object state)
```
mconnew: Nit: Extra line

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
iamjasonp: I'd suggest codeformatting later. There's no reason to litter this PR with even more stuff :)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
iamjasonp: Sounds good

iamjasonp: Sounds good

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/SamlAssertion.cs
roncain: Thanks Matt

---
In reply to: [78483831](https://github.com/dotnet/wcf/pull/1515#discussion_r78483831) [](ancestors = 78483831)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/EncryptedKey.cs
roncain: I'll can enable all of it except the call to ReadString which is in Net Standard 2.0.  I'm going to add "NET Standard 2.0" to all comments where we disable something pending taking a dependency on the new API's coming into CoreFx's NET Standard 2.0.  I may be able to use the dev/api feed once we create our own dev/api branch, but I'm hanging in master just to stay in sync with its changes.

---
In reply to: [78486674](https://github.com/dotnet/wcf/pull/1515#discussion_r78486674) [](ancestors = 78486674)

src/System.Private.ServiceModel/src/System/ServiceModel/Description/ClientCredentials.cs
roncain: ok

---
In reply to: [78485628](https://github.com/dotnet/wcf/pull/1515#discussion_r78485628) [](ancestors = 78485628)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/CryptoHelper.cs
roncain: I've been looking at CoreFx, and SuppressMessage is still widely used there.  I don't know what the plans are for them long term.  So for now I think I would like to preserve these in case there is any value in linking back to what FxCop was warning against.

---
In reply to: [78506207](https://github.com/dotnet/wcf/pull/1515#discussion_r78506207) [](ancestors = 78506207)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityChannelFactory.cs
roncain: Agree. I opened issue #1528 to do the all-up CodeFormatter pass.

---
In reply to: [78679251](https://github.com/dotnet/wcf/pull/1515#discussion_r78679251) [](ancestors = 78679251)

roncain:
Closing this PR now because PR #1530 copied all the files in this review verbatim into the new ws-trust branch.

PR #1519, opened 2016-09-13
Update PRService to deal with issues at merge

### LINKED ISSUE #1066: WcfService: Investigate how to force-quit badly behaved services on CI server
We've run into a couple of instances now where the CI server box has failed the prerequisite/cleanup steps because either git.exe was running for a long time and not proceeding, or w3wp.exe was holding onto files that was preventing the cleanup/sync necessary. 

This has necessitated manual intervention in these cases to fix and stabilize the services. 

We should find a way to force-quit and recover services/processes like this on the CI server to help keep our CI server system stable without the need for human intervention.

iamjasonp:
The merge step to get master into a good state was causing an indefinite hang in
certain cases, which would result in 1) git.exe to hang and 2) the PRService to return
a 500 error and cause CI to fail intermittently

This change removes the merge step and replaces it with a force checkout, deletes all
branches, and then checks out to origin/master during the cleanup step. This results in
a clean repo as well as sidesteps the merge step causing the hang

skip ci please

Fixes #1066

### DIFF
--- a/src/System.Private.ServiceModel/tools/PRService/pr.ashx
+++ b/src/System.Private.ServiceModel/tools/PRService/pr.ashx
@@ -1,5 +1,6 @@
- // Copyright (c) Microsoft. All rights reserved.
- // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
  
 <%@ WebHandler Language="C#" Class="PullRequestHandler" %>
 
@@ -279,10 +280,8 @@ public class PullRequestHandler : IHttpHandler
 
         string[] gitCommands = new string[]
         {
-            "checkout -f master",
-            "clean -fdx",
             "fetch --all --prune",
-            "merge --ff-only origin/master"
+            "checkout -f origin/master"
         };
 
         success = RunGitCommands(gitCommands, gitRepoPath, executionResult);
@@ -300,13 +299,14 @@ public class PullRequestHandler : IHttpHandler
             // we won't get branchesList == null because of the success check
             for (int i = 0; i < branchesList.Length; i++)
             {
-                // Don't try to delete "master" branch or detached since we're checked out to it
-                if (!branchesList[i].Equals("master"))
-                {
-                    commands.Add(string.Format("branch -D {0}", branchesList[i]));
-                }
+                // delete all branches - GetBranches will prevent deletion of the detached head at origin/master
+                commands.Add(string.Format("branch -D {0}", branchesList[i]));
             }
 
+            // Execute a checkout to put the branch back to a usable state at the master branch
+            // We do this to prevent successive origin/master checkouts from checking out to the same commit
+
+            commands.Add("checkout -b master --track origin/master");
             success = RunGitCommands(commands.ToArray(), gitRepoPath, executionResult);
         }
 
@@ -331,7 +331,7 @@ public class PullRequestHandler : IHttpHandler
         };
 
         bool success = true;
-        
+
         foreach (string gitCommand in gitCommands)
         {
             success = false;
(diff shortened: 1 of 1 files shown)


iamjasonp:
This change is currently in testing already on wcfcoresrv2

roncain:
Question only (don't block) -- do we know why the merge would hang periodically?  Some kind of git bug?  Prompting for UI?  One reason I ask is that occasionally I do a merge, git complains about some kind of mismatch and recommends a reboot.  It happens about once a month, and I wondered if it was a similar phenomenon.

---
In reply to: [246649041](https://github.com/dotnet/wcf/pull/1519#issuecomment-246649041) [](ancestors = 246649041)

src/System.Private.ServiceModel/tools/PRService/pr.ashx
```diff
@@ -279,10 +280,8 @@ public class PullRequestHandler : IHttpHandler
 
         string[] gitCommands = new string[]
         {
-            "checkout -f master",
-            "clean -fdx",
             "fetch --all --prune",
-            "merge --ff-only origin/master"
+            "checkout -f origin/master"
```
roncain: >checkout -f origin/master [](start = 13, length = 25)

Non-blocking question: could you help me understand exactly what git does in the new code you've added?  For example, I see that a forced checkout to origin/master puts the repro into a state with a detached HEAD.  Then it deletes *all* branches, including master, and creates a new tracking branch called 'master' that tracks origin/master and makes that the current branch.  So the very first time this runs (after someone did the first clone), it deletes the "normal" local master branch.  Subsequent runs delete the tracking branch called master.  In all cases, it works with a pristine copy of what is in origin/master that does not depend on the prior state of the master.  Do I have that right?

roncain:
LGTM.
I asked some questions in comments but none are blocking.  I tried these steps myself and they worked fine for me.

iamjasonp:
> do we know why the merge would hang periodically? Some kind of git bug? Prompting for UI? One reason I ask is that occasionally I do a merge, git complains about some kind of mismatch and recommends a reboot. It happens about once a month, and I wondered if it was a similar phenomenon.

I spent a little time trying to figure this out - it's strange as it seems to only happen in the context of me launching this process via w3wp, and there seems not to be any kind of pop-up UI or other message that blocks. Basically if you observed the machine after hitting some repos (and it's different each time), then git.exe is just stuck; any subsequent requests to the same repo on a different git.exe process works fine though, which is why the problems show up as intermittent problems as CI (i.e., a retry will likely fix things)

I've never seen any kind of "inconsistency, please reboot" message either - the worst I've ever seen in a message doing a merge is that the repo can't be fast-forwarded. 

Re: [comment](https://github.com/dotnet/wcf/pull/1519#discussion_r78551851):
> could you help me understand exactly what git does in the new code you've added? 

Yes, your description is correct. The advantage of this set of steps is that we end up with a pristine copy of the repo every time we run `CleanupBranches`, and all further checkouts, also run with `-f`, basically will again forcibly check out any tracked files to the correct commit in `origin` (i.e., dotnet/wcf) 

The reason why we re-create a local master branch is because I've observed odd occasions where `checkout -f origin/master` followed by a bunch of commits, fetch, followed by `checkout -f origin/master` caused nothing to happen as the repo believed it's already on the correct branch. This _shouldn't_ be a problem given the `-f`, but this extra step helps us be doubly sure. 

Also - I don't like using `pull` because should the local branch contain committed changes (not likely, but possible if someone is manually debugging something on said repo), a merge is automatically done and you end up with a messed up repo which no longer fast forwards. 

> So the very first time this runs (after someone did the first clone), it deletes the "normal" local master branch.

Actually, upon first clone, local master already tracks origin/master. Git is fairly intelligent about this - if you check out a branch which exists in your origin remote but not locally, it'll automatically set up a tracking branch. If you want to try this, make sure your local repo has no StressTests branch (git branch -D StressTests), and then `git checkout StressTests` - you'll see a message stating that the checkout created a new StressTests branch tracking origin/StressTests

I spell it out in this case because, you know, scripts. Verbosity is good :smile:

PR #1522, opened 2016-09-13
Miscellaneous test property fixes

roncain:
Trim test property and environment variable values to avoid
issues we found with accidental whitespace.

UPN_Available was not properly being passed into TestProperties,
so this Condition was always false unless an environment variable
was set for it on the machine where the tests were run.

IssueAttribute parses the IncludeTestsWithIssues TestProperty to
allow multiple issues to be specified.  But if semicolon separated,
it causes a code-gen error in TestProperties.  So change it to allow
a comma separated list.


hongdai:
LGTM.

PR #1545, opened 2016-09-16
Update caseing in dependencies.props

StephenBonikowsky:
*  RemoteDependencyBuildInfo "WCF" needs to match the CurrentRef property but it is case sensitive.


StephenBonikowsky:
This should fix the auto updating PR #1540

PR #1547, opened 2016-09-19
Fix issue where CommunicationObject.OnOpen was not called.

### LINKED ISSUE #1544: CommunicationObject.OnOpen and OnBeginOpen is never called

Thanks

roncain:
CommunicationObject.OnOpen() was not being called properly when the synchronous Open() method was used.   The fix was to make both Open and Close code paths be aware if they were started synchronously or asynchronously and to call their appropriate versions of OnOpen/OnBeginOpen or OnClose/OnBeginClose.

fixes #1544

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
@@ -29,6 +29,8 @@ namespace System.ServiceModel.Channels
         private CommunicationState _state;
         private bool _onOpenAsyncCalled;
         private bool _onCloseAsyncCalled;
+        private bool _isSynchronousOpen;
+        private bool _isSynchronousClose;
 
         protected CommunicationObject()
             : this(new object())
@@ -144,6 +146,7 @@ namespace System.ServiceModel.Channels
 
         public void Close(TimeSpan timeout)
         {
+            _isSynchronousClose = true;
             CloseAsyncInternal(timeout).WaitForCompletion();
         }
 
@@ -319,6 +322,7 @@ namespace System.ServiceModel.Channels
 
         public void Open(TimeSpan timeout)
         {
+            _isSynchronousOpen = true;
             OpenAsyncInternal(timeout).WaitForCompletion();
         }
 
@@ -767,17 +771,28 @@ namespace System.ServiceModel.Channels
                             !GetType().Namespace.StartsWith("System.ServiceModel"),
                             String.Format("Type '{0}' is required to override OnCloseAsync", GetType()));
 #endif
+            // The code below executes only for custom CommunicationObjects because
+            // all WCF product code is required to override OnCloseAsync() and not to call
+            // this base implementation.
 
             // If we have already executed this code path, just return a completed Task.
-            // It means this type did not override OnCloseAsync or OnBeginClose.
+            // This is a safeguard against product code that failed to override OnCloseAsync().
             if (_onCloseAsyncCalled)
             {
                 return TaskHelpers.CompletedTask();
             }
 
-            // External types cannot override this method because it is not public.
-            // Redirect this call to the APM path.
             _onCloseAsyncCalled = true;
+
+            // External types cannot override this method because it is not public.
+            // If the caller started this open synchronously, we must use the synchronous
+            // methods they may have overridden, but we call then asynchronously.
+            if (_isSynchronousClose)
+            {
+                return CallActionAsync<TimeSpan>(OnClose, timeout);
+            }
+
+            // This began as an asynchronous close, so redirect this call to the APM path.
             return Task.Factory.FromAsync(OnBeginClose, OnEndClose, timeout, TaskCreationOptions.RunContinuationsAsynchronously);
         }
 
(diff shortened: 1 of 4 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -769,15 +773,23 @@ internal protected virtual Task OnCloseAsync(TimeSpan timeout)
...
-            // External types cannot override this method because it is not public.
-            // Redirect this call to the APM path.
             _onCloseAsyncCalled = true;
+
+            // External types cannot override this method because it is not public.
+            // If the caller started this open synchronously, we must use the synchronous
+            // methods they may have overridden.
+            if (_isSynchronousClose)
+            {
+                return Task.Run(() => OnClose(timeout));
```
mconnew: This will cause a lot more threads to be used. In a channel factory with 3 channel layers, each layer will cause an extra thread to be used. This will cause thread pool starvation and complete process stalling as we wait for the threadpool to add more threads.

src/System.ServiceModel.Primitives/tests/Channels/CommunicationObjectTest.cs
```diff
@@ -0,0 +1,544 @@
...
+    public static void CommunicationObject_Abort_Close_Methods_Called()
+    {
+        MockCommunicationObject mco = new MockCommunicationObject();
+        List<string> closeMethodsCalled = new List<string>();
+
+        // *** SETUP *** \\
+        mco.OnClosingOverride = () =>
+        {
+            closeMethodsCalled.Add("OnClosing");
+            mco.DefaultOnClosing();
```
mconnew: Maybe create a helper class with all these overrides already populated with delegates and expose the lists to reduce code duplication.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
iamjasonp: Interesting - what's a good solution to this? 

---
In reply to: [79509865](https://github.com/dotnet/wcf/pull/1547#discussion_r79509865) [](ancestors = 79509865)

roncain: Yes, I am also interested in what you propose. This is actually the code that was there before I switched to using OnBeforeOpen. Perhaps there is a way to use AsyncLocal or something like it to detect we are already running in a layered async path?

src/System.ServiceModel.Primitives/tests/Channels/CommunicationObjectTest.cs
roncain: Not sure what you are suggesting. The mock has already prepopulated all of these with delegates. The test is just replacing them to override default behavior for test-specific purposes. I really wanted Moq here, so this is a handspun version.

src/System.ServiceModel.Primitives/tests/Channels/CommunicationObjectTest.cs
```diff
@@ -0,0 +1,544 @@
...
+        };
+
+        mco.OnCloseOverride = (TimeSpan t) =>
+        {
+            closeMethodsCalled.Add("OnClose");
+            mco.DefaultOnClose(t);
+        };
+
```
zhenlan: Missing `OnBeginCloseOverride`?

src/System.ServiceModel.Primitives/tests/Channels/CommunicationObjectTest.cs
zhenlan: I suppose we can have a common setup function that takes a `MockCommunicationObject` argument. It will hook up all the On*Overrides and probably events as well. Each test will call the same setup function so we don't worry about missing anything (also reduced duplicated code), but each test will have different execution block so is validation block.

roncain: We can chat, but MockCommunicationObject is fully initialized with delegates to behave as a CommunicationObject. It requires no further initialization. The tests decide which methods to override *for that test*. This is standard 'mock' behavior. I've modelled this after Moq. If some tests need to initialize in the same way, we can provide a helper if needed. Generally, however, it is better for each unit test to be self-contained. I'd prefer not to stall this PR on discussions of how to write mocks but am open to discussion post-merge. The main question is "Do these tests validate what we need to validate?"

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -789,19 +804,51 @@ internal protected virtual Task OnOpenAsync(TimeSpan timeout)
...
+                return CallActionAsync<TimeSpan>(OnOpen, timeout);
+            }
+
+            // This began as an asynchronous open, so just redirect this call to the APM path.
             return Task.Factory.FromAsync(OnBeginOpen, OnEndOpen, timeout, TaskCreationOptions.RunContinuationsAsynchronously);
         }
+
+        // Calls the given Action asynchronously.  It uses the IOThreadScheduler
+        // to avoid consuming an extra thread.
+        private Task CallActionAsync<TArg>(Action<TArg> action, TArg argument)
```
mconnew:
```csharp
private async Task CallActionAsync<TArg>(Action<TArg> action, TArg argument)
{
    using(TaskHelpers.RunTaskContinuationsOnOurThreads())
    {
        await Task.Yield(); // Move synchronous method off of thread pool
        action(argument);
    }
}
```

mconnew:
LGTM

zhenlan:
LGTM

Issue #10, opened 2015-05-13
HTTP does not support UpnEndpointIdentity


iamjasonp:
The ctor currently throws PlatformNotSupportedException

zhenlan:
With [this](https://github.com/dotnet/wcf/commit/4a9fe667bbe2eeb035306378bb9c3b09ba22db60), this should only be an issue for UWP.

iamjasonp:
This is a little bit more nuanced: 

* **On UWP,** the class can not be constructed 

* On other platforms, UpnEndpointIdentity can be constructed, and: 
  * In HTTP scenarios, the UPN is not passed on due to dotnet/ corefx#4502
  * In Windows TCP scenarios, the credential will be passed along on supported platforms.
  * In non-Windows TCP scenarios, #946, #1262 prevents cases from working at the moment.

roncain:
Tests disabled with Issue(10) now pass for me on Windows when run manually.

roncain:
Closing because #1502 appears to have fixed this.  I will open a new PR that re-enables the tests previously disabled by issue #10.

roncain:
Reopening -- this issue exists only in UWP, even after PR #1502.

roncain:
Retesting in UWP showed these tests pass now, so PR #1560 has been opened to re-enable them.

roncain:
Closing now that #1560 has been merged.  I manually tested this against NET Native running against a self-hosted service running on a 2nd machine, and passing explicit credential in TestProperties.

PR #1548, opened 2016-09-19
Re-enable tests disabled by issue 10

roncain:
PR #1502 fixed issue #10 by implementing UpnEndpointIdentity.
But 2 tests were still disabled by issue 10, so this PR just
re-enables those tests.  Both require manual setup to run, so
we will not see them pass or fail in CI or VSO.

Update: the tests still fail in UWP, so this PR only re-enables them for non-UWP frameworks.


roncain:
Please don't merge till testing against NET Native is complete.

PR #1552, opened 2016-09-21
Port to release/1.1.0 fix to issue #1544

### LINKED ISSUE #1544: CommunicationObject.OnOpen and OnBeginOpen is never called

roncain:
Original commit message in master was:
Fix issue where CommunicationObject.OnOpen was not called

Converting the synchronous CommunicationObject open/close paths
introduced a bug that prevented the OnOpen() method to be called
when the synchronous Open() method was used.

The fix is to be aware of whether the caller invoked the synchronous
Open or the asynchronous BeginOpen and invoke the appropriate
synchronous or asynchronous override.

Fixes #1544


roncain:
Given permission to merge by shiproom.

PR #1553, opened 2016-09-21
Fix TestTFM filtering so it works for all and not just netcoreapp

StephenBonikowsky:
* Porting dotnet/corefx#11819


hongdai:
LGTM.

PR #1568, opened 2016-09-27
Remove all the #if#defs needed for ToF

### LINKED ISSUE #1527: Remove all conditional test code used to run under ToF
We have maintained a conditional FULLXUNIT_NOTSUPPORTED that permitted our tests to run under both ToF and Helix.  But now that we can run against UWP directly from GitHub, there is no need to maintain this extra code for ToF.

All the conditional code inside FULLXUNIT_NOTSUPPORTED should be removed.

Note that issue #1526 is somewhat related -- updating VSO to make UWP runs nightly.  That issue does not need to be resolved before resolving this one.

StephenBonikowsky:
* Fixes #1527


fgreinacher:
Out of curiosity: What's ToF?

fgreinacher:
I see, thanks for the info!

PR #1571, opened 2016-09-28
Fix OnOpen/OnBeginOpen issues with custom channels

### LINKED ISSUE #1544: CommunicationObject.OnOpen and OnBeginOpen is never called

roncain:
Fixes the issue where OnOpen/OnBeginOpen and
OnClose/OnBeginClose were incorrectly called when
using custom channels.

Fixes the product issue by propagating knowledge of
sync and async open/close requests down the channel
stack so they stay sync or async based on the original
call.

Created mocks and unit tests for custom channels

**READY FOR REVIEW NOW**

Fixes #1544

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
@@ -27,10 +27,10 @@ namespace System.ServiceModel.Channels
         private bool _traceOpenAndClose;
         private object _eventSender;
         private CommunicationState _state;
-        private bool _onOpenAsyncCalled;
-        private bool _onCloseAsyncCalled;
-        private bool _isSynchronousOpen;
-        private bool _isSynchronousClose;
+        internal bool _isSynchronousOpen;
+        internal bool _isSynchronousClose;
+        private bool _supportsAsyncOpenClose;
+        private bool _supportsAsyncOpenCloseSet;
 
         protected CommunicationObject()
             : this(new object())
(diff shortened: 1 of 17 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs
```diff
@@ -798,12 +903,10 @@ internal protected virtual Task OnCloseAsync(TimeSpan timeout)
 
         internal protected virtual Task OnOpenAsync(TimeSpan timeout)
```
roncain: Reviewers: doing a code coverage run reveals this and the virtual OnCloseAsync here in the base class now appears to be dead code.  This is consistent with all WCF product code overriding this method, plus the fact the way we call external types doesn't use the code path that gets here.  I'm contemplating a hard assert here that we should never get here, and then remove the body of these methods.  Thoughts?

roncain:
Ping... any comments? From my perspective, this PR is complete.

/cc: @Zhenlan @iamjasonp @mconnew

iamjasonp:
In the cursory review I've done, I don't see anything that pops out... will take a closer look as time permits

roncain:
**Ping** -- any comments?  I rebased and squashed.  No changes have been made for the past 10 days.

Reminder: without this fix, the custom channel layer is broken.  I have tested manually against NET Native, and of course all tests pass, including the ones added by this PR.

/cc: @zhenlan @mconnew @hongdai @StephenBonikowsky @KKhurin

mconnew:
:shipit:

roncain:
Thanks @mconnew

PR #1572, opened 2016-09-28
Fix issue with S.Reflection.Extensions being a fat pkg.

StephenBonikowsky:
* See @ericstj commit for details...
https://github.com/dotnet/corefx/commit/3d7bdac4b02d42cbbf15d7ac164373c36d983a0f#diff-1cf61071aa9a8e298dd7972458ec9dcd

This should fix the issue we are seeing with PR #1540

### DIFF
--- a/src/Common/test-runtime/project.json
+++ b/src/Common/test-runtime/project.json
@@ -1,12 +1,9 @@
 {
   "dependencies": {
-    "Microsoft.NETCore.Platforms": "1.0.2-beta-24512-01",
-    "Microsoft.NETCore.Targets": "1.0.3-beta-24512-01",
+    "NETStandard.Library": "1.6.1-beta-24512-01",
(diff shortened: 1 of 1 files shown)


ericstj:
Why didn't you just make this change in the dotnet-bot PR?

StephenBonikowsky:
I was wanting to make sure it worked without all the noise from that PR and if merged today that PR would get updated tomorrow morning with the new builds. I can add it there and close it here if it makes more sense to do so.

ericstj:
You should do it there so that you can ensure it actually fixes the problem.

Also when doing this it's a good idea to manually run dotnet restore on the test-runtime project.json to confirm there are no warnings.

StephenBonikowsky:
Ok, will do.

StephenBonikowsky:
I got @roncain to edit the file with this change directly in PR #1540.

Issue #1574, opened 2016-09-30
Change to use of X509Certificate2 copy constructor


iamjasonp:
Currently, in various areas in the WCF codebase, we have code like the following to clone an X509Certificate2

``` csharp
    new X509Certificate2(certificate.Handle)
```

In Linux only - the private key does not get copied along. 

Additionally, the copy constructor is not yet part of netstandard1.1. The copy constructor is slated to be added in netstandard2 (per dotnet/corefx#12214, https://github.com/dotnet/corefx/pull/12240) 

In the meantime, we do some calisthenics to achieve the copy (and hence have scenarios listed in #1295 work) but we do not want to have this live in our codebase forever. This issue tracks: 
- [ ] Removal of the self-copy code we added (in the codebase, such instances are annotated with this bug number)
- [ ] Change code to use the X509Certificate2 copy constructor when it becomes available

PR #1575, opened 2016-10-01
Work around X509Certificate2 copy constructor omission 

iamjasonp:
In .NET Core, the X509Certificate2 copy constructor was missing; where WCF code
needed to clone a certificate, we replaced the references to the copy constructor
with code that uses the certificate.Handle instead

However, in Linux, OpenSSL does not provide the private key when using this handle,
meaning the private key goes missing in all subsequent copies; as such, client cert
authentication doesn't work.

This is a stopgap solution only, as we await the inclusion of the X509Certificate2
copy ctor in a future .NET Core version

Relates to #1574

### DIFF
--- a/src/System.Private.ServiceModel/src/Extensions/X509Certificate2Extensions.cs
+++ b/src/System.Private.ServiceModel/src/Extensions/X509Certificate2Extensions.cs
@@ -1,0 +1,40 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Security.Cryptography.X509Certificates;
+
+namespace System.ServiceModel
+{
+    internal static class X509Certificate2Extensions
+    {
+        // dotnet/wcf#1574 - This is a workaround for dotnet/corefx#12214
+        // The copy constructor was removed in .NET Core, this is the workaround to allow us to make copies of certificates
+
+        internal static X509Certificate2 CloneCertificateInternal(this X509Certificate2 certificateToClone)
+        {
+#if TARGETS_WINDOWS
+            return new X509Certificate2(certificateToClone.Handle);
+#else
+            X509Certificate2Collection collection = new X509Certificate2Collection(certificateToClone);
+            X509Certificate2Collection copyCollection = collection.Find(X509FindType.FindByThumbprint, certificateToClone.Thumbprint, false);
+            return copyCollection[0];
+#endif
+        }
+
+        // dotnet/wcf#1574 - This is a workaround for dotnet/corefx#12214
+        internal static X509Certificate2 CloneCertificateInternal(this X509Certificate certificateToClone)
+        {
+#if TARGETS_WINDOWS
+            return new X509Certificate2(certificateToClone.Handle);
+#else
+            byte[] certificateBytes = certificateToClone.Export(X509ContentType.Pfx, string.Empty);
+            X509Certificate2 certificateToClone2 = new X509Certificate2(certificateBytes, string.Empty);
+
+            X509Certificate2Collection collection = new X509Certificate2Collection(certificateToClone2);
+            X509Certificate2Collection copyCollection = collection.Find(X509FindType.FindByThumbprint, certificateToClone2.Thumbprint, false);
+            return copyCollection[0];
+#endif
+        }
+    }
+}
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.4.1.0.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.4.1.0.cs
@@ -172,7 +172,7 @@ public partial class HttpsTests : ConditionalWcfTest
                nameof(Client_Certificate_Installed),
                nameof(SSL_Available))]
     [Issue(959, Framework = FrameworkID.NetNative)] // Server certificate validation not supported in NET Native
-    [Issue(1295, OS = OSID.AnyUnix)] // A libcurl built with OpenSSL is required
+    [Issue(1295, OS = OSID.AnyCentOS | OSID.AnyFedora | OSID.AnyOpenSUSE | OSID.AnyOSX | OSID.AnyRHEL)] // A libcurl built with OpenSSL is required
     [OuterLoop]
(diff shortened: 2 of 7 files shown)


iamjasonp:
Reviewers: I had earlier talked about including some test that would signify the behaviour correction. Unfortunately, it seems like the `X509Certificate2(IntPtr handle)` won't ever provide us the private key in Linux because of OpenSSL restrictions on what handle comes back. 

Anyone else have a novel idea for what we could use as a sentinel test - without using reflection?

iamjasonp:
@dotnet-bot test OuterLoop CentOS7.1

iamjasonp:
@dotnet-bot test Outerloop Ubuntu16.04

src/System.Private.ServiceModel/src/Extensions/X509Certificate2Extensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        // dotnet/wcf#1574 - This is a workaround for dotnet/corefx#12214
+        // The copy constructor was removed in .NET Core, this is the workaround to allow us to make copies of certificates
+
+        internal static X509Certificate2 CloneCertificateInternal(this X509Certificate2 certificateToClone)
+        {
+#if TARGETS_WINDOWS
+            return new X509Certificate2(certificateToClone.Handle);
+#else
+            X509Certificate2Collection collection = new X509Certificate2Collection(certificateToClone);
```
roncain: Will the collection get GC'ed, or does the handle hold it memory? Should we remove it from the collection after finding it? I don't know the lifetime rules for collections.

src/System.Private.ServiceModel/src/Extensions/X509Certificate2Extensions.cs
```diff
@@ -0,0 +1,40 @@
...
+            X509Certificate2Collection copyCollection = collection.Find(X509FindType.FindByThumbprint, certificateToClone.Thumbprint, false);
+            return copyCollection[0];
+#endif
+        }
+
+        // dotnet/wcf#1574 - This is a workaround for dotnet/corefx#12214
+        internal static X509Certificate2 CloneCertificateInternal(this X509Certificate certificateToClone)
```
roncain: Nit -- maybe a comment why this clone is different than the one above it with the same name, due to different arg type.

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.4.1.0.cs
```diff
@@ -172,7 +172,7 @@ public static void SameBinding_Soap12_EchoString()
                nameof(Client_Certificate_Installed),
                nameof(SSL_Available))]
     [Issue(959, Framework = FrameworkID.NetNative)] // Server certificate validation not supported in NET Native
-    [Issue(1295, OS = OSID.AnyUnix)] // A libcurl built with OpenSSL is required
+    [Issue(1295, OS = OSID.AnyCentOS | OSID.AnyFedora | OSID.AnyOpenSUSE | OSID.AnyOSX | OSID.AnyRHEL)] // A libcurl built with OpenSSL is required
```
roncain: Just curious -- so all *nixes still fail except Ubuntu? So this fix is primarily to get Ubuntu working?

src/System.Private.ServiceModel/src/Extensions/X509Certificate2Extensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+#if TARGETS_WINDOWS
+            return new X509Certificate2(certificateToClone.Handle);
+#else
+            byte[] certificateBytes = certificateToClone.Export(X509ContentType.Pfx, string.Empty);
+            X509Certificate2 certificateToClone2 = new X509Certificate2(certificateBytes, string.Empty);
+
+            X509Certificate2Collection collection = new X509Certificate2Collection(certificateToClone2);
+            X509Certificate2Collection copyCollection = collection.Find(X509FindType.FindByThumbprint, certificateToClone2.Thumbprint, false);
+            return copyCollection[0];
```
roncain: Presumably Find can throw. Should we catch and throw something better?

iamjasonp: This find should never throw, because we're adding the cert and taking it straight back out. If there's an issue here then we might want to bubble it up... I did think about wrapping it in a nicer exception but I'm not sure what that'd be... I couldn't think of a builtin exception that works best

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Https/HttpsTests.4.1.0.cs
iamjasonp: I haven't tested this against the other *nixes, but given that these are http tests, I don't expect other *nixes to work based on the _curl+openSSL_ dependency.

src/System.Private.ServiceModel/src/Extensions/X509Certificate2Extensions.cs
iamjasonp: @bartonjs, any thoughts on this?

iamjasonp:
@dotnet-bot test OuterLoop Ubuntu16.04

iamjasonp:
@dotnet-bot test OuterLoop Ubuntu16.04

iamjasonp:
@roncain said: 
> Sure wish CI Ubuntu runs would pass,

Yep, they pass now :) The issues we saw were transient machine issues, but I have the Ubuntu 14 and 16 test suites passing properly

src/System.Private.ServiceModel/src/Extensions/X509Certificate2Extensions.cs
bartonjs: The certificate has no reference to any collection, so it won't keep the collection alive.  The collection does have a reference to the cert and thus would keep it alive... but the collection itself should get cleaned up very quickly.

bartonjs: > Just to be sure -- the collection does not dispose the certs it owns when it gets GC'ed, right?

Correct.  The collection is just a pre-generics typed collection; no finalizer, no dispose.  Just a longer-named List\<X509Certificate2\> minus some benefits of list and plus the Find method.

PR #1583, opened 2016-10-05
Update to buildtools 00820-01

ericstj:
This will redistribute refs if the API version doesn't change, as such
we need to update the index, and update test projects to pick up the
package build (with the older ref version).

/cc @StephenBonikowsky @zhenlan @mconnew


roncain:
It looks like tests are passing, but then an error occurs involving PackageIndex. @ericstj do you know the fix? Do we need to run with /t:UpdatePackageIndex?

ericstj:
Sorry, yeah I'll fix that.  It's happening because the new buildtools removed some of the refs so these no longer appear in the index.

ericstj:
Actually the problem was that the last stable packages weren't being restored (at least not all of them) so the harvesting wasn't happening on the build machine.  I've fixed this by ensuring the repo restores these.

StephenBonikowsky:
I think we will also need to update ScenarioTest.Common.csproj to use pkgproj references.

ericstj:
Looks like you are right.  I didn't see these failures when building locally, I'll see what I missed to configure build for these projects.

roncain:
LGTM
/cc: @StephenBonikowsky

StephenBonikowsky:
@ericstj I only just saw that we hadn't resolved this PR, is this something we still want/need in release/1.1.0?

ericstj:
It would be good to get WCF on the same build of buildtools as corefx, that way you have a way to update buildtools in your release branch.

StephenBonikowsky:
Could you clarify a couple things for us?

I don't understand 
> that way you have a way to update buildtools in your release branch
This is the release branch and the way to update it is just like this... right?

Also, since your last update of the buildtools version in the release/1.1.0 branch in CoreFx it has since been updated again to support testing some new Distros, so do we update this PR to this latest version?

ericstj:
>>that way you have a way to update buildtools in your release branch

>This is the release branch and the way to update it is just like this... right?

We forked buildtools for release/1.1.0.  Currently you're not on tip of that fork, so if you wanted to take a targeted fix to buildtools and not get any of the other fixes in 1.1.0 we'd have to fork again.

>Also, since your last update of the buildtools version in the release/1.1.0 branch in CoreFx it has since been updated again to support testing some new Distros, so do we update this PR to this latest version?

Yeah, I think that's a good idea.  /cc @mellinoe   I also needed to make another fix to the 1.1.0 fork to handle the stable-version transition: https://github.com/dotnet/buildtools/commit/39fbab964eb6fc41c733180457828ff9c51ed018

StephenBonikowsky:
Ah, that makes sense.
Do you want to update this PR then or do you want me to clone it, update and submit a new PR?

ericstj:
I've updated it.

StephenBonikowsky:

StephenBonikowsky:
@dotnet-bot test innerloop CentOS7.1

ericstj:
CentOS runs are failing due to https://github.com/dotnet/corefx/issues/13395

StephenBonikowsky:
Not sure what is wrong with the CentOS machines, I think we are ok to Merge.
LGTM

Ok with you @ericstj ?

StephenBonikowsky:
Are we ok to merge or do you want to wait until the CentOS machines get re-spun?

ericstj:

ericstj:
@dotnet-bot test innerloop CentOS7.1

ericstj:
I deleted the machines, we can see if it fixes it.

ericstj:
Appears to have fixed it.  It's progressed past restore.

ericstj:
This should be OK to merge, assuming you have whatever approval is required for 1.1.

StephenBonikowsky:
Not sure what approval is needed at this point for infrastructure changes, I just sent a normal approval request with the info.

StephenBonikowsky:
Have approval, waiting until day after we release to merge.

StephenBonikowsky:
1.1 has released, merging.

PR #1590, opened 2016-10-10
Fix the issue that we use WindowsIdentity.Current on *NIX

### LINKED ISSUE #1569: Workaround scenarios failed by S.Security.Principal.WindowsIdentity.GetCurrent
This is to find out a workaround for issue #946. The affected tests are:

NegotiateStream_Http_Tests.NegotiateStream_Http_With_ExplicitSpn [FAIL]
NegotiateStream_Http_Tests.NegotiateStream_Http_AmbientCredentials [FAIL]
Https_ClientCredentialTypeTests.WindowsAuthentication_RoundTrips_Echo [FAIL]

hongdai:
* We use try/catch to catch PlatformNotSupported exception
and use the same way to set default user on UWP.

Fix #1569

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
@@ -776,6 +776,8 @@ namespace System.ServiceModel.Security
 
         internal static class NetworkCredentialHelper
         {
+            static string s_currentUser = string.Empty;
+            const string DefaultCurrentUser = "____CURRENTUSER_NOT_AVAILABLE____";
             static internal bool IsNullOrEmpty(NetworkCredential credential)
             {
                 return credential == null ||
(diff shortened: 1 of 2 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
```diff
@@ -793,16 +794,33 @@ static internal bool IsDefault(NetworkCredential credential)
...
+                    //so returning a username which is very unlikely to be a real username;
+                    currentUser = "____CURRENTUSER_NOT_AVAILABLE____";
                 }
 #else
-                // There's no way to retrieve the current logged in user Id in UWP apps or in
-                // *NIX so returning a username which is very unlikely to be a real username;
-                return "____CURRENTUSER_NOT_AVAILABLE____";
-#endif
+                // There's no way to retrieve the current logged in user Id in UWP apps
+                currentUser = "____CURRENTUSER_NOT_AVAILABLE____";
```
roncain: >"____CURRENTUSER_NOT_AVAILABLE____" [](start = 30, length = 35)

Nit: this could be a private const in this class to avoid duplication.   Not blocking.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
```diff
@@ -776,6 +776,7 @@ internal static NetworkCredential GetNetworkCredentialOrDefault(NetworkCredentia
 
         internal static class NetworkCredentialHelper
         {
+            static string currentUser = string.Empty;
```
roncain: >currentUser [](start = 26, length = 11)

We should call this s_currentUser and declare it private to be consistent with recommended code formatting conventions.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
```diff
@@ -793,16 +794,33 @@ static internal bool IsDefault(NetworkCredential credential)
...
-                using (WindowsIdentity self = WindowsIdentity.GetCurrent())
+                if (!string.IsNullOrEmpty(currentUser))
                 {
-                    return self.User.Value;
+                    return currentUser;
+                }
+
+                // CurrentUser could be set muliple times
+                // This is fine because it does not affect the value returned.
+#if SUPPORTS_WINDOWSIDENTITY // NegotiateStream
```
hongdai: Agreed. I will remove it.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityUtils.cs
hongdai: Agreed.

Issue #1426, opened 2016-07-29
Add API T:System.ServiceModel.PeerHopCountAttribute


zhenlan:
The API is not used internally.

roncain:
Tasks are:
- [ ] Add implementation
- [ ] Add to contract
- [ ] Add unit tests
- [ ] Add scenario tests (may require adding add'l use of implementation throughout product)

roncain:
[Design Discussion]
We should discuss whether exposing this in the public API is necessary.  In the full framework, [PeerHopCount] could really appear in any [MessageContract] class to mark which fields are to be treated as the hop count.  But it is really intended just for peer-to-peer programming, and I can't find any special handling of it in WCF product code.  It is possible some [MessageContract] type containing this attribute might be used on the server, possibly leading to a compilation error if it is propagated to the client via Connected Services.  But I think that usage scenario is likely very rare and recommend we don't add this API until we know it is needed.

zhenlan:
Design discussion agrees to not add these APIs at this point. We will wait for customer feedback to see if there is a need. As a workaround, user can just use [MessageHeader] with explicitly specified XML namespace and name.

roncain:
Closing as agreed

PR #1596, opened 2016-10-17
Add PeerHopCountAttribute

roncain:
Adds the implementation of PeerHopCountAttribute.
Later PR's will add to public contract and add tests.
This is the first part of issue #1426.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/PeerHopCountAttribute.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/PeerHopCountAttribute.cs
@@ -1,0 +1,33 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Net.Security;
+using System.ServiceModel.Channels;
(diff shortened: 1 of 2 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/PeerHopCountAttribute.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public PeerHopCountAttribute()
+            : base()
+        {
+            base.Name = PeerStrings.HopCountElementName;
+            base.Namespace = PeerStrings.HopCountElementNamespace;
+            base.ProtectionLevel = ProtectionLevel.None;
+            base.MustUnderstand = false;
+        }
+        public new bool MustUnderstand
```
mconnew: These can use the new delegate syntactic sugar. e.g.:
```csharp
    public new bool MustUnderstand => base.MustUnderstand;
```
I've been using it lately and it really helps to make the code look cleaner, and the compiler emits identical instructions.

roncain: Will do.

Issue #1416, opened 2016-07-29
Add API T:System.ServiceModel.Channels.ChannelPoolSettings


zhenlan:

roncain:
Tasks are:
- [ ] Add implementation [in progress]
- [ ] Add to contract
- [ ] Add unit tests
- [ ] Add scenario tests (may require adding add'l use of implementation throughout product)

StephenBonikowsky:
Need to finish this next Monday. Going to complete it for this milestone.

PR #1597, opened 2016-10-17
Add ChannelPoolSettings

roncain:
First step of issue #1416 
Later PR's will add to contract and unit test


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ChannelPoolSettings.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            if (_maxOutboundChannelsPerEndpoint != channelPoolSettings._maxOutboundChannelsPerEndpoint)
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        internal bool InternalShouldSerialize()
```
mconnew: This method is only used when creating the config from a binding.

Issue #1417, opened 2016-07-29
Add API T:System.ServiceModel.Channels.ITransportTokenAssertionProvider


zhenlan:

roncain:
Tasks are:
- [ ] Add implementation
- [ ] Add to contract
- [ ] Add unit tests
- [ ] Add scenario tests (may require adding add'l use of implementation throughout product)

roncain:
[Design Discussion]
This API appears to be used only in exporting policy via MetadataExporter.  Each TransportBindingElement has the option to implement this to add an XmlElement to the policy.  This interface is currently not in the product, and consequently none of the TransportBindingElements currently implement it, as they do in the full framework:

Recommendation: do *not* add this API unless @mconnew sees utility in it for future development.

/cc: @zhenlan @mconnew @StephenBonikowsky @hongdai

mconnew:
I agree not to bring this API in. There is no client side utility for this.

zhenlan:
Design discussion agrees not to add this API. Please close this issue.

roncain:
Closing as agreed.  Also closed the PR that added this interface.

PR #1599, opened 2016-10-19
Add ITransportTokenAssertionProvider

roncain:
First step in issue #1417.  Other PR's will add this to the
public API, add tests, etc.


roncain:
Reviewers: please refer to the design discussion in #1417 to decide whether we want this PR to be merged or not.  I believe it is used only in building the export policy for MetadataExporter, and we probably don't want to add this to the API or even implement it, because nothing will consume it.  

Unless someone has a good argument to add it, I propose discarding this PR.

roncain:
Closing this PR because we've decided not to add support for it or add it to the public API.

PR #1610, opened 2016-10-24
Port to release/1.1.0 custom channel fix

### LINKED ISSUE #1544: CommunicationObject.OnOpen and OnBeginOpen is never called

roncain:
Cherry-pick of: 96e0c4d137ff583eda836c2448b737bd54a8fb73

Original commit comments:
Fix OnOpen/OnBeginOpen issues with custom channels

Fixes the product issue by propagating knowledge of
sync and async open/close requests down the channel
stack so they stay sync or async based on the original
call.

Created mocks and unit tests for custom channels

Fixes #1544


roncain:
Merging with permission from @Petermarcu and @zhenlan .

PR #1611, opened 2016-10-24
Fix test services for addding service reference

hongdai:
* We would like to leverge the same test server for WCF connected service
testing.
* I have verified the updated test service can be used for adding service references, both http and net.tcp
bindings.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Contract/Message/MessageContractCommon.4.1.0.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Contract/Message/MessageContractCommon.4.1.0.cs
@@ -15,7 +15,6 @@ namespace MessageContractCommon
     public class MessageContractConstants
     {
         // CONSTANTS
-        public const string wrapperName = "CustomWrapperName";
         public const string wrapperNamespace = "http://www.contoso.com";
         public const string dateElementName = "Date_of_Request";
         public static string dateElementValue = "";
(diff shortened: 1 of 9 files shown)


hongdai:
Ready for review.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Message/MessageContractCommon.4.1.0.cs
```diff
@@ -15,7 +15,6 @@ namespace MessageContractCommon
     public class MessageContractConstants
     {
         // CONSTANTS
-        public const string wrapperName = "CustomWrapperName";
```
roncain: Nit: could you have just changed the constant here instead of removing it?  The rest of the tests use these constants to avoid duplicating strings inside the tests. Not blocking.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Message/MessageContractCommon.4.1.0.cs
```diff
@@ -78,7 +76,7 @@ public interface IMessageContract
 
     public class MessageContractTypes
     {
-        [MessageContract(IsWrapped = true, WrapperName = "CustomWrapperName", WrapperNamespace = "http://www.contoso.com")]
```
roncain: [No change required -- observation only] It's too bad the original test didn't use the same constants in MessageContractConstants for these attribute named parameters.  It would have eliminated so much string duplication scattered in the contracts and tests.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Message/MessageContractCommon.4.1.0.cs
hongdai: I'd prefer not to have a const defined in the common code for wrapper name as it could be different per test, although it happens to be the same for the existing 3 tests.

PR #1617, opened 2016-10-27
Updating all tests to run using netcoreapp1.1 test TFM.

StephenBonikowsky:
* We are going to have to eventually, next step is to move product packages to NetStandard1.7


StephenBonikowsky:
@joperezr this was pretty simple, just updating the test project.json files. Look ok to you?

PR #1621, opened 2016-10-28
Enabling tests to run against netcore50 test tfm

StephenBonikowsky:
* WCF currently supports two TestTFMs, they need to be explicitly listed for test projects now so that the 'FilterToTestTFM' property used in Helix runs works as expected. This "may" fix the netcore50 runs in Helix.


StephenBonikowsky:
@joperezr this works as expected for me locally. Hopefully it also fixes the netcore50 issue in VSO.

joperezr:
if you want to repro running the tests locally, you should be able to run the tests by manually calling:
```cmd
msbuild src\tests.builds /p:FilterToTestTFM=netcore50 /p:TestNugetRuntimeId=win10-x64
```

PR #1624, opened 2016-11-02
Remove dead code

roncain:
Remove types not in public contract that are not used.
This is part of a cleanup pass for code coverage analysis.


roncain:
Reviewers: I'm analyzing code coverage pending adding more test coverage.  The code removed in this PR was not used (and not in any public API).

zhenlan:
@roncain can you check the usage data of these APIs? They may have high usage in Framework/Xamarin but not flagged in previous usage analysis because they are already in implementation.

roncain:
Good point @zhenlan -- I should verify usage before proceeding.  Labelled **PR DO NOT MERGE** to put this on the back burner while other higher priority tasks are handled first.

roncain:
I'm going to close this PR and not merge it. Some of the dead code could be reached by API's we have not yet exposed.  The cost of understanding it sufficiently does not warrant doing this work now.

PR #1641, opened 2016-11-04
Rename test properties and members to remove "Bridge"

roncain:
This is the 2nd of 2 steps to remove the old infrastructure.
The first step deleted all old infrastructure code.

This follow-up PR just renames variables, constants, member names,
etc. to remove "Bridge" from any names.


StephenBonikowsky:
The one failure is a buildtools restore failure that seems to happen fairly often.
I think it is good to merge.
Don't want to reset that build due to the log CI queues we have been suffering from today.

PR #1642, opened 2016-11-04
Updating the WCF package version to latest.

StephenBonikowsky:
* New APIs were just added, updating to the package version just built with these changes so Helix test projects can build.


StephenBonikowsky:
Only WCF package versions were updated in the project.json files. Without these changes our scheduled Helix runs over the weekend will fail due to the new APIs that were added earlier today by @roncain.

PR #1651, opened 2016-11-09
Fix Duplex Channel shaped tests using http.

StephenBonikowsky:
* One test was using the wrong binding, both needed seperate endpoints on the server configured to receive incoming messages using websockets.

### DIFF
--- a/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/NetHttpTestServiceHost.cs
+++ b/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/NetHttpTestServiceHost.cs
@@ -32,4 +32,30 @@ namespace WcfService
         {
         }
     }
+
+    public class NetHttpTestServiceHostUsingWebSocketsFactory : ServiceHostFactory
+    {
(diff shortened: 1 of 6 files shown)


src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/NetHttpTestServiceHost.cs
```diff
@@ -32,4 +32,30 @@ public NetHttpTestServiceHost(Type serviceType, params Uri[] baseAddresses)
...
+    }
+
+    public class NetHttpTestServiceHostUsingWebSockets : TestServiceHostBase<IWcfService>
+    {
+        protected override string Address { get { return "NetHttpWebSockets"; } }
+
+        protected override Binding GetBinding()
+        {
+            var binding = new NetHttpBinding();
+            binding.WebSocketSettings.TransportUsage = WebSocketTransportUsage.Always;
```
StephenBonikowsky: Because the test uses IDuplexSessionChannel the incoming message will be using WebSockets, and because the interface on the server side is not attributed with "CallbackContract" the binding must explicitly set the WebSocketTransportUsage to Always, otherwise the call fails.

PR #1663, opened 2016-11-17
WCF 1.1 Stress Tests

KKhurin:
Refactored tests to support more stress scenarios
Added numerous command line parameters to control both perf and stress tests
Added basic stress results reporting
Tuned perf scenario tests to achieve stable results
Addressed some of the previous PR feedback
Added various security bindings

### DIFF
--- a/src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
+++ b/src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
@@ -13,26 +13,33 @@ namespace SharedPoolsOfWCFObjects
     /// <summary>
     /// This class provides a few Call*AndRecordStats() utility methods which will store and aggregate the timing of the delegates they invoke 
     /// 
-    /// Note: all Call*AndRecordStats() methods catch all exceptions coming from the delegates
-    /// However we don't want to miss the case when an unexpected exception happens for the calls that should never have exceptions
+    /// Note that there are 2 separate mechanisms to control how the exceptions are handled by these utility methods.
+    /// The first one is the boolean "hideExceptions" parameter to all Call*AndRecordStats methods where 'true' means completely ignore exceptions.
+    /// This is used by the scenarios where the exceptions are expected and should be ignored no matter what kind of test is running.
+    /// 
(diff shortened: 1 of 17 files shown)


src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/CallStats.cs
```diff
@@ -13,26 +13,33 @@ namespace SharedPoolsOfWCFObjects
...
+            : this(samples, errorSamples, exceptionHandler, CallTimingStatsCollector.CalculateSomePercentiles, CallTimingStatsCollector.CalculateSomePercentiles)
         {
         }
 
-        public CallStats(int samples, int errorSamples, Func<long[], TimingPercentileStats> statsCalculator, Func<long[], TimingPercentileStats> errorStatsCalculator)
+        public CallStats(int samples, int errorSamples, Func<Exception, bool> exceptionHandler, Func<long[], TimingPercentileStats> statsCalculator, Func<long[], TimingPercentileStats> errorStatsCalculator)
         {
             SunnyDay = new CallTimingStatsCollector(samples, statsCalculator);
             RainyDay = new CallTimingStatsCollector(errorSamples, errorStatsCalculator);
+            _exceptionHandler = exceptionHandler;
```
roncain: >exceptionHandler [](start = 32, length = 16)

Nit: not required to merge this PR -- if a null exceptionHandler is passed in, we will get NullReferenceException trying to use it in the tests below.  I know your tests always pass it in, but it would be a good idea to Contract.Assert() it is not null (if built Debug) or throw an ArgumentException (if Retail).  Alternatively, if a null is passed in, silently set _exceptionHandler to a func that returns true.  Recommended only for future consideration -- don't make a new PR just for this.

src/System.Private.ServiceModel/tests/Stress/SharedPoolsOfWCFObjects/IService1.cs
```diff
@@ -5,6 +5,7 @@
 using System.IO;
 using System.ServiceModel;
 using System.Runtime.Serialization;
+using System;
```
roncain: >using System [](start = 0, length = 12)

Nit: we usually put these in sorted order.  Don't make a new PR just for this.  Consider for future only.

PR #1672, opened 2016-11-21
Fix UWP F5 runs.

StephenBonikowsky:
* Need to add AppXRunnerVersion to dependencies.props file.


hongdai:
LGTM, provided CI passes.

PR #1682, opened 2016-11-29
Update docs to explain how to run tests as UWP

### LINKED ISSUE #1670: Get back ability to run UWP tests locally
With recent updates, we're no longer able to run our tests in UWP locally with the existing scripts (through they run correctly in Helix).

Task is to find out why, fix it, and update any .md file where we describe how to run in UWP

roncain:
Due to changes in the build tools, the instructions for how to
run unit and scenario tests as UWP changed slightly.

Fixes #1670 

skip ci please

### DIFF
--- a/Documentation/developer-guide.md
+++ b/Documentation/developer-guide.md
@@ -114,7 +114,7 @@ To run a unit test as UWP, you need to 'cd' to the folder containing the test pr
 ```
 cd src\System.ServiceModel.Primitives\tests
 
-msbuild  /p:TestTFM=netcore50 /p:TestNugetRuntimeId=win10-x64-aot /t:RebuildAndTest /p:UseDotNetNativeToolchain=true /p:TestWithCore=true
+msbuild /t:BuildAndTest /p:TargetGroup=netstandard1.3 /p:TestTFM=netcore50aot /p:TestNugetRuntimeId=win10-x64-aot /p:UseDotNetNativeToolchain=true
 ```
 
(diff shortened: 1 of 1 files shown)


Documentation/developer-guide.md
```diff
@@ -114,7 +114,7 @@ To run a unit test as UWP, you need to 'cd' to the folder containing the test pr
 ```
 cd src\System.ServiceModel.Primitives\tests
 
-msbuild  /p:TestTFM=netcore50 /p:TestNugetRuntimeId=win10-x64-aot /t:RebuildAndTest /p:UseDotNetNativeToolchain=true /p:TestWithCore=true
+msbuild /t:BuildAndTest /p:TargetGroup=netstandard1.3 /p:TestTFM=netcore50aot /p:TestNugetRuntimeId=win10-x64-aot /p:UseDotNetNativeToolchain=true
```
roncain: Reviewers: I found out that setting TargetGroup to netstandard 1.3 like this was necessary when we use msbuild explicitly.  When this is not set, it appears some packages for some test projects come from netstandard1.3 and others from netstandard1.4, even though they are using the identical project.json.  Apparently the build.cmd and build-managed.cmd took care of that somehow.  Packages coming from random versions of netstandard caused errors at build time.  This is not strictly a UWP issue -- any use of msbuild explicitly can cause this.

StephenBonikowsky:
lgtm

PR #1700, opened 2016-12-12
Remove the stringbuilder and do some cleanup in DataContractResolverTest

huanwu:
Remove the stringbuilder
Replace the "catch" with a Finally only to do some channel cleanup.


StephenBonikowsky:
LGTM

Issue #1701, opened 2016-12-12
Workaround bug introduced in XmlReader.Dispose


StephenBonikowsky:
System.Xml was ported from the full framework containing a bug that exists there as well.
@mconnew will be opening an issue in CoreFx to get this fixed.

The Dispose method for the EncodingStreamWrapper should know whether it's inner stream is meant for reading or writing and shouldn't call Flush on a read stream. (This is a CoreFx problem)

On the WCF side of things, we can just make BufferedReadStream.Flush and FlushAsync no-ops instead of throwing, this is the immediate solution for WCF.

CoreFx Issue: dotnet/corefx#14454

zhenlan:
Do we understand why this was not a problem for WCF full framework?

mconnew:
Yes, on full framework we use a BufferedStream which can buffer reads and writes on a stream. This class doesn't/didn't exist on core so I created BufferedReadStream. We only need to read so I was able to make the class slightly more efficient by making it unable to write. I made Flush and FlushAsync throw as there's never a need to Flush a reading stream and it's better to throw from a method you expect to never be called than to do nothing as you can hide bugs otherwise. As BufferedStream can be used to buffer writes, it doesn't throw even though in our usage it would be a no-op. A recent change to XmlReader now causes Close to be called on EncodingStreamWrapper which then called flush on a reading only stream.

zhenlan:
We will not revert the code change @StephenBonikowsky made. Let's just clean up the comments.

PR #1702, opened 2016-12-13
Remove throws from Flush methods in BufferedReadStream.

StephenBonikowsky:
* Workaround for bug introduced by corefx, see dotnet/corefx #14454
* Fixes Issue #1701


StephenBonikowsky:
Not sure if we wanted to put the throws back in once the issue is fixed from the corefx side so added comments and just commented out the throws.

PR #1705, opened 2016-12-14
Fixing and updating test to use best practices.

### LINKED ISSUE #398: Fix ServiceRestart_Throws_CommunicationException test
The ServiceRestart_Throws_CommunicationException test is incorrect in several ways and should be corrected.  Main observable issues are:
1. It fails running multi-machine (either Windows or Linux)
2. It reports only the unexpected exception message and does not include the stack trace.

Problems to fix are:
1. It uses the HttpBaseAddress_Basic endpoint, which means it is talking to a ServiceHost used for other tests.  It should have a dedicated IResource endpoint on the Bridge that exists for no other purpose but this test.
2. It does some string replacement of the original URI thinking it contains "[HOST]" which is an incorrect assumption.  I'm not sure what the test is attempting to do but it should not need to alter the URL to the endpoint dedicated to a service restart
3. It uses the old style exception catching and reporting (rather than Assert.Throws()) and fails to capture the stack trace.

Note that this test should also technically inform the ResourceController it no longer has the resource for this endpoint before it closes the service host, otherwise the ResourceController will believe it still exists and continue to try to use a non-existent service host.

StephenBonikowsky:
* Removed usage of StringBuilder.
* Added comments to better explain what the test does.
* Fixes #398

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.4.1.0.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/ExpectedExceptionTests.4.1.0.cs
@@ -53,74 +53,68 @@ public partial class ExpectedExceptionTests : ConditionalWcfTest
 
     [WcfFact]
     [OuterLoop]
-    [Issue(398)]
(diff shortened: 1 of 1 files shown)


hongdai:
LGTM.

StephenBonikowsky:

PR #1735, opened 2017-01-09
Fix bug with sending Message with XmlElement MessageHeader

### LINKED ISSUE #702: SOAP Headers not working in NetCore

shmao:
XmlElementMessageHeader.OnWriteHeaderAttributes threw PNSE because one required API was not available on NetCore. Adding the implementation for the method as the missing API is available now.

Fix #702.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/OperationFormatter.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/OperationFormatter.cs
@@ -877,7 +877,14 @@ namespace System.ServiceModel.Dispatcher
 
             protected override void OnWriteHeaderAttributes(XmlDictionaryWriter writer, MessageVersion messageVersion)
             {
+#if NETStandard13
                 throw ExceptionHelper.PlatformNotSupported();
+#else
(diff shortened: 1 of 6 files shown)


hongdai:
LGTM.

StephenBonikowsky:
LGTM

shmao:
@dotnet-bot test OuterLoop OSX Debug, OuterLoop OSX Release

PR #1744, opened 2017-01-12
Adding ActiveIssue to ServiceRestart_Throws_CommunicationException test

StephenBonikowsky:
* This test failure is due to an underlying problem in WinRTHttpClient behavior.
* Until that gets fixed we can skip this test on netnative runs.
* Amended this commit to also add an ActiveIssue for test case 'Message_With_MessageHeaders_RoundTrips'.


StephenBonikowsky:
@zhenlan if you see the last check pass could you merge? It would be great for our nightly run to be clean.

Issue #1504, opened 2016-09-02
Investigate port exhaustion issue on Net.Tcp in stress


KKhurin:
When running stress against IIS over Net.Tcp transport we see port exhaustion. So far it does not look like any of the known issues. Need more investigation to figure out the cause.

zhenlan:
This still repros often.

@mconnew suggested to set a break point at socket closing for further investigation.

zhenlan:
@KKhurin made fix #1752 for this issue.

PR #1752, opened 2017-01-24
Fixing #1504 leaking connections in nettcp streaming.

KKhurin:
This is a fix for per request leak of nettcp connections in streaming scenario.


KKhurin:
cc: @mconnew @zhenlan

PR #1768, opened 2017-02-03
[WIP] Some initial build of S.P.ServiceModel against NS2.0 package

weshaggard:
@zhenlan 

Here is a starting point for getting S.P.ServiceModel building against the NS2.0 package. There are a number of hacking things that could be normalized but this should get things started. Also current the tests aren't passing so we would need to looking into that as well.

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -9,6 +9,7 @@
   <!-- Auto-upgraded properties for each build info dependency. -->
   <PropertyGroup>
     <CoreFxExpectedPrerelease>beta-24911-02</CoreFxExpectedPrerelease>
+    <StandardExpectedPrerelease>beta-25003-02</StandardExpectedPrerelease>
   </PropertyGroup>
 
   <!-- Full package version strings that are used in other parts of the build. -->
@@ -31,6 +32,7 @@
     <RemoteDependencyBuildInfo Include="CoreFx">
       <BuildInfoPath>$(BaseDotNetBuildInfo)corefx/$(DependencyBranch)</BuildInfoPath>
       <CurrentRef>$(CoreFxCurrentRef)</CurrentRef>
+      <DisabledPackages>NETStandard.Library</DisabledPackages>
     </RemoteDependencyBuildInfo>
     <RemoteDependencyBuildInfo Include="CoreClr">
       <BuildInfoPath>$(BaseDotNetBuildInfo)coreclr/$(DependencyBranch)</BuildInfoPath>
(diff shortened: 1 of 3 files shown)


zhenlan:
Thank you @weshaggard!

@StephenBonikowsky can you see if you can update our tests to pick this up?

StephenBonikowsky:
Will do!

StephenBonikowsky:
When I start fixing the errors it just leads to more and more. Between our old buildtools version and how behind we are compared to CoreFx with all the various scripts I think trying to get our tests working with this PR may not be worth the time.

weshaggard:
@StephenBonikowsky what sort of issues are you seeing?

StephenBonikowsky:
The initial error that we see here is...
`System.IO.FileLoadException : Could not load file or assembly 'System.Private.ServiceModel, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The located assembly's manifest definition does not match the assembly reference.`

So I fixed https://github.com/dotnet/wcf/blob/master/src/Common/test-runtime/project.json, adding NETStandard.Library and removing individual corefx packages then also removing uap10.0 dependency.

I would keep getting different errors as I tried running the Unit tests under our Facade locations, fixing those issues then involved updating the associated .builds files and the test projects.
https://github.com/dotnet/wcf/tree/master/src/System.ServiceModel.Http/tests

I didn't write down each error as I ran into it so I'm sorry I can't be more specific. It got to the point that I'm not sure it was worth making so many changes that wouldn't survive our move to the new engineering system in a few weeks.

weshaggard:
@StephenBonikowsky fair but I do think a number of things will survive the engineering system move. You guys are going to have some unique requirements above corefx. For example corefx doesn't build directly against the NETStandard.Library package nor does it restore and run against the NETCore.App package both of which I suspect you guys may end up doing.

I think it might be good to get together and talk about some of the changes that you guys will be doing with the engineering system switch and see how we can either start the transition or minimally patch stuff to enable library work to progress until we shift the engineering system.

StephenBonikowsky:
Ok, that sounds good to me. How about I reproduce some of the first changes I made, I'll scope them to separate commits and review them with you as I go.

StephenBonikowsky:
Working on this in PR #1779

weshaggard:
OK I'm closing this and you can continue on #1779. I will be OOF after today and not back until next tuesday and so I will have a look at more next week.

Issue #1693, opened 2016-12-08
Need to wrap calls to CoreClrSocketConnection.Abort() from a timer thread


KKhurin:
to avoid unhandled exceptions like this one:
~~~
!pe
Exception object: 000001650258d6e0
Exception type:   System.Net.Sockets.SocketException
Message:          An invalid argument was supplied
InnerException:   <none>
StackTrace (generated):
    SP               IP               Function
    0000001B6357EC80 00007FFE15BFD349 System_Net_Sockets!System.Net.Sockets.Socket.Shutdown(System.Net.Sockets.SocketShutdown)+0x279
    0000001B6357ECD0 00007FFE15C9B017 System_Private_ServiceModel!System.ServiceModel.Channels.CoreClrSocketConnection.Abort(Int32, System.String, TransferOperation)+0x177
    0000001B6357ED30 00007FFE7312D4AF System_Private_CoreLib_ni!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)+0x10f
    0000001B6357EDA0 00007FFE731F2B26 System_Private_CoreLib_ni!System.Threading.TimerQueueTimer.CallCallback()+0x186
    0000001B6357EE20 00007FFE731F2824 System_Private_CoreLib_ni!System.Threading.TimerQueueTimer.Fire()+0x94
    0000001B6357EE90 00007FFE731EFE44 System_Private_CoreLib_ni!System.Threading.ThreadPoolWorkQueue.Dispatch()+0x1a4
~~~

KKhurin:
This was discovered as part of #1676

PR #1770, opened 2017-02-07
Wrapping timeout callbacks to guard against unhandled exceptions …

KKhurin:
…on timer threads. This should fix the crash reported in #1693.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
@@ -92,13 +92,27 @@ namespace System.ServiceModel.Channels
 
         protected static void OnReceiveTimeout(SocketConnection socketConnection)
         {
-            socketConnection.Abort(SR.Format(SR.SocketAbortedReceiveTimedOut, socketConnection._receiveTimeout), TransferOperation.Read);
+            try
+            {
+                socketConnection.Abort(SR.Format(SR.SocketAbortedReceiveTimedOut, socketConnection._receiveTimeout), TransferOperation.Read);
+            }
+            catch (SocketException)
+            {
+                // Guard against unhandled SocketException in timer callbacks
+            }
         }
 
         protected static void OnSendTimeout(SocketConnection socketConnection)
         {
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
```diff
@@ -92,13 +92,27 @@ protected string RemoteEndpointAddressString
...
         protected static void OnReceiveTimeout(SocketConnection socketConnection)
         {
-            socketConnection.Abort(SR.Format(SR.SocketAbortedReceiveTimedOut, socketConnection._receiveTimeout), TransferOperation.Read);
+            try
+            {
+                socketConnection.Abort(SR.Format(SR.SocketAbortedReceiveTimedOut, socketConnection._receiveTimeout), TransferOperation.Read);
+            }
+            catch (SocketException)
+            {
+                // Guard against unhandled SocketException in timer callbacks
```
StephenBonikowsky: So you just swallow the exception?

KKhurin: any idea on what to do with this exception? 
The connection is gone so any attempt to use it will result in exceptions on the calling thread. There is a way to save the exception to (presumably) throw it later but this will eventually result in a much bigger change in the behavior.

KKhurin: any idea on what to do with this exception? 
The connection is gone so any attempt to use it will result in exceptions on the calling thread. There is a way to save the exception to (presumably) throw it later but this will eventually result in a much bigger change in the behavior.

KKhurin: any idea on what to do with this exception? 
The connection is gone so any attempt to use it will result in exceptions on the calling thread. There is a way to save the exception to (presumably) throw it later but this will eventually result in a much bigger change in the behavior.

StephenBonikowsky: Maybe just add a comment with what you just said?

StephenBonikowsky:
LGTM

PR #1778, opened 2017-02-15
Resource string cleanup

morganbr:
Automated cleanup of unused resources in System.Private.ServiceModel. This removes about 75% of the file or roughly 300kb from the binary. The automated tool scanned for source references to resource strings. Github's diffing is showing additions, but there are no actual additions or edits, just removals. Verification was build+unit tests.

### DIFF
--- a/src/System.Private.ServiceModel/src/Resources/Strings.resx
+++ b/src/System.Private.ServiceModel/src/Resources/Strings.resx
@@ -126,101 +126,14 @@
   <data name="RequiredAttributeMissing" xml:space="preserve">
     <value>Attribute '{0}' is required on element '{1}'.</value>
   </data>
-  <data name="UnsupportedCryptoAlgorithm" xml:space="preserve">
-    <value>Crypto algorithm {0} not supported in this context.</value>
-  </data>
-  <data name="CustomCryptoAlgorithmIsNotValidHashAlgorithm" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


joshfree:
nice!

/cc @zhenlan @shmao

morganbr:
cc @mconnew

zhenlan:
Thanks @morganbr! Did the automated tool scan against binary or the source code? If it's the binary, as @mconnew mentioned offline, the Windows/Linux/UWP builds many have different references to the resources. Just want to make sure we don't miss anything.

morganbr:
@zhenlan, the tooling ran against the source code in the System.Private.ServiceModel directory. It's just a regex, so it's not foolproof, but my hope is CI covered the various build flavors.

zhenlan:
CI does not cover UWP and our code coverage is only ~30% so I don't trust tests that much.

morganbr:
@Zhenlan, my base assumption here is that, like other CoreFX libraries, resources are always fetched via the SR class. If that's true, there should be a build failure if I removed any strings that were still needed. I did also do a local build as well and it looks like the netcore50 version of S.P.ServiceModel built correctly. Is that correct reasoning for System.Private.ServiceModel?

morganbr:
Thanks for approving, @Zhenlan. I don't have merge access on the WCF repo, so feel free to merge on my behalf.

PR #1814, opened 2017-03-21
Fixed timeout tracking for the tests that make multiple requests

KKhurin:
Fixed timeout tracking for the tests that make multiple requests per test call.
Added some useful logging and tracing for detected time outs.
Simplified tests and minor code cleanup.


StephenBonikowsky:
LGTM

PR #1820, opened 2017-03-24
Adding a doc on how to use WCF .NET Core in Windows and Linux containers

KKhurin:
The provided in-place sample code should be enough to create an .NET Core client WCF app and to run it in a container.

### DIFF
--- a/Documentation/HowToUseWCFInContainers.md
+++ b/Documentation/HowToUseWCFInContainers.md
@@ -4,11 +4,12 @@ Prerequisites:
 - Windows Server 2016 or Windows 10 with "Containers" feature enabled
 - Latest [Docker](https://www.docker.com) version installed
 - If running Docker inside a VM, make sure you [enabled nested virtualization for the VM](https://blogs.technet.microsoft.com/virtualization/2015/10/13/windows-insider-preview-nested-virtualization/) 
-
+- For getting started with Windows containers please refer to this document: https://docs.microsoft.com/en-us/virtualization/windowscontainers/quick-start/quick-start-windows-10
+- For running a sample (and similar simple service apps) in a container you can follow a blog post by Jeffrey T. Fritz at https://blogs.msdn.microsoft.com/webdev/2017/02/20/lets-try-wcf-self-hosted-services-in-a-container/
 
 
 ### 1. Sample app
-These are the steps for creating a sample WCF service and client. If you already have a sample WCF service app (or even host it in a container as described in an excellent blog post by Jeffrey T. Fritz at https://blogs.msdn.microsoft.com/webdev/2017/02/20/lets-try-wcf-self-hosted-services-in-a-container/), and a corresponding .NET Core WCF client project then you may skip the sample app creation and proceed to step 2. 
+These are the steps for creating a sample WCF service and client. If you already have a sample WCF service app, and a corresponding .NET Core WCF client project then you may skip the sample app creation and proceed to step 2. 
 For the purpose of this demo we’ll go ahead and just create a WCF service app from a default project template. Among other things this will generate a service and a data contract that look like the one below (IService1.cs):
 
 ~~~
@@ -141,3 +142,4 @@ $pass = ConvertTo-SecureString -AsPlainText -Force password
 Import-pfxcertificate -certStoreLocation Cert:LocalMachine\My -Password $pass -FilePath client_cert.pfx
 Import-Certificate -certStoreLocation Cert:LocalMachine\Root -FilePath cert_authority_public.cer
 ~~~
+
--- a/Documentation/HowToUseWCFInContainers.md
+++ b/Documentation/HowToUseWCFInContainers.md
@@ -1,0 +1,143 @@
+This document describes how to get started with WCF for .NET Core in containers. It will guide you through creating a sample WCF client & server apps, creating a Windows and Linux Docker containers and share additional details on how to install client certificates in Windows containers.
+
+Prerequisites:
+- Windows Server 2016 or Windows 10 with "Containers" feature enabled
+- Latest [Docker](https://www.docker.com) version installed
+- If running Docker inside a VM, make sure you [enabled nested virtualization for the VM](https://blogs.technet.microsoft.com/virtualization/2015/10/13/windows-insider-preview-nested-virtualization/) 
+
+
+
+### 1. Sample app
+These are the steps for creating a sample WCF service and client. If you already have a sample WCF service app (or even host it in a container as described in an excellent blog post by Jeffrey T. Fritz at https://blogs.msdn.microsoft.com/webdev/2017/02/20/lets-try-wcf-self-hosted-services-in-a-container/), and a corresponding .NET Core WCF client project then you may skip the sample app creation and proceed to step 2. 
+For the purpose of this demo we’ll go ahead and just create a WCF service app from a default project template. Among other things this will generate a service and a data contract that look like the one below (IService1.cs):
+
+~~~
+using System;
+using System.Runtime.Serialization;
+using System.ServiceModel;
+
+namespace HelloWcfService
+{
+    [ServiceContract]
+    public interface IService1
+    {
+        [OperationContract]
+        string GetData(int value);
+        [OperationContract]
+        CompositeType GetDataUsingDataContract(CompositeType composite);
+    }
+
+    [DataContract]
+    public class CompositeType
+    {
+        bool boolValue = true;
+        string stringValue = "Hello ";
+        [DataMember]
+        public bool BoolValue { get { return boolValue; } set { boolValue = value; } }
+        [DataMember]
+        public string StringValue  { get { return stringValue; } set { stringValue = value; } }
+    }
+}
+~~~
+
+For further simplicity let’s skip the “standard” path of service reference generation and simply make a copy of it for our future WCF core client. Next, we’re going to write our WCF client code (program.cs):
+
+~~~
+using HelloWcfService;
+using System;
+using System.ServiceModel;
+
+namespace ConsoleApp1
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            var url = Environment.GetEnvironmentVariable("url");
+            var helloEndpoint = new EndpointAddress(url);
+            var binding = new BasicHttpBinding();
+            var factory = new ChannelFactory<IService1>(binding, helloEndpoint);
+            var client = factory.CreateChannel();
+            var data = client.GetData(5);
+            Console.WriteLine(data);
+        }
+    }
+}
+~~~
+
+Finally, we’ll need a project file. As .NET Core evolves the project file may look different depending on the version of .NET core SDK you’re going to use in the container. One way to generate a project file compatible with a particular version of SDK (or Visual Studio) is to run the following (make sure the directory you’re running this in doesn’t already have a project file):
+
+~~~
+mkdir wcfClient
+cd wcfClient
+dotnet new console
+~~~
+
+and use the newly created project file as a template for your WCF client project. The project that we’re providing here is created by Visual Studio 2017. The last step is to add references to WCF .NET Core project like the one below. Again, depending on the SDK version, the format may change (wcfClient.csproj):
+
+~~~
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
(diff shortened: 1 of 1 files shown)


Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
+This document describes how to get started with WCF for .NET Core in containers.
+
+### 1. Sample app
+These are the steps for creating a sample WCF service and client. If you already have a sample WCF service app (or even host it in a container like described in an excellent Jeffrey�s blog post https://blogs.msdn.microsoft.com/webdev/2017/02/20/lets-try-wcf-self-hosted-services-in-a-container/), and a corresponding .NET Core WCF client project then you may skip the sample app creation and proceed to step 2.
```
StephenBonikowsky: replace "like" with "as"

StephenBonikowsky: replace "in an excellent Jeffrey’s blog post https://..."
with "in an excellent blog post by Jeffrey T. Fritz at https://..."

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+            var factory = new ChannelFactory<IService1>(binding, helloEndpoint);
+            var client = factory.CreateChannel();
+            var data = client.GetData(5);
+            Console.WriteLine(data);
+        }
+    }
+}
+~~~
+
+Finally, we�ll need a project file. Unfortunately the project file may look different depending on the version of .NET core SDK you�re going to use in the container. One way to generate a project file compatible with a particular version of SDK is to run
```
StephenBonikowsky: replace: "is to run"
with: "is to run the following..."

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+</Project>
+~~~
+
+#### To sum it up: you have a WCF default template (or your own) service running and a folder called �wcfClient� with the client app that has 3 files: wcfClient.csproj, program.cs, and IService1.cs.
+
+### 2. Container base image
+.NET Core Runtime Docker images are available for both released versions and for nightly builds (https://hub.docker.com/r/microsoft/dotnet-nightly/). One can get images that have only the .NET Core runtime or both the runtime and the SDK. We�re going to use Nano Server and Linux SDK images for the latest 1.1.1 nightly build (which are compatible with the project file that we created above).
+
+### 3. Dockerfile
+Our goal here is to have a docker file that restores necessary packages and builds the app at the time when we build the container. Should you pick up a different base layer and rebuild your container � the app will automatically use the new binaries. Having a container with a shared .NET Core framework helps to save space and to speed up lunching new containers with the same base image.
```
StephenBonikowsky: typo: "lunching"
corrected: "launching"

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+#### Dockerfile for Windows Nano Server
+~~~
+FROM microsoft/dotnet-nightly:1.1.1-sdk-nanoserver
+COPY wcfClient wcfClient
+RUN dotnet restore wcfClient\wcfClient.csproj
+RUN dotnet build wcfClient\wcfClient.csproj
+CMD dotnet run -p wcfClient\wcfClient.csproj 
+~~~
+### 4. Build and run the container
+Navigate to the folder that contains the dockerfile and a subfolder called �demo� and run the following commands. Since our sample app reads environment variable named "url" to create an endpoint we pass it via --env parameter:
```
StephenBonikowsky: replace: "reads environment"
with: "reads an environment"

StephenBonikowsky:
A few nits, otherwise LGTM

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+
+~~~
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.App">
+      <Version>1.0.1</Version>
```
zhenlan: Are you sure the version is 1.0.1? I expect the version is 1.1.1 since you have TargetFramework `netcoreapp1.1` above.

BTW, I remember in the final VS2017 build, the package reference to `Microsoft.NETCore.App` is not present anymore for projects created from default template. Can you please double check?

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+    </PackageReference>
+    <PackageReference Include="System.Private.ServiceModel">
+      <Version>4.3.0</Version>
+    </PackageReference>
```
zhenlan: Please don't include package reference to System.Private.ServiceModel. This is a private package. Customers are not recommended to reference it directly. We may change it in the future. Other WCF contract packages will bring it in anyway.

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
+This document describes how to get started with WCF for .NET Core in containers.
```
zhenlan: It will be nice to have a brief summary of what we are going to build.

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+
```
zhenlan: Can we add instructions of prerequisite, for example, what OS, what need to be installed and what are the links?

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+      <Version>4.3.0</Version>
+    </PackageReference>
+  </ItemGroup>
+</Project>
+~~~
+
+#### To sum it up: you have a WCF default template (or your own) service running and a folder called �wcfClient� with the client app that has 3 files: wcfClient.csproj, program.cs, and IService1.cs.
+
+### 2. Container base image
+.NET Core Runtime Docker images are available for both released versions and for nightly builds (https://hub.docker.com/r/microsoft/dotnet-nightly/). One can get images that have only the .NET Core runtime or both the runtime and the SDK. We�re going to use Nano Server and Linux SDK images for the latest 1.1.1 nightly build (which are compatible with the project file that we created above).
```
zhenlan: Feels like this section can be just merged with the section below as I am not sure what I am supposed to do by this step. Looks like it just provided some information that may be useful later to create dockerfile.

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+
+### 3. Dockerfile
+Our goal here is to have a docker file that restores necessary packages and builds the app at the time when we build the container. Should you pick up a different base layer and rebuild your container � the app will automatically use the new binaries. Having a container with a shared .NET Core framework helps to save space and to speed up launching new containers with the same base image.
```
zhenlan: Can we provide instructions like what you did in step 1, for example, what's the name of dockerfile, where it should live etc.?

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+#### Dockerfile for Windows Nano Server
+~~~
+FROM microsoft/dotnet-nightly:1.1.1-sdk-nanoserver
+COPY wcfClient wcfClient
+RUN dotnet restore wcfClient\wcfClient.csproj
+RUN dotnet build wcfClient\wcfClient.csproj
+CMD dotnet run -p wcfClient\wcfClient.csproj 
+~~~
+### 4. Build and run the container
+Navigate to the folder that contains the dockerfile and a subfolder called �demo� and run the following commands. Since our sample app reads an environment variable named "url" to create an endpoint we pass it via --env parameter:
```
zhenlan: I found nowhere in the instruction we have a folder called "demo" created. Can you please provide more details?

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+
+~~~
+docker build -t wcf_client_in_container .
+docker run --env url=http://server_name/app_name/Service1.svc -it --rm wcf_client_in_container
+~~~
```
zhenlan: The docker command won't just work. Please provide instructions of where to install it and any difference for windows and linux etc.

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,142 @@
...
+
+### 5. Certificates 
+Windows Nano Server has powershell cmdlets to install client certificates. If your certificate authority public key is in cert_authority_public.cer  and client certificate is client_cert.pfx (protected by pwd �password�) then the dockerfile will have the following powershell commands:
+~~~
+$pass = ConvertTo-SecureString -AsPlainText -Force password 
+Import-pfxcertificate -certStoreLocation Cert:LocalMachine\My -Password $pass -FilePath client_cert.pfx
+Import-Certificate -certStoreLocation Cert:LocalMachine\Root -FilePath cert_authority_public.cer
+~~~
```
zhenlan: So this works only in windows container, right? How about Linux? Where should these cert files live? How did files shared between the host and the container?

It's a little bit weird to see this section as there is no previous scenario using certificate. How about you update the sample code in this section to show how certs are used?

Documentation/HowToUseWCFInContainers.md
KKhurin: Added prerequisites. Although there are differences between Linux and Windows they are not significant for this demo and go beyond the scope of this document.

Documentation/HowToUseWCFInContainers.md
KKhurin: One benefit of containers is that you can quickly try latest & greatest version without a need to install anything. In this section you're picking the right container image for your app. The next section shows examples of the dockerfiles that will use the image you chose here. The choice of 1.1.1 image is in line with the spirit of "quickly trying latest and greatest" via Docker images - in this case - the latest update for a shipped product version.

Documentation/HowToUseWCFInContainers.md
KKhurin: Correct. To cover client certs in Linux we would have to go into a lot more details here which are better described in the corresponding Linux documentation. 

Since Windows containers are so new I thought I would mention a quick "how to" to save reader's time on searching and experimenting.

Documentation/HowToUseWCFInContainers.md
KKhurin: Fixed, thanks! :)

Documentation/HowToUseWCFInContainers.md
KKhurin: Fixed. Thanks!

Documentation/HowToUseWCFInContainers.md
KKhurin: Fixed. Thanks!

Documentation/HowToUseWCFInContainers.md
```diff
@@ -0,0 +1,143 @@
...
+
+Prerequisites:
+- Windows Server 2016 or Windows 10 with "Containers" feature enabled
+- Latest [Docker](https://www.docker.com) version installed
+- If running Docker inside a VM, make sure you [enabled nested virtualization for the VM](https://blogs.technet.microsoft.com/virtualization/2015/10/13/windows-insider-preview-nested-virtualization/) 
+
+
+
+### 1. Sample app
+These are the steps for creating a sample WCF service and client. If you already have a sample WCF service app (or even host it in a container as described in an excellent blog post by Jeffrey T. Fritz at https://blogs.msdn.microsoft.com/webdev/2017/02/20/lets-try-wcf-self-hosted-services-in-a-container/), and a corresponding .NET Core WCF client project then you may skip the sample app creation and proceed to step 2.
```
zhenlan: Is service in container something you have tried? People may expect, for example, cert auth to work when the service is hosted in container after reading this doc. If we did not get a chance to validate scenarios with both client and service in containers, I would skip mentioning the service container blog for now.

KKhurin: Only tried cert authentication with the client running in Windows containers so far. 

The service container was mentioned here simply is the spirit of the containers story - try something quick without disrupting the host. All kinds of advanced scenarios go beyond this quick start readme. I hope this sets the expectations.

PR #1830, opened 2017-03-27
Adding SecurityNegotiationException to the public contract.

StephenBonikowsky:
* It is already fully implemented, just needed to be added to the public contract.


zhenlan:
Can we add at least one unit test to make sure the API is accessible from public contract? Other than that, LGTM.

StephenBonikowsky:
Looks like I am going to need new packages first before I can get the tests to compile. I'll merge this, build new packages and put out a new PR with the unit test.

PR #1857, opened 2017-04-19
Post 1.1 package / version updates

ericstj:
WCF should be harvesting assets from the last shipped packages, those from the 1.1 release with package versions 4.3.0.

In addition, the libraries need to increment the bugfix portion of the version to give them a unique assembly version over the last release.  I didn't need to do this for Primitives since that already has a newer minor version.

/cc @StephenBonikowsky @MattGal @weshaggard


StephenBonikowsky:
lgtm

ericstj:
Hold off on merging this.

PR #1869, opened 2017-04-25
Removing extra slash from URLs generated by ServiceUtilHelper

KKhurin:
BuildBaseUri() already leaves a trailing slash so we don't need to add an extra one when generating the endpoint address.


zhenlan:
@dotnet-bot test OuterLoop Selfhost Windows_NT Debug please

PR #1871, opened 2017-04-26
Building ns2.0 harvesting ns1.3

StephenBonikowsky:


StephenBonikowsky:
@ericstj as I dealt with errors and made changes to fix them things only seemed to get worse... :(

At this point, after trying to harvest ns1.3 System.ServiceModel.Primitives.pkgproj it was complaining about a conflict between the live build version of 1.3 being 4.1 versus the version it was trying to harvest in the package folder being version 4.2.

I then thought I had to update https://github.com/dotnet/wcf/blob/c60f05411443786fae6f0a5f8c33b7901aa16189/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.csproj as well since we shouldn't be building a live ns1.3 right?

If you have any ideas it would be great :) will pick this back up tomorrow.

src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.csproj
```diff
@@ -13,5 +13,11 @@
   <ItemGroup>
     <None Include="project.json" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetGroup)' == 'netstandard2.0'">
```
ericstj: Don't do this.  API compat should be getting the right references.  If its not fix the references don't hack around with content.

ericstj:
Note that you have added surface area to System.ServiceModel.Primitives since your last public release so you do need to deliver live builds of that library for netstandard1.3 if you intend to actually ship that new API to netstandard1.3 consumers.  Presumably that API would also need to be implemented somewhere so perhaps it was incorrect to do that if you were not building a new live version of System.Private.ServiceModel for netstandard1.3.

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <ItemGroup>
     <ProjectReference Include="..\ref\System.ServiceModel.Duplex.csproj">
-      <SupportedFramework>net45;win8;netcoreapp1.0;$(AllXamarinFrameworks)</SupportedFramework>
+      <SupportedFramework>net45;win8;netcoreapp2.0;$(AllXamarinFrameworks)</SupportedFramework>
```
ericstj: Only update these if you are *retargeting* your reference assemblies.  These are supposed to represent the minimum frameworks you expect to be supported by the API version of that reference.  See https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/package-projects.md#applicability-validation

zhenlan:
@ericstj can we include the new surface area of S.SM.Primitives only for NS2.0 and not include them for NS1.3?

ericstj:
@zhenlan yeah, you can do that.  You'll want to retarget the ref and make sure you harvest both the old ref and old lib that didn't have the API changes.  This will be a build-to-build breaking change so folks building against an old package will break when running on a new one (due to assembly version going down from 4.2.0.0 to whatever you shipped last release), but that may be what you want.

You should do a review of all the supported framework entries for your pkgproj's to make sure they are listing only the frameworks supported by your live build.  That should cover most of the validation errors.

Alternatively if you wanted to be more explicit you could list out the API versions you expect to be supported by every framework using the alternative syntax. The metadata on the project reference is just a shortcut for using the version of that project for validation.
https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/package-projects.md#applicability-validation

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
StephenBonikowsky: Ok, so leaving netcoreapp1.0 means we support the higher versions as well, in this case netcoreapp1.1 and netcoreapp2.0, yes?
In your email you mentioned that we 

> should update the SupportedFrameworks you expect cover with your build

 so how do we need to update them then?

ericstj: > so leaving netcoreapp1.0 means we support the higher versions as well, in this case netcoreapp1.1 and netcoreapp2.0, yes?

Yes.

> so how do we need to update them then?

Only update them if you retarget the reference projects.  See my [below explanation](https://github.com/dotnet/wcf/pull/1871#issuecomment-297492616) for more context.

zhenlan:
Thanks @ericstj. Yes, I understand it's going to be a build-to-build breaking change. This is why I would like us to make this change before any public release.

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
StephenBonikowsky: Thanks, based on your additional details we do want to 

> retarget the ref and make sure you harvest both the old ref and old lib that didn't have the API changes.

For the first part of that in my local repo I just updated PackageTargetFramework and NuGetTargetMoniker like so...
```
<PackageTargetFramework>netstandard2.0</PackageTargetFramework>
<NuGetTargetMoniker>.NETStandard,Version=v2.0</NuGetTargetMoniker>
```
And I updated the associated project.json to...
```
{
  "frameworks": {
    "netstandard2.0": {
      "dependencies": {
        "NETStandard.Library": "2.0.0-preview1-25131-01",
        "System.Security.Principal.Windows": "4.4.0-beta-24911-02",
        "System.ServiceModel.Primitives": "4.4.0-beta-25221-02",
        "System.Reflection.DispatchProxy": "4.4.0-beta-24911-02",
        "System.Runtime.InteropServices.RuntimeInformation": "4.4.0-beta-24911-02"
      }
    }
  }
}
```
Does that look correct?

StephenBonikowsky: Updated with additional changes.

ericstj:
These build errors look to me like you aren't getting the Netstandard.Library targets [imported](https://github.com/dotnet/wcf/blob/20d881032f5339e16c2e35627b48e57a43adadc8/dir.targets#L112)

That's because the condition there is using TargetGroup and you don't set that on your reference projects.

Perhaps you should change the condition from `'$(TargetGroup)'=='netstandard2.0'` to `'$(NuGetTargetMoniker)'=='.NETStandard,Version=v2.0'`.  /cc @MattGal

StephenBonikowsky:
I'll try that, thanks.

StephenBonikowsky:
Didn't make any difference, still trying things but I updated the PR

ericstj:
In the NETStandard.Library targets they assume some properties have been created by msbuild/sdk targets:
https://github.com/dotnet/standard/blob/master/netstandard/pkg/targets/netstandard/NETStandard.Library.targets#L6

So you may want to include the following in dir.targets
```
<PropertyGroup Condition="$(NuGetTargetMoniker.StartsWith('.NETStandard,'))">
  <TargetFrameworkIdentifier>.NETStandard</TargetFrameworkIdentifier>
</PropertyGroup>
```

ericstj:
After doing this I start hitting issues with your old versions of the corefx packages.  For example:
```
CSC : error CS1705: Assembly 'System.Security.Principal.Windows' with identity 'System.Security.Principal.Windows, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' uses 'System.Runtime, Version=4.2.0.0, Culture=
neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'System.Runtime' with identity 'System.Runtime, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' [F:\dotnet\wcf\src\S
ystem.Private.ServiceModel\src\System.Private.ServiceModel.csproj]
```
You'll need to update those to a more recent build that is based off of netstandard.dll.

StephenBonikowsky:
Had just figured out that `<TargetFrameworkIdentifier>` needed to be set to `.NETStandard`, made the change in "C:\OSS\WCF\src\dir.props" instead for now.

Working on the version conflicts now.

StephenBonikowsky:
Updated the PR with the next commit.

Looks like now the errors all have to do with what should or should not be supported and a harvesting error.

StephenBonikowsky:
@ericstj are many of these errors because I need to add runtime assets here?
https://github.com/dotnet/wcf/blob/c60f05411443786fae6f0a5f8c33b7901aa16189/src/System.Private.ServiceModel/pkg/System.Private.ServiceModel.pkgproj#L9-L10
If so can you give me an example?

dependencies.props
```diff
@@ -1,14 +1,14 @@
...
-    <CoreFxCurrentRef>2718aebf9298724ee788f871287bdc82ec950722</CoreFxCurrentRef>
+    <CoreFxCurrentRef>6de42fd82cb7b8db12fe9339e4304f77343edc83</CoreFxCurrentRef>
     <CoreClrCurrentRef>2718aebf9298724ee788f871287bdc82ec950722</CoreClrCurrentRef>
     <WCFCurrentRef>43f8d65aadc9dd4b30cb976652961bbbdcd15f5f</WCFCurrentRef>
   </PropertyGroup>
 
   <!-- Auto-upgraded properties for each build info dependency. -->
   <PropertyGroup>
-    <CoreFxExpectedPrerelease>beta-24911-02</CoreFxExpectedPrerelease>
+    <CoreFxExpectedPrerelease>preview2-25226-01</CoreFxExpectedPrerelease>
```
zhenlan: I really want us to get preview1 build first so please take dependency on corefx preview1 build for now. Once we get good WCF preview1 build we will move to preview2 shortly.

src/Common/test-runtime/project.json
```diff
@@ -11,7 +11,7 @@
         "Microsoft.DotNet.BuildTools.TestSuite": "1.0.1-prerelease-01001-04",
         "xunit.console.netcore": "1.0.3-prerelease-00925-01",
         "Microsoft.xunit.netcore.extensions": "1.0.1-prerelease-01001-04",
-        "System.Reflection.DispatchProxy": "4.4.0-beta-24911-02",
+        "System.Reflection.DispatchProxy": "4.4.0-preview2-25226-01",
```
zhenlan: Let's use latest preview1 build across the board.

dependencies.props
StephenBonikowsky: Ok, done locally.

StephenBonikowsky:
Found that in [packageIndex.json](https://github.com/dotnet/wcf/blob/master/pkg/baseline/packageIndex.json#L66) we were missing the 4.3.0 stable version.

StephenBonikowsky:
@ericstj I have two types of errors now, both look like validation errors, working on them but not having much success.
`C:\OSS\WCF\Tools\Packaging.targets(1097,5): error : System.ServiceModel.Duplex should support API version 4.0.3.0 on .NETCoreApp,Version=v1.0/win7-x86 but C:\OSS\WCF\packages\System.ServiceModel.Duplex\4.3.0\ref\netstandard1.1\System.ServiceModel.Duplex.dll was found to support 4.0.2.0. [C:\OSS\WCF\src\System.ServiceModel.Duplex\pkg\System.ServiceModel.Duplex.pkgproj]
`

and

`C:\OSS\WCF\Tools\Packaging.targets(1097,5): error : System.ServiceModel.Duplex should be supported on .NETCoreApp,Version=v1.1/ubuntu.16.10-x64 but has no runtime assets. [C:\OSS\WCF\src\System.ServiceModel.Duplex\pkg\System.ServiceModel.Duplex.pkgproj]`

StephenBonikowsky:
@ericstj @MattGal 
Is this correct? [packageIndex.json](https://github.com/dotnet/wcf/blob/ac4ab235e552ecd2be06d9c9e9b622eff60cfc52/pkg/baseline/packageIndex.json#L62)

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
```diff
@@ -6,7 +6,7 @@
       <SupportedFramework>net45;win8;netcoreapp1.0;$(AllXamarinFrameworks)</SupportedFramework>
```
ericstj: You retargeted the reference assembly but didn't update the supportedFramework entries.

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
```diff
@@ -6,7 +6,7 @@
...
     </ProjectReference>
     <ProjectReference Include="..\src\System.ServiceModel.Duplex.builds" />
```
ericstj: You retargeted the src assembly but didn't harvest the old asset: lib/netstandard1.3

pkg/baseline/packageIndex.json
```diff
@@ -66,7 +66,8 @@
       "StableVersions": [
         "3.9.0",
         "4.0.0",
-        "4.1.0"
+        "4.1.0",
+        "4.3.0"
```
ericstj: 👍

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
StephenBonikowsky: When I include that I get...
`error : Path 'lib/netstandard1.1' was specified in PathsToInclude but was not found in the package System.ServiceModel.Duplex/4.3.0. [C:\OSS\WCF\src\System.ServiceModel.Duplex\pkg\System.ServiceModel.Duplex.pkgproj]`

Why is it looking under `4.3.0` and not `4.4.0-beta-25221-02`?

In my packages folder I also see that I have `packages\System.ServiceModel.Duplex\4.4.0-beta-25221-02\ref\netstandard1.1` but not `netstandard1.3`
and
`C:\OSS\WCF\packages\System.ServiceModel.Duplex\4.4.0-beta-25221-02\lib\netstandard1.3` but not `netstandard1.1
`Is this correct?

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
StephenBonikowsky: Simply don't understand what the supportedFramework entries is supposed to be updated with. We retargeted the ref assembly from netstandard1.1 to netstandard2.0, adding netstandard1.1 to the supportedFramework entries throws an error.

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
ericstj: You should harvest whatever you're old assembly was.  It is supposed to use the stable version: that's the whole point of harvesting.  I guessed that it matched the ref but it'll actually match whatever you were actually targeting (probably the same thing the src project was targeting).  Look inside the 4.3.0 package to see what's there.  You'll also see log spew around this if you check the msbuild log.

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
ericstj: The supportedFramework entry should be updated with what conrete frameworks you expect that projectreference to support.  Since the project is supporting netstandard2.0 that means... the frameworks that support netstandard2.0: netcoreapp2.0;net461;$(AllXamarinFrameworks).

It's important that you understand this part if you intend to use it.  The purpose of SupportedFramework is for the developer of the package to record in source control where he/she expects that asset to apply.  It's a *safegaurd* so that when folks retarget something and it changes the applicability of API versions in their package they do so with the explicit knowledge of what that change will do.  Please read more here: https://github.com/dotnet/wcf/pull/1871#issuecomment-297492616

StephenBonikowsky:
I think we are close now, do we also harvest net46? I was getting some validation errors for it, when I added it `<HarvestIncludePaths Include="ref/net46;lib/net46" />` I then get more validation errors for 4.6.1, 4.6.2 and 4.6.3.
```
error : System.ServiceModel.Primitives should support API version 4.2.0.0 
on .NETFramework,Version=v4.6.3/win7-x64 
but C:\OSS\WCF\packages\System.ServiceModel.Primitives\4.3.0\lib\net46\System.ServiceModel.Primitives.dll 
was found to support 4.1.1.0. 
[C:\OSS\WCF\src\System.ServiceModel.Primitives\pkg\System.ServiceModel.Primitives.pkgproj]
```

StephenBonikowsky:
I just pushed update 3 of this PR.
Everything is green and all tests are running and passing. (one known exception)

@ericstj We removed net461 from the SupportedFramework as it was pulling in whole new set of errors.

We would love a review at this point, we need to be able to build against the planned build of netstandard library for preview 1 tomorrow.

StephenBonikowsky:
The one test failure is expected.

src/System.Private.ServiceModel/src/windows/project.json
```diff
@@ -3,11 +3,11 @@
   "frameworks": {
     "netcore50": {
```
zhenlan: Don't need to change now. I think we can remove netcore50 section in the future since we are harvesting netcore50 now.

src/Common/test-runtime/project.json
```diff
@@ -37,10 +37,10 @@
         "System.Linq": "4.4.0-beta-24911-02",
         "System.Linq.Parallel": "4.4.0-beta-24911-02",
         "System.Linq.Queryable": "4.4.0-beta-24911-02",
-        "System.Net.Http": "4.4.0-beta-24911-02",
+        "System.Net.Http": "4.4.0-preview1-25218-03",
```
zhenlan: Don't need to change now. If we don't run our tests against NS1.3 anymore, I think we can remove this whole "netstandard1.3" section in the future.

src/dir.targets
```diff
@@ -2,6 +2,10 @@
   <PropertyGroup>
     <ErrorIfBuildToolsRestoredFromIndividualProject Condition="!Exists('$(ToolsDir)')">true</ErrorIfBuildToolsRestoredFromIndividualProject>
   </PropertyGroup>
+
+  <PropertyGroup Condition="$(NuGetTargetMoniker.StartsWith('.NETStandard,'))">
```
ericstj: Add a comment about why this is needed:
> The NETStandard.Library targets expect the TargetFrameworkIdentifier property to be set.  This is normally set the SDK targets by the but it isn't set by BuildTools.

I filed https://github.com/dotnet/buildtools/issues/1465 to try to get clarity on this moving forward.

ericstj:
This is OK for now.

> We removed net461 from the SupportedFramework as it was pulling in whole new set of errors.

All of those are valid errors, and you're hiding them now.

The scenario you've broken is someone building against netstandard2.0, then trying to run in net461.

The fix will be to ensure you have a net461 build for all of your contracts.

I'm ok with you handling that in a separate commit.

I've filed https://github.com/dotnet/buildtools/issues/1466 to track catching this in BuildTools even if you get SupportedFramework wrong like you've done.

StephenBonikowsky:
Opened #1875 to fix the net461 errors.

PR #1882, opened 2017-05-02
Add test for xsfa

StephenBonikowsky:


StephenBonikowsky:
We need to update our setup scripts for OSX10.12, error seems to indicate it has to do with the certificates we install.
Not blocking merging because of this.

PR #1885, opened 2017-05-03
To support soap encoded message.

### LINKED ISSUE #1549: Add OperationFormatUse to support soap encoded WSDL
1. The Enum type OperationFormatUse  is missing from runtime.

There is a customer ask for OperationFormatUse.Encoded to be supported. See #1517 for details.

### LINKED ISSUE #1797: Expose XmlSerializerOperationBehavior.GetXmlMappings
Need expose XmlSerializerOperationBehavior class and GetXmlMappings method to be used in svcutil

shmao:
The PR enabled WCF to support soap encoded message. WCF now supports `OperationFormatUse.Encoded`.

Exposed XmlSerializerOperationBehavior.GetXmlMappings.

Fix #1549.
Fix #1797.

### DIFF
--- a/src/System.Private.ServiceModel/src/Resources/Strings.resx
+++ b/src/System.Private.ServiceModel/src/Resources/Strings.resx
@@ -1932,4 +1932,22 @@
   <data name="Xml_InvalidNodeType" xml:space="preserve">
     <value>'{0}' is an invalid XmlNodeType.</value>
   </data>
+  <data name="SFxRpcMessageMustHaveASingleBody" xml:space="preserve">
+    <value>RPC Message {1} in operation {0} must have a single MessageBodyMember.</value>
+  </data>
+  <data name="SFxBodyObjectTypeCannotBeInherited" xml:space="preserve">
+    <value>Type {0} cannot inherit from any class other than object to be used as body object in RPC style.</value>
+  </data>
+  <data name="SFxBodyObjectTypeCannotBeInterface" xml:space="preserve">
+    <value>Type {0} implements interface {1} which is not supported for body object in RPC style.</value>
+  </data>
+  <data name="SFxHeadersAreNotSupportedInEncoded" xml:space="preserve">
+    <value>Message {0} must not have headers to be used in RPC encoded style.</value>
+  </data>
+  <data name="SFxMultiplePartsNotAllowedInEncoded" xml:space="preserve">
+    <value>Part {1}:{0} is repeating and is not supported in Soap Encoding.</value>
+  </data>
+  <data name="SFxInvalidSoapAttribute" xml:space="preserve">
+    <value>XmlSerializer attribute {0} is not valid in {1}. Only SoapElement attribute is supported.</value>
+  </data>
 </root>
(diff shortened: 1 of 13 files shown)


src/System.Private.ServiceModel/src/Resources/Strings.resx
```diff
@@ -1932,4 +1932,22 @@
   <data name="Xml_InvalidNodeType" xml:space="preserve">
     <value>'{0}' is an invalid XmlNodeType.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="SFxRpcMessageMustHaveASingleBody" xml:space="preserve">
+    <value>RPC Message {1} in operation {0} must have a single MessageBodyMember.</value>
+  </data>
+  <data name="SFxBodyObjectTypeCannotBeInherited" xml:space="preserve">
+    <value>Type {0} cannot inherit from any class other than object to be used as body object in RPC style.</value>
```
StephenBonikowsky: Add the article "the" like so...
`Type {0} cannot inherit from any class other than the object to be used as the body object in the RPC style.`

src/System.Private.ServiceModel/src/Resources/Strings.resx
```diff
@@ -1932,4 +1932,22 @@
...
+  </data>
+  <data name="SFxBodyObjectTypeCannotBeInterface" xml:space="preserve">
+    <value>Type {0} implements interface {1} which is not supported for body object in RPC style.</value>
```
StephenBonikowsky: "in the RPC style."

src/System.Private.ServiceModel/src/Resources/Strings.resx
```diff
@@ -1932,4 +1932,22 @@
...
+  </data>
+  <data name="SFxHeadersAreNotSupportedInEncoded" xml:space="preserve">
+    <value>Message {0} must not have headers to be used in RPC encoded style.</value>
```
StephenBonikowsky: the RPC

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatSoapTest.cs
```diff
@@ -0,0 +1,61 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using System.Threading.Tasks;
+using Infrastructure.Common;
+using Xunit;
+
+public static partial class XmlSerializerFormatTests
```
StephenBonikowsky: I know this Is nit-picky but we would like to have a consistent 'feel' across all the tests, look at some of the other tests for examples but could you...
- Add some comment blocks for sections of the tests like...
             // *** SETUP *** \\
            // *** EXECUTE *** \\
           // *** CLEANUP *** \\
- Use try/catch/finally in these tests it doesn't look like you need the 'Catch' but in the finally we make sure we cleanup by calling `ScenarioTestHelpers.CloseCommunicationObjects`

src/System.Private.ServiceModel/src/Resources/Strings.resx
shmao: `RPC style` seems to be a term that people might use. For example, [this article](https://www.javatpoint.com/difference-between-rpc-and-document) uses `RPC style` in this way. I'm going to leave the message as is. Thanks for the suggestion, @StephenBonikowsky .

PR #1894, opened 2017-05-05
Adding support for net 4.6.1

StephenBonikowsky:

Issue #1839, opened 2017-03-30
System.ServiceModel.Duplex and System.ServiceModel.Security shipped reference assembly revisions over what was in desktop


ericstj:
These two packages contain reference assemblies with revisions over what was first shipped in desktop.

The result is that you compile a netstandard1.1 library then try to use that in a net45 project and the versions don't match.

Roslyn appears to be tolerant of the mismatch, and using dotnet CLI I don't see a warning or error from RAR, but I do see the following in the log:
```
                     Dependency "System.ServiceModel.Security, Version=4.0.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a". (TaskId:52)
                         Could not resolve this reference. Could not locate the assembly "System.ServiceModel.Security, Version=4.0.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. (TaskId:52)
                             For SearchPath "F:\wcfTest\lib\bin\Debug\netstandard1.1". (TaskId:52)
                             Considered "F:\wcfTest\lib\bin\Debug\netstandard1.1\System.ServiceModel.Security.winmd", but it didn't exist. (TaskId:52)
                             Considered "F:\wcfTest\lib\bin\Debug\netstandard1.1\System.ServiceModel.Security.dll", but it didn't exist. (TaskId:52)
                             Considered "F:\wcfTest\lib\bin\Debug\netstandard1.1\System.ServiceModel.Security.exe", but it didn't exist. (TaskId:52)
                             For SearchPath "{TargetFrameworkDirectory}". (TaskId:52)
                             Considered "C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\System.ServiceModel.Security.winmd", but it didn't exist. (TaskId:52)
                             Considered "C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\System.ServiceModel.Security.dll", but it didn't exist. (TaskId:52)
                             Considered "C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5\System.ServiceModel.Security.exe", but it didn't exist. (TaskId:52)
[...]

PR #1897, opened 2017-05-09
Fix s.sm.duplex and security refs

StephenBonikowsky:
This is regarding #1839
Until PR #1894 is merged and I re-base this PR will also show all of those changes.

### DIFF
--- a/src/System.ServiceModel.Security/ref/System.ServiceModel.Security.csproj
+++ b/src/System.ServiceModel.Security/ref/System.ServiceModel.Security.csproj
@@ -2,6 +2,7 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
+    <AssemblyVersion>4.0.2.0</AssemblyVersion>
     <OutputType>Library</OutputType>
     <PackageTargetFramework>netstandard2.0</PackageTargetFramework>
     <NuGetTargetMoniker>.NETStandard,Version=v2.0</NuGetTargetMoniker>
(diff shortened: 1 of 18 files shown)


StephenBonikowsky:
@ericstj with my changes based on looking at your old CoreFx PR #[corefx\6902](https://github.com/dotnet/corefx/pull/6902) I'm getting...

`C:\OSS\WCF\Tools\Packaging.targets(1097,5): error : PackageIndex from C:\OSS\WCF\pkg\baseline\\packageIndex.json, C:\OSS\WCF\packages\Microsoft.Private.PackageBaseline\2.0.0-preview2-25304-02\build\packageIndex.json is has an assembly version entry(s) for 4.0.3.0 which are no longer in package System.ServiceModel.Duplex.  Please run /t:UpdatePackageIndex on this project to commit an update. [C:\OSS\WCF\src\System.ServiceModel.Duplex\pkg\System.ServiceModel.Duplex.pkgproj]`

@zhenlan and I discussed this but we don't understand enough of the 'why' this is happening to figure out what the solution is.

src/System.ServiceModel.Security/ref/System.ServiceModel.Security.csproj
```diff
@@ -2,6 +2,7 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
+    <AssemblyVersion>4.0.2.0</AssemblyVersion>
```
ericstj: You could just change https://github.com/StephenBonikowsky/wcf/blob/45229d8763582d6e8ff8edb1bd1fc12db33af575/src/System.ServiceModel.Security/dir.props#L5

zhenlan:
> you can either continue to move forward on versioning and continue to omit the desktop ref since it will unify at runtime so long as it stays at 4.0.x.y

Thanks @ericstj! At this point, I am leaning towards this option. We have multiple desktop ref versions out there already. Customers who ran into this can work around by removing binding redirect. So far we did not hear any complaint about this and we can revisit it if this becomes an issue for people in the future.

StephenBonikowsky:
Thanks for your comments on this Eric, understanding this a little better now. We realized that we've actually changed the Duplex contract which we believe requires a minor version update.
Since what I was initially trying to do with this PR is pretty tightly connected to PR #1894 and what is currently here is completely wrong :) I am going to abandon this PR and add the changes we thing are necessary to PR #1894.

Issue #1694, opened 2016-12-08
CallOnceManager ends up completing the async result twice


KKhurin:
This was found with short Send timeout tests:
~~~
    SP               IP               Function
    0000000B3037ED30 00007FFE15C6FDA7 System_Private_ServiceModel!System.Runtime.InternalSR.AsyncResultCompletedTwice(System.Object)+0x27
    0000000B3037ED60 00007FFE15B19ECD System_Private_ServiceModel!System.Runtime.AsyncResult.Complete(Boolean)+0x13d
    0000000B3037EDE0 00007FFE15C6E42A System_Private_ServiceModel!System.ServiceModel.Channels.ServiceChannel+CallOnceManager+AsyncWaiter.TimerCallback(System.Object)+0x5a
    0000000B3037EE10 00007FFE7312D4AF System_Private_CoreLib_ni!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)+0x10f
    0000000B3037EE80 00007FFE731F2B26 System_Private_CoreLib_ni!System.Threading.TimerQueueTimer.CallCallback()+0x186
    0000000B3037EF00 00007FFE731F2824 System_Private_CoreLib_ni!System.Threading.TimerQueueTimer.Fire()+0x94
    0000000B3037EF70 00007FFE731EFE44 System_Private_CoreLib_ni!System.Threading.ThreadPoolWorkQueue.Dispatch()+0x1a4

~~~

KKhurin:
Found by #1676

zhenlan:
Triage discussed options:
1. Do interlocked.compareexchage to avoid the race, but this option potentially has perf impact for non-short-timeout scenarios.
2. Add unhandled exception frame.

We should change exception throw to debug.assert.

KKhurin:
resolved with https://github.com/dotnet/wcf/pull/1904

PR #1904, opened 2017-05-10
Fix for #1694

KKhurin:
The fix is to add an UnhandledExceptionFrame for timer callback in CallOnceManager.AsyncWaiter

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannel.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannel.cs
@@ -2281,7 +2281,7 @@ namespace System.ServiceModel.Channels
 
             internal class AsyncWaiter : AsyncResult, IWaiter
             {
-                private static Action<object> s_timerCallback = new Action<object>(TimerCallback);
+                private static TimerCallback s_timeoutCallback = new TimerCallback(Fx.ThunkCallback<object>(TimeoutCallback));
 
                 private CallOnceManager _manager;
                 private TimeSpan _timeout;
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ServiceChannel.cs
```diff
@@ -2263,7 +2263,7 @@ private void CloseWaitHandle()
 
             internal class AsyncWaiter : AsyncResult, IWaiter
             {
-                private static Action<object> s_timerCallback = new Action<object>(TimerCallback);
+                private static Action<object> s_timerCallback = Fx.ThunkCallback<object>(TimerCallback);
```
mconnew: Can you rename the static method TimerCallback as it took me a moment to realize TimerCallback is both a local static method and the delegate definition System.Threading.TimerCallback. Either that, or change line 2281 to read:
```csharp
_timer = new Timer(new System.Threading.TimerCallback(s_timerCallback), this, timeout, TimeSpan.FromMilliseconds(-1));
```
Right now the code is confusingly ambiguous.
Also, although you don't need to change this, you could have created an overload of ThunkCallback which takes and returns a System.Threading.TimerCallback and you could avoid the new operation on line 2281. You could consider changing this line to be:
```
private static System.Threading.TimerCallback s_timerCallback = new System.Threading.TimerCallback(Fx.ThunkCallback<object>(TimerCallback));
```
That way you avoid the new operator on line 2281 on each instantiation. All suggestions, nothing required.

KKhurin: renaming the method

PR #1910, opened 2017-05-10
Fix 1574 and enable tests - Ready to review

### LINKED ISSUE #1574: Change to use of X509Certificate2 copy constructor

hongdai:
* Use X509Certificate2 copy constructor as it becomes available in 2.0.
* Enable those TCP tests require service certificate. They pass by using
the X509Certificate2 copy constructor.
* Update a test to matches what it's intended to do.
Fix #1574


hongdai:
@mconnew @StephenBonikowsky @zhenlan This is ready for review. The only failure is apparently a CI issue.

zhenlan:
@dotnet-bot test OuterLoop OSX10.12 Release please

hongdai:
@zhenlan   There are multiple issues around. This change fixes the below exception:
Interop+AppleCrypto+AppleCommonCryptoCryptographicException : One or more parameters passed to a function were not valid. You can see the callstack [here](https://ci.dot.net/job/dotnet_wcf/job/master/job/outerloop_osx10.12_debug_prtest/33/testReport/junit/(root)/CustomBindingTests/DefaultSettings_Tcp_Binary_Echo_RoundTrips_String/
).

The callstack shows .X509Certificate2Extensions.CloneCertificateInternal fails.

Issue #1913, opened 2017-05-11
Breaking change from cofefx causing Peer trust scenarios to fail on Linux


hongdai:
Corefx started to block opening the Disallowed X509Store on Linux from 5/8. See [commit](https://github.com/dotnet/corefx/commit/0c6dc1d6a4a0cba584ad0d77e558b439b26f03fc).

hongdai:
Setup tests are failing as well. The error message is apparent:
"MESSAGE:
System.PlatformNotSupportedException : The Disallowed store is not supported on this platform."

hongdai:
Peer trust test affected: cp_ClientCredentialTypeTests.NetTcp_SecModeTrans_CertValMode_PeerTrust_Succeeds_In_TrustedPeople

zhenlan:
This should be fixed by dotnet/corefx#19844 and dotnet/corefx#19945.

@StephenBonikowsky, can you please update our dependency to the latest corefx build and reenable the related tests?

zhenlan:
Reopen as the fix needs to make in release/2.0.0 branch as well.

zhenlan:
@StephenBonikowsky tests that are disabled by 1913 should be renabled after you updated the corefx dependency version.

StephenBonikowsky:
Verified that this test ran and did not fail on AnyUnix machines in latest release/2.0.0 build in MC, 20170526.11.

PR #1914, opened 2017-05-11
Disable tests due to CoreFx breaking change.

StephenBonikowsky:
* Issue #1913: Breaking change from cofefx causing Peer trust scenarios to fail on Linux


hongdai:
LGTM.

PR #1931, opened 2017-05-17
Rev package version to 4.5

StephenBonikowsky:
Eric and Wes, I particularly want your review for this.
I'm not sure that my changes to packageIndex.json are correct, especially for S.Private.ServiceModel, I got it to pass validation, but...

### DIFF
--- a/pkg/baseline/packageIndex.json
+++ b/pkg/baseline/packageIndex.json
@@ -9,8 +9,9 @@
       "BaselineVersion": "4.4.0",
       "InboxOn": {},
       "AssemblyVersionInPackageVersion": {
-        "4.1.1.0": "4.4.0",
-        "4.1.2.0": "4.4.0"
+        "4.1.1.0": "4.5.0",
+        "4.1.2.0": "4.4.0",
+        "4.1.3.0": "4.5.0"
       }
     },
     "System.ServiceModel.Duplex": {
@@ -19,13 +20,14 @@
         "4.0.1",
         "4.3.0"
       ],
-      "BaselineVersion": "4.4.0",
+      "BaselineVersion": "4.5.0",
       "InboxOn": {},
       "AssemblyVersionInPackageVersion": {
         "4.0.0.0": "4.0.0",
         "4.0.1.0": "4.0.1",
         "4.0.2.0": "4.3.0",
-        "4.1.0.0": "4.4.0"
+        "4.1.0.0": "4.4.0",
+        "4.1.1.0": "4.5.0"
       }
     },
     "System.ServiceModel.Http": {
@@ -36,7 +38,7 @@
         "4.1.0",
         "4.3.0"
       ],
-      "BaselineVersion": "4.4.0",
+      "BaselineVersion": "4.5.0",
       "InboxOn": {},
       "AssemblyVersionInPackageVersion": {
         "3.9.0.0": "3.9.0",
@@ -44,7 +46,8 @@
         "4.0.10.0": "4.0.10",
         "4.1.0.0": "4.1.0",
         "4.1.1.0": "4.3.0",
-        "4.1.2.0": "4.4.0"
+        "4.1.2.0": "4.4.0",
+        "4.1.3.0": "4.5.0"
       }
     },
     "System.ServiceModel.NetTcp": {
@@ -53,13 +56,14 @@
         "4.1.0",
         "4.3.0"
       ],
-      "BaselineVersion": "4.4.0",
+      "BaselineVersion": "4.5.0",
       "InboxOn": {},
       "AssemblyVersionInPackageVersion": {
         "4.0.0.0": "4.0.0",
         "4.1.0.0": "4.1.0",
         "4.1.1.0": "4.3.0",
-        "4.1.2.0": "4.3.0"
+        "4.1.2.0": "4.4.0",
+        "4.1.3.0": "4.5.0"
       }
     },
     "System.ServiceModel.Primitives": {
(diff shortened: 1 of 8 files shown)


pkg/baseline/packageIndex.json
```diff
@@ -9,8 +9,9 @@
       "BaselineVersion": "4.4.0",
```
zhenlan: @ericstj I know you have signed off and thanks... but should this be 4.5.0 like others?

pkg/baseline/packageIndex.json
```diff
@@ -9,8 +9,9 @@
...
       "InboxOn": {},
       "AssemblyVersionInPackageVersion": {
-        "4.1.1.0": "4.4.0",
-        "4.1.2.0": "4.4.0"
+        "4.1.1.0": "4.5.0",
```
zhenlan: I know @StephenBonikowsky made this change to make validation happy but I just have a hard time to understand the magic here :) - both 4.1.1.0 and 4.1.2.0 were mapping to package version 4.4.0 but now 4.1.1.0 has to be changed to be mapped to 4.5.0 but not for 4.1.2.0. @ericstj is this expected?

pkg/baseline/packageIndex.json
StephenBonikowsky: It doesn't help explain but I just updated this locally and it worked, as in no validation errors.

StephenBonikowsky:
I noticed in https://github.com/dotnet/corefx/pull/19842 that `<PlatformPackageVersion Condition="'$(PlatformPackageVersion)' == ''">1.2.0</PlatformPackageVersion>` in 'Packaging.props' was updated as well. Should we as well? Not sure what this version number affects.

ericstj:
`PlatformPackageVersion` is used for two things in CoreFx: the version of Microsoft.NETCore.Platforms that we build and the version that gets referenced for any RID-specific package.  You shouldn't update that now, but consider updating when you ingest a CoreFx build that has a new version of Microsoft.NETCore.Platforms.  That package contains the RID graph so its really only necessary to update your dependency if you need the newer version or are trying to make sure your corefx dependencies all came from the same build.

pkg/baseline/packageIndex.json
ericstj: This list is meant to indicate the first version in which an assembly appeared. You shouldn't be changing a mapping unless you decided not to ship that assembly in the previous release.

pkg/baseline/packageIndex.json
ericstj: BaselineVersion is used to determine the minimum version used for all packages in the repo.

By default we use the AssemblyVersion in packageVersion map to understand what version of a package to reference.  Then we'll raise the package version of the reference so that its at least the baseline version.

pkg/baseline/packageIndex.json
zhenlan: Thanks @ericstj. Based on this 4.1.1.0 should be mapped to 4.3.0 package version as 4.1.1.0 was first released in 4.3.0. @StephenBonikowsky can you try to see if this solves the validation error?

> You shouldn't be changing a mapping unless you decided not to ship that assembly in the previous release.

@ericstj, just to understand, if an assembly is already released, how can I decide not to ship it in a *previous* release?

ericstj: > if an assembly is already released, how can I decide not to ship it in a previous release?

You can't but the index could have been wrong.  For example you were building a new version of that assembly but then decided to harvest it but the index wasn't updated.

zhenlan: I see. Does this affect the building of package or is it only for validation? If it's former, we should make the 4.1.1.0->4.3.0 mapping change in release/2.0.0 branch too, correct?

ericstj: It's primarily for validation.  Especially if your BaseLineVersion is higher than these mappings, the mappings are irrelevant based on https://github.com/dotnet/wcf/pull/1931#discussion_r117287353.

pkg/baseline/packageIndex.json
zhenlan: I see. I think we should change this to 4.5.0 in the case as we want all others (products and tests) to reference the latest package we are newly building. Thanks @ericstj.

StephenBonikowsky:
Hmm, apparently we don't even use it, so is there ever a reason to update this property then? Is it being commented out a mistake?
https://github.com/dotnet/wcf/blob/master/Packaging.props#L7

ericstj:
You don't need that line anymore.  We carry a runtime.json in buildtools itself and use that for validation.

StephenBonikowsky:
Updated.

pkg/baseline/packageIndex.json
```diff
@@ -1,3 +1,7 @@
+// StableVersions: Released versions.
+// BaselineVersion: Used to determine the minimum version used for all packages in the repo.
+// AssemblyVersionInPackageVersion: Used to understand what version of a package to reference.
+//                                  Then raise the package version of the reference so that its at least the baseline version.
```
ericstj: So long as our JSON parser is ok with this I am fine.  Another pattern folks use is to create properties in the json.

EG:
```json
  "__comment" : [
    "StableVersions: Released versions.",
    "BaselineVersion: Used to determine the minimum version used for all packages in the repo."
    ...
  ]

StephenBonikowsky: If that is the accepted pattern for comments I'm happy to use it.

PR #1937, opened 2017-05-23
Updated baseline for non-shared framework CoreFx packages ns1.3

StephenBonikowsky:
* Using 4.3.0 baseline for System.Reflection.DispatchProxy and System.Security.Principal.Windows in our NetStandard 1.3 dependency tree.

### DIFF
--- a/src/System.Private.ServiceModel/pkg/System.Private.ServiceModel.pkgproj
+++ b/src/System.Private.ServiceModel/pkg/System.Private.ServiceModel.pkgproj
@@ -10,4 +10,15 @@
     <HarvestIncludePaths Include="runtimes/win7/lib/netstandard1.3" />
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+
+  <Target Name="SetDependencyVersion" AfterTargets="GetNuGetPackageDependencies">
+    <ItemGroup>
+      <Dependency Condition="'%(Identity)' == 'System.Reflection.DispatchProxy' AND '%(TargetFramework)' == 'netstandard1.3'">
+        <Version>4.3.0</Version>
+      </Dependency>
+      <Dependency Condition="'%(Identity)' == 'System.Security.Principal.Windows' AND '%(TargetFramework)' == 'netstandard1.3'">
+        <Version>4.3.0</Version>
+      </Dependency>
+    </ItemGroup>
+  </Target>
 </Project>


ericstj:
Can you check in a full baseline rather than a partial one?  That way you aren't forced to keep this in sync with your dependencies if they should grow.

See https://github.com/dotnet/corefx/commit/22fbf8edabe1daf260d5cfe8de8628650f882d02.

pkg/baseline/packageIndex.json
```diff
@@ -113,6 +113,12 @@
         "4.0.3.0": "4.4.0",
         "4.0.4.0": "4.5.0"
       }
+    },
```
weshaggard: Note that this is also going to change the package references for your netstandard2.0 dependencies which you might not want. Please verify that it doesn't cause any issues.

StephenBonikowsky:
@ericstj if I add the full baseline and use PackageIndex to pull it in from our pkgprojs I then have to add [dir.targets](https://github.com/dotnet/corefx/blob/master/pkg/dir.targets#L13) which still doesn't pull it in because I assume that I have to then pull in this new target file. Not sure how far we want to fix this, we really only want it in our release 2.0 branch short term, in our Master I am beginning to work on moving to buildtoolsv2, not sure if that will impact this.

StephenBonikowsky:
@weshaggard You are correct, in our release 2.0 branch it looks like so with the partial baseline...
```
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.NETCore.Platforms" version="2.0.0-preview2-25304-02" />
        <dependency id="System.Net.Http.WinHttpHandler" version="4.4.0-preview2-25304-02" />
        <dependency id="System.Reflection.DispatchProxy" version="4.4.0-preview2-25315-0" />
        <dependency id="System.Security.Principal.Windows" version="4.4.0-preview2-25315-0" />
      </group>
```
In the Master branch is looks like...
```
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.NETCore.Platforms" version="2.0.0-preview2-25304-02" />
        <dependency id="System.Net.Http.WinHttpHandler" version="4.4.0-preview2-25304-02" />
        <dependency id="System.Reflection.DispatchProxy" version="4.4.0-preview1-25323-0" />
        <dependency id="System.Security.Principal.Windows" version="4.4.0-preview1-25323-0" />
      </group>
```
@zhenlan is this minor version change ok?

zhenlan:
The release/2.0.0 branch one looks ok.

The master branch one is unexpected but we have more time to fix it later.

weshaggard:
@StephenBonikowsky perhaps I'm missing something but it looks like both master and release have the same dependencies. Did you make a copy/paste error there?

StephenBonikowsky:
I created #1939 for the release/2.0.0 branch.

@weshaggard I updated my release/2.0.0 branch and re-ran and the only change was the minor version, I pasted the output in the new PR for release/2.0.0 branch.

ericstj:
> @ericstj if I add the full baseline and use PackageIndex to pull it in from our pkgprojs I then have to add dir.targets 

Naw, you just import it in the same place you already import your current index.
https://github.com/dotnet/wcf/blob/8b8829a5a4f6e19758e5cd2cacb218bcaea95dd3/pkg/baseline/baseline.props#L20

StephenBonikowsky:
Thanks @ericstj, that worked but as @weshaggard pointed out it affects 
>  the package references for your netstandard2.0 dependencies

Do we want this?
```
<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
  <metadata minClientVersion="2.12">
    <id>System.Private.ServiceModel</id>
    <version>4.5.0-preview1-25323-0</version>
    <title>System.Private.ServiceModel</title>
    <authors>Microsoft</authors>
    <owners>microsoft,dotnetframework</owners>
    <requireLicenseAcceptance>true</requireLicenseAcceptance>
    <licenseUrl>https://github.com/dotnet/wcf/blob/master/LICENSE.TXT</licenseUrl>
    <projectUrl>https://dot.net/</projectUrl>
    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
    <description>Internal implementation package not meant for direct consumption.  Please do not reference directly.  Provides implementation of System.ServiceModel packages. 
d9cefdc55ed92e4bd7a02daa22624fbe822d3fad 
When using NuGet 3.x this package requires at least version 3.4.</description>
    <releaseNotes>https://go.microsoft.com/fwlink/?LinkID=799421</releaseNotes>
    <copyright>© Microsoft Corporation.  All rights reserved.</copyright>
    <serviceable>true</serviceable>
    <dependencies>
      <group targetFramework=".NETStandard1.3">
        <dependency id="Microsoft.NETCore.Platforms" version="1.2.0-preview1-25323-0" />
        <dependency id="NETStandard.Library" version="1.6.1" />
        <dependency id="System.Collections.NonGeneric" version="4.3.0" />
        <dependency id="System.Collections.Specialized" version="4.3.0" />
        <dependency id="System.ComponentModel.EventBasedAsync" version="4.3.0" />
        <dependency id="System.Linq.Queryable" version="4.3.0" />
        <dependency id="System.Net.NameResolution" version="4.3.0" />
        <dependency id="System.Net.Security" version="4.3.0" />
        <dependency id="System.Net.WebHeaderCollection" version="4.3.0" />
        <dependency id="System.Net.WebSockets" version="4.3.0" />
        <dependency id="System.Net.WebSockets.Client" version="4.3.0" />
        <dependency id="System.Reflection.DispatchProxy" version="4.3.0" />
        <dependency id="System.Reflection.TypeExtensions" version="4.3.0" />
        <dependency id="System.Runtime.Serialization.Primitives" version="4.3.0" />
        <dependency id="System.Runtime.Serialization.Xml" version="4.3.0" />
        <dependency id="System.Security.Claims" version="4.3.0" />
        <dependency id="System.Security.Principal" version="4.3.0" />
        <dependency id="System.Security.Principal.Windows" version="4.3.0" />
        <dependency id="System.Xml.XmlDocument" version="4.3.0" />
        <dependency id="System.Xml.XmlSerializer" version="4.3.0" />
      </group>
      <group targetFramework=".NETStandard2.0">
        <dependency id="Microsoft.NETCore.Platforms" version="1.2.0-preview1-25323-0" />
        <dependency id="System.Net.Http.WinHttpHandler" version="4.4.0-preview1-25323-0" />
        <dependency id="System.Reflection.DispatchProxy" version="4.4.0-preview1-25323-0" />
        <dependency id="System.Security.Principal.Windows" version="4.4.0-preview1-25323-0" />
      </group>
    </dependencies>
  </metadata>
</package>
```

pkg/baseline/packageIndex.json
ericstj: Not necessarily.  Baseline is a minimum version.  Since we update the versions of the references with every release the minimum won't come into the picture here.

ericstj:
I replied to Wes's comment.  The dependencies look ok to me.  With respect to the new platforms package: you don't necessarily need to update that for ns1.3 but it also doesn't hurt.  I explained the purpose of that here: https://github.com/dotnet/wcf/pull/1931#issuecomment-302447990.

StephenBonikowsky:
Should we not worry about fixing this in Master until after we have moved to using buildtoolsv2?

ericstj:
I'd just put the smaller fix in master same as release then remove it once you can use the baseline.

>The dependencies look ok to me. 

Just to connect the two threads, there is a problem in the dependencies that I missed before.  Because you're importing a baseline in your repo without specifying pre-release all those packages are getting the pre-release version of your repo.  We were going to fix that by changing the order of import to bring in the 1.1 index last, then make a buidtools fix to always prefer the last baseline.

ericstj:
👍

PR #1948, opened 2017-05-25
Remove unnecessary content from the common project.json in 2.1.

StephenBonikowsky:
* Since we are currently only building and running for ns2.0 and everything else is being harvested we don't need several sections.


StephenBonikowsky:
PR #1947 is the PR with the same fix for Release2.0 branch.

PR #1958, opened 2017-05-30
Dispose HttpClient instances when closing HttpChannelFactory

mconnew:


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -571,11 +570,13 @@ protected internal override Task OnCloseAsync(TimeSpan timeout)
         protected override void OnClosed()
         {
             base.OnClosed();
-            var httpClientToDispose = _httpClient;
-            if (httpClientToDispose != null)
+            if (_httpClientCache != null && !_httpClientCache.IsDisposed)
             {
-                _httpClient = null;
-                httpClientToDispose.Dispose();
+                lock (_httpClientCache)
```
zhenlan: Do you need a null check after the lock?

zhenlan:
Add @KKhurin for code review.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
KKhurin: Looks like OnClosed could be called by multiple threads and while the base.OnClosed guarantees to notify the handler only once the derived classes have to take care of the concurrency as well.

KKhurin: Also there appears to be a get here before _httpClientCache is even created and then create _httpClientCache on another thread.
The window for this is small though and as far as I can see the worst case is we'll end up with a closed factory holding a reference to MRU cache (which is not worse than before and it would require another lock here which I am ware about).

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -571,11 +570,13 @@ protected internal override Task OnCloseAsync(TimeSpan timeout)
         protected override void OnClosed()
         {
             base.OnClosed();
-            var httpClientToDispose = _httpClient;
-            if (httpClientToDispose != null)
+            if (_httpClientCache != null && !_httpClientCache.IsDisposed)
```
shmao: `_httpClientCache` could become null after the null check and right before the check on `_httpClientCache.IsDisposed` . I think we can use `Interlocked.CompareExchange` here to get the instance and set `_httpClientCache` to null. Then we can also omit the lock below.

mconnew: That's not possible. `OnClosed()` is called from `CommunicationObject.Close` which under a lock saves the current state and transitions it to the Closing state. Only if the previous state was Opened will OnClosed get called. Because of the lock, only a single thread will have the previous state as Opened which means OnClosed will only ever be called once. This is the only place where _httpClientCache is set to null. The null check is there to protect against the scenario of opening the ChannelFactory and closing it before a single request is made. The lock below is there to protect against concurrent modification of _httpClientCache as another thread might be trying to add an HttpClient while Dispose is being called.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
mconnew: As I detailed in an earlier comment, OnClosed can't be called by multiple threads due to protections in CommunicationObject. If another request is started on another thread, it will add the request to a list of outstanding requests before _httpClientCache would be created. The Close call will block waiting for that outstanding request to complete and then Dispose() after that. The factory state is checked before adding the request to the outstanding requests list so a race condition shouldn't be possible.

PR #1971, opened 2017-06-13
Updating NetStandard version for 2.0.0 release.

StephenBonikowsky:
Don't need the dependency on CoreClr, removing it.
Added XmlUpdate step for Standard.


dependencies.props
```diff
@@ -1,15 +1,18 @@
...
     <WCFCurrentRef>c9a03fe0c3309508cb509a4ffdd6dc8b79ed525e</WCFCurrentRef>
+    <StandardCurrentRef>8f8313cbe4b0f4d89c4d0835e08e3e340a18eec0</StandardCurrentRef>
   </PropertyGroup>
 
   <!-- Auto-upgraded properties for each build info dependency. -->
   <PropertyGroup>
-    <CoreFxExpectedPrerelease>preview2-25309-01</CoreFxExpectedPrerelease>
-    <StandardExpectedPrerelease>preview2-25309-01</StandardExpectedPrerelease>
+    <CoreFxExpectedPrerelease>preview2-25405-01</CoreFxExpectedPrerelease>
+    <StandardExpectedPrerelease>preview2-25401-01</StandardExpectedPrerelease>
```
dagood: There isn't an `XmlUpdateStep` with `ElementName` pointing at this property (and `BuildInfoName` specified), so this property won't get updated. I suggest removing it and using `NETStandardPackageVersion` everywhere instead.

dependencies.props
```diff
@@ -31,14 +34,14 @@
       <CurrentRef>$(WCFCurrentRef)</CurrentRef>
     </RemoteDependencyBuildInfo>
     <RemoteDependencyBuildInfo Include="CoreFx">
-      <BuildInfoPath>$(BaseDotNetBuildInfo)corefx/$(DependencyBranch)</BuildInfoPath>
+      <BuildInfoPath>$(BaseDotNetBuildInfo)corefx/$(CoreDependencyBranch)</BuildInfoPath>
       <CurrentRef>$(CoreFxCurrentRef)</CurrentRef>
       <!-- manually pick up a new baseline in order to get package dependencies updated -->
       <DisabledPackages>Microsoft.Private.PackageBaseline;NETStandard.Library</DisabledPackages>
```
dagood: I don't think `NETStandard.Library` is needed here since it's moved to dotnet/standard.

dependencies.props
```diff
@@ -1,15 +1,18 @@
...
 
   <!-- Auto-upgraded properties for each build info dependency. -->
   <PropertyGroup>
-    <CoreFxExpectedPrerelease>preview2-25309-01</CoreFxExpectedPrerelease>
-    <StandardExpectedPrerelease>preview2-25309-01</StandardExpectedPrerelease>
+    <CoreFxExpectedPrerelease>preview2-25405-01</CoreFxExpectedPrerelease>
+    <StandardExpectedPrerelease>preview2-25401-01</StandardExpectedPrerelease>
+    <NETStandardPackageVersion>2.0.0-preview2-25401-01</NETStandardPackageVersion>
+    <NETStandardPackageId>NETStandard.Library</NETStandardPackageId>
+    <MicrosoftNETCoreAppPackageVersion>2.0.0-preview2-25407-01</MicrosoftNETCoreAppPackageVersion>
```
dagood: I'd prefer `MicrosoftNETCoreAppPackageVersion` to be in the next PropertyGroup until it's actually auto-upgraded, just for clarity. (I'm ok with `NETStandardPackageId` because it's obviously static and just associated with `NETStandardPackageVersion`.)

joperezr:
@StephenBonikowsky random question: do you guys plan to move soon to the newer CLI tools? If so, you could get rid of all of the project.jsons that you have on the repo plus you would be able to update buildtools.

dagood:
@joperezr Yep, https://github.com/dotnet/wcf/pull/1960 is some existing work on that, and https://github.com/dotnet/core-eng/issues/793 is tracking me helping get it working to the end.

weshaggard:
Looks reasonable to me other then @dagood's feedback.

dependencies.props
StephenBonikowsky: It is used further down... https://github.com/dotnet/wcf/pull/1971/files#diff-54339bb49aae3cad78638955e8cda9f7L78
And if I remove that DependencyBuildInfo the build fails with tons of types not found.

dagood: I think you should try changing `<Version>2.0.0-$(StandardExpectedPrerelease)</Version>` to `<Version>$(NETStandardPackageVersion)</Version>`, rather than removing the whole DependencyBuildInfo. (This will help in the future because 2.0.0 is no longer hard-coded.)

StephenBonikowsky: That still doesn't work with the same errors.

dagood: I see, it's being used in a few other places that need to be replaced too:
https://github.com/dotnet/wcf/blob/71d7060ebe62dd590b2c35df44b40b56e6885cbd/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj#L68
https://github.com/dotnet/wcf/blob/71d7060ebe62dd590b2c35df44b40b56e6885cbd/dir.targets#L119

I did a github search in `master` and changed the branch to `release/2.0.0` to find those two, so there may be more lurking somewhere.

An alternative is to add a `XmlUpdateStep` for `StandardExpectedPrerelease` so it at least stays in sync with `NETStandardPackageVersion`, but replacing is better in the long run. (The 2.0.0 is specifically something that will change with servicing releases, so even in the servicing branch, it's worth fixing.)

StephenBonikowsky: I think those were it. Amended the PR with those changes. It built locally just fine.

dagood: Cool, LGTM!

PR #1973, opened 2017-06-13
Update the release 2.0.0 branch to reference WCF pkgs built in same b…

StephenBonikowsky:
…ranch.

* Also adding CoreSetup for updating "Microsoft.NETCore.App".
* Commenting out all RemoteDependencyBuildInfo and their matching XmlUpdateSteps.
* As of this PR these dependencies are the official candidates for release 2.0.0.

        - CoreFx packages: 4.4.0-preview2-25405-01
        - "NETStandard.Library": 2.0.0-preview2-25401-01
        - "Microsoft.NETCore.App": 2.0.0-preview2-25407-01
        - WCF Packages: 4.4.0-preview2-25413-01

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -2,61 +2,76 @@
   <!-- Source of truth for dependency tooling: the commit hash of the dotnet/versions master branch as of the last auto-upgrade. -->
   <PropertyGroup>
     <CoreFxCurrentRef>534770cbd98c4083de4d93b809f2e0968f0f856a</CoreFxCurrentRef>
-    <WCFCurrentRef>c9a03fe0c3309508cb509a4ffdd6dc8b79ed525e</WCFCurrentRef>
+    <WCFCurrentRef>a2ba4d6141ef25bdb4c2d2c014b9e1f305a0ea90</WCFCurrentRef>
     <StandardCurrentRef>8f8313cbe4b0f4d89c4d0835e08e3e340a18eec0</StandardCurrentRef>
+    <CoreSetupCurrentRef>c2c072242cb0bdea99efb7349d69492f83200a7d</CoreSetupCurrentRef>
   </PropertyGroup>
 
   <!-- Auto-upgraded properties for each build info dependency. -->
   <PropertyGroup>
     <CoreFxExpectedPrerelease>preview2-25405-01</CoreFxExpectedPrerelease>
+    <WcfExpectedPrerelease>preview2-25413-01</WcfExpectedPrerelease>
     <NETStandardPackageVersion>2.0.0-preview2-25401-01</NETStandardPackageVersion>
     <NETStandardPackageId>NETStandard.Library</NETStandardPackageId>
+    <MicrosoftNETCoreAppPackageVersion>2.0.0-preview2-25407-01</MicrosoftNETCoreAppPackageVersion>
   </PropertyGroup>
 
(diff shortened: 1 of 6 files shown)


StephenBonikowsky:
@weshaggard The only version I'm not completely sure about is for CoreFx packages, I used the last one with preview2 in the name, the most recent one has moved to "preview3".

I also commented out the RemoteDependencyBuildInfos and XmlUpdateSteps as I see done in CoreFx release 2.0.0, I assume this is how you ensure that you don't continue to auto-update the versions of the various dependencies.

If this PR looks good this should provide WCF's official candidates for 2.0.0 release.

dependencies.props
```diff
@@ -2,13 +2,15 @@
...
+    <CoreSetupCurrentRef>c2c072242cb0bdea99efb7349d69492f83200a7d</CoreSetupCurrentRef>
   </PropertyGroup>
 
   <!-- Auto-upgraded properties for each build info dependency. -->
   <PropertyGroup>
     <CoreFxExpectedPrerelease>preview2-25405-01</CoreFxExpectedPrerelease>
+    <WcfExpectedPrerelease>preview2-25413-01</WcfExpectedPrerelease>
     <NETStandardPackageVersion>2.0.0-preview2-25401-01</NETStandardPackageVersion>
     <NETStandardPackageId>NETStandard.Library</NETStandardPackageId>
   </PropertyGroup>
```
dagood: `MicrosoftNETCoreAppPackageVersion` should be moved into this property group now that it's auto-updated.

weshaggard:
>@weshaggard The only version I'm not completely sure about is for CoreFx packages, I used the last one with preview2 in the name, the most recent one has moved to "preview3".

That's correct.

dependencies.props
```diff
@@ -2,13 +2,15 @@
...
-    <WCFCurrentRef>c9a03fe0c3309508cb509a4ffdd6dc8b79ed525e</WCFCurrentRef>
+    <WCFCurrentRef>a2ba4d6141ef25bdb4c2d2c014b9e1f305a0ea90</WCFCurrentRef>
     <StandardCurrentRef>8f8313cbe4b0f4d89c4d0835e08e3e340a18eec0</StandardCurrentRef>
+    <CoreSetupCurrentRef>c2c072242cb0bdea99efb7349d69492f83200a7d</CoreSetupCurrentRef>
   </PropertyGroup>
 
   <!-- Auto-upgraded properties for each build info dependency. -->
   <PropertyGroup>
     <CoreFxExpectedPrerelease>preview2-25405-01</CoreFxExpectedPrerelease>
+    <WcfExpectedPrerelease>preview2-25413-01</WcfExpectedPrerelease>
```
dagood: Technically you can leave out this property (and its upgrader), and project.json upgrades would still happen. Up to you if you think this is useful info to have in dependencies.props! E.g. CoreFX has some \*ExpectedPrerelease properties, but Core-Setup doesn't have any.

dependencies.props
StephenBonikowsky: Will do

dependencies.props
StephenBonikowsky: Yeah, I added it because I though it was handy to have one place we could dependably come to, to know what versions we are using for all package references. I'd like to keep it.

PR #1974, opened 2017-06-15
Move to preview3 prerelease label.

StephenBonikowsky:


zhenlan:
Thanks @StephenBonikowsky. Should we move our corefx/netstandard dependencies to preview3 as well?

StephenBonikowsky:
After we merge this we should see one of the auto generated PRs get triggered to do that.

PR #1975, opened 2017-06-16
Removing throws from no-op code in BufferedReadStream.

### LINKED ISSUE #1701: Workaround bug introduced in XmlReader.Dispose

StephenBonikowsky:
Fixes #1701


StephenBonikowsky:
I don't understand why but apparently #1702 was never actually merged.

PR #1976, opened 2017-06-16
Revert "Move to preview3 prerelease label."

StephenBonikowsky:
Reverts dotnet/wcf#1974


StephenBonikowsky:
Reverting so we can make another fix to the release/2.0.0 using the preview2 label.

PR #1986, opened 2017-06-23
[dev/buildtools-v2.0] Enable all configurations and build packages

dagood:
This fixes the netfx configuration and gets all packages to build.

I bumped System.ServiceModel.Security from `4.0.4.0` to `4.1.0.0`: it uses `PermitInboxRevsion`, which "Permits a higher revision/build to still be considered as a match for an inbox assembly". Increasing major version indicates that net461+ is *not* inbox and makes package validation successful. (See https://github.com/dotnet/buildtools/pull/1411.)

I disabled netfx tests, because it doesn't look like wcf/master is running them. I think this makes sense, since they're only forwarding types to Desktop in netfx. I reran the netcoreapp, uap, and uapaot tests and they were the same as usual.

Here's a diff of the files in the packages compared to `4.5.0-preview1-25420-01`: https://gist.github.com/dagood/de24be43a23e6092e8bfb83add4d6af1. `uap10.1` is now in there--not completely sure if that's what you want or not.


StephenBonikowsky:
This looks great! I have PR #1987 out as well for the scenario tests.
If you merge this I can rebase and update my PR, I also have an issue for you to look at in my PR.

dagood:
Cool, I'll go ahead and merge and take a look.

PR #1987, opened 2017-06-23
Enable Scenario tests for buildtoolsv2.

StephenBonikowsky:
* Make sure the service gets triggered and setup for the tests.
* All the scenario test projects should build and run.


StephenBonikowsky:
@dagood All the test projects build and run in my local Windows environment for netcoreapp.
To trigger the OuterLoop tests I had to use the following command...
`build-managed.cmd -tests --outerloop`

I get three test case failures for 'Security.TransportSecurity.Tests', which I think may be due to how things are being bin-placed...
`System.IO.FileLoadException : Could not load file or assembly 'System.Security.Principal.Windows, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The located assembly's manifest definition does not match the assembly reference.`

Scanning the bin folder I find seven 'System.Security.Principal.Windows.dll' and it looks like 2 distinct versions, 4.1.1.0 and 4.1.0.0, my guess is that it is compiling against 4.1.1.0 and trying to run against 4.1.0.0.

Any ideas on how to fix this?

dagood:
I did have assembly load problems working on other tests, but I don't have a solid way to approach them other than tracking down which packages (and therefore `external/` projects) they come from. I did write a quick powershell to associate which packages the files in bin came from:

`$n = 'System.Security.Principal.Windows.dll'; ((ls -Recurse -Include $n C:\git\wcf\bin) + (ls -Recurse -Include $n C:\git\wcf\packages)) | %{ "" + (Get-FileHash $_ -Algorithm MD5).Hash +" "+ $_.FullName } | sort`

Your guess does sound reasonable. I'll have some time to look at it myself tomorrow.

dagood:
Here's what I found:

```
B52CBF5C431C30F4EF746A5A68680516 (4.1.0.0)
bin\runtime\netcoreapp-Windows_NT-Debug-x64\System.Security.Principal.Windows.dll
bin\testhost\netcoreapp-Windows_NT-Debug-x64\shared\Microsoft.NETCore.App\9.9.9\System.Security.Principal.Windows.dll
bin\Windows_NT.AnyCPU.Debug\runtime\netcoreapp\System.Security.Principal.Windows.dll
packages\runtime.win-x64.microsoft.netcore.app\2.0.0-preview2-25319-02\runtimes\win-x64\lib\netcoreapp2.0\System.Security.Principal.Windows.dll

A6E731B7C83B405E8DB2BD48B329398C (4.1.1.0)
bin\AnyOS.AnyCPU.Debug\core-setup\netcoreapp\System.Security.Principal.Windows.dll
bin\AnyOS.AnyCPU.Debug\core-setup\netstandard\System.Security.Principal.Windows.dll
bin\ref\netcoreapp\System.Security.Principal.Windows.dll
bin\ref\netstandard\System.Security.Principal.Windows.dll
packages\system.security.principal.windows\4.5.0-preview2-25319-04\ref\netstandard2.0\System.Security.Principal.Windows.dll
packages\system.security.principal.windows\4.5.0-preview2-25319-04\ref\uap10.1\System.Security.Principal.Windows.dll
```

This shows me that the copy of `System.Security.Principal.Windows.dll` we're getting from CoreFX mismatches the NETCoreApp package, and the mismatch goes over an assembly version bump. Fixing it is like https://github.com/dotnet/wcf/commit/2a67a58740a4ea7708e43aa805466a84f0b6c735: in core-setup.depproj, change the CoreFX package versions to 4.4.0-preview2-25316-01.

A few things are rough:
 * You might want to split these CoreFX packages in core-setup.depproj into a new external/corefx/corefx.depproj for clarity.
 * The CoreFX package versions should definitely be centralized in dependencies.props.

There's a problem with auto-upgrading these, though... upgrading takes the latest build of each product, so there will usually be a small mismatch between the CoreFX refs and Core-Setup runtimes you get. This will be a problem when a CoreFX assembly version is revved and Core-Setup hasn't completed a build with that package yet.

A fix would be switching from depending on Core-Setup to only depending on CoreFX. That way, there's a coherent latest build to upgrade to. This is technically possible, but puts WCF in a strange place where it depends on the `Microsoft.Private.CoreFx.NETCoreApp/UAP` packages which we don't ship on the NuGet gallery--only in MyGet for daily builds.

A better fix, I think, would be making the upgrade smarter: instead of taking latest CoreFX, find which build the current Microsoft.NETCore.App/UWP package depends on, and use those versions. The info is available, so I'd say this is feasible. (Check out `packages\microsoft.netcore.app\2.0.0-preview2-25319-02\Microsoft.NETCore.App.versions.txt`: it has the dotnet/versions commit hash for the CoreFX version it used.) This would be be BuildTools work.

StephenBonikowsky:
I rebased and now none of the facades are resolving "netstandard" or "System.Runtime", for example...

`Could not resolve this reference. Could not locate the assembly "System.Runtime". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. [C:\OSS\WCF\src\System.ServiceModel.Http\src\System.ServiceModel.Http.csproj]`

StephenBonikowsky:
Ok, locally I can...
build-managed.cmd
build-managed.cmd -tests (runs unit tests)
build-managed.cmd -tests --outerloop (runs outerloop tests)

Running for netcoreapp I have clean builds and runs.

@dagood could you run for uap? I would expect some errors but the tests should at least run.

PR #1990, opened 2017-06-23
Updating various scripts used for Unix.

StephenBonikowsky:


StephenBonikowsky:
Just updating these files to the latest content from CoreFx.
Need these to be updated for CI in master.

PR #2011, opened 2017-06-28
Depend on packages from CoreFX, not Core-Setup

dagood:
WCF outputs are used in the Core-Setup build, so it makes sense to use CoreFX and CoreCLR "private" packages even though they aren't published to the NuGet gallery. Even though the WCF bits in the Core-Setup packages weren't being used, the Core-Setup dependency was circular from the perspective of product outputs.

Upgrading CoreFX stops us from using an old version of `System.Net.Http.WinHttpHandler` that was breaking UAP builds with some unwanted dependencies.

After getting netcoreapp, uap, and uapaot tests running, I also made sure `-allconfigurations` still worked and produced the expected package contents.

@StephenBonikowsky


StephenBonikowsky:
@dotnet-bot test this please

Issue #2019, opened 2017-06-30
[uapaot] Method 'InvalidEnumArgumentException..ctor' was not included in compilation


shmao:
Ran tests with command `.\build.cmd -framework:uapaot -tests -outerloop`. Many tests failed due to errors like the one shown below,

> System.MissingMethodException : Method 'InvalidEnumArgumentException..ctor(string, int, Type)' was not included in compilation, but was referenced in WebSocketTransportUsageHelper.Validate(WebSocketTransportUsage). There may have been a missing assembly.

shmao:
The issue may be related to https://github.com/dotnet/wcf/pull/2011. 

/cc: @dagood

dagood:
Yes, I think I noticed this go away when CoreFX runtime assemblies are copied from the private package into the ILCInputFolder. (I'm working through Jose's list of corrections.)

shmao:
@dagood sounds good. Thanks!

Issue #1480, opened 2016-08-17
Vs 2015 MS Test Report error  Run Common Language Runtime detected an invalid program.


som-nitjsr:
while  Debuging  the test it works OK. But if I Run the Test It give me this Error.

Result StackTrace:  
at System.ServiceModel.Security.X509CertificateRecipientClientCredential..ctor(X509CertificateRecipientClientCredential other)
   at System.ServiceModel.Description.ClientCredentials..ctor(ClientCredentials other)
   at System.ServiceModel.Description.ClientCredentials.CloneCore()
   at System.ServiceModel.Description.ClientCredentials.Clone()
   at System.ServiceModel.Description.ClientCredentials.CreateSecurityTokenManager()
   at System.ServiceModel.Channels.HttpChannelFactory`1.InitializeSecurityTokenManager()
   at System.ServiceModel.Channels.HttpChannelFactory`1.OnOpen(TimeSpan timeout)
   at System.ServiceModel.Channels.HttpsChannelFactory`1.OnOpen(TimeSpan timeout)
   at System.ServiceModel.Channels.CommunicationObject.Open(TimeSpan timeout)
   at System.ServiceModel.Channels.ServiceChannelFactory.TypedServiceChannelFactory`1.OnOpen(TimeSpan timeout)
   at System.ServiceModel.Channels.CommunicationObject.Open(TimeSpan timeout)
   at System.ServiceModel.ChannelFactory.OnOpen(TimeSpan timeout)
   at System.ServiceModel.Channels.CommunicationObject.Open(TimeSpan timeout)
   at System.ServiceModel.ChannelFactory.EnsureOpened()
   at System.ServiceModel.ChannelFactory`1.CreateChannel(EndpointAddress address, Uri via)
   at System.ServiceModel.ChannelFactory`1.CreateChannel()
   at System.ServiceModel.ClientBase`1.CreateChannel()
   at System.ServiceModel.ClientBase`1.CreateChannelInternal()
   at System.ServiceModel.ClientBase`1.get_Channel()
   at System.ServiceModel.ClientBase`1.get_InnerChannel()
   at VMS.SF.Gateway.Tests.GatewayTests.GetSimpleResponseUsingSoap(String userName, String password) in C:\vms\sourcecode\SFA\Server\Dev\Src\WSF2\SFServer\Tests\VMS.SF.Gateway.Tests\GatewayTests.cs:line 225
[...]

PR #2021, opened 2017-06-30
depproj fixes for uapaot framework testing

dagood:
* Get clrcompression.dll from CoreFX package
 * Create corefx-runtime.depproj for specific CoreFX assembly placement

Applied @joperezr's suggestions, and they worked: `build.cmd -framework=uapaot -tests` runs cleanly on my machine.

Something I didn't do is move the `Microsoft.Private.CoreFx.UAP` ref copying into a new depproj: I left it in `netstandard.depproj` for now. For clarity, it makes sense to me to move it to a new depproj (the `uap;` in `netstandard/Configurations.props` seems weird) but it didn't seem needed. I couldn't figure out how to make a single `corefx.depproj` that deploys both ref and runtimes, otherwise I would have done that.

Fixes https://github.com/dotnet/wcf/issues/2019

### DIFF
--- a/external/corefx-runtime/corefx-runtime.depproj
+++ b/external/corefx-runtime/corefx-runtime.depproj
@@ -1,0 +1,30 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <PropertyGroup>
+    <!-- support cross-targeting by choosing a RID to restore when running on a different machine that what we're build for -->
+    <NugetRuntimeIdentifier Condition="'$(OSGroup)' == 'Windows_NT' AND '$(RunningOnUnix)' == 'true'">win7-x64</NugetRuntimeIdentifier>
+    <NugetRuntimeIdentifier Condition="'$(OSGroup)' == 'Unix' AND '$(RunningOnUnix)' != 'true'">ubuntu.14.04-x64</NugetRuntimeIdentifier>
+    <NugetRuntimeIdentifier Condition="'$(TargetGroup)' == 'uap'">win10-$(ArchGroup)</NugetRuntimeIdentifier>
+    <NugetRuntimeIdentifier Condition="'$(TargetGroup)' == 'uapaot'">win10-$(ArchGroup)-aot</NugetRuntimeIdentifier>
+    <RidSpecificAssets>true</RidSpecificAssets>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="@(RefAndRuntimePackageReference)" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetGroup)' == 'netstandard' or
+                        '$(TargetGroup)' == 'netcoreapp'">
+    <PackageReference Include="System.Security.Principal.Windows">
+      <Version>$(MicrosoftPrivateCoreFxNETCoreAppPackageVersion)</Version>
+    </PackageReference>
+
+    <PackageReference Include="System.Net.Http.WinHttpHandler"
+                      Condition="'$(TargetsWindows)'=='true'">
+      <Version>$(MicrosoftPrivateCoreFxNETCoreAppPackageVersion)</Version>
+    </PackageReference>
+  </ItemGroup>
+
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
+</Project>
--- a/external/corefx-runtime/corefx-runtime.depproj
+++ b/external/corefx-runtime/corefx-runtime.depproj
@@ -14,8 +14,7 @@
     <PackageReference Include="@(RefAndRuntimePackageReference)" />
   </ItemGroup>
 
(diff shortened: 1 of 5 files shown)


external/corefx-runtime/corefx-runtime.depproj
```diff
@@ -0,0 +1,30 @@
...
+    <NugetRuntimeIdentifier Condition="'$(TargetGroup)' == 'uap'">win10-$(ArchGroup)</NugetRuntimeIdentifier>
+    <NugetRuntimeIdentifier Condition="'$(TargetGroup)' == 'uapaot'">win10-$(ArchGroup)-aot</NugetRuntimeIdentifier>
+    <RidSpecificAssets>true</RidSpecificAssets>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="@(RefAndRuntimePackageReference)" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetGroup)' == 'netstandard' or
```
joperezr: Small NIT: you can remove the netstandard check as that is not one of your configurations

external/runtime/runtime.depproj
```diff
@@ -48,22 +48,6 @@
     </PackageReference>
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="@(RefAndRuntimePackageReference)" />
```
joperezr: question: Do you even use this RefAndRuntimePackageReference any longer? If not, you can probably clean it up now. from external/dir.props

external/test-runtime/XUnit.Runtime.depproj
```diff
@@ -157,19 +157,6 @@
...
-    <!--
-      Copy clrcompression.dll into ILCInputFolder. This is needed in WCF and not
-      CoreFX: in CoreFX, clrcompression.dll is a build output.
-
-      Without clrcompression.dll, the xunit runner no-ops when using RunTests.
-      It works during a vertical build because Run.cmd adds environment
-      variables that allow the runner to find the BuildTools clrcompression.dll.
-    -->
-    <Copy Condition="'$(TestILCFolder)'!=''"
-          SourceFiles="$(TestILCFolder)\ExternalFiles\Framework\clrcompression.dll"
```
joperezr: Thanks for fixing this! Note that now, you should be able to add a CI that makes sure that this vertical build is not broken since by removing this dependency we don't require restoring the optional.json for building anymore. We only need it now for running tests.

external/runtime/runtime.depproj
dagood: It's shared by `corefx-runtime.depproj` and `netstandard.depproj` now. If you know how to make a single `corefx.depproj` that deploys both refs and runtimes, then I'd love to get rid of it--wasn't able to find a way.

I looked at having two `BinPlaceConfiguration` items with different metadata, but I didn't see what I could set to have them resolve refs vs. runtimes.

shmao:
@dagood after your fix on Wednesday, #2011, running tests on `uap` worked fine on my dev machine. Would you mind trying `build.cmd -framework=uap -tests` locally and see if it still works with this fix? Thanks!

external/corefx-runtime/corefx-runtime.depproj
```diff
@@ -0,0 +1,30 @@
...
+                        '$(TargetGroup)' == 'netcoreapp'">
+    <PackageReference Include="System.Security.Principal.Windows">
+      <Version>$(MicrosoftPrivateCoreFxNETCoreAppPackageVersion)</Version>
+    </PackageReference>
+
+    <PackageReference Include="System.Net.Http.WinHttpHandler"
+                      Condition="'$(TargetsWindows)'=='true'">
```
zhenlan: Should we add the condition that this is not uap or uapaot?

external/runtime/runtime.depproj
joperezr: I believe there is no way to do it with BinPlaceConfigurations, but I have done this in the past by adding a target to the depproj that runs after RestorePackages that would copy the one that is not copied automatically. After RestorePackages, `@(Reference)` will contain the ref assemblies and `@(ReferenceCopyLocal)` will contain the runtime assemblies, so after just running one restore you could theoretically copy both to the right folder.

external/corefx-runtime/corefx-runtime.depproj
dagood: It's not the clearest thing, but this entire ItemGroup is conditioned on `'$(TargetGroup)' == 'netcoreapp'` so it's already checked.

dagood:
@shmao Yep, `uap` still works, also `netcoreapp` and `-allconfigurations`, thanks for the reminder.

I like that idea, @joperezr. I'm not familiar with CI, though--@StephenBonikowsky, is this something you could do? (After this gets merged?)

StephenBonikowsky:
Already working on it, we only have one test failure right now though so this PR is good.
The failure is an infrastructure test that verifies the runtime OS but we are using a property that no longer exists in BuildTools v2.
See #1480

StephenBonikowsky:
I can work on adding the runs Jose mentioned as well.

dagood:
Thanks, merging!

external/corefx-runtime/corefx-runtime.depproj
zhenlan: Got it. Thanks!

external/runtime/runtime.depproj
dagood: Adding a custom target for that sounds uncomfortable to me. I gave just adding a new depproj a try in https://github.com/dotnet/wcf/compare/master...dagood:corefx-ref-depproj and it seems to work. Does it make sense to factor it that way?

PR #2027, opened 2017-07-03
Split corefx ref depproj from netstandard.depproj

dagood:
See https://github.com/dotnet/wcf/pull/2021#discussion_r125087923

This is an alternative way I thought of to get those corefx references separate. I figure it would be better to discuss the idea in a fresh PR rather than that review thread. :smile:


external/corefx/ref/corefx.ref.depproj
```diff
@@ -0,0 +1,43 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build">
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <PropertyGroup>
+    <BinPlaceRef>true</BinPlaceRef>
+    <NuGetDeploySourceItem>Reference</NuGetDeploySourceItem>
+    <NugetRuntimeIdentifier>None</NugetRuntimeIdentifier>
```
joperezr: Nit: I believe you don't need this as this is the default.

dagood:
Hmm, I didn't consider speed.

On maintainability, I was thinking this would actually be *more* maintainable, because there's no custom target with new logic to understand. Right now, I don't know what that custom target would contain, but it was easy to make these two projects.

joperezr:
We have done that custom target in corefx in the past, actually we have an example where we do exactly this same thing for SharedFrameworkValidation. [This is how the target](https://github.com/dotnet/corefx/blob/master/external/SharedFrameworkValidation/harvestPackages.SharedFramework.depproj#L22-L28) looks like.

Given it's a simple target, that's why I thought one project should be enough, but again, if you feel strongly about the design you have now, it's fine with me.

dagood:
Thanks for the pointer! Tried it out, tested netcoreapp/uap/uapaot: https://github.com/dotnet/wcf/compare/master...dagood:combined-corefx-depproj. Less code and better perf (although I didn't measure)--I think it's better. I'm not happy that it needs a custom target, but agreed, it's tiny. Closing this in favor of that.

PR #2029, opened 2017-07-03
Restore CoreFX runtimes and refs in one depproj

dagood:
Take the CoreFX ref-copy code out of netstandard.depproj and put it in the CoreFX depproj. The CoreFX depproj now copies runtime bits, but also copies ref bits with a custom target.

See https://github.com/dotnet/wcf/pull/2027


StephenBonikowsky:
LGTM

joperezr:
Merging this now since I have to do some work on top of it.

PR #2041, opened 2017-07-06
Adding uapvNext support to System.Private.ServiceModel

joperezr:
System.Private.ServiceModel didn't support uapvNext, which was causing the restore of that TFM to pick the netcore50 assets instead of the netstandard2.0 ones. This change fixes that.

cc: @dagood @StephenBonikowsky @ericstj


StephenBonikowsky:
LGTM

PR #2043, opened 2017-07-06
Enable NegotiateStream use in uap and enable tests

mconnew:
This removes the UAP specific code paths and uses the CoreClr code path on UAP


mconnew:
Fixes #1235

PR #2044, opened 2017-07-07
Fixing System.Private.ServiceModel package for uap consumption with the right asset

joperezr:
cc: @dagood @StephenBonikowsky @zhenlan 

Ok so this addresses several issues for the System.Private.ServiceModel package:

- References for uap10.0.15138 TFM were wrong, includding things it shouldn't
- We used to have an asset for unix/uap10.0.15138, which should never be the  case, so I removed it.
- We were basically copying the asset from netstandard2.0 into uap, which is wrong, as uap has it's own implementation

With these fixes in, I believe we now should be good to go to consume WCF from core-setup and produce the UWP metapackage.

### DIFF
--- a/src/System.Private.ServiceModel/src/Configurations.props
+++ b/src/System.Private.ServiceModel/src/Configurations.props
@@ -2,13 +2,13 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <PackageConfigurations>
+      uap-Windows_NT;
       netstandard-Windows_NT;
       netstandard-Unix;
     </PackageConfigurations>
     <BuildConfigurations>
       $(PackageConfigurations);
       netstandard;
-      uap-Windows_NT;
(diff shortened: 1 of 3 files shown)


src/System.Private.ServiceModel/src/Configurations.props
```diff
@@ -2,13 +2,13 @@
...
   <PropertyGroup>
     <PackageConfigurations>
+      uap-Windows_NT;
       netstandard-Windows_NT;
       netstandard-Unix;
     </PackageConfigurations>
     <BuildConfigurations>
       $(PackageConfigurations);
       netstandard;
-      uap-Windows_NT;
```
shmao: Just question: why don't we need `uap` in BuildConfigurations?

joperezr: Because BuildConfigurations imports PackageConfigurations in like 10

shmao: Ah, I didn't see that. Thanks.

StephenBonikowsky:
LGTM

PR #2048, opened 2017-07-08
Change to use S.Net.Socket and SslStream instead of StreamSocket on UAP

### LINKED ISSUE #2035: SocketConnection.Write Throw Different Exception on uapaot
See https://github.com/dotnet/wcf/pull/2033.

`System.ServiceModel.Channels.SocketConnection.Write` calls different methods on different platforms: 
On NetCore it calls `CoreClrSocketConnection.WriteCore`; 
On uap/uapaot, it calls `RTSocketConnection.WriteCore`. 

The two `WriteCore` use different networking APIs. As the networking APIs throw different exceptions, we get different exceptions in error cases. Do we want to consider to make the exceptions consistent across platforms?

mconnew:
Fixes #2035


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpsChannelFactory.cs
```diff
@@ -237,10 +237,23 @@ private static void SetCertificate(ServiceModelHttpMessageHandler handler, Secur
...
         }
 
+        private static void ValidateClientCertificate(X509Certificate2 certificate)
+        {
+#if FEATURE_NETNATIVE
+            bool found;
+            X509CertificateInitiatorClientCredential.TryGetUapCertificate(certificate, out found);
+            if (!found)
+            {
+                throw ExceptionHelper.PlatformNotSupported("Directly setting the Certificate is not supported for HTTP yet. Use X509CertificateInitiatorClientCredential.SetCertificate instead");
```
zhenlan: Should we use resource instead?

mconnew: We generally haven't used resources for PNSE exceptions.

mconnew:
I spoke with @zhenlan in person and he's okay with this going in as is.

PR #2049, opened 2017-07-09
Cherry pick set up tests fixes

hongdai:
…issue

* The setup tests do not add much value today as we output certification
installation trace that help identify certificate setup failures.
* The set up tests are also fragile because it does not use the same way
we install the certificates on Linux/Osx machines, and it's non trial to make
them the same way.


hongdai:
I have tested the fix locally. Since there is no CI for the release 2.0.0 branch, I will check it in once someone approves it.

PR #2064, opened 2017-07-13
Move target needed for test.

StephenBonikowsky:
* Target named GenerateTestSharedFrameworkDepsFile is needed for test but was located in src.builds which is the product build path.
* Moving it to tests.builds which is where it should be.
* This fixes an issue in VSTS where we have to build product and test in separate build definitions.

### DIFF
--- a/src/GenerateDepsJson.targets
+++ b/src/GenerateDepsJson.targets
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <UsingTask TaskName="GenerateDepsJson" AssemblyFile="$(BuildToolsTaskDir)Microsoft.DotNet.Build.Tasks.dll"/>
+
+  <!-- After we build all the source libraries we need to generate a deps.json file for the shared test framework -->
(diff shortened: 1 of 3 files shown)


dagood:
My interpretation of the CI is that it's actually needed in both places to work in all scenarios. Maybe it should live in `dir.traversal.targets`?

StephenBonikowsky:
That seems to be the next best choice.

StephenBonikowsky:
Actually doing it differently, I think it makes for a cleaner design and doesn't clutter up dir.traversal.targets

chaseWillden:
@StephenBonikowsky This might have been a breaking commit...

StephenBonikowsky:
@chaseWillden what build is broken?

chaseWillden:
@StephenBonikowsky. The one released on here: https://marketplace.visualstudio.com/items?itemName=WCFCORETEAM.VisualStudioWCFConnectedService. 

See issues #2079 #2078 #2077 #2076 

Good question, it might have been a VS build. On 7/14/2017 it worked great and today (7/17/2017) it broke down. I haven't read through the code to identify anything further.

StephenBonikowsky:
Adding @zhenlan

zhenlan:
@chaseWillden sorry for running into the issue with our latest update of WCF Connected Service. We are looking into it and will update in the issues you mentioned. This PR is not related to the WCF connected service issue.

PR #2071, opened 2017-07-14
Fix RssReader state machine when there isn't white-space node separation.

drago-draganov:


drago-draganov:
The pull request is by mistake

PR #2110, opened 2017-07-27
Build single scenario test project for uapaot.

### LINKED ISSUE #2089: Merge test projects into less number of projects to reduce build overhead
Currently we have so many different test projects, which caused big overhead in building them individually. Especially in uapaot build, the overhead is huge. Current uapaot build can take half hour to an hour while the build time could be reduced to a few minutes if we had merged all test projects into one.

StephenBonikowsky:
* Because of how long it takes to run a project through the ILC toolchain, collapse all the scenario test projects into a single test project for uapaot builds.
* Allow this option to be ignored in VSTS so that test result reporting in MC still looks good.
* Can still override locally if separate projects are desired.
* Had to update the expected test ILC package version.
* Fixes #2089

### DIFF
--- a/src/tests.builds
+++ b/src/tests.builds
@@ -10,11 +10,8 @@
     <GenerateCodeCoverageReportForAll>true</GenerateCodeCoverageReportForAll>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(RunningSharedFrameworkValidation)'=='true'">
-    <TestProjectExclusions Include="$(MSBuildThisFileDirectory)*\tests\**\System.Data.Odbc.Tests.csproj" />
-    <TestProjectExclusions Include="$(MSBuildThisFileDirectory)*\tests\**\System.IO.Pipes.Tests.csproj" />
-    <TestProjectExclusions Include="$(MSBuildThisFileDirectory)*\tests\**\Microsoft.XmlSerializer.Generator.Tests.csproj" />
-    <TestProjectExclusions Include="$(MSBuildThisFileDirectory)*\tests\**\System.Reflection.Context.Tests.csproj" />
+  <ItemGroup>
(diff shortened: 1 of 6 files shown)


StephenBonikowsky:
@dagood I had to update the test ilc package version in the dependencies.props file to match what was getting pulled down by...
`
image
`
Is this expected?

src/tests.builds
```diff
@@ -32,7 +29,17 @@
   </PropertyGroup>
   
   <ItemGroup>
-    <Project Include="$(MSBuildThisFileDirectory)*\tests\**\*$(_ProjectPattern).csproj" Exclude="@(TestProjectExclusions)" />
+    <!--If a uapaot run use the single scenario test project by default-->
+    <Project Include="$(MSBuildThisFileDirectory)System.Private.ServiceModel\tests\Scenarios\AllScenarioTests.csproj" Condition="'$(TargetGroup)'=='uapaot' and '$(BuildSeparateScenarioProjects)'=='false'" />
```
zhenlan: I would use '$(BuildSeparateScenarioProjects)'!='true' so BuildSeparateScenarioProjects=blah will be considered false.

dagood:
> I had to update the test ilc package version in the dependencies.props file to match

Based on the way dependencies.props is currently set up, this is expected: there are some `XmlUpdateStep`s currently commented that would upgrade the properties automatically. Uncommenting them makes sense: https://github.com/StephenBonikowsky/wcf/blob/13d9a443be2ef304a07bebaf29ecf87c4159ee4e/dependencies.props#L101-L115. Ideally they would have been uncommented when the `RemoteDependencyBuildInfo`s were uncommented.

What's happening now is that the `RemoteDependencyBuildInfo`s upgrade `optional.json`, but the properties aren't being updated because the `XmlUpdateStep`s don't exist.

StephenBonikowsky:
Thanks @dagood , I'll get another commit going to enable auto-update.

src/tests.builds
StephenBonikowsky: Updated.

PR #2127, opened 2017-07-28
Fix tests hanging waiting for async responses.

### LINKED ISSUE #2124: Client.ChannelLayer.Tests hanging

StephenBonikowsky:
* This is a cherry-pick of the fix that went into Master branch (PR #2125)
* Force these tests to not run in parallel.
* Fixes #2124


StephenBonikowsky:
Approved by .NET Native Shiproom

Issue #1965, opened 2017-06-06
The content type text/xml; charset=\"utf-8\" of the response message does not match the content type of the binding (text/xml; charset=utf-8)


crackedcornjimmy:
I am calling a SOAP WCF Service from a new .Net Core WebApi application and am getting the following error:

"The content type text/xml; charset=\"utf-8\" of the response message does not match the content type of the binding (text/xml; charset=utf-8).
If using a custom encoder, be sure that the IsContentTypeSupported method is implemented properly. 
The first 1024 bytes of the response were:..."

If I run the application and watch it on Fiddler, I see proper XML data come back. We're good there. Sure enough, the Request Content-Type header is text/xml; charset=utf-8 and the response Content-Type Header is Content-Type: text/xml; charset="utf-8". So yes...there are quotation marks there.

If I peruse the internet, I find the following Stack Overflow question:

https://stackoverflow.com/questions/41684342/the-content-type-text-xml-charset-utf-8-of-the-response-message-does-not-matc

This is exactly my problem and the user of the SO site utilized the following solution:

https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/CustomTextMessageEncoder.cs

The thing is, I don't know how to wire that up into my application to get it to work. Can someone provide examples?

Please let me know if there is an easier way or if more information is needed.

crackedcornjimmy:
This same question was asked [here](https://github.com/dotnet/wcf/issues/1684) with no luck.

anupmo:
Is there any solution yet for the above issue?

Kcaxtanhu:
Does any one succeeded on this attempt? I followed @KKhurin thread #2137, but after implementing the Costum Handler I've got the bellow error:

`System.ServiceModel.FaultException: 
         
   at System.ServiceModel.Channels.ServiceChannel.HandleReply(ProxyOperationRuntime operation, ProxyRpc& rpc)
   at System.ServiceModel.Channels.ServiceChannel.EndCall(String action, Object[] outs, IAsyncResult result)
   at System.ServiceModel.Channels.ServiceChannelProxy.TaskCreator.<>c__DisplayClass1_0.<CreateGenericTask>b__0(IAsyncResult asyncResult)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #2137, opened 2017-08-01
Adding support for quoted char set in Content-Type

KKhurin:
The fix is to simply match the quotes handing logic in IsCharSetSupported() to the one already implemented in GetEncodingFromContentType().

Tests to follow shortly...


KKhurin:
This will fix #1965.

PR #2155, opened 2017-08-04
Fixes to tests to enable UWP certificate scenarios to succeed

mconnew:
Move forward the dependent versions to pick up latest networking libraries and modify test security helper to account for differences in behavior on UWP.


mconnew:
We still have one more failing test which is a product issue. When certificate validation fails, we aren't interpreting the error code from the HttpRequestException and are throwing a CommunicationException when a SecurityNegotiationException was expected.

StephenBonikowsky:
LGTM

PR #2177, opened 2017-08-09
Ensure new UAP CI Job has unique name.

StephenBonikowsky:
The name of the new job as is conflicted with an existing job, need to make it unique so it creates a separate job.


StephenBonikowsky:
test ci please

PR #2195, opened 2017-08-14
Use specific package references rather than metapackage umbrella.

jimmyca15:


ElJerry:
LGTM 👍

lab/src/Microsoft.SyndicationFeed/src/Microsoft.SyndicationFeed.csproj
```diff
@@ -3,7 +3,15 @@
     <VersionPrefix>1.0.0</VersionPrefix>
     <VersionSuffix>preview</VersionSuffix>
```
zhenlan: I know this is not part if this PR... I am wondering if this should be "preview1" instead. Will this be part of package version?

jimmyca15: Yes it will. We could make it _preview1_ if it is more in line with other packages, but _preview_ should be fine as well.

PR #2239, opened 2017-08-25
Exlude test failed by active issue 2147

hongdai:
* The UPN test on Linux/Osx is failing because of issue 2147


StephenBonikowsky:
LGTM

PR #2250, opened 2017-08-29
Implement AtomWriter.

drago-draganov:


lab/src/Microsoft.SyndicationFeed/src/Atom/AtomFeedReader.cs
```diff
@@ -14,7 +14,7 @@ public class AtomFeedReader : SyndicationFeedReaderBase
         private bool _knownFeed;
 
         public AtomFeedReader(XmlReader reader)
-            : this(reader, new AtomFeedParser())
```
jimmyca15: Perhaps AtomFeedReader should be renamed AtomReader to follow the same pattern.

jimmyca15:
The Atom writer implementation looks good. Tests as well.

PR #2285, opened 2017-09-13
Sample certificate auth on OSX similar to peer channel

mconnew:
Also improved OS detection on OSX and fixed client certificate installation on OSX

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/CertificateManager.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/CertificateManager.cs
@@ -1,7 +1,9 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
 
 using System;
+using System.IO;
 using System.Security.Cryptography; 
 using System.Security.Cryptography.X509Certificates;
 using System.Text;
@@ -58,6 +60,22 @@ namespace Infrastructure.Common
             }
         }
 
+        internal static string OSXCustomKeychainFilePath
+        {
+            get
+            {
+                return Path.Combine(Environment.CurrentDirectory, "wcfLocal.keychain");
+            }
+        }
+
+        internal static string OSXCustomKeychainPassword
+        {
+            get
+            {
+                return "WCFKeychainFilePassword";
+            }
+        }
+
         // Adds the given certificate to the given store unless it is
         // already present.  Returns the  certificate either already in
         // the store or the one requested.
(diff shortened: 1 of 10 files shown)


mconnew:
This still needs a little bit of cleanup so not quite ready for merging/review

mconnew:
This is ready for review now. I have verified the latest changes run on OSX as the OSX certificate tests aren't currently running correctly on OSX ci tests. We have some failures, but they aren't due to this change.

src/System.Private.ServiceModel/tests/Common/Infrastructure/CertificateManager.cs
```diff
@@ -104,6 +122,41 @@ internal static StoreLocation PlatformSpecificRootStoreLocation
             return resultCert;
         }
 
+        // Adds the given certificate to the given keychain unless it is
+        // already present.  Returns the  certificate either already in
+        // the store or the one requested.
+        public static X509Certificate2 AddToOSXKeyChainIfNeeded(SafeKeychainHandle keychain,
+                                                          X509Certificate2 certificate)
```
hongdai: Seems it's only used to install cert to a custom keychain, could we rename the method to something like AddToOSXCustomKeyChainIfNeeded?

mconnew: I don't understand, the method is already called `AddToOSXKeyChainIfNeeded`?

hongdai: There is a system KeyChain on OSX and this method is not for system KeyChain. I want to reflect it in the method name.

hongdai:
>             new Tuple<string, OSID>("osx", OSID.AnyOSX),

Could you add below to detect osx.10.12 from test runtime? This is for consistence.

            new Tuple<string, OSID>("osx.10.12", OSID.OSX_10_12),


---
Refers to: src/System.Private.ServiceModel/tests/Common/Infrastructure/OSHelper.cs:48 in 24a367b. [](commit_id = 24a367b24df0813cfbf822e654c411d04d63ca20, deletion_comment = False)

hongdai:
None of my comments are blocking.

src/System.Private.ServiceModel/tests/Common/Infrastructure/CertificateManager.cs
mconnew: You can create an instance of SafeKeychainHandle for the system keychain and use this method if you want and it will work.

src/System.Private.ServiceModel/tests/Common/Infrastructure/ServiceUtilHelper.cs
```diff
@@ -211,6 +217,60 @@ public static void EnsureClientCertificateInstalled()
...
+
+                    s_peerCertificate = peerCertificate;
+                    s_osXPeerCertAvailabilityChecked = true;
+                }
+            }
+        }
+
+        // If the installation failed, throw an exception everytime
+        // this method is called.
+        ThrowIfOSXKeychainCertificateInstallationError();
```
shmao: The method is used here only. I don't think we need a separate method for throwing the error. That way `s_osXKeychainCertInstallErrorMessage` need not to be a static member.

mconnew:
@dotnet-bot test outerloop OSX10.12

hongdai:

hongdai:

PR #2286, opened 2017-09-15
Update BuildTools version, and fix for new roslyn compiler.

StephenBonikowsky:
* to prerelease-02015-01
* Also fixing an issue caused by the new roslyn compiler used in the updated buildtools that will cause errors without this fix.
* See https://github.com/dotnet/corefx/pull/24076

### DIFF
--- a/BuildToolsVersion.txt
+++ b/BuildToolsVersion.txt
@@ -1,1 +1,1 @@
-2.0.0-prerelease-02007-01
+2.0.0-prerelease-02015-01
--- a/dir.props
+++ b/dir.props
@@ -267,6 +267,6 @@
   </PropertyGroup>
 
   <!-- Use Roslyn Compilers to build -->
-  <Import Project="$(RoslynPropsFile)" Condition="'$(RunningOnUnix)'!='true' and Exists('$(RoslynPropsFile)') and '$(UseRoslynCompilers)'!='false'" />
-  <Import Project="$(RoslynPropsFile)" Condition="'$(RunningOnUnix)'=='true' and Exists('$(RoslynPropsFile)')" />
+  <Import Project="$(RoslynPropsFile)" Condition="'$(RunningOnUnix)'!='true' and Exists('$(RoslynPropsFile)') and '$(UseRoslynCompilers)'!='false' and '$(RoslynIncompatibleMsbuildVersion)' != 'true'" />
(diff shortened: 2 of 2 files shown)


joperezr:
Why not bump up buildtools as well in this PR? that way we can actually see in CI if this will work on all configurations.

StephenBonikowsky:
True

joperezr:
oh I see, you have a maestro PR that is trying to update buildtools. Would it make more sense to add these two changes to that PR to verify CI passes? That said, I'm ok to have the changes separated in two PRs as these changes will not affect the current state of things.

StephenBonikowsky:
I'll update this PR with the changes from the maestro PR and if it looks good merge here and cancel that PR.

joperezr:
do you not have access to that PR so that you can add the changes in there directly? Or alternatively, you could just update buildtools in here and not the rest of the dependencies, and let that other things for the other PR.

StephenBonikowsky:
Woops, already done all in here.

joperezr:
mm that should be ok I guess too. I would just update the Title of this PR then since now it's a more robust update.

joperezr:
For future reference, you should be able to just push commits into the update PRs opened by dotnet bot, and if for whatever reason you can't you can talk to @karajas

StephenBonikowsky:
Sounds good.

StephenBonikowsky:
Hmm, looks like we are hitting another breaking change, have you guys encountered this in CoreFx?

```
System.MissingMethodException : Method not found: 'System.ReadOnlySpan`1<Char> System.Span.AsReadOnlySpan(System.String)'.
14:15:11         Stack Trace:
14:15:12               at System.SpanExtensions.AsReadOnlySpan(String text)
14:15:12            E:\A\_work\1342\s\corefx\src\System.Net.Primitives\src\System\Net\IPAddress.cs(230,0): at System.Net.IPAddress.TryParse(String ipString, IPAddress& address)
14:15:12            E:\A\_work\1342\s\corefx\src\System.Net.NameResolution\src\System\Net\DNS.cs(315,0): at System.Net.Dns.HostResolutionBeginHelper(String hostName, Boolean justReturnParsedIp, Boolean includeIPv6, Boolean throwOnIIPAny, AsyncCallback requestCallback, Object state)
```

joperezr:
mm no, that doesn't seem like it is related to buildtools change, that looks more like a different ingestion change. That's why I think it would be better if we leave this PR just for buildtools update, that way we can be mor certain of what is causing the problem

StephenBonikowsky:
Ok, I'll update this in a bit.

StephenBonikowsky:
Everything looks good now, merging.

PR #2296, opened 2017-09-19
Release 1.1.0 resolving the common projects ref issue.

StephenBonikowsky:
* A buildtools issue is blocking the WCF common projects from working correctly, when building test against packages the Common projects are not using the correct version of the product causing test build errors.
* After several weeks and not getting sufficient support for fixing this in buildtools this PR works around the issue by doing without common projects.
* It's ugly but un-blocks us in a servicing branch.


StephenBonikowsky:
I have verified with a private branch in VSO that these changes work and there are no build errors.

hongdai:
LGTM.

PR #2297, opened 2017-09-20
Master update dependencies

eerhardt:
Replacement for https://github.com/dotnet/wcf/pull/2251 since I can't update @dotnet-maestro-bot's fork.


StephenBonikowsky:
LGTM, waiting on last few checks.

dependencies.props
```diff
@@ -41,7 +42,7 @@
     <XunitPerfAnalysisPackageVersion>1.0.0-beta-build0007</XunitPerfAnalysisPackageVersion>
     <XunitNetcoreExtensionsVersion>1.0.1-prerelease-01616-05</XunitNetcoreExtensionsVersion>
 
-    <CoreClrPackageVersion>2.1.0-preview2-25624-02</CoreClrPackageVersion>
+    <CoreClrPackageVersion>2.1.0-preview1-25720-03</CoreClrPackageVersion>
```
dagood: This should be moved to the earlier PropertyGroup because it's auto-upgraded now. (The NETStandard props should have been moved down here when Standard wasn't being upgraded... whoops.)

eerhardt: Why are there sections based on whether the property is auto-updated or not?  That seems like an odd grouping.

dagood: Just continuing the existing way it's documented at this point. A comment in between two properties might be mistaken to just document the upcoming property rather than the following group.

PR #2300, opened 2017-09-21
First commit to release/1.0.0 after creating new VSO 1.0.0 branch.

StephenBonikowsky:
* I get VSO git failures due to using a git command that expects there to be a 'HEAD' which does not yet exist until after the first commit.
* Making a non-intrusive commit to get passed this.


StephenBonikowsky:
This can't break CI, merging without waiting.

PR #2301, opened 2017-09-21
Updating the init tools scripts with better error checking.

StephenBonikowsky:
* Matches changes made in CoreFx in release/1.0.0


hongdai:
LGTM.

PR #2303, opened 2017-09-23
Release/1.0.0 update build tools to 530-03

StephenBonikowsky:
The commits are all ports from CoreFx that were either part of the build tools update or preceded it.
Need this changes to get this branch building in VSO.


src/System.ServiceModel.Primitives/src/project.json
```diff
@@ -7,8 +7,10 @@
       "dependencies": {
         "System.Collections.NonGeneric": "4.0.0",
         "System.Collections.Specialized": "4.0.0",
+        "System.ComponentModel.EventBasedAsync": "4.0.10",
```
hongdai: Why do we need to add the explicit dependency? I prefer not to change product code in any way to minimize the risk.

StephenBonikowsky: Without the explicit reference we get validation errors, new validation checks added in buildtools.

zhenlan: Good catch @hongdai. According to the 1.0 version of [System.ServiceModel.Primitives](https://www.nuget.org/packages/System.ServiceModel.Primitives/4.1.0) on nuget.org, this package already has a dependency on System.ComponentModel.EventBasedAsync package. In the code, we do reference types defined in that package too. It's a question why it ever built before. This reference was also [explicitly added in 1.1](https://github.com/dotnet/wcf/blob/release/1.1.0/src/System.ServiceModel.Primitives/src/project.json#L10). So I think it's okay to add it here.

PR #2308, opened 2017-09-26
Update connected services images

mlacouture:


zhenlan:
Can we change to use high resolution icon in the screen shots?

PR #2310, opened 2017-09-27
Update UAP package version.

StephenBonikowsky:
* The UAP package in corefx was recently changed to be version 4.6 but the NETCore.App package is still 4.5.
* We were using a single variable for the version of all CoreFx packages, now need to split it so there is a separate one for the UAP package.

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -29,6 +29,7 @@
       versions. Make more properties for CoreFX if split versions are required.
     -->
     <MicrosoftPrivateCoreFxNETCoreAppPackageVersion>4.5.0-preview1-25718-03</MicrosoftPrivateCoreFxNETCoreAppPackageVersion>
+    <MicrosoftPrivateCoreFxUAPPackageVersion>4.6.0-preview1-25718-03</MicrosoftPrivateCoreFxUAPPackageVersion>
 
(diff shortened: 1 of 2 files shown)


StephenBonikowsky:
Thanks @eerhardt for pointing out the UAP package versioning change.

dependencies.props
```diff
@@ -108,6 +109,11 @@
       <ElementName>MicrosoftPrivateCoreFxNETCoreAppPackageVersion</ElementName>
       <PackageId>Microsoft.Private.CoreFx.NETCoreApp</PackageId>
     </XmlUpdateStep>
+    <XmlUpdateStep Include="CoreFx">
+      <Path>$(MSBuildThisFileFullPath)</Path>
+      <ElementName>MicrosoftPrivateCoreFxUAPPackageVersion</ElementName>
+      <PackageId>Microsoft.Private.CoreFx.NETCoreApp</PackageId>
```
eerhardt: This needs to be the UAP package ID.  `Microsoft.Private.CoreFx.UAP`

StephenBonikowsky: Ah, ty

StephenBonikowsky:
I'm verifying the fix in VSO right now.

StephenBonikowsky:
VSO is working, merging.

PR #2312, opened 2017-09-28
fixed wrong image

mlacouture:
Just realized the service endpoint screenshot was from a newer version of the tool and not the one to be released next.


zhenlan:
Let's hold on this. If we agree to include the UI change in the release, we don't need to fix this.

zhenlan:
No need for this now.

PR #2330, opened 2017-10-05
Updating runtime version of Suse being used for tests.

StephenBonikowsky:
Had to update the version.


mconnew:
Should we add suse to the groovy ci config too?

StephenBonikowsky:
I will be going through all the netci.groovy files at some point and bringing them up to date.

PR #2339, opened 2017-10-11
Skip the "UploadToAzure" step if the build didn't produce any packages.

StephenBonikowsky:
* Version forward Microsoft.NETCore.Platforms

This is based on dotnet/corefx#12353

@ericstj and @chcosta WCF never actually build the 1.0.0 packages we released out of VSO, at the time we were having issues with publishing and so I believe we did it in a combination of TFS and manually.

And then the VSO build defs and pipeline were used going forward for vNext, so I am only now creating the pipeline and setting-up the build defs in order to service release/1.0.0.

The fixes in this PR based on Chris' old PR in corefx get me passed the final product build error...
`PublishContent.targets(22,5): error : No items were found matching pattern 'E:\A\_work\1129\s\wcf\bin/packages/**\*.nupkg'. [E:\A\_work\1129\s\wcf\src\publish.proj]`

I would like your review, I'm not sure that I need to version forward Microsoft.NETCore.Platforms.

### DIFF
--- a/Packaging.props
+++ b/Packaging.props
@@ -1,10 +1,11 @@
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
+    <PlatformPackageVersion Condition="'$(PlatformPackageVersion)' == ''">1.0.2</PlatformPackageVersion>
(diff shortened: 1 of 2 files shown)


StephenBonikowsky:
Obsolete.

PR #2341, opened 2017-10-17
XmlSerlizerFormatAttribute test

imcarolwang:


imcarolwang:
@shmao , please help review the code, it's ported from desktop but simplified (only cover http binding and a few basic tool option),  svcutil code is not in master branch yet so related test code is commented out so test not run on core yet.

imcarolwang:
@shmao , could you please help check if there's problem in test code if you have time?

IntParams is a class contain two public int type fields , it is used as wcf operation parameter transmitted between service and client in test code, but the client test failed because value for both int fields of IntParam become 0 after passed to service, so sum method return unexpected calculate result 0 to client. 

I added a service test method sum2 which accept two int values as parameter and return sum result, client test just passed.

Run desktop Client against the service sum(IntParams) also failed, but I tried set up one of the services individually in a desktop app then run my desktop client, test passed.

shmao:
I guessed you may need to add `XmlSerializerFormatAttribute` attribute on https://github.com/dotnet/wcf/pull/2341/files#diff-07b37d926b6bffa8ab14ad0ae2cc7524R651.

imcarolwang:
Thanks! You are right, add `XmlSerializerFormatAttribute ` in contract referenced by client fix the issue.
Do you have any idea why for same service set up individually on desktop, without adding the attribute client test still pass?

imcarolwang:
Note: when running desktop client against service setup by wcf.git, if client referenced contract is not adding XmlSerializerFormat attribute, test also fail. So wcf core client and desktop client behave the same against wcf service set up in wcf.git.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/XmlSerFormatAttrServices.cs
```diff
@@ -0,0 +1,349 @@
...
+        public string Concatenate(IntParams par)
+        {
+            return string.Format("{0}{1}", par.p1, par.p2);
+        }
+
+        [OperationBehavior]
+        [XmlSerializerFormat(Style = OperationFormatStyle.Document, Use = OperationFormatUse.Literal)]
+        public void DoSomething(IntParams par)
+        {
+            Console.WriteLine("From DocLitSingleNsService inside DoSomething method, params: {0} {1}", par.p1, par.p2);
```
shmao: Console output on server side won't help much. I'd prefer to remove all `Console.WriteLine`.

Is this test testing return type being `void`? I think we can do something like this,
```c#
public void SetIntParamsProperty(IntParams par);
public IntParams GetIntParamsProperty();
```
We first call `SetIntParamsProperty` to set the value on server side and then call `GetIntParamsProperty` to retrieve the value and do verification.

shmao: This applies to other test methods with `void` return type.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/BasicHttpXmlSerFormatAttrServiceHost.cs
```diff
@@ -0,0 +1,162 @@
...
+            return new BasicHttpBinding();
+        }
+
+        public RpcEncSingleNsServiceHost(Type serviceType, params Uri[] baseAddresses)
+            : base(serviceType, baseAddresses)
+        {
+            //multiple binding namespaces test: need to figure out if to add new servicehost for it, code would be like below.
+            //Binding binding = GetBinding();
+            //binding.Namespace = "http://some.Namespace.org/";
+            //ServiceEndpoint endpoint2 = this.AddServiceEndpoint(typeof(ICalculator), binding, Address+"/extra");
```
shmao: Do we need the above lines?

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/BasicHttpXmlSerFormatAttrServiceHost.cs
```diff
@@ -0,0 +1,162 @@
...
+    public class RpcEncMultiNsServiceHostFactory : ServiceHostFactory
+    {
+        protected override ServiceHost CreateServiceHost(Type serviceType, Uri[] baseAddresses)
+        {
+            var serviceHost = new RpcEncMultiNsServiceHost(serviceType, baseAddresses);
+            return serviceHost;
+        }
+    }
+
+    public class RpcEncMultiNsServiceHost : TestServiceHostBase<ICalculator, IHelloWorld>
```
shmao: Can we rename this to `XmlSerializerDualContractServiceHost` and reuse it for all service implementing `TestServiceHostBase<ICalculator, IHelloWorld>` at https://github.com/dotnet/wcf/pull/2341/files#diff-7befb2b82d7343194e5888fe3775a510R50?

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/BasicHttpXmlSerFormatAttrServiceHost.cs
```diff
@@ -0,0 +1,162 @@
...
+    public class RpcEncSingleNsServiceHostFactory : ServiceHostFactory
+    {
+        protected override ServiceHost CreateServiceHost(Type serviceType, Uri[] baseAddresses)
+        {
+            var serviceHost = new RpcEncSingleNsServiceHost(serviceType, baseAddresses);
+            return serviceHost;
+        }
+    }
+
+    public class RpcEncSingleNsServiceHost : TestServiceHostBase<ICalculator>
```
shmao: Can we rename this to `XmlSerializerICalculatorServiceHost` and reuse it for all service implementing `TestServiceHostBase<ICalculator>`?

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
```diff
@@ -69,4 +69,71 @@ public static void XmlSerializerFormatAttribute_SupportFaults()
...
+            var intParam = new IntParams() { p1 = 5, p2 = 10 };
+            var floatParam = new FloatParams() { p1 = 5.0f, p2 = 10.0f };
+            var byteParam = new ByteParams() { p1 = 5, p2 = 10 };
+
+            Assert.Equal(3, serviceProxy1.Sum2(1, 2));
+            Assert.Equal(intParam.p1 + intParam.p2, serviceProxy1.Sum(intParam));
+            Assert.Equal(string.Format("{0}{1}", intParam.p1, intParam.p2), serviceProxy1.Concatenate(intParam));
+            Assert.Equal((float)(floatParam.p1 / floatParam.p2), serviceProxy1.Divide(floatParam));
+            Assert.Equal((new byte[] { byteParam.p1, byteParam.p2 }), serviceProxy1.CreateSet(byteParam));
+            Assert.Equal(DateTime.Now.Date, serviceProxy1.GetCurrentDateTime().Date);
```
shmao: Can we pass the date object to the server and get it back? My concern is that if the client and server are not in the same time zone, it's possible the date to be different. It may cause issue on CI or helix runs.

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
```diff
@@ -69,4 +69,71 @@ public static void XmlSerializerFormatAttribute_SupportFaults()
...
+            serviceProxy1.DoSomething(intParam);
```
shmao: Per my comments on `DoSomething` below, we need to do verification on this.

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
```diff
@@ -69,4 +69,71 @@ public static void XmlSerializerFormatAttribute_SupportFaults()
...
+
+            if (isMultiNs)
+            {
+                serviceProxy2.SayHello("test string");
```
shmao: We need to do verification on this.

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
```diff
@@ -69,4 +69,71 @@ public static void XmlSerializerFormatAttribute_SupportFaults()
...
+    [WcfFact]
+    [OuterLoop]
+    public static void XmlSFAttributeWsdlTest()
+    {
+        RunVariation(Endpoints.BasciHttpRpcEncSingleNs_Address);
+        RunVariation(Endpoints.BasicHttpRpcLitSingleNs_Address);
+        RunVariation(Endpoints.BasicHttpDocLitSingleNs_Address);
+        RunVariation(Endpoints.BasicHttpRpcEncMultiNs_Address, true);
+        RunVariation(Endpoints.BasicHttpRpcLitMultiNs_Address, true);
+        RunVariation(Endpoints.BasicHttpDocLitMultiNs_Address, true);
```
shmao: Can we split this into 6 tests? It would be easier to know which one failed if any of them failed.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/IXmlSerFormatAttrServices.cs
```diff
@@ -0,0 +1,60 @@
...
+    [ServiceContract, XmlSerializerFormat]
+    public interface IHelloWorld
+    {
+        [OperationContract, XmlSerializerFormat]
+        void SayHello(string name);
+    }
+
+    public class IntParams
+    {
+        public int p1;
```
shmao: `p1` -> `P1`

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
imcarolwang: Split into 6 tests.

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
imcarolwang: Make sense. Updated.

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
imcarolwang: Updated, replaced with SetIntParamsProperty(IntPrams) and GetIntParamsProperty().

src/System.Private.ServiceModel/tests/Scenarios/Contract/XmlSerializer/XmlSerializerFormatTest.cs
imcarolwang: Replaced with SetStringField(string) and string GetStringField().

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/IXmlSerFormatAttrServices.cs
imcarolwang: Updated.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/XmlSerFormatAttrServices.cs
imcarolwang: Test updated.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/BasicHttpXmlSerFormatAttrServiceHost.cs
imcarolwang: Updated.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/BasicHttpXmlSerFormatAttrServiceHost.cs
imcarolwang: No need. Deleted.

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/testhosts/BasicHttpXmlSerFormatAttrServiceHost.cs
imcarolwang: Updated.

PR #2363, opened 2017-11-02
Fix XmlSerializerFormatAttribute test issue

imcarolwang:
**Issue**
In the original implementation , to verify the return void service method in client, a static variable in service was used in the void set method to set the variable and call a get method to return the static variable for result verification.
It caused some problem which looks due to multithreading access to the static variable, test fail intermittently in helix run as the GetIntParamsProperty not always return expected result.

**Fix**
Thanks for Shin's advice, use ConcurrentDictionary(Guid, Object) to set and get value fixes the muti-thread access issue (without involving WCF session).

cc: @shmao , @StephenBonikowsky, @huanwu


imcarolwang:
@dotnet-bot help

PR #2364, opened 2017-11-03
Release2.0.0/add service model shim

StephenBonikowsky:
@weshaggard and @danmosemsft 

I'm currently blocked on restoring the netfx packages specified in the src\shims\netfxreference.props file.

We used to use project.json files but now use *.depproj's to restore packages, I'm not sure if that change is tied to moving to buildtoolsv2.

The current error I'm getting is...
```
C:\OSS\WCF\Tools\depProj.targets(79,5): error : Error no assets were resolved from NuGet packages. [C:\OSS\WCF\external\netfx\netfx.depproj]
C:\OSS\WCF\dir.traversal.targets(139,5): error : (No message specified) [C:\OSS\WCF\external\dir.proj]
```

I'm looking into it but assistance from anybody with more buildtools knowledge than I have would be very helpful.


external/netfx/Configurations.props
```diff
@@ -0,0 +1,18 @@
...
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- Include netcoreapp since we need to generate facades between desktop and netcoreapp -->
+    <BuildConfigurations>
+      net45;
+      net451;
+      net46;
+      net461;
+      net462;
+      net463;
```
zhenlan: >net463 [](start = 6, length = 6)

There is no .NET Framework 4.6.3. Should we have net471?

StephenBonikowsky: It is what CoreFx is using...
https://github.com/dotnet/corefx/blob/release/2.0.0/external/netfx/Configurations.props

weshaggard: I believe net461 is the only one you guys need.

weshaggard:
@StephenBonikowsky did you get this working with PackageReference's? If not then you will need to create a project.json file for this. You can look at the history of the directory in corefx to find the one you need.

src/shims/netfxreference.props
```diff
@@ -0,0 +1,37 @@
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup>
+    <NetFxReference Include="mscorlib">
```
weshaggard: Do you really need all these references?

StephenBonikowsky:
No, couldn't get Package References to work. Working on the project.json solution.

src/shims/netfxreference.props
StephenBonikowsky: Once I get it building I'll check to see that we have the minimum needed.

StephenBonikowsky:
Ok, I amended my previous commit and updated everything to more closely mimic the changes in CoreFx from back in February of this year before they moved to Configurations.props files.
After clearing up a few issues I have it building and packaging the shim into the System.ServiceModel.Primitives package in the netstandard2.0 folder.

StephenBonikowsky:
Tested this with my locally build packages.
Created a NETCore project with a project reference to a netfx project doing some basic WCF stuff including a call to a separate WCF service.
This initially resulted in an error.
Then added my locally created packages with the new ServiceModel shim to the NETCore project and it worked.

weshaggard:
In order to service this package you will need to bump up the version otherwise when you do an official build it will fail to publish because it will clash.

StephenBonikowsky:
Thanks Wes, we were planning on doing that next.

Joebeazelman:
This doesn't work on UWP see issue #2438

zhenlan:
@Joebeazelman, this change is for .net core 2.0. It is not for UWP. In UWP, you should already have the shim that ships in `Microsoft.NETCore.UniversalWindowsPlatform` v6.0.x package used in your UWP app.

PR #2370, opened 2017-11-10
KeepAliveEnabled property implementation for HttpTransportBindingElement

BogdanovKirill:
This property could be used to disable for each HTTP query keep-alive header.

### DIFF
--- a/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfService.cs
+++ b/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfService.cs
@@ -424,6 +424,15 @@ namespace WcfService
             return value;
         }
 
+        public bool IsHttpKeepAliveDisabled()
+        {
+            MessageProperties properties = new MessageProperties(OperationContext.Current.IncomingMessageProperties);
+            var property = (HttpRequestMessageProperty)properties[HttpRequestMessageProperty.Name];
+            WebHeaderCollection collection = property.Headers;
+            string connectionValue = collection.Get(Enum.GetName(typeof(HttpRequestHeader), HttpRequestHeader.Connection));
+            return connectionValue.Equals("Close", StringComparison.OrdinalIgnoreCase);
+        }
+
         private static string StreamToString(Stream stream)
         {
             var reader = new StreamReader(stream, Encoding.UTF8);
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
@@ -50,6 +50,7 @@ namespace System.ServiceModel.Channels
         private bool _useDefaultWebProxy;
         private Lazy<string> _webSocketSoapContentType;
         private SHA512 _hashAlgorithm;
+        private bool _keepAliveEnabled;
 
         internal HttpChannelFactory(HttpTransportBindingElement bindingElement, BindingContext context)
             : base(bindingElement, context, HttpTransportDefaults.GetDefaultMessageEncoderFactory())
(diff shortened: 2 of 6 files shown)


akhterul-mustafa:
👍 LGTM

zhenlan:
@BogdanovKirill Thanks for your interest in contributing to the project.

As noted in our [contributing doc](https://github.com/dotnet/wcf/blob/master/Documentation/contributing.md), WCF repo follows the [.NET Core Contribution](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/contributing.md) guidelines. For API additions, we would recommend not to submit PRs until the APs have been approved via the [API Review Process](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/api-review-process.md). Can you please open an issue and follow the said process?

BogdanovKirill:
@zhenlan 

Ok, thank you for your feedback. https://github.com/dotnet/wcf/issues/2389

mconnew:
Your implementation looks correct. Would you be able to add a test? You can add a variation to this [source file](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.4.0.0.cs). Add a new method to [this](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/IWcfService.cs) interface and [this](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfService.cs) class. In the implementation, get the request headers using the same mechanism as used [here](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfService.cs#L61) and check the Connection header is present with the value "close".

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfService.cs
```diff
@@ -424,6 +424,23 @@ public string EchoTimeAndSetCookie(string name)
...
+            HttpRequestMessageProperty property;
+            object obj;
+            MessageProperties properties = new MessageProperties(OperationContext.Current.IncomingMessageProperties);
+
+            if (properties.TryGetValue(HttpRequestMessageProperty.Name, out obj))
+            {
+                property = obj as HttpRequestMessageProperty;
+                WebHeaderCollection collection = property.Headers;
+                string connectionValue = collection.Get(Enum.GetName(typeof(HttpRequestHeader), HttpRequestHeader.Connection));
+                return connectionValue.Equals("Close", StringComparison.OrdinalIgnoreCase);
```
mconnew: Would nameof(HttpRequestHeader.Connection) work here?

src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/WcfService.cs
```diff
@@ -424,6 +424,23 @@ public string EchoTimeAndSetCookie(string name)
...
+            }
```
mconnew: Doesn't this return true if the Connection header has the value `Close` meaning KeepAlive is disabled? Also this method doesn't echo anything back so having Echo in the name doesn't make sense. Can you either rename to `IsKeepAliveEnabled` and flip the logic or rename to `IsKeepAliveDisabled`. You are also returning true if there is a missing message property. Instead throw an exception as in this case we can't tell. This will be converted to a FaultException on the client so will signal something has gone wrong with the test scenario.

BogdanovKirill:
@mconnew 

Thank you very much for help. Need some time, I will fix my mistakes

BogdanovKirill:
@mconnew

Could you check new iteration please ?

mconnew:
Sorry, Christmas got away from me. The code looks good, I'm going to manually verify this locally rebased against master and if everything looks good, then I'll merge it.

mconnew:
Fixes #2389

PR #2371, opened 2017-11-10
Updating packager and assembly versions for servicing.

StephenBonikowsky:
Getting the branch ready for servicing.


pkg/baseline/packageIndex.json
```diff
@@ -22,11 +22,12 @@
         "4.3.0",
         "4.4.0"
       ],
-      "BaselineVersion": "4.4.0",
```
weshaggard: You shouldn't update the baselineversions.

Packaging.props
```diff
@@ -20,7 +20,7 @@
     <PlatformPackageVersion Condition="'$(PlatformPackageVersion)' == ''">1.2.0</PlatformPackageVersion>
 
     <!-- by default all packages will use the same version which revs with respect to product version -->
-    <PackageVersion Condition="'$(PackageVersion)' == ''">4.4.0</PackageVersion>
+    <PackageVersion Condition="'$(PackageVersion)' == ''">4.4.1</PackageVersion>
```
weshaggard: Do you really want to update all packages? I would suggest just updating the one you want to ship and not build the others.

src/System.ServiceModel.Primitives/dir.props
```diff
@@ -2,6 +2,6 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="..\dir.props" />
   <PropertyGroup>
-    <AssemblyVersion>4.2.0.0</AssemblyVersion>
+    <AssemblyVersion>4.2.0.1</AssemblyVersion>
```
weshaggard: If you aren't changing anythink you don't really need to update the versions.

Packaging.props
weshaggard: That is what we do in corefx. We only add the packages we want to ship to the packages.builds file https://github.com/dotnet/corefx/blob/release/2.0.0/src/packages.builds#L19.

zhenlan: I think we should respin all WCF packages because people may not use the S.SM.Primitives package directly.

weshaggard: Of course that is your choice.

StephenBonikowsky:
After speaking with @zhenlan will amend this PR shortly.

Packaging.props
zhenlan: After discussed with @weshaggard offline, I think we can go with the plan to rebuild System.ServiceModel.Primitives only as the compat pack will be updated to reference the newest version of this package. The compat pack is what customers will use in the end.

StephenBonikowsky:
This now only builds a new package of System.ServiceModel.Primitives but same assembly version since nothing changed.

StephenBonikowsky:
Just realized I need to figure out how to make the use a fully stable version, currently it is...
System.ServiceModel.Primitives.4.4.1-servicing-25910-0.nupkg

weshaggard:
>Just realized I need to figure out how to make the use a fully stable version, currently it is...
System.ServiceModel.Primitives.4.4.1-servicing-25910-0.nupkg

That is what you want until you are ready to stabilize it. I would suggest get a least one official build before you do that.

Once you do that to stabilize just add the version https://github.com/dotnet/wcf/pull/2371/files#diff-c5acbfc4baa444e6a7321d2a8b07ccbaL85.

StephenBonikowsky:
Ok, will stabilize when ready for an actual servicing release.

PR #2385, opened 2017-11-15
Package new servicing versioned packages.

StephenBonikowsky:
* Building non-stable package versions.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
@@ -686,12 +686,15 @@ public static partial class ServiceContractTests
         ChannelFactory<IWcfService> factory = null;
         IWcfService serviceProxy1 = null;
         IWcfService serviceProxy2 = null;
+        string expectedEcho1 = "first";
+        string expectedEcho2 = "second";
 
         try
         {
             // *** SETUP *** \\
             binding = new BasicHttpBinding(BasicHttpSecurityMode.None);
             binding.CloseTimeout = ScenarioTestHelpers.TestTimeout;
+            binding.SendTimeout = ScenarioTestHelpers.TestTimeout;
             factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
             serviceProxy1 = factory.CreateChannel();
             serviceProxy2 = factory.CreateChannel();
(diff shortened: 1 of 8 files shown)


StephenBonikowsky:
@zhenlan not sure what this is about....

```
 500 - Internal server error.
 There is a problem with the resource you are looking for, and it cannot be displayed.
```

StephenBonikowsky:
Fixing the one test seems to have cleared everything up.

StephenBonikowsky:
@safern Could I have you take a look at these changes as well. It is sufficiently different in 1.0.0 that I would like to make sure I'm not missing anything.

src/System.ServiceModel.Duplex/pkg/System.ServiceModel.Duplex.pkgproj
```diff
@@ -1,6 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <PropertyGroup>
+    <Version>4.0.2</Version>
```
safern: Did you verify that Version actually works for the package? Because in the latest repos we changed it to be `<PackageVersion>` could you please confirm that?

StephenBonikowsky: Yes, the packages it builds locally are versioned as I would expect.

safern: Thanks for validating.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
```diff
@@ -705,8 +708,42 @@ public static void BasicHttp_Abort_ChannelFactory_Operations_Active()
...
+            try
+            {
+                actualEcho2 = t2.GetAwaiter().GetResult();
+            }
+            catch (Exception e)
+            {
+                exception2 = e;
+            }
+
+            Assert.True(exception1 != null || actualEcho1 != null, "First operation should have thrown Exception or returned an echo");
```
shmao: Under what condition would the expression be `false`?

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
```diff
@@ -705,8 +708,42 @@ public static void BasicHttp_Abort_ChannelFactory_Operations_Active()
...
+            Assert.True(exception2 != null || actualEcho2 != null, "Second operation should have thrown Exception or returned an echo");
+
+            Assert.True(actualEcho1 == null || String.Equals(expectedEcho1, actualEcho1),
+                        String.Format("First operation returned '{0}' but expected '{1}'.", expectedEcho1, actualEcho1));
+
+            Assert.True(actualEcho2 == null || String.Equals(expectedEcho2, actualEcho2),
+            String.Format("Second operation returned '{0}' but expected '{1}'.", expectedEcho2, actualEcho2));
```
shmao: Nit: the indent is different from the Assert for actualEcho1.

src/System.Private.ServiceModel/tests/Scenarios/Contract/Service/ServiceContractTests.cs
StephenBonikowsky: Good question, Ron's code, I assume it wasn't intentional.
Will open a bug to look at this more carefully.
For now won't block this PR as it is a straight port.

PR #2387, opened 2017-11-15
Version forward release/1.1.0 packages for servicing.

StephenBonikowsky:
* At this point no assembly changes, just making sure we can package and publish new package versions end-to-end including to MyGet via VSTO.

@weshaggard we would like to confirm through our VSTO build pipeline that we can successfully build and publish these packages to MyGet. This gets us ready for the upcoming servicing release.


weshaggard:
Looks correct. @safern can also help review these changes.

StephenBonikowsky:
Thank you!

PR #2422, opened 2017-12-04
Were not bin placing tests in Linux folder when building for Linux.

StephenBonikowsky:


mconnew:
Can you add an xml comment briefly explaining that this is used to override the OSGroup when creating official builds.

StephenBonikowsky:
Sure, I'll wait until CI passes...hopefully and then amend it with the comment and merge.

StephenBonikowsky:
I should also point out that I tested this in VSTS and confirmed that it fixed the problem.

mconnew:
LGTM functionally.

PR #2426, opened 2017-12-05
Build all WCF contracts for servicing releases.

StephenBonikowsky:
Zhenlan, I reved the package versions by one which leaves only NetTcp not having the same package version, should I change NetTcp to also be package version 4.4.1 so they are all the same?


zhenlan:
We should include `System.Private.ServiceModel` package in this change too and increment the assembly version of S.P.SM.dll.

src/System.Private.ServiceModel/pkg/System.Private.ServiceModel.pkgproj
```diff
@@ -2,6 +2,7 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
+    <PackageVersion>4.4.1</PackageVersion>
     <PreventImplementationReference>true</PreventImplementationReference>
     <!-- workaround for https://github.com/dotnet/buildtools/issues/1529 -->
     <HarvestVersion>4.3.0</HarvestVersion>
```
zhenlan: Just add a note, we may need to change this to 4.3.1 somehow. Next round.

PR #2447, opened 2017-12-13
Harvest latest version for 2.0.0 fat package.

StephenBonikowsky:


pkg/baseline/packageIndex.json
```diff
@@ -20,12 +20,13 @@
         "4.0.0",
         "4.1.0",
         "4.3.0",
+        "4.3.1",
         "4.4.0"
       ],
       "BaselineVersion": "4.4.1",
       "InboxOn": {},
       "AssemblyVersionInPackageVersion": {
-        "4.1.1.0": "4.3.0",
```
zhenlan: we don't need to remove this line.

StephenBonikowsky: fixed.

PR #2448, opened 2017-12-14
Workaround for path not supported error when syncing sources.

StephenBonikowsky:
* See issue #4094 at NuGet: https://github.com/NuGet/Home

Bug in 2.0.0 preview versions of CLI do not like it when the first item in a list of sources to restore is a file path and not a url. By making sure a url is first we work around the issue.

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -64,15 +64,15 @@
 
   <!-- list of nuget package sources passed to nuget restore -->
   <ItemGroup Condition="'$(ExcludeInternetFeeds)' != 'true'">
-    <!-- Example to consume local CoreCLR package:
-         /p:OverridePackageSource=C:\coreclr\bin\Product\Windows_NT.x64.Debug\.nuget\pkg
-    -->
-    <DotnetSourceList Include="$(OverridePackageSource)" />
     <!-- Need to escape double forward slash (%2F) or MSBuild will normalize to one slash on Unix. -->
     <!-- Including buildtools to pull in TestSuite packages and repackaged xunit dependencies-->
     <DotnetSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-buildtools/api/v3/index.json" />
     <DotnetSourceList Include="https:%2F%2Fdotnet.myget.org/F/dotnet-core/api/v3/index.json" />
     <DotnetSourceList Include="https:%2F%2Fapi.nuget.org/v3/index.json" />
+    <!-- Example to consume local CoreCLR package:
+         /p:OverridePackageSource=C:\coreclr\bin\Product\Windows_NT.x64.Debug\.nuget\pkg
+    -->
+    <DotnetSourceList Include="$(OverridePackageSource)" />
   </ItemGroup>
 
   <PropertyGroup>


StephenBonikowsky:
This is the error we are seeing in VSTS...
`2017-12-14T08:15:02.3091055Z C:\Users\dlab\.dotnet\x64\sdk\2.2.0-preview1-007522\NuGet.targets(477,5): error : The format of the path 'E:\A\_work\1972\s\Packages\AzureTransfer\Release;https://dotnet.myget.org/F/dotnet-buildtools/api/v3/index.json;https://dotnet.myget.org/F/dotnet-core/api/v3/index.json;https://api.nuget.org/v3/index.json;https://dotnet.myget.org/F/roslyn/api/v3/index.json' is not supported. [E:\A\_work\1972\s\wcf\Tools\OptimizationData.msbuild] [E:\A\_work\1972\s\wcf\build.proj]`

StephenBonikowsky:
@dagood @weshaggard 
@zhenlan and I were chatting and realized that the ordering might matter...? If the tooling restores from OverridePackageSource first does it "win" thereby preventing the same package from one of the other sources from getting restored? If so my putting OverridePackageSource could result in the unexpected package getting restored.

I think this would only be the case when the packages in different sources are using the exact same package version.

weshaggard:
>realized that the ordering might matter...? I

Ordering does matter but as you point out there really shouldn't be conflicting packages with the same version across the feeds so this should be OK. We've had to do similar fixes in other repo's

zhenlan:
> We've had to do similar fixes in other repo's

@StephenBonikowsky, can we make sure we do the fix in the same way as other repo's? I'd like us to be consistent with other repo's, so we don't present a unique issue in the future.

StephenBonikowsky:
Well, I just looked through a bunch of repos. I only saw one that was using 'OverridePackageSource' all the rest are only using urls so they wouldn't have this issue and the one exception hasn't been changed...
https://github.com/dotnet/core-setup/blob/2a19751530a6abcad593883c4381a6b4893f809e/dir.props#L47

So unless the workaround is in a different area of the scripts I don't see anything to compare to.
@weshaggard any advice re Zhenlan's desire to do the fix in the same way as other repos?

dagood:
Looks like you have the same fix as CoreFX https://github.com/dotnet/corefx/commit/910e50de871cb36628d3acf8542bee00c766ffe9, but it was undone in https://github.com/dotnet/corefx/pull/24924 it looks like. I do think I'm right about the "build tests from packages" scenarios not being used anymore, which might explain why nobody has hit it again in CoreFX.

StephenBonikowsky:
Yes, you are correct Davis, one of my biggest headaches is that we seem to be the only team that does product and build in separate build definitions, this is continuously causing breaking changes for us as the assumption in build tools seems to be that you build and run on the same platform at the same time.

In case you wonder why we don't follow the same model... :) A primary scenario for us is the ability to build on Windows and run on any platform. We should also be able to build on any platform but early on that wasn't very easy and so here we are...

dagood:
Yeah... this is something that would be great to cover once we can get the long-awaited BuildTools minimal template/test repo to make sure we aren't actively breaking it with upgrades.

zhenlan:
I probably missed something. I am wondering why corefx repo does not need the workaround. As of now, it still has `$(OverridePackageSource)` as the first entry for `ResotreSources`. I did not double check but I assume we are on the same version of buildtools (by auto update PR).

Just to be clear, I don't think how wrong this workaround is. I just don't like these small discrepancies, which can cause big headaches for likely corner cases in the future but super hard to diagnose. I would rather be wrong but consistent so the "wrong" can be easily identified when it went wrong.

dagood:
> I am wondering why corefx repo does not need the workaround. As of now, it still has `$(OverridePackageSource)` as the first entry for `ResotreSources`.

CoreFX did need it, but it was undone (for no stated reason):

> [me] "Looks like you have the same fix as CoreFX dotnet/corefx@910e50d, but it was undone in dotnet/corefx#24924."

> I just don't like these small discrepancies

Stephen pointed out that this is actually a large discrepancy, and actively causes headaches like this one:

> one of my biggest headaches is that we seem to be the only team that does product and build in separate build definitions, this is continuously causing breaking changes for us as the assumption in build tools seems to be that you build and run on the same platform at the same time.

I think the change in dotnet/corefx#24924 that reverts the workaround should be reverted (to put the workaround back in place--@weshaggard @karajas), but there's no guarantee that matching CoreFX will always work in WCF because this whole feature is not used by normal CoreFX scenarios. (Hence the comment about the minimal template/test repo, which would explicitly test during BuildTools development.)

weshaggard:
corefx didn't do it this way because we don't use OverridePackageSource as part of any normal build definitions. However with that said I suspect no one can use it currently at all for local file paths, so I agree with @dagood that we should probably fix this in corefx as well.

zhenlan:
Thanks for your input, @dagood and @weshaggard! As corefx will revert dotnet/corefx#24924, I will go ahead and merge this PR.

PR #2453, opened 2017-12-18
Test syncing fails to pull in this needed assembly.

StephenBonikowsky:
* System.ComponentModel.Composition
* Looks like package Microsoft.Private.CoreFx.NETCoreApp is failing to pull this in, manually adding this package reference.

### DIFF
--- a/external/corefx/corefx.depproj
+++ b/external/corefx/corefx.depproj
@@ -26,6 +26,9 @@
     <PackageReference Include="Microsoft.Private.CoreFx.NETCoreApp">
       <Version>$(MicrosoftPrivateCoreFxNETCoreAppPackageVersion)</Version>
     </PackageReference>
+    <PackageReference Include="System.ComponentModel.Composition">
+      <Version>$(MicrosoftPrivateCoreFxNETCoreAppPackageVersion)</Version>
+    </PackageReference>
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetGroup)' == 'uap' or


StephenBonikowsky:
@dagood @weshaggard 
All our tests were failing on MC with...
`Error: assembly specified in the dependencies manifest was not found -- package: 'runtime.win-x64.microsoft.netcore.app', version: '2.0.0-preview1-002111-00', path: 'runtimes/win-x64/lib/netcoreapp2.0/System.ComponentModel.Composition.dll'`
I believe this dependency should have been getting pulled in via our package reference to `Microsoft.Private.CoreFx.NETCoreApp` but wasn't.

This is another example of us failing only in the situation where product building and test building are separated. The product building pulls this package down but just syncing and building test does not pull this down so we didn't catch it in CI or normal local building.

I think this is something that should be fixed in CoreFx with the `Microsoft.Private.CoreFx.NETCoreApp` package right?

dagood:
LGTM, and I'm pretty sure this is intentional: CoreFX only builds individual packages for projects that aren't part of the main (NETCoreApp) package. I'm pretty sure `IsNETCoreApp` is controlling the inclusion: 
[System.ComponentModel.Composition dir.props](https://github.com/dotnet/corefx/blob/3a9896802161e86f68345a5cb09abb9fe5ec5b1d/src/System.ComponentModel.Composition/dir.props)
[System.Threading.Thread dir.props](https://github.com/dotnet/corefx/blob/3a9896802161e86f68345a5cb09abb9fe5ec5b1d/src/System.Threading.Thread/dir.props#L7)

dagood:
Didn't see your comment before posting mine--my intuition is that the DLL shouldn't be depended on, and it does seem like something to fix.

dagood:
`runtime.win-x64.microsoft.netcore.app` `2.0.0-preview1-002111-00` is very old (05 May 2017) and does have `System.ComponentModel.Composition.dll` in it. A newer version I checked doesn't: [`2.1.0-preview1-26018-05`](https://dotnet.myget.org/feed/dotnet-core/package/nuget/runtime.win-x64.Microsoft.NETCore.App/2.1.0-preview1-26018-05) (18 Dec 2017). It sounds like some unexpected packages are being restored?

weshaggard:
System.ComponentMOdel.Composition.dll used to only be a facade with a single type-forward so it was part of NETCoreApp with all the other facades. Since we implemented it fully we put it into its own package. So your fix here is correct.

PR #2455, opened 2017-12-21
Update some dependencies that aren't done by the auto updates.

StephenBonikowsky:
* CoreFxBaseLinePackageVersion, this was causing us to pull in some old packages of CoreFx.
* XunitNetcoreExtensionsVersion, this older version needed to be updated to work with newer CoreFx packages.
* A couple packages that XUnit depends on also needed to be updated.


zhenlan:
Thanks, @StephenBonikowsky. We should keep this updated for the master branch on regular basis going forward.

PR #2456, opened 2017-12-22
Updating the assembly version of S.P.SM for servicing.

StephenBonikowsky:
@zhenlan made these changes in VSTS to create servicing packages with stable versions while having incremented the S.P.SM assembly version and package version.


zhenlan:
We shouldn't increment the package version or make package stable, which will conflict with what we have in the master branch. For UWP6.0, we ship S.P.SM.dll directly with meta-package. We don't ship ServiceModel packages.

zhenlan:
@StephenBonikowsky has made necessary change in a separate PR.

PR #2484, opened 2018-01-10
Version forward S.P.SM

StephenBonikowsky:
* Skipping forward to 4.1.3.2 because the servicing release of uwp6.0 used version 4.1.3.1

### DIFF
--- a/src/System.Private.ServiceModel/dir.props
+++ b/src/System.Private.ServiceModel/dir.props
@@ -2,7 +2,7 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="..\dir.props" />
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


src/System.Private.ServiceModel/dir.props
```diff
@@ -2,7 +2,7 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="..\dir.props" />
   <PropertyGroup>
-    <AssemblyVersion>4.1.3.0</AssemblyVersion>
+    <AssemblyVersion>4.1.3.2</AssemblyVersion>
```
zhenlan: To follow current versioning style, this should be `4.1.4.0` because UWP may have another servicing release in the future and that will need version `4.1.3.2`. So we need to reserve `4.1.3.x` for UWP.

StephenBonikowsky: Makes sense.

PR #2490, opened 2018-01-10
Updated the S.SM.Syndication.dll for Shim.

shmao:
The checked-in S.SM.Syndication.dll didn't have correct digital
signature.


StephenBonikowsky:
LGTM

PR #2506, opened 2018-01-18
Fix support for using proxies with HTTP transport

mconnew:


mconnew:
I have tested this manually with both setting a system wide proxy as well as explicitly setting a proxy.
Fixes #2506 1592

PR #2519, opened 2018-01-25
Increment assembly versions to match package versions.

StephenBonikowsky:
* Keep the label as Preview 2.


StephenBonikowsky:
Verified locally that the assemblies produced were versioned correctly.

StephenBonikowsky:
@zhenlan Can I get an LGTM?

PR #2522, opened 2018-01-26
Update master branch packaging to harvest latest release of 1.1.0

StephenBonikowsky:
* This updates the harvested S.P.SM assembly for netcore50 and netstandard1.3 to the most recently released version that was released in package 4.3.1
* The harvested S.P.SM assembly version should now be 4.1.1.1


StephenBonikowsky:
I verified locally that version 4.1.1.1 of S.P.SM is now packaged under \runtimes\win7\lib for both netstandard1.3 and netcore50.

StephenBonikowsky:
If this is good, I'll cherry-pick it to the release/2.1.0 branch

pkg/baseline/packageIndex.json
```diff
@@ -16,12 +16,14 @@
       "StableVersions": [
         "4.0.0",
         "4.1.0",
-        "4.3.0"
+        "4.3.0",
```
StephenBonikowsky: @ericstj Not sure if I needed to keep 4.3.0 or simply override it with 4.3.1.

ericstj: Never delete versions from the index.  It's meant to be a private copy of package history and mirror nuget.org releases.

zhenlan: +1

StephenBonikowsky: Ok, thanks. I'll amend this PR.

PR #2533, opened 2018-01-31
Building non-Windows builds using the CLI's MSBuild

joperezr:
cc: @StephenBonikowsky @zhenlan @weshaggard 

With these changes now the non-Windows builds will be built using the MSBuild that comes with the CLI instead of the one we restore manually using buildtools.


joperezr:
cc: @huanwu

PR #2554, opened 2018-02-10
Update test infrastructure to detect what OS it is running on.

StephenBonikowsky:
* This has been broken since moving to buildtoolsv2 as the 'TestNugetRuntimeId' property that was used for this purpose was removed.
* Also removing a few OS versions that are EOL and adding a couple new ones.
* Adding a test case that can be enabled whenever a new OS version is added to the lab so the RID can be confirmed in a temp run and then added to Infrastructure.Common

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/OSHelper.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/OSHelper.cs
@@ -142,27 +142,6 @@ namespace Infrastructure.Common
             return osid;
         }
 
-        // The $(TestNugetRuntimeId) property was deprecated in build tools v2.
-        // No longer calling this method.
-        private static OSID OSIDfromTestRuntime()
-        {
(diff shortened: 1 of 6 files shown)


StephenBonikowsky:
Verified works as expected in VSTS runs.

src/System.Private.ServiceModel/tests/Common/Infrastructure/OSHelper.cs
```diff
@@ -147,8 +142,8 @@ private static OSID DetectCurrentOS()
             return osid;
         }
 
-        // Maps from the $(TestNugetRuntimeId) property to corresponding OSID.
-        // Returns OSID.None if cannot determine the OSID.
+        // The $(TestNugetRuntimeId) property was deprecated in build tools v2.
+        // No longer calling this method.
         private static OSID OSIDfromTestRuntime()
```
zhenlan: Let's clean this up too so we don't cause confusion for the future 😄

StephenBonikowsky: Done

PR #2603, opened 2018-02-26
Disable auto update of ProjectNtfs

StephenBonikowsky:
* Support for UWP is currently broken, disabling the sections of the dependencies.props file that auto updated versions of files needed for UWP.


StephenBonikowsky:
Just commenting it out in the same way that CoreFx did, it will eventually be re-enabled.

StephenBonikowsky:
@dotnet-bot test outerloop OSX10.12

PR #2610, opened 2018-02-27
[release/2.1.0] Update BuildTools to 2.1.0-preview1-02523-01

dagood:
This version includes a fix to unblock dependency auto-update flow after GitHub stopped allowing weak cryptographic standards: dotnet/core-eng#2734.

This will also fix the build failing when it tries to update dotnet/versions.


dagood:
This is missing an update from https://github.com/dotnet/wcf/pull/2611, closing as duplicate.

PR #2611, opened 2018-02-27
Update to latest buildtools version.

StephenBonikowsky:
* This will fix the System.Net.WebException when trying to update the Versions repo in our Publishing definition.

See dotnet/core-eng#2734


dependencies.props
```diff
@@ -41,7 +41,7 @@
     <AppXRunnerVersion>1.0.3-prerelease-00921-01</AppXRunnerVersion>
     <CoreFxBaseLinePackageVersion>2.2.0-preview1-26216-02</CoreFxBaseLinePackageVersion>
     <XunitPerfAnalysisPackageVersion>1.0.0-beta-build0015</XunitPerfAnalysisPackageVersion>
-    <XunitNetcoreExtensionsVersion>2.1.0-preview1-02431-03</XunitNetcoreExtensionsVersion>
+    <XunitNetcoreExtensionsVersion>2.1.0-preview1-02523-01</XunitNetcoreExtensionsVersion>
```
dagood: My attempt didn't include this update--I think dependencies.props is missing an updater for this package?

StephenBonikowsky: Sorry, didn't see your PR, yes you are right, it doesn't have the updater section for 'XunitNetcoreExtensionsVersion'. I have a couple questions about that, I'll follow-up with you offline.

StephenBonikowsky: No reason not to include the updater, amending the PR to include this change.

PR #2620, opened 2018-02-28
Move win7 harvested binaries under runtimes/win/lib

StephenBonikowsky:
* On Win10 machines NuGet is picking win7 instead of win which is technically correct.
* Moving the win7 assets under win/lib/<TFM> fixes the problem.


StephenBonikowsky:
I also ran this in VSO and confirmed the package output was what we wanted.

PR #2630, opened 2018-03-01
Increment assembly versions.

StephenBonikowsky:
* Also changed the PreReleaseLabel to differentiate the packages produced by uwp6.0 and 2.1


StephenBonikowsky:
I built locally and checked one of the packages to confirm that the new assembly version was correctly applied as well as the updated pre-release label.

PR #2649, opened 2018-03-06
Updating UAPvNext TFM

joperezr:
cc: @zhenlan @weshaggard @ericstj 

Updating the UAP TFM on WCF to match corefx.

@ericstj I didn't update the packageIndex inboOn for uap RS3, since these guys are a bit special given that we still had assets that applied to RS3 on the packages. Do we want to update the package index and have these guys have a placeholder for RS3 instead? or should we do the TreatAsOutOfBox supression?

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -11,7 +11,7 @@
 
   <!-- Define vNext UAP Moniker -->
   <PropertyGroup>
-    <UAPvNextVersion>10.0.15138</UAPvNextVersion>
+    <UAPvNextVersion>10.0.16300</UAPvNextVersion>
     <UAPvNextTFMFull>UAP,Version=v$(UAPvNextVersion)</UAPvNextTFMFull>
     <UAPvNextTFM>uap$(UAPvNextVersion)</UAPvNextTFM>
   </PropertyGroup>
--- a/pkg/baseline/packageIndex.json
+++ b/pkg/baseline/packageIndex.json
@@ -36,7 +36,9 @@
         "4.3.1"
       ],
       "BaselineVersion": "4.5.0",
-      "InboxOn": {},
+      "InboxOn": {
+        "uap10.0.16299": "4.1.1.0"
+      },
       "AssemblyVersionInPackageVersion": {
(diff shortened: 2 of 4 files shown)


ericstj:
I would expect that WCF wants to be treated as out of box so that their API isn't locked by what was inbox in UAP.  We can either do that by updating the package index and adding a suppression, or just leave the index alone.  I'd feel a bit better about updating the index since it keeps it in sync with how we do things elsewhere.

StephenBonikowsky:
@zhenlan Could you clarify this a bit for us? Our binaries were part of the UAP metapackage that shipped inbox on RS3. Is the discussion here about how we are shipping for RS4, as in In-Box versus Out-Of-Box for RS4? Why would we want to do it differently than we did for RS3?

Also, should whatever the end result of this PR be ported to the uwp6.1 branch?

joperezr:
> I'd feel a bit better about updating the index since it keeps it in sync with how we do things elsewhere.

Sounds good, I'll do that then.

joperezr:
> Could you clarify this a bit for us?

The intention of this is basically to prepare the new changes going in to master to target the next release of UWP.

> Also, should whatever the end result of this PR be ported to the uwp6.1 branch?

Yes, I believe this work should be ported to that branch once it's in master.

zhenlan:
> I would expect that WCF wants to be treated as out of box so that their API isn't locked by what was inbox in UAP

I am still a bit confused about this. What makes something considered in-box vs. out-out-box?

joperezr:
If the package index states that this assembly was shipped inbox on UAP RS3, then package validation will fail when when building the packages next time because it will say that the package index shows that assemblies should be inbox (which means packages shouldn't have an asset for them in that framework) but the package did have an asset resolved for that framework. What @ericstj suggests is to update the index, but add a validation suppression in order to keep the assets for UAP on the package without validation failing, this way, you will be able to do API additions after we ship the next version of UAP.

StephenBonikowsky:
This looks good to me now, any questions @zhenlan ?

PR #2652, opened 2018-03-07
Workaround for GitHub disabling weak crypto.

StephenBonikowsky:
* Needed to publish build info to dotnet/versions.
* See https://github.com/dotnet/core-eng/issues/2734
* See https://github.com/dotnet/core-eng/issues/2772


StephenBonikowsky:
Switching to using dotnet.exe msbuild instead of build-managed.cmd because TLS1.2 is already enabled in Core.

PR #2665, opened 2018-03-09
Fix http timeout on windows

mconnew:
Also adds test that's only enabled when setting the Long_Running property. This is to enable running the test in CI and nightly's but not slow down individual local runs.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.4.0.0.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.4.0.0.cs
@@ -8,7 +8,7 @@ using System.ServiceModel.Channels;
 using Infrastructure.Common;
 using Xunit;
 
-public static class Binding_Http_BasicHttpBindingTests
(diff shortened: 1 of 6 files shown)


zhenlan:
Add @KKhurin for code review.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.4.0.0.cs
```diff
@@ -42,4 +42,45 @@ public static void DefaultSettings_Echo_RoundTrips_String()
...
+        try
+        {
+            // *** SETUP *** \\
+            binding = new BasicHttpBinding();
+            binding.SendTimeout = bindingSendTimeout;
+            factory = new ChannelFactory<IWcfService>(binding, new EndpointAddress(Endpoints.HttpBaseAddress_Basic));
+            serviceProxy = factory.CreateChannel();
+
+            // *** EXECUTE *** \\
+            string result = serviceProxy.EchoWithTimeout("Hello", serviceOperationTimeout);
```
KKhurin: NIT, use testString here?

zhenlan:
@mconnew can this be merged?

PR #2668, opened 2018-03-10
Add new exception handling logic for SocketsHttpHandler differences

mconnew:
Added changes needed for failure modes with SocketsHttpHandler as it throws different exceptions than the previous HttpClientHandler implementations.
Switched to using CredentialCache to restrict credentials to a single authentication mechanism. This matches the full framework behavior.
Improved the DigestAuthentication request validation to fail if the client sends authentication headers which aren't digest.


mconnew:
Only test failing now looks to be digest authentication. This is likely a protocol error in either the WCF service side digest authentication component I wrote or the digest protocol support in SocketsHttpHandler. I'll investigate to work out which it is.

mconnew:

PR #2702, opened 2018-03-21
Stebon/2.0.0/update versions

StephenBonikowsky:


src/System.Private.ServiceModel/pkg/System.Private.ServiceModel.pkgproj
```diff
@@ -2,7 +2,7 @@
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
   <PropertyGroup>
-    <PackageVersion>4.4.2</PackageVersion>
+    <PackageVersion>4.4.3</PackageVersion>
```
zhenlan: If we remove this line, will it pick up PackageVersion from Packaging.props file automatically?

StephenBonikowsky: It looks like when it is set in the pkgproj it wins, I can remove the lines in the individual pkgprojects and then it honors what is set in Packaging.props.

Issue #1888, opened 2017-05-03
Test Case Failure: NetTcp_TransportSecurity_Streamed_TimeOut_Long_Running_Operation


StephenBonikowsky:
StreamingTests.NetTcp_TransportSecurity_Streamed_TimeOut_Long_Running_Operation

Has been failing in CI recently and needs to be investigated.

Stacktrace
MESSAGE:
Assert.Throws() Failure\r\nExpected: typeof(System.TimeoutException)\r\nActual: typeof(System.ServiceModel.CommunicationException): The socket connection was aborted because an asynchronous send to the socket did not complete within the allotted timeout of 00:00:04.9949997. The time allotted to this operation may have been a portion of a longer timeout.
+++++++++++++++++++
STACK TRACE:
[...]

PR #2704, opened 2018-03-22
Fixing WCF 1888 by eliminating timing issue

KKhurin:
#1888 happened due to 2 timeout timers competing to set which operation was aborted. The fix is to simplify the logic that decides whether to throw a timeout exception.


StephenBonikowsky:
If I understand the change correctly it looks like now it will always return the timeout exception.

LGTM

StephenBonikowsky:
@KKhurin can this be merged, are you waiting on another approval or review? I don't see anybody else added to review it.

StephenBonikowsky:
@mconnew maybe you can take a look at it? It would be great to get this in and not have it sit here.

KKhurin:
Matt wanted to take a deeper look this week.

PR #2709, opened 2018-03-23
Using local variable to avoid null ref

### LINKED ISSUE #2708: NullRef when using short timeouts for http binding
This NullRef doesn't happen on the timer thread (so if the user code handles the exception it won't crash the process) but it is so frequent that it blocks entire stress testing of other fixes for time out issues.

KKhurin:
This would fix #2708 and unblock further testing of short timeouts

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
@@ -1012,11 +1012,11 @@ namespace System.ServiceModel.Channels
                     _timeoutHelper = timeoutHelper;
                     _factory.ApplyManualAddressing(ref _to, ref _via, message);
                     _httpClient = await _channel.GetHttpClientAsync(_to, _via, _timeoutHelper);
-                    _httpRequestMessage = _channel.GetHttpRequestMessage(_via);
 
                     // The _httpRequestMessage field will be set to null by Cleanup() due to faulting
                     // or aborting, so use a local copy for exception handling within this method.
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -999,7 +999,7 @@ public HttpClientChannelAsyncRequest(HttpClientRequestChannel channel)
 
                         if (!suppressEntityBody)
                         {
-                            _httpRequestMessage.Content = MessageContent.Create(_factory, request, _timeoutHelper);
+                            httpRequestMessage.Content = MessageContent.Create(_factory, request, _timeoutHelper);
```
zhenlan: Just wondering if line 981 should be changed to 
```
httpRequestMessage = _channel.GetHttpRequestMessage(_via);
```

KKhurin: Unfortunately not really :(  _httpRequestMessage is used in a few other code paths that can run simultaneously (and like you probably suspect, that code can run into the same kind of race condition that we're fixing here). So this fix does not completely eliminate all the issues - it just mitigates the most frequent one.

StephenBonikowsky:
@KKhurin If you rebase and force push we can get a clean CI, Matt's PR merged a couple days ago fixes the test failures we are seeing.

StephenBonikowsky:
The test case failures in OSX started before this PR, I think we might have taken an auto-update that caused it. So don't block this PR on those test failures.

LGTM - Merge it!

PR #2718, opened 2018-03-26
Stebon/2.1.0/port sockets http handler changes

StephenBonikowsky:
Cherry-picking of #2668 from Master to release/2.1.0


StephenBonikowsky:
This PR contains the test fix for the test failures we are currently seeing. It needs to be merged before it will take effect.

PR #2752, opened 2018-04-11
Remove reference to UwpTfm for vNext.

StephenBonikowsky:


StephenBonikowsky:
@ericstj this should fix the issue with NuGet restore trying to restore 16300.

ericstj:
Yep, just to be clear about what was happening before, you had a package that had 
lib\uap1\foo.dll
lib\uap2\bar.dll
lib\other\foo.dll
lib\other\bar.dll

The expectation of the package author was that both foo and bar were in every folder.  I do believe package validation would have helped here, but it looks like it was turned off.  It's probably worth your time to turn that back on and find out what (if any) things you need to do to make it understand and validate your scenario.

PR #2755, opened 2018-04-11
Enable package signing

StephenBonikowsky:
Based on dotnet/corefx#28991

@weshaggard in VSTS I will also add the "Sign Packages" step in our Publish build definition.
Are there any other changes I have missed that need to be made in order to enable signing our packages?


weshaggard:
>@weshaggard in VSTS I will also add the "Sign Packages" step in our Publish build definition.
Are there any other changes I have missed that need to be made in order to enable signing our packages?

No as long as your build definition is updated.

StephenBonikowsky:
The test run succeeded in VSTS, just waiting for CI to complete.

PR #2756, opened 2018-04-11
Remove System.ServiceModel shim from ref bin.

### LINKED ISSUE #2748: System.ServiceModel.dll in System.ServiceModel.Primitives package has wrong referenced assembly versions
In System.ServiceModel.Primitives package, under both ref and lib, open system.servicemodel.dll in ildasm, you can see that the following references in its manifest. The referenced assembly versions are all 0.0.0.0. We expect the referenced assembly versions should exactly match with corresponding façade assembly versions. Looks like this is a regression from v2.0 release. Also, we should evaluate if system.servicemodel.dll is needed under ref folder. If not needed, will moving it will cause breaking change?
.assembly extern System.Runtime
{
  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )                         // .?_....:
  .ver 0:0:0:0
}
.assembly extern System.ServiceModel.Primitives
{
  .publickeytoken = (CC 7B 13 FF CD 2D DD 51 )                         // .{...-.Q
  .ver 0:0:0:0
}
.assembly extern System.ServiceModel.Http
{
  .publickeytoken = (CC 7B 13 FF CD 2D DD 51 )                         // .{...-.Q
  .ver 0:0:0:0
}
.assembly extern System.ServiceModel.Duplex
{
  .publickeytoken = (CC 7B 13 FF CD 2D DD 51 )                         // .{...-.Q
  .ver 0:0:0:0
}
.assembly extern System.ServiceModel.NetTcp
{
  .publickeytoken = (CC 7B 13 FF CD 2D DD 51 )                         // .{...-.Q
  .ver 0:0:0:0
}
.assembly extern System.ServiceModel.Security
{
  .publickeytoken = (CC 7B 13 FF CD 2D DD 51 )                         // .{...-.Q
  .ver 0:0:0:0
}
.assembly extern System.ServiceModel.Syndication
{
  .publickeytoken = (CC 7B 13 FF CD 2D DD 51 )                         // .{...-.Q
  .ver 0:0:0:0
}

StephenBonikowsky:
@joperezr and @weshaggard  We shouldn't have the shim in a ref dir correct, this is only for runtime redirects?

Fixes #2748

### DIFF
--- a/src/System.ServiceModel.Primitives/pkg/System.ServiceModel.Primitives.pkgproj
+++ b/src/System.ServiceModel.Primitives/pkg/System.ServiceModel.Primitives.pkgproj
@@ -41,14 +41,6 @@
     <File Include="$(ServiceModelFacade)">
       <TargetPath Condition="'$(_targetRuntime)' == ''">lib/netstandard2.0</TargetPath>
     </File>
-    <!-- Ref shim of System.ServiceModel.dll for UAP -->
-    <File Include="$(ServiceModelFacade)">
-      <TargetPath Condition="'$(_targetRuntime)' == ''">ref/$(UAPvNextTFM)</TargetPath>
-    </File>
-    <!-- Ref shim of System.ServiceModel.dll for netstandard2.0 -->
-    <File Include="$(ServiceModelFacade)">
-      <TargetPath Condition="'$(_targetRuntime)' == ''">ref/netstandard2.0</TargetPath>
-    </File>
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
 </Project>


weshaggard:
We should have it in the ref as well otherwise people cannot consume a library at build time.

StephenBonikowsky:
I thought this shim was for runtime compatibility only? Maybe I don't understand the scenario and am missing something.

I did notice that shims built by CoreFx in the package "Microsoft.Windows.Compatibility.Shims"  don't seem to have ref dirs either?

StephenBonikowsky:
Thanks @joperezr We had a discussion and worked through your scenario and agree, you are right it is needed.

Is the reason why the shims in Microsoft.Windows.Compatibility.Shims do not have a separate ref directory because of some rule that in the case of the ref dir being absent it will assume that the assemblies in the lib folder are also the ref assemblies?

weshaggard:
>Is the reason why the shims in Microsoft.Windows.Compatibility.Shims do not have a separate ref directory because of some rule that in the case of the ref dir being absent it will assume that the assemblies in the lib folder are also the ref assemblies?

Yes that is correct.

PR #2757, opened 2018-04-12
Remove exclude of private packages from signing.

StephenBonikowsky:
* In corefx they don't publish private packages to NuGet, WCF does publish to NuGet our one private package.


StephenBonikowsky:
I verified that all the other packages that were signed yesterday were correctly signed.

src/publish.proj
```diff
@@ -29,15 +29,14 @@
...
-    <FinalPublishPrivatePattern>$(PackageDownloadDirectory)\**\*Private*.nupkg</FinalPublishPrivatePattern>
     <FinalSymbolsPackagesPattern>$(PackageDownloadDirectory)\**\*.symbols.nupkg</FinalSymbolsPackagesPattern>
     <!-- The SignFiles target needs OutDir to be defined -->
     <OutDir>$(PackageDownloadDirectory)</OutDir>
   </PropertyGroup>
 
   <Target Name="GetPackagesToSign">
     <ItemGroup>
-      <FilesToSign Include="$(FinalPublishPattern)" Exclude="$(FinalPublishPrivatePattern);$(FinalSymbolsPackagesPattern)">
+      <FilesToSign Include="$(FinalPublishPattern)" Exclude="$(FinalSymbolsPackagesPattern)">
```
weshaggard: Do you have any non-shipping packages?

If not you should remove the usage below in PublishToAzureBlobFeed as well.

StephenBonikowsky: We don't, will do.

StephenBonikowsky:
Confirmed it passed in VSTS.

PR #2769, opened 2018-04-17
Use Expect 100-Continue header

mconnew:
Now that SocketsHttpHandler is the HTTP client being used, we can add the Expect header to avoid sending the body twice when there's an authentication challenge.

Also improved error handling for when DNS lookup fails. This was causing a test failure (wrong exception) on win7.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelHelpers.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelHelpers.cs
@@ -136,9 +136,12 @@ namespace System.ServiceModel.Channels
             var innerSocketException = exception.InnerException as SocketException;
             if (innerSocketException != null)
             {
-                var socketErrorCode = innerSocketException.SocketErrorCode;
+                SocketError socketErrorCode = innerSocketException.SocketErrorCode;
                 switch (socketErrorCode)
                 {
+                    case SocketError.TryAgain:
+                    case SocketError.NoRecovery:
+                    case SocketError.NoData:
                     case SocketError.HostNotFound:
                         return new EndpointNotFoundException(SR.Format(SR.EndpointNotFound, request.RequestUri.AbsoluteUri), exception);
                     default:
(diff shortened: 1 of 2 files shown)


zhenlan:
Add @KKhurin for code review.

This is targeting 2.1 so it will be great to get this in asap.

zhenlan:
@mconnew I didn't see tests. Can we add a test?

KKhurin:
Do we need any kind of perf sign off for this?

mconnew:
@zhenlan, this isn't something you can test as it's handled by the underlying transport code and nothing about it is exposed up the call stack. Functional testing to make sure nothing has broken is about the only thing we can do in an automated way. Perf tests should give you an increase in throughput when using authentication.

zhenlan:
@mconnew I understand it's a bit tricky for E2E functional testing. I am thinking about a navie unit test to just make sure 100-continue header actually sent out with requests. You can do that either in this PR or a separate one.

@KKhurin please only do a code review sign-off at this point. After the change is cherry-picked to the 2.1 branch, we can do perf testing against official packages, which should also be easier for you.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelHelpers.cs
```diff
@@ -136,9 +136,12 @@ public static Exception ConvertHttpRequestException(HttpRequestException excepti
...
             if (innerSocketException != null)
             {
-                var socketErrorCode = innerSocketException.SocketErrorCode;
+                SocketError socketErrorCode = innerSocketException.SocketErrorCode;
                 switch (socketErrorCode)
                 {
+                    case SocketError.TryAgain:
+                    case SocketError.NoRecovery:
+                    case SocketError.NoData:
                     case SocketError.HostNotFound:
```
KKhurin: is this an exhaustive list of all socket error codes where we could have returned "endpoint not found"?

mconnew: According to the NCL team it is exhaustive at least on Windows.

Issue #2753, opened 2018-04-11
Remove extra HEAD request and use Expect Header


mconnew:
Now that SocketsHttpHandler is the default HttpClientHandler, we can remove the extra HEAD request and use the Expect header. This can only be done on non-UAP for now as SocketsHttpHandler isn't used for UAP.

PR #2782, opened 2018-04-21
Adding Expect 100-Continue header for client certificates plus tests

mconnew:
Added scenario tests for no auth, digest auth and client certificates.


StephenBonikowsky:
It would be great if we could get this in today.

KKhurin:
@mconnew do you see any probable case when someone wants to explicitly disable 100-continue? If so should we make it easier than overriding IsExpectContinueHeaderRequired in a custom binding?

mconnew:
Fixes #2753

PR #2788, opened 2018-04-23
Stebon/2.1.0/port expect100 continue header

StephenBonikowsky:
Cherry-picks of commits in...
#2782


StephenBonikowsky:
@mconnew FYI, your new test case is also failing on the OSX10.12 platform with the same exact error as the other two failures. Can you confirm that this is expected?

PR #2807, opened 2018-04-30
Make svctuilcore as dotnetcore app. And add script to modify the package.

huanwu:
Make svcutilcore as dotnetcore app.
Draft version of script to modify the package.

@StephenBonikowsky could you help put this changeset in the server to test if it works?

cc: @mconnew @Lxiamail

### DIFF
--- a/UpdateSvcutilCorePackage.ps1
+++ b/UpdateSvcutilCorePackage.ps1
@@ -1,0 +1,122 @@
+$logfile = Join-Path $PSScriptRoot 'UpdateSvcutilCorePackage.log'
+"Log file for ModifySvcutilCorePackage............."  | Out-File $logfile
+
+if ($args.Length -ne 1)
+{
+    "Invalid number of input parameters" | Out-File $logfile -Append
+    exit;
+}
+
+$path = $args[0]
+if (![IO.Directory]::Exists($path))
+{
+    "The path " + $path + " not exist" | Out-File $logfile -Append
+    exit
+}
+
+$target = Get-Childitem –Path $path | Where-Object Name -like svcutilcore*nupkg | Select-Object -First 1
+
+#Write-Output $target
+If (!$target)
+{
+    $errormessage = "Package file doesn't exist"
+    Write-Output $errormessage
+    $errormessage | Out-File $logfile -Append
+    exit;
+}
+
+"The target is " + $target | Out-File $logfile -Append
+
+$zipfileName = Join-Path $path $target
+if(![System.IO.File]::Exists($zipfileName)){
+    $errormessage = "Package File doesn't exist"
+    Write-Output error
+    $errormessage | Out-File $logfile
+    exit;
+}
+
+$message = "Modifying Package: " + $zipfileName
+
+Write-Output $message
+$message | Out-File $logfile
+
+Add-Type -assembly  System.IO.Compression.FileSystem
+
+Try
+{
+    #Open the zip file
+    "Open the package" | Out-File $logfile -Append
+    $zip =  [System.IO.Compression.ZipFile]::Open($zipfileName,"Update")
+
+    #find svcutilcore.nuspec in the package
+    "Search for svcutilcore.nuspec" | Out-File $logfile -Append
+    $myfile = $zip.Entries.Where({$_.name -eq 'svcutilcore.nuspec'})
+    if(!$myfile)
+    {
+        $errormessage = "svcutilcore.nuspec doesn't exist in the package"
+        Write-Output $errormessage
+        $errormessage | Out-File $logfile -Append
+        exit;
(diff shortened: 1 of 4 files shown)


UpdateSvcutilCorePackage.ps1
```diff
@@ -0,0 +1,122 @@
+$logfile = Join-Path $PSScriptRoot 'UpdateSvcutilCorePackage.log'
```
StephenBonikowsky: Can we move this script out of the root?
We currently put scripts used for setup and other purposes under...
wcf\src\System.Private.ServiceModel\tools\scripts
It could go there or if you prefer another similar tools\scripts under svcutilcore.

src/svcutilcore/pkg/svcutilcore.pkgproj
```diff
@@ -18,12 +18,18 @@
...
       <TargetPath>build\</TargetPath>
     </PackageFile>
   </ItemGroup>
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.targets))\dir.targets" />
   <Import Project=".\GenerateNupkgProps.targets" />
+
+    <Target Name="ModifyPackage" AfterTargets="CreatePackage">
+        <Message Text="Call $(WcfRootFolder)UpdateSvcutilCorePackage.ps1 to modify the svcutiolcore.nuspec" Importance="high"/>
+        <Exec Command="powershell set-ExecutionPolicy Unrestricted" />
+        <Exec Command="powershell $(WcfRootFolder)UpdateSvcutilCorePackage.ps1 $(OutputPath)"/>
```
mconnew: I'm pretty certain this is bad as I think you modify the execution policy system wide. Each Exec is a new process launch so this wouldn't work unless this was persisting. Instead modify the second Exec to use the `-ExecutionPolicy UnRestricted` parameter. So it would look something like this:
```xml
<Exec Command="powershell -ExecutionPolicy UnRestricted -File $(WcfRootFolder)UpdateSvcutilCorePackage.ps1 $(OutputPath)"/>
```

Having said that, the UpdateSvcutilCorePackage.ps1 script should be signed and then we don't need to override the ExecutionPolicy.

mconnew: I just realized we don't distribute the script, so ignore the signing idea, but you should override the policy on the command line like my xml snippet shows.

---
In reply to: [188427393](https://github.com/dotnet/wcf/pull/2807#discussion_r188427393) [](ancestors = 188427393)

UpdateSvcutilCorePackage.ps1
```diff
@@ -0,0 +1,122 @@
...
+
+$path = $args[0]
+if (![IO.Directory]::Exists($path))
+{
+    "The path " + $path + " not exist" | Out-File $logfile -Append
+    exit
+}
+
+$target = Get-Childitem �Path $path | Where-Object Name -like svcutilcore*nupkg | Select-Object -First 1
+
```
mconnew: The dash before Path is an "en dash" and not a hypen like you intended. This is unicode character U+2013.

UpdateSvcutilCorePackage.ps1
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    "Open svcutilcore.nuspec for read" | Out-File $logfile -Append
+    $streamreader = [System.IO.StreamReader]($myfile).Open()
+    $filecontent = $streamreader.ReadToEnd()
+    "Before Update ............" | Out-File $logfile -Append
+    $filecontent | Out-File $logfile -Append
+    "`n" | Out-File $logfile -Append
+    $streamreader.Close()
+    $streamreader.Dispose()
```
mconnew: StreamReader.Close just calls StreamReader.Dispose so you don't need to call both.

UpdateSvcutilCorePackage.ps1
```diff
@@ -0,0 +1,122 @@
...
+    $doc = [xml] $doc.OuterXml.Replace(" xmlns=`"`"", "")
+
+    "Open svcutilcore.nuspec for write" | Out-File $logfile -Append
+    $streamwriter = [System.IO.StreamWriter]($myfile).Open()
+    $streamwriter.BaseStream.SetLength(0)
+    $doc.Save($streamwriter)
+
+    $streamwriter.Flush()
+    $streamwriter.Close()
+    $streamwriter.Dispose()
```
mconnew: Same here, only need one of Close and Dispose

huanwu:
@Lxiamail @mconnew Thanks for reviewing the script for me. This PR is just for test purpose. I will send another PR for final review. I will correct all issues based on your PR.

PR #2812, opened 2018-05-02
Change the public key of WCF assemblies to MSFT.

StephenBonikowsky:


StephenBonikowsky:
@ericstj Is this all that is needed?
Is it better to set it in each of the library's dir.props or is ok to just do it in our root dir.props (and remove the one in S.P.SM)?

ericstj:
Totally up to you for how you want to manage your repo.  In corefx we want new libraries to default to using the open key so that folks don't need us to build it to get fully signed copy.  For WCF you can decide what you want your default to be.

Have you double-checked that this matches everything that you previously shipped?  That's the most important thing to do and deserves calling out as an explicit requirement to merging this.

StephenBonikowsky:
Using S.SM.Primitives I examined the released packages for 4.4.1 and 4.4.2 and confirmed that they have the same public key.
I then verified that our 4.5.0 packages are using a different public key as @ericstj already pointed out.
I then built new packages in VSTS after setting the AssemblyKey value to 'MSFT' in our root dir.props and can confirm that the assembly in this package is back to being the same as what we shipped in 4.4.1 and 4.4.2
https://dotnet.myget.org/feed/dotnet-core/package/nuget/System.ServiceModel.Primitives/4.5.0-preview3-26503-01

StephenBonikowsky:

StephenBonikowsky:
Updating this PR with a commit to remove the AssemblyKey in the S.P.SM dir props. We only need it set once in the root dir.props.

PR #2817, opened 2018-05-03
Change the public key of WCF assemblies to MSFT. (#2812)

StephenBonikowsky:
* Change the public key of WCF assemblies to MSFT.

* Remove AssemblyKey from library specific dir.props

* We only need to set this once in the root dir.props

(cherry picked from commit 39e11fac133c972ffb97a7e54544fdc1a46bdc78)


StephenBonikowsky:
Approved by 2.1 Shiproom.

Issue #2827, opened 2018-05-07
Update wcf/pkg/baseline/packageIndex.json


StephenBonikowsky:
Update this file with all the assembly to package relationships.

StephenBonikowsky:
Completed in Master and release/2.1.0

PR #2828, opened 2018-05-07
Updating Master branch versions for next major release.

StephenBonikowsky:
I opened #2827 to fully update the packageIndex.json file, this PR is just to get packages and assemblies building out of Master correctly versioned.


zhenlan:
I suggest we split this into two PRs. One for common infrastructure changes that should go to 2.1 branch too eventually. Another for versioning change that is for master branch only. I'd like our master and 2.1 branches have as little difference as possible for infrastructure.

StephenBonikowsky:
Will do.

StephenBonikowsky:
Updated, I realized that the changes I removed were not intended for this PR in the first place, I had made them in the wrong private branch.

StephenBonikowsky:
Now that we have normalized all assemblies to have the same version I should be able to move this to the root dir.props.

PR #2830, opened 2018-05-07
Update release-notes for WCF Core 2.1 RC1 release

Lxiamail:


mconnew:
LGTM
I verified the links at least were valid links. I'm presuming there is only one rc1 version of packages on nuget in which case the links are correct.

Issue #2561, opened 2018-02-13
Test case failure on SLES - WebSocketTests/WebSocket_Https_Duplex_TextStreamed 


StephenBonikowsky:
Test Case: WebSocket_Https_Duplex_TextStreamed
Located: ...\WCF\src\System.Private.ServiceModel\tests\Scenarios\Extensibility\WebSockets\WebSocketTests.4.1.0.cs

Message
```
System.InvalidOperationException : An error occurred when creating the WebSocket with the factory of type 'CoreClrClientWebSocketFactory'. See the inner exception for details.
---- System.Net.WebSockets.WebSocketException : Unable to connect to the remote server
-------- System.Net.Http.HttpRequestException : The SSL connection could not be established, see inner exception.
------------ System.Security.Authentication.AuthenticationException : The remote certificate is invalid according to the validation procedure.
```
Stack Trace:
```
   at System.ServiceModel.Channels.ClientWebSocketTransportDuplexSessionChannel.CreateWebSocketWithFactoryAsync(X509Certificate2 certificate, TimeoutHelper timeoutHelper) in /root/wcf-1374341/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ClientWebSocketTransportDuplexSessionChannel.cs:line 225
   at System.ServiceModel.Channels.ClientWebSocketTransportDuplexSessionChannel.OnOpenAsync(TimeSpan timeout) in /root/wcf-1374341/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ClientWebSocketTransportDuplexSessionChannel.cs:line 94
   at System.ServiceModel.Channels.CommunicationObject.OnOpenAsyncInternal(TimeSpan timeout) in /root/wcf-1374341/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs:line 461
   at System.ServiceModel.Channels.CommunicationObject.System.ServiceModel.IAsyncCommunicationObject.OpenAsync(TimeSpan timeout) in /root/wcf-1374341/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs:line 436
   at System.ServiceModel.Channels.CommunicationObject.OpenAsyncInternal(TimeSpan timeout) in /root/wcf-1374341/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/CommunicationObject.cs:line 410
[...]

PR #2837, opened 2018-05-09
Expected exception changed on Linux machines

StephenBonikowsky:
* Linux machines used to throw a PNSE when trying to cert store location LocalMachine which is when we knew to use the CurrentUser location instead.
* At some point it began wrapping the PNSE inside of a CryptographicException, just need our infra logic to account for this change.


StephenBonikowsky:
Have a test run going in VSTS, if it works we can also remove the active issue #2561 from tests.

StephenBonikowsky:
The fix works, the test run in VSTS showed the test case I enabled passed.

mconnew:
LGTM

PR #2843, opened 2018-05-09
Support stable builds

StephenBonikowsky:
This is a port of dotnet/corefx PRs...

dotnet/corefx#27582
dotnet/corefx#29543

Following are the scenarios this enables, these are the comments in the original PR from @weshaggard 

This change supports building corefx packages as stable by
passing in the StabilizePackageVersions=true property.

This does edit the package index in-place but that shouldn't
cause any problems for official builds are the only  onces that
might do this. After we ship a stable set of packages the stable
versions should be commited to the repo in the package index.

There are a few scenarios:

1. Neither StabilizePackageVersions or PackageVersionStamp set
- Default dev state and all packages will have the prerelease label
  that is committed in the repo and a build number.

2. StabilizePackageVersions=false, PackageVersionStamp='label'
- All packages will have prerelease label 'label' and build number

3. StabilizePackageVersions=true, PackageVersionStamp='label'
- Packages not marked BlockStable will have prelease label 'label'
  and no build number
- BlockStable packages will have a prelease label 'label' and no build
  number

4. StabilizePackageVersions=true, PackageVersionStamp=''
- Packages not marked BlockStable will not have a prelease label
  or build number, and thus have a stable version.
- BlockStable packages will have the prelease label commited in the
  repo and no build number


StephenBonikowsky:
@weshaggard I confirmed all the scenarios locally, but I want to clarify one of your scenario descriptions....

Scenario 3: 

> StabilizePackageVersions=true, PackageVersionStamp='label'
The description says...

> BlockStable packages will have a prelease label 'label' and no build
> number
That should say "... and **a** build number" correct?

weshaggard:
>That should say "... and a build number" correct?

No it shouldn't have a build number in that scenario. That scenario is for producing final prerelease packages with no build number.

StephenBonikowsky:
I think maybe I was unclear, that scenario has two outcomes based on whether or not a package is marked BlockStable or not, its just that the description would indicate the same outcome...

> Packages not marked BlockStable will have prelease label 'label'
> and no build number
> BlockStable packages will have a prelease label 'label' and no build
> number

weshaggard:
For better or worse for things marked "BlockStable" we still enable them to produce a build without a build number so I think the test is still correct.

StephenBonikowsky:
Ok, thanks.

PR #2857, opened 2018-05-14
Fix user-agent headers

### LINKED ISSUE #2673: User-Agent has incorrectly formatted header value
See [this StackOverflow post](https://stackoverflow.com/q/49262393/2957232).  

Basically, I've been working with a .NET Standard library with a WCF reference to a SOAP web service. When I add the library as a reference to a project using the .NET framework, and set the user agent using the method outlined by [this MSDN blog post](https://blogs.msdn.microsoft.com/mohamedg/2012/12/13/adding-http-headers-to-wcf-calls/), everything is just fine. When I reference the library in a .NET Core 2.0 project, I get a System.FormatException for the value I set to the "User-Agent" header.  

For example, something like the following works in the framework, but throws an exception in Core (see the SO link for full code, this is only the implementation of IClientMessageInspector.BeforeSendRequest): 
 
    public object BeforeSendRequest(ref Message request, IClientChannel channel)
    {
        HttpRequestMessageProperty property = new HttpRequestMessageProperty();
        
        //Bad, causes an exception
        property.Headers["User-Agent"] = "Mozilla/4.0 (compatible; MSIE 6.0; MS Web Services Client Protocol 4.0.30319.42000)";

        //Works, but insufficient
        //property.Headers["User-Agent"] = "Mozilla/4.0";

        request.Properties.Add(HttpRequestMessageProperty.Name, property);

        return null;
    }

In the above mentioned SO link, user [Jeroen Mostert](https://stackoverflow.com/users/4137916/jeroen-mostert) says it is [definitely a bug.](https://stackoverflow.com/questions/49262393/user-agent-has-incorrectly-formatted-header-in-net-core#comment85563745_49262393)

mconnew:
We were incorrectly directly parsing the User-Agent header when it's a multi-valued header. This change causes `HttpRequestHeaders` to handle the special multi-value parsing that it required for the User-Agent header.

Fixes #2673

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
@@ -1264,7 +1264,7 @@ namespace System.ServiceModel.Channels
                             }
                             else if (string.Compare(name, "user-agent", StringComparison.OrdinalIgnoreCase) == 0)
                             {
-                                _httpRequestMessage.Headers.UserAgent.Add(ProductInfoHeaderValue.Parse(value));
+                                _httpRequestMessage.Headers.Add(name, value);
                             }
                             else if (string.Compare(name, "if-modified-since", StringComparison.OrdinalIgnoreCase) == 0)
                             {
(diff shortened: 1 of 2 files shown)


StephenBonikowsky:
LGTM !!

Issue #1332, opened 2016-06-24
Digest authentication on Linux/OSX doesn't work for HTTP transport


mconnew:
This is due to an issue in HttpClientHandler in corefx. See dotnet/corefx#9672

mconnew:
This is only an issue when communicating with a server which isn't using digest authentication against a domain. With a WCF service using the built-in digest authentication there should be no problem.

mconnew:
The issue is that the username is sent as realm\\username on some platforms, and as realm\username on others, even when the realm specified by the server matches. IIS is tolerant of this when using integrated authentication, but other platforms may not be.

roncain:
I see the CoreFx issue https://github.com/dotnet/corefx/issues/9672 is slated for 1.2, so this will probably not be available in 1.1

Lxiamail:
@mconnew Can you verify the issue has been fixed.

PR #2860, opened 2018-05-14
Revert earlier test change which made username validation less strict

mconnew:
With the new SocketsHttpHandler, the username is now correctly formed when using Digest authentication. This change reverts the workaround in our custom Digest authentication component in our WCF services.
This change validates that #1332 has been fixed.


StephenBonikowsky:
LGTM

PR #2874, opened 2018-05-16
Override OSX to use machine store for root certificate

mconnew:


StephenBonikowsky:
I made the changes in VSTS as well and still getting...

> Error opening StoreName: 'My' certificate store from StoreLocation 'CurrentUser' in ReadWrite mode while attempting to install cert

StephenBonikowsky:
@mconnew I believe we determined that this change won't fix the problem, it required a property being set on the machines. I don't believe this was ever done by the engineering team anyway.
I think we should close this PR.

StephenBonikowsky:
@mconnew Can we close this PR?

PR #2877, opened 2018-05-17
Remove unused code in svcutil.xmlserializer.

### LINKED ISSUE #2834: Remove XmlSerializer parameter from svcutil.xmlserializer tool
We pull out the XmlSerializer part from svcutil and make a separate tool. So we need remove this parameter in the tool and the related unnecessary code.

huanwu:
Fix #2834 
@mconnew @Lxiamail


src/svcutilcore/pkg/files/dotnet-svcutil.xmlserializer.targets
```diff
@@ -14,7 +14,7 @@
     <Delete Condition="Exists('$(_SerializerPdbIntermediateFolder)') == 'true'" Files="$(_SerializerPdbIntermediateFolder)" ContinueOnError="true" />
     <Delete Condition="Exists('$(_SerializerCsIntermediateFolder)') == 'true'" Files="$(_SerializerCsIntermediateFolder)" ContinueOnError="true" />
     <Message Text="Running SvcUtil Serialization Tool" Importance="normal" />
-    <Exec Command="dotnet svcutil.xmlserializer /t:xmlserializer $(IntermediateOutputPath)$(AssemblyName)$(TargetExt) /out:$(IntermediateOutputPath)$(_SerializationAssemblyName)" ContinueOnError="true" />
+    <Exec Command="dotnet svcutil.xmlserializer $(IntermediateOutputPath)$(AssemblyName)$(TargetExt) /out:$(IntermediateOutputPath)$(_SerializationAssemblyName)" ContinueOnError="true" />
```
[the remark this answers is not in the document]
huanwu: Yes, we need change convention so it is consistent with dotnet cli. But it is not in the scope of this PR. I will file a separate issue to track.

src/svcutilcore/tests/SvcutilTests.cs
```diff
@@ -20,10 +20,8 @@ public static void SvcutilTest1()
         {
             string outputFile = Path.GetTempFileName() + ".cs";
             Assert.False(File.Exists(outputFile));
-
-            Tool.Main(new string[] { "/t:xmlserializer", Assembly.GetExecutingAssembly().Location, $"/out:{outputFile}" });
+            Tool.Main(new string[] { Assembly.GetExecutingAssembly().Location, $"/out:{outputFile}" });
```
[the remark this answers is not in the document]
huanwu: Same as above. I will file a separate issue to track.

src/svcutilcore/tests/dotnet-svcutil.xmlserializer.Tests.csproj
```diff
@@ -80,7 +80,7 @@
     <PropertyGroup>
       <CorrectedOutputPath>$(OutputPath.Replace('/', '\'))</CorrectedOutputPath>
     </PropertyGroup>
-    <Exec Command="$(GeneratorCliPath)dotnet $(CorrectedOutputPath)dotnet-svcutil.xmlserializer.dll /t:xmlserializer $(CorrectedOutputPath)$(InputAssembly).dll /out:$(CorrectedOutputPath)$(SerializerName).cs"  />
+    <Exec Command="$(GeneratorCliPath)dotnet $(CorrectedOutputPath)dotnet-svcutil.xmlserializer.dll $(CorrectedOutputPath)$(InputAssembly).dll /out:$(CorrectedOutputPath)$(SerializerName).cs"  />
```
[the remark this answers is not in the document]
huanwu: Same as above. I will file a separate issue to track.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/OptionsStatics.cs
```diff
@@ -80,65 +80,24 @@ internal static class Targets
```
Lxiamail: Looks like the Cmd class and Abbr class needs some clean up as well

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/ToolRuntime.cs
```diff
@@ -31,29 +31,13 @@ internal ToolExitCodes Run()
                 InputModule inputModule = InputModule.LoadInputs(_options);
 
                 Tool.Assert(_options.GetToolMode().HasValue, System.SR.Format(System.SR.AmbiguousToolUseage, Options.Cmd.Target, Options.Cmd.Validate));
-                ToolMode toolMode = _options.GetToolMode().Value;
-
-                return ExecuteToolMode(toolMode, inputModule);
+                return ExecuteToolMode(inputModule);
```
Lxiamail: Do we want to keep the ExecuteToolMode still taking two params for potentially adding more mode in future?

huanwu: We can easily add it back if it is necessary in the future. But we don't have any plan or thoughts now.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/OptionsStatics.cs
```diff
@@ -8,65 +8,33 @@ internal partial class Options
...
         }
 
         internal static class Abbr
         {
-            internal const string Async = "a";
             internal const string Target = "t";
             internal const string Help = "?";
             internal const string Out = "o";
             internal const string Directory = "d";
-            internal const string Language = "l";
```
Lxiamail: Should we keep Language? it is kept in Cmd class.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/InputModule.cs
```diff
@@ -150,15 +150,6 @@ private bool LoadFileAsAssembly(FileInfo fileInfo)
 
                 _newInputModule.Assemblies.Add(assembly);
 
-                try
-                {
```
mconnew: Why was this code removed? ?This looks like validation code and SetAllowedModes has side effects (it will store the value of the last parameter).

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/Options.cs
```diff
@@ -553,10 +284,9 @@ private void ParseReferenceAssemblies()
                 bool nostdlib = _arguments.ContainsArgument(Options.Cmd.Nostdlib);
 
                 if (excludeTypesArgs != null && excludeTypesArgs.Count > 0)
-                    SetAllowedModesFromOption(ToolMode.ProxyGeneration, ToolMode.MetadataFromAssembly | ToolMode.ProxyGeneration | ToolMode.DataContractImport | ToolMode.XmlSerializerGeneration | ToolMode.ServiceContractGeneration, Options.Cmd.ExcludeType, null);
+                    SetAllowedModesFromOption(ToolMode.XmlSerializerGeneration, ToolMode.XmlSerializerGeneration, Options.Cmd.ExcludeType, null);
 
                 AddReferencedTypes(referencedAssembliesArgs, excludeTypesArgs, referencedCollectionTypesArgs, nostdlib);
-                _parent._typeResolver = CreateTypeResolver(_parent);
             }
```
mconnew: Why don't we need to add a type resolver any more?

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/ToolConsole.cs
```diff
@@ -183,7 +157,7 @@ internal static void WriteXmlSerializerTypeGenerationHelp()
                 HelpCategory helpCategory = new HelpCategory(SR.Format(SR.HelpXmlSerializerTypeGenerationCategory));
 
                 helpCategory.Description = SR.Format(SR.HelpXmlSerializerTypeGenerationDescription, ThisAssembly.Title);
-                helpCategory.Syntax = SR.Format(SR.HelpXmlSerializerTypeGenerationSyntax, ThisAssembly.Title, Options.Abbr.Target, Options.Targets.XmlSerializer, SR.Format(SR.HelpInputAssemblyPath));
+                helpCategory.Syntax = SR.Format(SR.HelpXmlSerializerTypeGenerationSyntax, ThisAssembly.Title, SR.Format(SR.HelpInputAssemblyPath));
```
mconnew: You don't need to wrap SR.HelpInputAssemblyPath with SR.Format. SR.Format does the equivalent of string.Format when you have parameters, but no parameters are being passed. Just use SR.HelpInputAssemblyPath without the SR.Format wrapping call.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/ToolConsole.cs
```diff
@@ -422,7 +248,7 @@ internal static void WriteExamples()
                 HelpCategory helpCategory = new HelpCategory(SR.Format(SR.HelpExamples));
                 helpCategory.WriteHelp();
 
-                WriteExample(SR.Format(SR.HelpExamples18), SR.Format(SR.HelpExamples19));
+                WriteExample(SR.Format(SR.HelpExamples1), SR.Format(SR.HelpExamples2));
             }
```
mconnew: Neither SR.Format is needed here.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/ToolRuntime.cs
```diff
@@ -30,30 +30,14 @@ internal ToolExitCodes Run()
...
-                case ToolMode.ServiceContractGeneration:
-                case ToolMode.DataContractExport:
-                case ToolMode.Validate:
-                case ToolMode.MetadataFromAssembly:
-                case ToolMode.WSMetadataExchange:
-                    throw new System.ArgumentException(System.SR.ErrInvalidTarget);
-            }
-
+            GenerateSerializer(inputModule);
             return ToolExitCodes.Success;
```
mconnew: You should keep the switch statement or at least an if statement here. If the tool mode is DisplayHelp but we somehow get into this method, we will run GenerateSerializer instead of it being a noop.

src/svcutilcore/tests/SvcutilTests.cs
```diff
@@ -20,10 +20,8 @@ public static void SvcutilTest1()
...
             Assert.True(File.Exists(outputFile));
-
             File.Delete(outputFile);
```
mconnew: We have code coverage tooling which generates a report showing you which methods get called. It would be good to compare that report before and after this change to make sure there isn't any code paths which are now getting skipped which we don't want to skip.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/Options.cs
huanwu: I can only see this typeresolver used in ServiceLoader in wsdlexporter.

huanwu: I of course can add it back if you feel it is more safe.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/OptionsStatics.cs
huanwu: fixed

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/InputModule.cs
huanwu: I don't think the code is useful anymore. We only have two mode. One is display, which will only be called when you pass help parameter. And the other is XmlSerializer. The call here is to set the default allowed mode is MetadataFromAssembly, and the allowed mode is ToolMode.MetadataFromAssembly , ToolMode.DataContractExport, ToolMode.Validate ,ToolMode.XmlSerializerGeneration. The validate here is to validate the service and not related to xmlserializer.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/OptionsStatics.cs
huanwu: Thanks for finding it. I will fix. :)

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/ToolConsole.cs
huanwu: Will fix.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/ToolRuntime.cs
huanwu: DisplayHelp won't invoke the code here. If the parameter contains help, it will set ToolMode as DisplayHelp and it will then be called before initiate the input model as below.

if (_options.GetToolMode() == ToolMode.DisplayHelp)
            {
                ToolConsole.WriteHelpText();
                return ToolExitCodes.Success;
            }
            else
            {
                InputModule inputModule = InputModule.LoadInputs(_options);

                Tool.Assert(_options.GetToolMode().HasValue, System.SR.Format(System.SR.AmbiguousToolUseage));
                return ExecuteToolMode(inputModule);
            }

src/svcutilcore/tests/SvcutilTests.cs
huanwu: The code coverage report for svcutil.xmlserializer is empty. What shall we do to enable it?

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/Options.cs
mconnew: _parent._typeResolver may only be used in ServiceLoader, but the method CreateTypeResolver also installs an AppDomain type resolver and assembly resolver so you don't need to explicitly call it for it to be used. This is the main logic for CreateTypeResolver:
```csharp
                TypeResolver typeResolver = new TypeResolver(options);
                AppDomain.CurrentDomain.TypeResolve += new ResolveEventHandler(typeResolver.ResolveType);
                AppDomain.CurrentDomain.AssemblyResolve += new ResolveEventHandler(typeResolver.ResolveAssembly);
```
If _parent._typeResolver isn't used, you could change the method to return void and not save a reference to the type resolver.

src/svcutilcore/tests/SvcutilTests.cs
mconnew: We need to do this as a follow up task as the code coverage run isn't configured to record coverage for svcutil. Can you create an issue for this.

mconnew:
Once you've added the type resolver back, LGTM

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/Options.cs
huanwu: To be more safe, I already added it back. I will review later and see what kind of change is required.

huanwu:
I already added TypeResolver back.

src/svcutilcore/tests/SvcutilTests.cs
huanwu: Sure.

huanwu:
The failure on OSX is unrelated to my fix.

PR #2899, opened 2018-05-24
Update WCF pks and assemblies to stable versions for 2.1 release.

StephenBonikowsky:
* Added the feed containing the final packages of dependencies.


StephenBonikowsky:
@weshaggard Any reason we shouldn't include the packageIndex.json as it was updated when setting /p:StabilizePackageVersion=true?

weshaggard:
You should only include the packageindex.json changes which mark the packages stable once the stable packages are all done and ready to release and there are no plans to respin them. You will want to do that along with a change to disable the building of packages, as you don't want to rebuild the stable versions. At that point you essentially start servicing mode and you only build packages which you plan to service and if you plan to service them you need to bump their versions.

mconnew:
LGTM

PR #2932, opened 2018-06-07
Update svcutil.xmlserializer Header.

huanwu:
Need use escaped character for \r\n in the resource. Also update the Header content for the tool.

@mconnew @Lxiamail

### DIFF
--- a/src/svcutilcore/src/Resources/Strings.resx
+++ b/src/svcutilcore/src/Resources/Strings.resx
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="Logo" xml:space="preserve">
-    <value>Microsoft (R) Service Model Metadata Tool\r\n[{0}, Version {1}]\r\n{2}\r\n</value>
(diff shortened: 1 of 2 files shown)


StephenBonikowsky:
@huanwu Is this blocked on anything? Can we either merge or close it?

huanwu:
@mconnew could you review my change?

src/svcutilcore/src/Resources/Strings.resx
```diff
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="Logo" xml:space="preserve">
-    <value>Microsoft (R) Service Model Metadata Tool\r\n[{0}, Version {1}]\r\n{2}\r\n</value>
+    <value>Microsoft (R) dotnet-svcutil.xmlserializer tool, Version {0}.&#13;&#10;[{1}]&#13;&#10;{2}&#13;&#10;</value>
   </data>
```
mconnew: We have many other instances of this problem in this resource file. I've checked all the SR code and there's nothing in there which would cause \r and \n in a resource string to be unescaped which is why this problem exists. I believe the other instances of this problem in this resource file also need to be fixed.

huanwu: Is existing change ok? I will file a separate issue to track it.

mconnew: Yes, please file an issue to fix the other resource strings. With that, this change is fine as is.

mconnew:
LGTM

PR #2943, opened 2018-06-12
Give more options for deciding which packages to build.

StephenBonikowsky:
***EDITED***

I included comments with the code, should be pretty self explanatory, here are the scenarios...

1. Nightly builds, local builds not setting any flags.
               All WCF and SvcUtil packages get built.

2. Stable packages are needed for either release or for release candidate testing.
               -If the flag for stable packages is set but the WCF or SvcUtil flags are not then no packages are built.
               -The WCF and/or SvcUtil flags must be set in conjunction with the flag for building stable packages in order to produce the desired packages.
               -The reason for this is because we don't want stable packages getting built unintentionally which would then mean we need to clean-up MyGet.

### DIFF
--- a/src/packages.builds
+++ b/src/packages.builds
@@ -4,13 +4,39 @@
 
   <PropertyGroup>
     <PackageReportDir Condition="'$(PackageReportDir)' == ''">$(BinDir)pkg/reports/</PackageReportDir>
+    <BuildAllPackages>true</BuildAllPackages>
+    <BuildWcfPackages>true</BuildWcfPackages>
   </PropertyGroup>
 
-  <ItemGroup>
-    <Project Include="$(MSBuildThisFileDirectory)..\pkg\*\*.builds" Condition="'$(SkipManagedPackageBuild)' != 'true'">
+  <ItemGroup Condition="'$(BuildAllPackages)' == 'true' AND '$(SkipManagedPackageBuild)' != 'true'" >
+    <Project Include="$(MSBuildThisFileDirectory)..\pkg\*\*.builds" >
       <AdditionalProperties>$(AdditionalProperties)</AdditionalProperties>
     </Project>
-    <Project Include="*\pkg\**\*.pkgproj" Condition="'$(SkipManagedPackageBuild)' != 'true' AND '$(BuildAllConfigurations)' == 'true'">
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(BuildAllPackages)' == 'false' AND '$(BuildWcfPackages)' == 'true'" >
+    <Project Include="$(MSBuildThisFileDirectory)\System.Private.ServiceModel\pkg\System.Private.ServiceModel.pkgproj">
+      <AdditionalProperties>$(AdditionalProperties)</AdditionalProperties>
+    </Project>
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
I verified this works by making this change in VSTS and producing the stable packages for WCF release 2.1.1.

zhenlan:
I am a little worried this further complicates the already complex build system. I am not sure if we can still tell the meaning of all combinations of `BuildAllPackages` and `BuildWcfPackages` a month later without looking at the source.

If we really have to introduce something like this, I would recommend us to be speific, for example, one flag `BuildSvcutilPackage` to control whether the build of svcutil package is included or not. Then the question becomes - are we planning to change this flag to false for a build and then change back every time? It feels weried to use source control for build control like this.

Alternatively, we can use publishing process to control things like this. I am sure you can come come ideas to use flags in VSTS build definitions. One even simpler option is to just filter the packages in the script when we download the pacakges to the file share, for example, 
`drop-download.cmd 2.1.0 26612.01 *.ServiceModel.*`

mconnew:
My preference would be to have a simple property called BuildScvutil which controls whether svcutil is even build or not. Make the default to be false and anyone can build it by using `build.cmd -- /p:BuildSvcUtilPackage=true` and similarly VSTS build definitions can easily flip that flag. Obviously the package building would also switch on this property.

zhenlan:
> Make the default to be false and anyone can build it by using build.cmd -- /p:BuildSvcUtilPackage=true

Experience tells me If we don't include something in regular build, bad thing happens over the time... :)

Corefx manages a lot more packages than us. Do they have global flags like this? How do they solve the problem that packages have different release cadence?

StephenBonikowsky:
The latest commit is based on offline conversations and I believe best reflects everybody's opinions.

- Default works like always.
- Being selective about which package to build only matters when building stable packages, so only care when that is happening.
- CoreFx uses flags just like this except they only need one flag to either build all packages or just the ones being serviced. (which are added individually to the [conditioned clause](https://github.com/dotnet/corefx/blob/release/2.1/src/packages.builds#L19))
- The build stable package flag is manually set in VSTS at queue time, this works the same way.

src/packages.builds
```diff
@@ -4,13 +4,38 @@
 
   <PropertyGroup>
     <PackageReportDir Condition="'$(PackageReportDir)' == ''">$(BinDir)pkg/reports/</PackageReportDir>
+    
+    <!-- 'StabilizePackageVersion' is used in conjunction with 'PackageVersionStamp' to determine whether or not a label and build number should be appended to a package name. -->
+    <!-- Default behavior when not building stable packages is to build both WCF and SvcUtil packages. -->
+    <!-- 'StabilizePackageVersion' is a flag set manually when stable packages are needed. -->
+    <!-- 'BuildWcfPackages' and 'BuildSvcUtilPackage' should also be explicitly set when stable packages are desired, otherwise no packages will be built. -->
+    <BuildWcfPackages Condition="'$(StabilizePackageVersion)' != 'true' AND '$(BuildWcfPackages)' == ''">true</BuildWcfPackages>
```
zhenlan: It's probably clearer to call this `BuildServiceModelPackages" since conceptually, everything in WCF repo are WCF packages.

StephenBonikowsky: True

PR #2949, opened 2018-06-13
Fix packaging issues for release of WCF 4.5.1 packages.

StephenBonikowsky:
* For target frameworks netstandard 1.0, 1.1 and 1.3 we should be harvesting assemblies from the latest package released from the 2.0.0 branch, in this case package version 4.3.1. This is because the 2.0.0 branch will continue to be serviced and we should be harvesting those most recent servicing updates.

* The package dependencies between WCF packages should always be of the same version since we now release all the WCF packages in sync.


StephenBonikowsky:
Sorry for the formatting change to the S.P.SM.pkgproj
The change is actually minimal, harvesting from the old `win7` location of the old packages and moving it to the new `win` location in the new packages.

PR #2955, opened 2018-06-18
Harvest the correct wcf package.

StephenBonikowsky:
* If not explicitly set `<HarvestVersion>` will default to the last listed stable version for each package in packageIndex.json.
* WCF needs to harvest from package version 4.3.1 for ns1.0, ns1.1 and ns 1.3.


StephenBonikowsky:
I'd like to keep the commented-out lines in the packageIndex.json as a complete record of the assemblies and packages we've shipped.
I'm doing a test run in VSTS to make sure the comments don't break anything.

pkg/baseline/packageIndex.json
```diff
@@ -43,8 +57,12 @@
       "AssemblyVersionInPackageVersion": {
         "4.0.0.0": "4.0.0",
         "4.0.1.0": "4.0.1",
-        "4.0.2.0": "4.3.1",
```
zhenlan: I would prefer keeping the mapping of 4.3.1 (the latest release) and leaving 4.3.0 in comments.

pkg/baseline/packageIndex.json
```diff
@@ -9,19 +9,28 @@
...
         "4.0.0",
         "4.1.0",
+        "4.1.1",
         "4.3.0",
-        "4.3.1"
+        "4.3.1",
+        "4.4.0",
+        "4.4.1",
+        "4.4.2",
+        "4.5.0"
```
zhenlan: Is the plan to add `4.5.1` later?

StephenBonikowsky: No reason not to do it right now, I'll amend the commit.

pkg/baseline/packageIndex.json
StephenBonikowsky: Checking with @ericstj to make sure there wasn't a specific reason for using the first pkg where the assembly first appeared. It may have some impact on package validation.

StephenBonikowsky: According to Eric...
`The purpose of the version table in the package index is to tell the packaging infrastructure the minimum package version needed for a particular assembly version.`

PR #2959, opened 2018-06-19
Explicitly harvest the desired package.

StephenBonikowsky:
* Based on PR 2955 in Master
* If not explicitly set `<HarvestVersion>` will default to the last listed stable version for each package in packageIndex.json.
* WCF needs to harvest from package version 4.3.1 for ns1.0, ns1.1 and ns 1.3.


StephenBonikowsky:

pkg/baseline/packageIndex.json
```diff
@@ -45,8 +61,12 @@
             "AssemblyVersionInPackageVersion": {
                 "4.0.0.0": "4.0.0",
                 "4.0.1.0": "4.0.1",
-                "4.0.2.0": "4.3.1",
+                //"4.0.1.0": "4.0.2",
+                "4.0.2.0": "4.3.0",
+                //"4.0.2.0": "4.3.1",
```
Lxiamail: When assembly version is the same, but package version updated. Seems like make sense to keep the latest package version. This is applicable to all the other assembly-> package version mapping in this file

StephenBonikowsky: It would seem so but if you take a look at PR #2955 @zhenlan made the same point but according to Eric this has some impact on package validation...

`The purpose of the version table in the package index is to tell the packaging infrastructure the minimum package version needed for a particular assembly version.`

Lxiamail: I see

Issue #3004, opened 2018-07-07
dotnet-svcutil.xmlserializer Could not load file or assembly System.ServiceModel.Primitives, Version=4.5.0.1


ctaggart:
I am trying to generate `VMware.vSphere.Management.Serialization` using `dotnet-svcutil.xmlserializer` to speed up serialization https://github.com/dotnet/wcf/issues/3002, but it fails with an error about not being able to load `'System.ServiceModel.Primitives, Version=4.5.0.1` if I reference the new nuget packages. Everything is building using the `microsoft/dotnet:2.1-sdk` docker image.

```
[18:46:55][Step 1/1]   Restore completed in 2.77 sec for /build/VMware.vSphere.Management.Serialization/VMware.vSphere.Management.Serialization.csproj.
[18:46:57][Step 1/1] Microsoft (R) Service Model Metadata Tool\r\n[Microsoft (R) Windows (R) Communication Foundation, Version 1.0.0-preview1]\r\nCopyright (c) Microsoft Corporation.  All rights reserved.\r\n
[18:46:57][Step 1/1] Generating XML serializers...
[18:46:57][Step 1/1] Warning: There were errors loading types in an assembly loaded from '/build/VMware.vSphere.Management/bin/release/netstandard2.0/VMware.vSphere.Management.dll' some types in the assembly could not be loaded and will not be available to the tool.
[18:46:57][Step 1/1] 
[18:46:57][Step 1/1]   Could not load file or assembly 'System.ServiceModel.Primitives, Version=4.5.0.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
[18:46:57][Step 1/1] 
[18:46:57][Step 1/1]   Could not load file or assembly 'System.ServiceModel.Primitives, Version=4.5.0.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
[18:46:57][Step 1/1] 
[18:46:57][Step 1/1] Error: An error occurred in the tool.
```

VMware.vSphere.Management.csproj
``` xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <DotNetCliToolReference Include="dotnet-svcutil" Version="1.0.1" />
  </ItemGroup>
  <ItemGroup>
[...]

PR #3013, opened 2018-07-10
example of switch to Project Sdk="Microsoft.NET.Sdk"

ctaggart:
This isn't meant to be merged. It just is an example of what was needed for me to update one the of the projects to use the newer .NET Core project system. I recommend you update. I'm was having issues with dotnet-svcutil.xmlserializer.csproj in #3004 and wanted to troubleshoot. I'll put inline comments for the main issues.


src/svcutilcore/src/dotnet-svcutil.xmlserializer.csproj
```diff
@@ -1,61 +1,39 @@
...
-    <TargetExt>.dll</TargetExt>
     <NoWarn>$(NoWarn);1634;1691;649;168;169</NoWarn>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <FileAlignment>512</FileAlignment>
-    <DebugSymbols>true</DebugSymbols>
-    <PlatformTarget>AnyCPU</PlatformTarget>
-    <ProjectGuid>{EAB2959D-0A16-4F60-A453-765491E1C069}</ProjectGuid>
+    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+    <PackAsTool>True</PackAsTool>
```
ctaggart: It would be cool if you just switched it to a .NET Core Global tool. Here is how.

PR #3022, opened 2018-07-12
Roll assemblies forward after uwp6.0 servicing release.

StephenBonikowsky:
* Increment the assembly versions.


StephenBonikowsky:
@zhenlan I didn't change the package version since the package doesn't matter, but now when looking at MyGet it's not possible to connect a uwp6.0 package to the assemblies it contains for uwp or the meta-package it ships with.
I'm wondering if we should change the pre-release label to contain a little more info, for instance "uwp6.0-MetaPkg<version>"

PR #3050, opened 2018-07-23
Fix the path so it works on linux.

huanwu:


StephenBonikowsky:
lgtm

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/InputModule.cs
```diff
@@ -166,7 +166,7 @@ private static bool TryFindFiles(string path, out FileInfo[] fileInfos)
                 // Figure out the directory part
                 string dirPath = Path.GetDirectoryName(path);
                 if (string.IsNullOrEmpty(dirPath))
-                    dirPath = @".\";
+                    dirPath = string.Concat(@".", Path.DirectorySeparatorChar);
                 DirectoryInfo dirInfo = new DirectoryInfo(dirPath);
```
mconnew: You don't need the literal @ anymore as there is no \ which would be treated as an escape character. Also `Path.DirectorySeparatorChar` is a static readonly char populated from PathInternal.DirectorySeparatorChar which is a const. If you use `"." + Path.DirectorySeparaterChar` instead of explicitly calling string.Concat, I think (only 98% sure) that the JIT will combine them into a literal constant. Either way, string.Concat is usually not needed as the compiler will emit that if necessary. The main exception being passing an array or IEnumerable of strings to be concatenated.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/OutputModule.cs
```diff
@@ -60,7 +60,7 @@ internal static string GetFilepathWithExtension(string outputFile, string extens
             internal static string TryGetDirectoryPath(string directory)
             {
                 if (directory == null || directory.Length == 0)
-                    return ".\\";
+                    return string.Concat(@".", Path.DirectorySeparatorChar);
                 else
```
mconnew: Remove the string.Concat here too.

kant2002: @huanwu would you able to make this change ?

huanwu: @kant2002 yes, just commited a new version.

StephenBonikowsky:
@huanwu Is there an update on this PR? Can we either merge or close it?

huanwu:
@mconnew could you review my change?

huanwu:
@mconnew could you review my change?

mconnew:
Lgtm

PR #3052, opened 2018-07-24
Version forward pkgs and assemblies for 1.0.0 branch.

StephenBonikowsky:
* For next servicing release.


StephenBonikowsky:
Lab machines are really flaky, there is no test or product issue here.

PR #3067, opened 2018-07-27
Change the public key of WCF assemblies to MSFT.

StephenBonikowsky:
* Change the public key of WCF assemblies to MSFT.
* Remove AssemblyKey from library specific dir.props
* We only need to set this once in the root dir.props

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -80,7 +80,7 @@
     <Platform>AnyCPU</Platform>
     <OutputType>Library</OutputType>
     <!-- Default any assembly not specifying a key to use the Open Key -->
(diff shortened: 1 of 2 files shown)


StephenBonikowsky:
Just need an LGTM

StephenBonikowsky:
@Lxiamail I believe we are planning on removing the uwp6.1 branch since we have never needed it? The uwp6.2 branch is already correct.

Lxiamail:
@StephenBonikowsky If we have confirmed that we can remove uwp6.1 branch, that sound good plan.

PR #3076, opened 2018-08-01
The exe does not exist, point to the dll instead.

StephenBonikowsky:
Minor fix to get the script working. This allows our build definitions to update the versions repo with new builds.


StephenBonikowsky:
This change is actually incorrect.
In release/2.0.0 the exe is missing and instead there is a dll, but in release/1.1.0 and release/1.0.0 there is in fact an msbuild.exe.

PR #3096, opened 2018-08-06
Update dependencies.props file in release/2.1.0

StephenBonikowsky:
* Only auto-update WCF, NETStandard.Library and buildtools package updates.
* I will also update the WCF build definition to only update the Versions repo on-demand.

### DIFF
--- a/dependencies.props
+++ b/dependencies.props
@@ -9,12 +9,9 @@
        These ref versions are pulled from https://github.com/dotnet/versions.
   -->
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
@dagood I also removed "CoreSetup" and "CoreClr" which together with "CoreFx" are used in the [runtime.depproj](https://github.com/dotnet/wcf/blob/release/2.1.0/external/runtime/runtime.depproj#L32) I'm hoping they all should be updated or not all together.

PR #3097, opened 2018-08-06
Update dependencies.props file in Master

StephenBonikowsky:
* Only auto-update WCF, NETStandard.Library and buildtools package updates.
* I will also update the WCF build definition to only update the Versions repo on-demand.
* I updated the commit hashes for buildtools and standard and then ran 'UpdateDependencies'.

note*: This makes the Master branch use buildtools from the buildtools/release/2.1.0 branch.


dagood:
I noticed this in the file although you aren't changing it in this PR:

https://github.com/dotnet/wcf/blob/eb91f2f018daf3cfe33db92d6b4c290463a114c3/dependencies.props#L32

Looks like it should be `1.0.0-beta-26226-00` based on https://github.com/dotnet/wcf/commit/85ddd35f8d14f56890cf06092a344c667d532323. (But if it's not causing problems... could remove the property.)

StephenBonikowsky:
Thanks Davis, I'm planning on going through all the dependencies.props more carefully and making sure that all the dependencies and versions are what we actually want.

ViktorHofer:
> note*: This makes the Master branch use buildtools from the buildtools/release/2.1.0 branch.

What is the reason that you are not using a live build? Asking because of some related work with xunit changes.

StephenBonikowsky:
We wanted to avoid all the churn related to buildtools and other infrastructure changes until its all a bit more settled and we have a better feel for the next release. We are also looking into moving to Arcade.

StephenBonikowsky:
Thanks for that info @ViktorHofer I am meeting with @chcosta to discuss Arcade, that should get us a better understanding on how we will be moving forward in our Master branch. I'll follow-up with you about this.

ViktorHofer:
@danmosemsft fyi

PR #3123, opened 2018-08-16
Update assembly and package versions for UWP only release of uwp6.2 f…

StephenBonikowsky:
…or Dev15.9 Preview2

* Assemblies and packages had already been rolled forward in anticipation of the next "full" servicing release.
* Since this is going to be for a UWP6.2 only release we want to avoid confusion by going back to the previous package version and roll forward only the 'revision' digit of the assembly version.


StephenBonikowsky:
Changes verified locally.

PR #1342, opened 2016-06-29
Modify digest authenticator to accept realm\\username username

mconnew:
Works around Issue #1332

Issue #3121, opened 2018-08-16
Fix digest test case for UWP


StephenBonikowsky:
Test case: "Http_ClientCredentialTypeTests.DigestAuthentication_Echo_RoundTrips_String_No_Domain"

Is failing on UWP due to strict username and pwd implementation that is not strictly necessary.
Update the test to be more username/pwd flexible so it will work on both netcoreapp and uwp frameworks.

StephenBonikowsky:
After some discussion with @mconnew the fix should be as simple as setting "_useStrictUsername" to false. I did so and still get the digest authentication failure.

Closing PR #3128 and adding @mconnew to this issue for additional investigation.

https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/tools/IISHostedWcfService/App_code/DigestServiceAuthorizationManager.cs#L164

PR #3128, opened 2018-08-21
Simplifying digest authenticator.

StephenBonikowsky:
* I believe the fix is to revert the changes from #1342.


mconnew:
You also need to remove the _useStrictUsername field and assignment. Have you validated this works for UAP as well?

StephenBonikowsky:
Closing this PR, the fix should be much simpler but the error persists, updated issue #3121 with additional info.

Issue #2711, opened 2018-03-23
Not able to consume a WCF service with CustomBinding endpoint with messageversion as Soap12


ranjitkrishnan:
I have to consumer a WCF service from my Core2.0 API. The WCF service has custombiding with messageVersion as Soap12.

Please find the wcf service binding configuration.

```xml   
<customBinding>
        <binding closeTimeout="00:10:00" openTimeout="00:10:00" sendTimeout="00:10:00">
          <textMessageEncoding  messageVersion ="Soap12">
            <readerQuotas maxDepth="2147483647" maxStringContentLength="2147483647" maxArrayLength="2147483647" maxBytesPerRead="2147483647" maxNameTableCharCount="2147483647" />
          </textMessageEncoding>
          <httpTransport maxReceivedMessageSize="2147483647" maxBufferSize="2147483647" useDefaultWebProxy="true" transferMode="Buffered" />
        </binding>
      </customBinding>
```
I am using System.ServiceModel.Primitives to create the CustomBinding from my .NET Core Code.

Please find the code below

```C#
           
            var binding = new CustomBinding();
TextMessageEncodingBindingElement textBindingElement = new TextMessageEncodingBindingElement
            {
                MessageVersion = MessageVersion.CreateVersion(EnvelopeVersion.Soap12, AddressingVersion.None),
                WriteEncoding = System.Text.Encoding.UTF8,
                ReaderQuotas = System.Xml.XmlDictionaryReaderQuotas.Max
            };

            binding.Elements.Add(textBindingElement);

            HttpTransportBindingElement httpBindingElement = new System.ServiceModel.Channels.HttpTransportBindingElement
            {
                AllowCookies = false,
                MaxBufferSize = int.MaxValue,
                MaxReceivedMessageSize = int.MaxValue,
                TransferMode= TransferMode.Buffered
            };
            binding.Elements.Add(httpBindingElement);
```
I am getting the below error from this call.

>ActionNotSupportedException: The message with Action &#x27;&#x27; cannot be processed at the receiver, due to a ContractFilter mismatch at the EndpointDispatcher. This may be because of either a contract mismatch (mismatched Actions between sender and receiver) or a binding/security mismatch between the sender and the receiver.  Check that sender and receiver have the same contract and the same binding (including security requirements, e.g. Message, Transport, None)

[...]

PR #3132, opened 2018-08-24
Add MessageVersion scenario tests.

### LINKED ISSUE #2864: Create scenario tests for all supported MessageVersion's
Not all supported `MessageVersion`'s have a public static property on the `MessageVersion` class but the `MessageVersion` can still be used with the `MessageVersion.CreateVersion(EnvelopeVersion envelopeVersion, AddressingVersion addressingVersion)` static method. We already have some tests for some of the supported message versions, but we have some missing. Priority should be to write tests for Soap12 and Soap11WSAddressing10 as we have an open issue (#2711) for these not working.  

We need tests for HTTP and NetTcp as some of the message versions have special handling with HTTP (action gets placed in headers for example). The actual test workload only needs to be calling `Echo()` as it's the SOAP envelope and transport framing we're testing and not the body.  

Here is a list of the supported message versions and how they are created on Core and the full framework. The server side should use the full framework method although the Core version would also work:

Message Version | WCF Core | Full Framework
---|---|---
none|MessageVersion.None|MessageVersion.None
Soap11|MessageVersion.Soap11|MessageVersion.Soap11
Soap12|MessageVersion.CreateVersion(EnvelopeVersion.Soap12, AddressingVersion.None)|MessageVersion.Soap12
Soap11Addressing10|MessageVersion.CreateVersion(EnvelopeVersion.Soap11, AddressingVersion.WSAddressing10)|MessageVersion.Soap11WSAddressing10
Soap12Addressing10|MessageVersion.Soap12WSAddressing10|MessageVersion.Soap12WSAddressing10
Soap11Addressing200408|MessageVersion.Soap11WSAddressingAugust2004|MessageVersion.Soap11WSAddressingAugust2004
Soap12Addressing200408|MessageVersion.Soap12WSAddressingAugust2004|MessageVersion.Soap12WSAddressingAugust2004

StephenBonikowsky:
* Fixes #2864

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
@@ -53,8 +53,13 @@ public static partial class Endpoints
     {
         get { return GetEndpointAddress("HttpSoap11.svc//http-soap11"); }
     }
-	
-	public static string HttpSoap11WSA2004_Address
+
+    public static string HttpSoap11WSA10_Address
+    {
+        get { return GetEndpointAddress("HttpSoap11WSA10.svc//http-Soap11WSA10"); }
+    }
+
+    public static string HttpSoap11WSA2004_Address
 	{
 		get { return GetEndpointAddress("HttpSoap11WSA2004.svc//http-Soap11WSA2004"); }
 	}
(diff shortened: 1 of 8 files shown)


StephenBonikowsky:
Two of these tests fail. The http variation for soap12 reproduces the customer reported failure from #2711 The tcp variation for soap12 throws an InvalidOperationException.

Details are in #2864

PR #3141, opened 2018-08-29
Update and clarify dependencies.

StephenBonikowsky:
* Distinguish between product versus test and infrastructure dependencies.
* Auto-update test and infrastructure related dependencies.
* Removed several package references that do not appear to have any impact on building or running either product or test.


StephenBonikowsky:
Also doing a test run in VSTS to verify nothing breaks on all supported platforms.

dependencies.props
```diff
@@ -10,40 +10,37 @@
...
   <PropertyGroup>
-    <WCFExpectedPrerelease>preview3-26429-01</WCFExpectedPrerelease>
-    <CoreFxExpectedPrerelease>preview3-26430-04</CoreFxExpectedPrerelease>
     <NETStandardPackageVersion>2.0.3</NETStandardPackageVersion>
     <NETStandardPackageId>NETStandard.Library</NETStandardPackageId>
-    <!--
-      This property is used for many CoreFX packages, because they have the same
-      versions. Make more properties for CoreFX if split versions are required.
-    -->
+    <CoreFxProductDependencyVersion>4.5.0</CoreFxProductDependencyVersion>
```
dagood: I think it should be pointed out that this version isn't auto-updated.

Also, since this isn't auto-updated, and it's just one version for all the CoreFX packages, the removed comment above actually applies to this situation.

If there are only two dependencies using this (using find only shows me two, at least) then making an auto-upgraded property for each individual package would actually be good IMO for future servicing.

dependencies.props
```diff
@@ -10,40 +10,37 @@
...
+  </PropertyGroup>
+
+  <!-- Tests/infrastructure dependency versions. -->
+  <PropertyGroup>
+    <WCFExpectedPrerelease>preview3-26429-01</WCFExpectedPrerelease>
```
dagood: Can this be removed (in favor of the full-version props)? Prerelease spec properties aren't ideal, since usually they're combined with some static `x.x.x` and this causes restore problems when moving to a new major/minor/patch.

dependencies.props
```diff
@@ -10,40 +10,37 @@
...
     <MicrosoftPrivateCoreFxNETCoreAppPackageVersion>4.5.0-preview3-26430-04</MicrosoftPrivateCoreFxNETCoreAppPackageVersion>
     <MicrosoftPrivateCoreFxUAPPackageVersion>4.6.0-preview3-26430-04</MicrosoftPrivateCoreFxUAPPackageVersion>
-
+    <MicrosoftNETCoreRuntimeCoreCLRPackageVersion>2.1.3-servicing-26724-06</MicrosoftNETCoreRuntimeCoreCLRPackageVersion>
+    <MicrosoftNETCorePlatformsPackageVersion>2.1.1-servicing-26724-03</MicrosoftNETCorePlatformsPackageVersion>
     <ProjectNTfsExpectedPrerelease>beta-26413-00</ProjectNTfsExpectedPrerelease>
     <ProjectNTfsTestILCExpectedPrerelease>beta-26413-00</ProjectNTfsTestILCExpectedPrerelease>
```
dagood: Same with these, it would be great to remove them and only use full-version props.

dependencies.props
StephenBonikowsky: The reason for not auto-updating those two CoreFx product dependencies is because we don't want to update those two dependencies unless absolutely necessary. So even if there are newer versions of those packages we don't want our released WCF packages forcing an upgrade of those two packages unless we have some specific reason for doing so.

However, I would like our tests to run against the latest version of these two dependencies and I was going to ask you about how to do that.

dependencies.props
StephenBonikowsky: Sure, I'll take a look at that.

dependencies.props
StephenBonikowsky: Dito

dependencies.props
dagood: Ah, that makes sense. I'd do something like this:

```xml
<!--
  Summary of the description you just gave. 😄 Plus the note
  about splitting CoreFxProductDependencyVersion if needed.
-->
<SystemReflectionDispatchProxyPackageVersion>4.5.0</SystemReflectionDispatchProxyPackageVersion>
<SystemNetHttpWinHttpHandlerPackageVersion>4.5.0</SystemNetHttpWinHttpHandlerPackageVersion>

... (in the tests section) ...
<SystemReflectionDispatchProxyTestPackageVersion>4.5.1 (or whatever)</SystemReflectionDispatchProxyTestPackageVersion>
<SystemNetHttpWinHttpHandlerTestPackageVersion>4.5.2</SystemNetHttpWinHttpHandlerTestPackageVersion>
```

Then hook up the `TestPackageVersion` props to auto-update and have those in the test runtime, while using `CoreFxProductDependencyVersion` to compile against/depend on.

(Putting the updated dependencies in the test runtime is the part I don't know as much about. It sounds right but I'm not completely sure how that works out when running the tests.)

StephenBonikowsky:
@dagood Sorry for the delay in updating this PR. You will see there are significant changes, I'll follow-up with you next week, I will be OOF tomorrow.

PR #3148, opened 2018-08-31
Update buildtools in 2.0.0

wtgodbe:
For dotnet/buildtools#2144 - takes the buildtools change to fix the copyright string.

CC @weshaggard @eerhardt 

@StephenBonikowsky I'm using a 1.1.0 version of buildtools for 2.0.0 here, as I see we were previously on 1.0.27, which I believe came from master at around the time we forked 1.1.0. Is this correct, or should I use a 2.0.0 version of buildtools?


StephenBonikowsky:
That is correct @wtgodbe thanks!

PR #3169, opened 2018-09-12
Version forward assemblies and packages for next servicing release.

StephenBonikowsky:
* Changing our versioning pattern from updating the revision number to updating the build number instead.
* This allows us to use the revision number for uwp only updates.


StephenBonikowsky:
@dotnet-bot test this please

PR #3181, opened 2018-09-24
Stebon/master/vs2017 infra changes for ci

StephenBonikowsky:
These changes are in preparation for taking the latest 2.1 buildtools update that includes a change that requires us to use machines with VS 2017 installed.

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -115,6 +115,7 @@ wcfUtilities = new WcfUtilities()
     wcfUtilities.addWcfOuterloopTestServiceSync(newJob, os, branch, isPR)
     
     newJob.with {
+    label('windows.10.amd64.clientrs4.devex.15.8.open')
         steps {
             batchFile("build.cmd -coverage -outerloop -${configurationGroup} -- /p:ShouldGenerateNuSpec=false /p:OSGroup=${osGroupMap[os]} /p:ServiceUri=%WcfServiceUri%")
         }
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
@dotnet-bot test ci please

StephenBonikowsky:
The check for testing dotnet-ci is the only check that will validate these changes.
Will merge as soon as that completes successfully.

netci.groovy
```diff
@@ -283,9 +286,10 @@ def supportedFullCycleInnerloopPlatforms = ['Windows_NT', 'Ubuntu14.04', 'Ubuntu
             if (osGroupMap[os] == 'Windows_NT')
             {
                 newJob.with {
+                label('windows.10.amd64.clientrs4.devex.15.8.open')
                     steps {
-                        batchFile("call \"C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\vcvarsall.bat\" x86 && build.cmd -${configurationGroup} -os:${osGroupMap[os]} -framework:${targetGroup}")
-                        batchFile("call \"C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\vcvarsall.bat\" x86 && build-tests.cmd -${configurationGroup} -os:${osGroupMap[os]} -framework:${targetGroup} -- /p:IsCIBuild=true")
```
Lxiamail: Do we know why we needed to call vcvarsall.bat before, but not needed now?

StephenBonikowsky:
We execute our scripts from a VS command environment, as new versions of VS come out infrastructure changes to account for it. In this case it looks like this has been here for a very long time, at some point the logic was made run.cmd (https://github.com/dotnet/wcf/blob/master/run.cmd#L26).

So this became an artifact that should have been updated/removed a long time ago but using the old VS environment didn't beak anything.

PR #3274, opened 2018-11-22
Change svcutil.xmlserializer's command line argument from "/option" to "--option"

### LINKED ISSUE #2882: change the svcutil.xmlserializer parameter convention
Need follow dotnet cli command line convention e.g. change "/out" to "--out" convention

yujayee:
Fixes #2882
With this code change, we only want to support "--" before name and "-" before abbreviation. 
Take command line option "out" for an example:
Before change, argument should start with "/out", "-out", "/o" or "-o"
After change, we only allow "--out" and "-o"

### DIFF
--- a/src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
+++ b/src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
@@ -25,13 +25,14 @@ namespace Microsoft.Tools.ServiceModel.SvcUtil.XmlSerializer
 
         internal CommandSwitch(string name, string abbreviation, SwitchType switchType)
         {
-            //ensure that either name doesn't start with '/' or '-'
+            //ensure that either name doesn't start with '--' or '-'
             //also convert to lower-case
-            if ((name[0] == '/') || (name[0] == '-'))
-                _name = (name.Substring(1)).ToLower(CultureInfo.InvariantCulture);
+            if (name.Length > 1 && name[0] == '-' && name[1] == '-')
+                _name = (name.Substring(2)).ToLower(CultureInfo.InvariantCulture);
             else
                 _name = name.ToLower(CultureInfo.InvariantCulture);
-            if ((abbreviation[0] == '/') || (abbreviation[0] == '-'))
+
+            if (abbreviation[0] == '-')
                 _abbreviation = (abbreviation.Substring(1)).ToLower(CultureInfo.InvariantCulture);
             else
                 _abbreviation = abbreviation.ToLower(CultureInfo.InvariantCulture);
@@ -60,12 +61,17 @@ namespace Microsoft.Tools.ServiceModel.SvcUtil.XmlSerializer
         {
             string temp;
 
-            //ensure that compare doesn't start with '/' or '-'
+            //ensure that compare doesn't start with '--' or '-'
             //also convert to lower-case
-            if ((other[0] == '/') || (other[0] == '-'))
-                temp = (other.Substring(1)).ToLower(CultureInfo.InvariantCulture);
-            else
-                temp = other.ToLower(CultureInfo.InvariantCulture);
+            if (other[0] == '-')
+            {
+                if (other.Length > 1 && other[1] == '-')
+                    other = other.Substring(2);
+                else
+                    other = other.Substring(1);
+            }
+            temp = other.ToLower(CultureInfo.InvariantCulture);
+
 
             //if equal to name, then return the OK
             if (_name.Equals(temp))
@@ -166,18 +172,21 @@ namespace Microsoft.Tools.ServiceModel.SvcUtil.XmlSerializer
                 bool argIsFlag = true;
 
                 //if argument does not start with switch indicator, place into "default" arguments
-                if ((arg[0] != '/') && (arg[0] != '-'))
+                if (arg[0] != '-')
                 {
                     arguments.Add(String.Empty, arg);
                     continue;
                 }
 
                 //if we have something which begins with '/' or '-', throw if nothing after it
-                if (arg.Length == 1)
+                if (arg.Length == 1 || (arg.Length ==2 && arg[1] == '-'))
                     throw new ArgumentException(SR.Format(SR.ErrSwitchMissing, arg));
 
-                //yank switch indicator ('/' or '-') off of command argument
-                arg = arg.Substring(1);
+                //yank switch indicator ('--' or '-') off of command argument
+                if (arg[1] != '-')
+                    arg = arg.Substring(1);
+                else
+                    arg = arg.Substring(2);
 
                 //check to make sure delimiter does not start off switch
                 delim = arg.IndexOfAny(new char[] { ':', '=' });
--- a/src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
+++ b/src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
@@ -25,17 +25,17 @@ namespace Microsoft.Tools.ServiceModel.SvcUtil.XmlSerializer
 
         internal CommandSwitch(string name, string abbreviation, SwitchType switchType)
         {
-            //ensure that either name doesn't start with '--' or '-'
+            //ensure name doesn't start with '--' and abbreviation doesn't start with '-'
             //also convert to lower-case
-            if (name.Length > 1 && name[0] == '-' && name[1] == '-')
-                _name = (name.Substring(2)).ToLower(CultureInfo.InvariantCulture);
-            else
-                _name = name.ToLower(CultureInfo.InvariantCulture);
+            if (name.StartsWith("--"))
+                name = name.Substring(2);
+          
+            _name = name.ToLower(CultureInfo.InvariantCulture);
 
-            if (abbreviation[0] == '-')
-                _abbreviation = (abbreviation.Substring(1)).ToLower(CultureInfo.InvariantCulture);
-            else
-                _abbreviation = abbreviation.ToLower(CultureInfo.InvariantCulture);
+            if (abbreviation.StartsWith("-"))
+                abbreviation = abbreviation.Substring(1);
+
+            _abbreviation = abbreviation.ToLower(CultureInfo.InvariantCulture);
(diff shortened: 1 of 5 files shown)


src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
```diff
@@ -25,13 +25,14 @@ internal class CommandSwitch
 
         internal CommandSwitch(string name, string abbreviation, SwitchType switchType)
         {
-            //ensure that either name doesn't start with '/' or '-'
+            //ensure that either name doesn't start with '--' or '-'
             //also convert to lower-case
-            if ((name[0] == '/') || (name[0] == '-'))
-                _name = (name.Substring(1)).ToLower(CultureInfo.InvariantCulture);
+            if (name.Length > 1 && name[0] == '-' && name[1] == '-')
```
Lxiamail: Looks like this statement didn't handled "name" starts with "-"

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
```diff
@@ -25,13 +25,14 @@ internal class CommandSwitch
...
+                _name = (name.Substring(2)).ToLower(CultureInfo.InvariantCulture);
             else
                 _name = name.ToLower(CultureInfo.InvariantCulture);
-            if ((abbreviation[0] == '/') || (abbreviation[0] == '-'))
+
+            if (abbreviation[0] == '-')
```
Lxiamail: Looks like this statement didn't handle the case if abbreviation starts with "--"

src/svcutilcore/src/Resources/Strings.resx
```diff
@@ -169,7 +169,7 @@
     <value>Description: {0} can generate code for service contracts, clients and data types from metadata documents. These metadata documents can be on disk or retrieved online. Online retrieval follows either the WS-Metadata Exchange protocol or the DISCO protocol.</value>
   </data>
   <data name="HelpCodeGenerationSyntax" xml:space="preserve">
-    <value>Syntax: {0} [/{1}:{2}]  {3}* | {4}* | {5}</value>
+    <value>Syntax: {0} [-{1}:{2}]  {3}* | {4}* | {5}</value>
```
Lxiamail: Should "-" be "--" instead?

src/svcutilcore/src/Resources/Strings.resx
```diff
@@ -208,13 +208,13 @@
     <value>The path to an assembly containing Service Contract types. Serialization types will be generated for all Xml Serializable types in each contract</value>
   </data>
   <data name="HelpXmlSerializerTypeGenerationSyntaxInput2" xml:space="preserve">
-    <value>Add the specified assembly to the set of assemblies used for resolving type references. (Short Form: /{0})</value>
+    <value>Add the specified assembly to the set of assemblies used for resolving type references. (Short Form: -{0})</value>
```
Lxiamail: Should "-" be "--" instead?

src/svcutilcore/src/Resources/Strings.resx
```diff
@@ -274,55 +274,55 @@
...
+    <value>Add the specified assembly to the set of assemblies used for resolving type references. If you are exporting or validating a service that uses 3rd-party extensions (Behaviors, Bindings and BindingElements) registered in config use this option to locate extension assemblies that are not in the GAC.  (Short Form: -{0})</value>
   </data>
   <data name="HelpNostdlib" xml:space="preserve">
     <value>Do not reference standard libraries. By default mscorlib.dll and system.servicemodel.dll are referenced.</value>
   </data>
   <data name="HelpExcludeTypeCodeGeneration" xml:space="preserve">
-    <value>A fully-qualified or assembly-qualified type name to exclude from referenced contract types. (Short Form: /{0})</value>
+    <value>A fully-qualified or assembly-qualified type name to exclude from referenced contract types. (Short Form: -{0})</value>
```
Lxiamail: Should "-" be "--" instead? This comment is applicable to all the places where "/" are replaced with "-"

src/svcutilcore/src/Resources/Strings.resx
yujayee: @Lxiamail This is called in ToolConsole.cs line 205, {1} is Options.Abbr.Target, which is the abbreviation of "target", so we should use "-" here.

src/svcutilcore/src/Resources/Strings.resx
yujayee: @Lxiamail With this code change, we only want to support "--" before name (--out) and "-" before abbreviation (-o). So this short form should start with "-".

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
```diff
@@ -166,18 +172,21 @@ internal static ArgumentDictionary ParseCommand(string[] cmd, CommandSwitch[] sw
...
-                if ((arg[0] != '/') && (arg[0] != '-'))
+                if (arg[0] != '-')
                 {
                     arguments.Add(String.Empty, arg);
                     continue;
                 }
 
                 //if we have something which begins with '/' or '-', throw if nothing after it
-                if (arg.Length == 1)
+                if (arg.Length == 1 || (arg.Length ==2 && arg[1] == '-'))
```
mconnew: Fix comment as it says we accept '/'

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
yujayee: @Lxiamail This constructor is only called in OptionsStatics.cs line 52-62. All names and abbreviations are passed in correctly. So I feel the check and conversion is not necessary here. Anyway, this code is to remove the "--" before name and "-" before abbrevation.

mconnew: This code can be made a lot more readable. It's not in a hot code path so no need to use individual character comparisons.
```c#
if (name.StartsWith("--"))
    name = name.Substring(2);

_name = name.ToLower(CultureInfo.InvariantCulture);

if (abbreviaton.StartsWith('-'))
    abbreviation = abbreviation.Substring(1);

_abbreviation = abbreviation.ToLower(CultureInfo.InvariantCulture);

File: src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs [hunk 7]
```diff
@@ -60,12 +61,17 @@ internal bool Equals(string other)
...
             string temp;
 
-            //ensure that compare doesn't start with '/' or '-'
+            //ensure that compare doesn't start with '--' or '-'
             //also convert to lower-case
-            if ((other[0] == '/') || (other[0] == '-'))
-                temp = (other.Substring(1)).ToLower(CultureInfo.InvariantCulture);
-            else
-                temp = other.ToLower(CultureInfo.InvariantCulture);
+            if (other[0] == '-')
```
mconnew: Code can be made more readable:
```c#
if (other.StartsWith("--"))
    other = other.Substring(2);
else if (other.StartsWith('-'))
    other = other.Substring(1);

if (_name.Equals(other, StringComparison.InvariantCultureIgnoreCase))
    return true;

return _abbreviation.Equals(other, StringComparison.InvariantCultureIgnoreCase));
```

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
```diff
@@ -166,18 +172,21 @@ internal static ArgumentDictionary ParseCommand(string[] cmd, CommandSwitch[] sw
...
+                if (arg[0] != '-')
                 {
                     arguments.Add(String.Empty, arg);
                     continue;
                 }
 
                 //if we have something which begins with '/' or '-', throw if nothing after it
-                if (arg.Length == 1)
+                if (arg.Length == 1 || (arg.Length ==2 && arg[1] == '-'))
                     throw new ArgumentException(SR.Format(SR.ErrSwitchMissing, arg));
```
mconnew: Can be made easier to read:
```c#
if (arg == "-" || arg == "--")
```

src/svcutilcore/src/Resources/Strings.resx
mconnew: Could you rename the string resource to make it clearer? Maybe HelpCodeGenerationAbbreviationSyntax?

---
In reply to: [236860575](https://github.com/dotnet/wcf/pull/3274#discussion_r236860575) [](ancestors = 236860575)

mconnew:
>   </data>

Does this need to be updated too? If so, are there any others in this file?

---
Refers to: src/svcutilcore/src/Resources/Strings.resx:191 in dfbfd4d. [](commit_id = dfbfd4d79862629248f9a08f04b473bbb7ab6f65, deletion_comment = False)

yujayee:
@mconnew 

> Does this need to be updated too? If so, are there any others in this file?

I didn't check the strings one by one in the resource file because there are lots of strings not in use anywhere, the one you mentioned is one of them. But I searched in ToolConsole.cs for all the places that calls resource file and fixed the corresponded string.

src/svcutilcore/src/Microsoft/Tools/ServiceModel/SvcUtil/CommandLineParser.cs
```diff
@@ -57,21 +58,19 @@ internal SwitchType SwitchType
...
+        { 
+            //ensure that compare doesn't start with '--' or '-'
             //also convert to lower-case
-            if ((other[0] == '/') || (other[0] == '-'))
-                temp = (other.Substring(1)).ToLower(CultureInfo.InvariantCulture);
-            else
-                temp = other.ToLower(CultureInfo.InvariantCulture);
-
+            if(other.StartsWith("--"))
+                other = other.Substring(2);
```
mconnew: Nit, formatting. There should be a space between the if and the open parenthesis. Same on the else if.

PR #3312, opened 2018-12-04
Fix string escaping.

### LINKED ISSUE #3186: Fix string escaping in src/System.Private.ServiceModel/src/Resources/Strings.resx
There's a bit of history with how the strings were handled in this repo as build tools and project systems evolved. When this code base was initially created, resource strings were moved to be in a json file. There was a code generation step where the literal resource strings were read, placed into quotes and emitted into generated code. This meant that the json file needed to have the text pre-escaped for C#.  An example of this is the resource string `UnableToResolveHost`.
https://github.com/dotnet/wcf/blob/9e19d2a21504595d0bfcbcb615403c47736982e5/src/System.Private.ServiceModel/src/Resources/Strings.resx#L862-L862
As the resource value is being parsed from XML, the resulting string contains a backslash followed by a quotation mark. This escape sequence is relevant when compiling code and shouldn't be in the resultant string object. The end result will be that you will see the escape sequence in output, e.g. if an exception is thrown and not caught, when the process is exiting it writes the exception message out to console. You will see the escape sequence on the console output.

StephenBonikowsky:
* Fixes #3186

### DIFF
--- a/src/System.Private.ServiceModel/src/Resources/Strings.resx
+++ b/src/System.Private.ServiceModel/src/Resources/Strings.resx
@@ -412,7 +412,7 @@
     <value>The fault reason does not contain any text translations.</value>
   </data>
   <data name="CannotDetermineSPNBasedOnAddress" xml:space="preserve">
-    <value>Client cannot determine the Service Principal Name based on the identity in the target address '{0}' for the purpose of SspiNegotiation/Kerberos. The target address identity must be a UPN identity (like acmedomain\\alice) or SPN identity (like host/bobs-machine).</value>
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
@mconnew could I get an ack if this looks good?

KKhurin:
Seems like the majority of quotes throughout the resources are single quotes. Should we also change the remaining (now un-escaped) ones to single for consistency?

StephenBonikowsky:
Opened #3315 for the consistent quotes question raised by @KKhurin

PR #3329, opened 2019-01-09
Add unit tests to validate the value of properties set in ISecurityCa…

StephenBonikowsky:
…pabilities.

* WCF uses ISecurityCapabilities and since we just made it part of the public contract adding tests to verify the expected values for its properties in all the cases where WCF uses it.

### DIFF
--- a/src/System.ServiceModel.Security/tests/ServiceModel/ISecurityCapabilitiesTest.cs
+++ b/src/System.ServiceModel.Security/tests/ServiceModel/ISecurityCapabilitiesTest.cs
@@ -1,0 +1,82 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
+using System;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using Infrastructure.Common;
+using Xunit;
+
+public static class ISecurityCapabilitiesTest
+{
+    [WcfFact]
+    public static void SecurityCapabilities_HttpsTransportBindingElement()
+    {
+        ISecurityCapabilities capability = CreateISecurityCapabilities(new HttpsTransportBindingElement());
+
+        Assert.Equal("EncryptAndSign", capability.SupportedRequestProtectionLevel.ToString());
+        Assert.Equal("EncryptAndSign", capability.SupportedResponseProtectionLevel.ToString());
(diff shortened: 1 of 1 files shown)


src/System.ServiceModel.Security/tests/ServiceModel/ISecurityCapabilitiesTest.cs
```diff
@@ -0,0 +1,82 @@
...
+public static class ISecurityCapabilitiesTest
+{
+    [WcfFact]
+    public static void SecurityCapabilities_HttpsTransportBindingElement()
+    {
+        ISecurityCapabilities capability = CreateISecurityCapabilities(new HttpsTransportBindingElement());
+
+        Assert.Equal("EncryptAndSign", capability.SupportedRequestProtectionLevel.ToString());
+        Assert.Equal("EncryptAndSign", capability.SupportedResponseProtectionLevel.ToString());
+        Assert.Equal("False", capability.SupportsClientAuthentication.ToString());
```
mconnew:
```suggestion
        Assert.False(capability.SupportsClientAuthentication);
```

mconnew: And similarly there's an Assert.True you can use too.

src/System.ServiceModel.Security/tests/ServiceModel/ISecurityCapabilitiesTest.cs
```diff
@@ -0,0 +1,82 @@
...
+using Xunit;
+
+public static class ISecurityCapabilitiesTest
+{
+    [WcfFact]
+    public static void SecurityCapabilities_HttpsTransportBindingElement()
+    {
+        ISecurityCapabilities capability = CreateISecurityCapabilities(new HttpsTransportBindingElement());
+
+        Assert.Equal("EncryptAndSign", capability.SupportedRequestProtectionLevel.ToString());
```
mconnew: You should be able to do
```c#
    Assert.Equal(ProtectionLevel.EncryptAndSign, capability.SupportedRequestProtectionLevel);
```

src/System.ServiceModel.Security/tests/ServiceModel/ISecurityCapabilitiesTest.cs
StephenBonikowsky: Thanks for the feedback, PR updated.

PR #3340, opened 2019-01-17
Stebon/master/add client base constructor to public contract

StephenBonikowsky:


StephenBonikowsky:
Yup, I didn't look closely at the core test I was using, generally all of our scenario tests are supposed to make a service call.
I'll pull this PR back and create a new one.

PR #3376, opened 2019-01-31
ongoing changes.

StephenBonikowsky:


src/System.Private.ServiceModel/Directory.Build.props
```diff
@@ -3,5 +3,9 @@
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
   <PropertyGroup>
     <WcfCommonDir Condition="'$(WcfCommonDir)' == ''">..\..\Common\src</WcfCommonDir>
+    <!-- We define our own implementation of GetResourceString -->
+    <GenerateResxSourceOmitGetResourceString>true</GenerateResxSourceOmitGetResourceString>
```
mconnew: What did corefx do here? Did they omit the GetResourceString method generation too or did they remove their implementation from SR.cs?

StephenBonikowsky: Yes, they also omit the generated version of this method in favor of the one they (and us) include in the checked-in SR.cs. I am going to follow-up with them to find out if we really need to do this, I would prefer just using the generated method if it works for us.

mconnew: If the generated one works for us, do we need to do the same as corefx?

PR #3426, opened 2019-03-02
Fix Soap12 for HTTP

### LINKED ISSUE #2711: Not able to consume a WCF service with CustomBinding endpoint with messageversion as Soap12

mconnew:
Fixes #2711 
Remove test for NetTcp as not valid config for NetTcp, only HTTP


StephenBonikowsky:
Signing-off but would like to understand more about the fix, was Soap12 never supposed to be supported in the NetTcp code path?

mconnew:
When I used the same binding config for Soap12 on .NET Framework, I got the exact same exception as we were seeing in the test on .NET Core. I checked the code path and there's an unavoidable exception (no other conditions can prevent it) when using AddressingVersion.None with NetTcp. I did a quick look for any documentation about it and couldn't find it, but the code is very straightforward.

PR #3489, opened 2019-03-22
Add ISerializationSurrogateProvider api to DataContractSerializerOperationBehavior

### LINKED ISSUE #3476: DataContractSerializerOperationBehavior missing DataContractSurrogate
While porting a large scale .net framework based application, we have been blocked by the missing property of DataContactSurrogate on the DataContractSerializerOperationBehavior . 

We have our own IDataContractSurrogate implementation that we use to do dynamic object serialization / deserialization of objects from WCF. 

We require a means to provided a custom bidirectional serializer for the DataContractSerializerOperationBehavior,  can this interface be added to the .net core / standard implementation.

yujayee:
Fixes #3476

### DIFF
--- a/src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
+++ b/src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
@@ -3,9 +3,12 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.Serialization;
 using System.ServiceModel;
 using System.ServiceModel.Description;
+using System.Text;
 using System.Threading.Tasks;
 using Infrastructure.Common;
 using Xunit;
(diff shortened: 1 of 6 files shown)


src/System.ServiceModel.Primitives/tests/Description/OperationBehaviorTest.cs
```diff
@@ -87,6 +90,37 @@ public static void XmlSerializerOperationBehavior_BasicUsage()
...
+        var members = new NonSerializableType[2];
+        members[0] = new NonSerializableType("name1", 1);
+        members[1] = new NonSerializableType("name2", 2);
+
+        using (MemoryStream ms = new MemoryStream())
+        {
+            SurrogateTestType obj = new SurrogateTestType { Members = members };
+
+            dcs.WriteObject(ms, obj);
+            ms.Position = 0;
```
mconnew: DCS is very good at serializing types which have no attributes on them. I suspect this test would work if you didn't set the surrogate provider. Can you add a bool field to the provider which gets set to true when the provider is called and then verify that it's true at the end of the test?

yujayee: @mconnew If I comment out line 99, the test will fail with an `InvalidDataContractException: Type 'NonSerializableType' cannot be serialized. Consider marking it with the DataContractAttribute attribute...`

mconnew: Good to know. It would still be nice to have an affirmative unambiguous check to know that the class has been used.

mconnew:
LGTM :shipit:

PR #3538, opened 2019-04-02
Transport security with message credential implementation

### LINKED ISSUE #8: Support security mode TransportWithMessageCredential
Message level security is not currently supported.
Attempting to use it will throw PlatformNotSupportedException

mconnew:
HTTP implementation for #8
NetTcp still to come.


mconnew:
Locally verified IIS hosted works so running again.

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
```diff
@@ -61,7 +97,305 @@ internal static bool IsAsymmetricAlgorithm(string algorithm)
...
+        }
+
+        /// <summary>
+        /// This generates the entropy using random number. This is usually used on the sending 
+        /// side to generate the requestor's entropy.
+        /// </summary>
+        /// <param name="data">The array to fill with cryptographically strong random nonzero bytes.</param>
+        public static void GenerateRandomBytes(byte[] data)
+        {
+            RandomNumberGenerator.GetNonZeroBytes(data);
```
bartonjs: Seems weird to want random non-zero.  That indicates that zero is a delimiter, and is introducing a bias on your data.

bartonjs: What target are you compiling for?  If it's netcoreapp2.1 (or higher) or netstandard2.1 then you can use the new static method `RandomNumberGenerator.Fill(data)`.  (Assuming you don't need the "non-zero" part)

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
```diff
@@ -61,7 +97,305 @@ internal static bool IsAsymmetricAlgorithm(string algorithm)
...
+        }
+
+        /// <summary>
+        /// This method generates a random byte array used as entropy with the given size. 
+        /// </summary>
+        /// <param name="sizeInBits"></param>
+        /// <returns></returns>
+        public static byte[] GenerateRandomBytes(int sizeInBits)
+        {
+            int sizeInBytes = sizeInBits / 8;
```
bartonjs: Are your bits guaranteed multiples of 8?  I recommend either asserting they are, or using `(sizeInBits + 7) / 8)`, so 0 => 0, 1 => 1, 7 => 1, 8 => 1, 9 => 2, etc.

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
```diff
@@ -61,7 +97,305 @@ internal static bool IsAsymmetricAlgorithm(string algorithm)
...
+            return data;
+        }
+
+        internal static RandomNumberGenerator RandomNumberGenerator
+        {
+            get
+            {
+                if (s_random == null)
+                {
+                    s_random = new RNGCryptoServiceProvider();
```
bartonjs: `RandomNumberGenerator.Create()`

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
```diff
@@ -61,7 +97,305 @@ internal static bool IsAsymmetricAlgorithm(string algorithm)
...
+        {
+            if (string.IsNullOrEmpty(algorithm))
+            {
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentNullException("algorithm"));
+            }
+
+            switch (algorithm)
+            {
+                //case SecurityAlgorithms.RsaSha1Signature:
+                //case SecurityAlgorithms.DsaSha1Signature:
```
bartonjs: These two should be resolvable.  http://www.w3.org/2000/09/xmldsig#rsa-sha1 => System.Security.Cryptography.Xml.RSAPKCS1SHA1SignatureDescription; http://www.w3.org/2000/09/xmldsig#dsa-sha1 => System.Security.Cryptography.Xml.DSASignatureDescription

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
```diff
@@ -61,7 +97,305 @@ internal static bool IsAsymmetricAlgorithm(string algorithm)
...
+                    return SHA256.Create();
+                case SecurityAlgorithms.Sha512Digest:
+                    return SHA512.Create();
+                case SecurityAlgorithms.Aes128Encryption:
+                case SecurityAlgorithms.Aes192Encryption:
+                case SecurityAlgorithms.Aes256Encryption:
+                case SecurityAlgorithms.Aes128KeyWrap:
+                case SecurityAlgorithms.Aes192KeyWrap:
+                case SecurityAlgorithms.Aes256KeyWrap:
+                    return Rijndael.Create();
```
bartonjs: Aes.Create(), not Rijndael.Create()

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
```diff
@@ -61,7 +97,305 @@ internal static bool IsAsymmetricAlgorithm(string algorithm)
...
+                case SecurityAlgorithms.TripleDesEncryption:
+                case SecurityAlgorithms.TripleDesKeyWrap:
+#pragma warning disable CA5350 // Do Not Use Weak Cryptographic Algorithms
+                    return TripleDES.Create();
+#pragma warning restore CA5350 // Do Not Use Weak Cryptographic Algorithms
+                case SecurityAlgorithms.HmacSha1Signature:
+                    byte[] key = new byte[64];
+                    new RNGCryptoServiceProvider().GetBytes(key);
+#pragma warning disable CA5350 // Do not use insecure cryptographic algorithm SHA1.
+                    return new HMACSHA1(key);
```
bartonjs: Seems weird that for HMACSHA1 you're doing manual-random, but HMACSHA256 you let the ctor do it.  Why not just let the HMACSHA1 ctor do it?

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
```diff
@@ -61,7 +97,305 @@ internal static bool IsAsymmetricAlgorithm(string algorithm)
...
+                    return null;
+            }
+        }
+
+
+        internal static object GetAlgorithmFromConfig(string algorithm)
+        {
+            if (string.IsNullOrEmpty(algorithm))
+            {
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentNullException("algorithm"));
```
bartonjs: Is dotnet/wcf not using `nameof()`? => `... new ArgumentNullException(nameof(algorithm))`

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
```diff
@@ -61,7 +97,305 @@ internal static bool IsAsymmetricAlgorithm(string algorithm)
...
+                case SHA256String:
+                case SecurityAlgorithms.Sha256Digest:
+                    return SHA256.Create();
+                case SecurityAlgorithms.Sha1Digest:
+#pragma warning disable CA5350 // Do not use insecure cryptographic algorithm SHA1.
+                    return SHA1.Create();
+#pragma warning restore CA5350 // Do not use insecure cryptographic algorithm SHA1.
+                case SecurityAlgorithms.HmacSha1Signature:
+#pragma warning disable CA5350 // Do not use insecure cryptographic algorithm SHA1.
+                    return new HMACSHA1(GenerateRandomBytes(64));
```
bartonjs: The earlier invocation was 64 bytes.  This is, I think, 64 bits?

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/KerberosSecurityTokenProvider.cs
```diff
@@ -59,16 +59,16 @@ public NetworkCredential NetworkCredential
             get { return _networkCredential; }
         }
 
-        internal SecurityToken GetToken(CancellationToken cancellationToken, ChannelBinding channelbinding)
+        internal SecurityToken GetToken(TimeSpan timeout, ChannelBinding channelbinding)
```
bartonjs: Aren't these breaking changes?

src/System.Private.ServiceModel/src/System/IdentityModel/SignatureResourcePool.cs
```diff
@@ -2,17 +2,56 @@
...
+                }
+
+                _hashAlgorithm = CryptoHelper.CreateHashAlgorithm(algorithm);
+            }
+            else
+            {
+                _hashAlgorithm.Initialize();
+            }
+
+            return _hashAlgorithm;
```
bartonjs: This doesn't look like it's "taking" so much as "borrowing" or "sharing".

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            if (xmlElement == null)
+                return false;
+
+            return CompareNodes(ReferenceXml, xmlElement);
+        }
+
+        private bool CompareNodes(XmlNode originalNode, XmlNode newNode)
+        {
+            if (originalNode.OuterXml == newNode.OuterXml)
```
bartonjs: OuterXml and InnerXml are expensive properties (they're uncached, for one).  Unless this check is expected to usually be true, some reordering might be warranted, such as comparing LocalName values (which are cached strings) and HasChildNodes values before doing any deep content analysis.

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityToken.cs
```diff
@@ -0,0 +1,166 @@
...
+                    // spec is closed.  SAML 2.0 is xs:ID
+                    id = tokenXml.GetAttribute("AssertionID");
+
+                    // if we are still null, "Id"
+                    if (string.IsNullOrEmpty(id))
+                    {
+                        id = tokenXml.GetAttribute("Id");
+                    }
+
+                    //This fixes the unecnrypted SAML 2.0 case. Eg: <Assertion ID="_05955298-214f-41e7-b4c3-84dbff7f01b9"
```
bartonjs: Typo: `unecnrypted` => `unencrypted`

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
```diff
@@ -0,0 +1,364 @@
...
+                    }
+                    catch (Exception e)
+                    {
+                        if (Fx.IsFatal(e))
+                            throw;
+
+                        if (!ShouldWrapException(e))
+                        {
+                            throw;
+                        }
```
bartonjs: This could be rewritten with exception filters.  `catch (Exception e) when (!Fx.IsFatal(e) && ShouldWrapException(e)) { throw Diagnostic... }`

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/X509AsymmetricSecurityKey.cs
```diff
@@ -0,0 +1,450 @@
...
+
+                                csp.Flags |= CspProviderFlags.UseExistingKey;
+                                _privateKey = new RSACryptoServiceProvider(csp);
+                            }
+                        }
+                        else
+                        {
+                            throw ExceptionHelper.PlatformNotSupported();
+                            // Need DSACertificateExtensions which is in netstandard2.1 and netcore2.0. As we target netstandard2.0, we don't
+                            // have access to DSACertificateExtensions
```
bartonjs: No ECDSA support?

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/X509AsymmetricSecurityKey.cs
```diff
@@ -0,0 +1,450 @@
...
+                {
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgument(algorithm, SR.Format(SR.EmptyOrNullArgumentString, nameof(algorithm)));
+                }
+
+                switch (algorithm)
+                {
+                    case SignedXml.XmlDsigDSAUrl:
+                        if ((this.PrivateKey as DSA) != null)
+                        {
+                            return (this.PrivateKey as DSA);
```
bartonjs: Save the value of the `as` to return it here?  `if ((this.PrivateKey as DSA dsa) != null) { return dsa; }` ?

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/X509AsymmetricSecurityKey.cs
```diff
@@ -0,0 +1,450 @@
...
+                    if (rsaSha256 == null)
+                        throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new NotSupportedException(SR.PrivateKeyNotRSA));
+                    return new RSAPKCS1SignatureFormatter(rsaSha256);
+
+                default:
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new NotSupportedException(SR.Format(SR.UnsupportedCryptoAlgorithm, algorithm)));
+            }
+        }
+
+        private static AsymmetricAlgorithm LevelUpRsa(AsymmetricAlgorithm asymmetricAlgorithm, string algorithm)
```
bartonjs: If I understand correctly, this code can't ever do anything, because you have a copy of it in the PrivateKey property getter; and the only thing you ever pass in is the PrivateKey property value.

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/X509SubjectKeyIdentifierClause.cs
```diff
@@ -0,0 +1,79 @@
...
+            return GetBuffer();
+        }
+
+        public bool Matches(X509Certificate2 certificate)
+        {
+            if (certificate == null)
+                return false;
+
+            byte[] data = GetSkiRawData(certificate);
+            return data != null && Matches(data, SkiDataOffset);
```
bartonjs: FWIW this would fail if anyone used a Subject Key Identifier in excess of 127 bytes (because then the prefix grows to a 3rd byte, or more).  Large values are not technically illegal, but will probably never happen.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReliableChannelBinder.cs
```diff
@@ -0,0 +1,2231 @@
...
+
+            if (CloseCore(out channel))
+            {
+                return;
+            }
+
+            try
+            {
+                this.OnShutdown();
+                await this.OnCloseAsync(timeoutHelper.RemainingTime());
```
bartonjs: Assuming you don't have some sort of IL rewriter running, you probably want `.ConfigureAwait(false)` on all of your awaitables.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecurityProtocol.cs
```diff
@@ -30,41 +27,619 @@ namespace System.ServiceModel.Security
...
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new SecurityNegotiationException(SR.Format(SR.TokenProviderCannotGetTokensForTarget, target), sne));
+            }
+
+            return token;
+        }
+
+        public abstract Task<Message> SecureOutgoingMessageAsync(Message message, TimeSpan timeout);
+
+        // subclasses that offer correlation should override this version
+        public virtual async Task<(SecurityProtocolCorrelationState, Message)> SecureOutgoingMessageAsync(Message message, TimeSpan timeout, SecurityProtocolCorrelationState correlationState)
```
bartonjs: I know that the type is internal, so this isn't really public, but FWIW we've come to rough conclusions for Framework Design Guidelines regarding Tuples.

* Prefer rich/named types over tuples
* Don't write the word (System.)Tuple
* Don't write the word (System.)ValueTuple
* If you feel a tuple is warranted, use named tuples with PascalCase identifiers.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/WSSecurityOneDotZeroSendSecurityHeader.cs
```diff
@@ -2,20 +2,819 @@
...
+
+        private void AddSignatureReference(MessageHeader header, string headerId, IPrefixGenerator prefixGenerator, XmlDictionaryWriter writer)
+        {
+            byte[] hashValue;
+            headerId = GetSignatureHash(header, headerId, prefixGenerator, writer, out hashValue);
+            var reference = new Reference();
+            reference.DigestMethod = AlgorithmSuite.DefaultDigestAlgorithm;
+            reference.DigestValue = hashValue;
+            reference.Id = headerId;
+            _signedXml.AddReference(reference);
```
bartonjs: I don't know your model, so I'll say "The other AddSignatureReference added an exc-c14n transform, should this one, too?" -- I'm having trouble aligning this with netfx to understand if there are two different types of thing.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/WSSecurityOneDotZeroSendSecurityHeader.cs
```diff
@@ -2,20 +2,819 @@
...
+            XmlDictionaryWriter utf8Writer = TakeUtf8Writer();
+            utf8Writer.StartCanonicalization(stream, false, null);
+            elementToSign.WriteTo(utf8Writer, ServiceModelDictionaryManager.Instance);
+            utf8Writer.EndCanonicalization();
+            stream.Position = 0;
+            var reference = new System.Security.Cryptography.Xml.Reference(stream);
+            reference.Uri = "#" + elementToSign.Id;
+            reference.DigestMethod = this.AlgorithmSuite.DefaultDigestAlgorithm;
+            reference.AddTransform(new XmlDsigExcC14NTransform());
+            _signedXml.AddReference(reference);
```
bartonjs: It really feels like perhaps you want a helper like

```C#
private void AddReference(string id, MemoryStream contents)
{
    contents.Position = 0;
    ...
    _signedXml.AddReference(reference);
}
```

because you seem to have a lot of copies of these 5 lines of code.

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
mconnew: I agree. All this code was copied from .NET Framework and I thought it was unusual, especially as there's a FillRandomBytes method immediately above which includes zeros. I understand this reduces entropy but it was intentional and I don't know why. I will open an issue to track investigating this and change it if possible but for now I'm going to match the behavior of .NET Framework.

---
In reply to: [273995601](https://github.com/dotnet/wcf/pull/3538#discussion_r273995601) [](ancestors = 273995601)

mconnew: That method doesn't look to provide any benefit over GetBytes. The benefit looks to be creating a Span for a part of an array and passing the Span to do a partial fill as that will be more efficient than GetBytes(byte[] data, int offset, int count) which does a copy. But when filling the entire array, this seems equivalent. But nice to know cleaner api's are available.

---
In reply to: [273996432](https://github.com/dotnet/wcf/pull/3538#discussion_r273996432) [](ancestors = 273996432)

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
mconnew: Original code used the parameter which specifies whether to use FIPS or not, and there wasn't an overload where you specify that without specifying FIPS. I missed that you just let it do it's own thing in the default constructor. I'll fix.

---
In reply to: [274001237](https://github.com/dotnet/wcf/pull/3538#discussion_r274001237) [](ancestors = 274001237)

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
bartonjs: Well, `Fill` is static, so you don't have to decide how you want to manage the object described as not thread-safe :smile:

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
mconnew: We can't new either of those classes as they are internal so we can't return them from this method. They will be created by our call to CryptoConfig elsewhere so we still support them, just not via this code path.

---
In reply to: [274000137](https://github.com/dotnet/wcf/pull/3538#discussion_r274000137) [](ancestors = 274000137)

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
mconnew: I fixed as many as I could do with regex find/replace but I didn't get them all. Do you know of a Roslyn analyzer which can fix this through an entire project?

---
In reply to: [274001822](https://github.com/dotnet/wcf/pull/3538#discussion_r274001822) [](ancestors = 274001822)

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
mconnew: I believe the else block covers this as it multiplies by 8 and checks you get the same value as the original parameter.

---
In reply to: [273996915](https://github.com/dotnet/wcf/pull/3538#discussion_r273996915) [](ancestors = 273996915)

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
mconnew: I'll switch the the parameterless constructor

---
In reply to: [274005077](https://github.com/dotnet/wcf/pull/3538#discussion_r274005077) [](ancestors = 274005077)

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
mconnew: I just realized I had fixed this in the second commit which has all the cleanup such as what you mentioned.

---
In reply to: [274725688](https://github.com/dotnet/wcf/pull/3538#discussion_r274725688) [](ancestors = 274725688,274001822)

src/System.Private.ServiceModel/src/System/IdentityModel/CryptoHelper.cs
mconnew: I just tried to change it and realized we're targeting netstandard2.0 still. I checked the implementation of what's returned by RandomNumberGenerator.Create() and the Fill method and they do the same thing, they all eventually call the same static method in RandomNumberGeneratorImplementation.

---
In reply to: [274725326](https://github.com/dotnet/wcf/pull/3538#discussion_r274725326) [](ancestors = 274725326)

src/System.Private.ServiceModel/src/System/IdentityModel/Selectors/KerberosSecurityTokenProvider.cs
mconnew: These are internal only methods. Although the class says it's public, it's not in our contract. We've got a lot of public implementations not in our contract because of copying code from .NET Framework.

---
In reply to: [274009300](https://github.com/dotnet/wcf/pull/3538#discussion_r274009300) [](ancestors = 274009300)

src/System.Private.ServiceModel/src/System/IdentityModel/SignatureResourcePool.cs
mconnew: I agree. Copied code from .NET Framework and that was the name there. I wanted to maintain the code the same as much as possible to reduce mental gymnastics for servicing.

---
In reply to: [274010183](https://github.com/dotnet/wcf/pull/3538#discussion_r274010183) [](ancestors = 274010183)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/GenericXmlSecurityKeyIdentifierClause.cs
mconnew: I don't know if LocalName can be different which OuterXml would be the same. I'd be worried about introducing an obscure difficult to track down bug because false is returned by a later comparison when the first one returns true.

---
In reply to: [274012427](https://github.com/dotnet/wcf/pull/3538#discussion_r274012427) [](ancestors = 274012427)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/KeyInfoSerializer.cs
mconnew: That's a good idea. This pattern exists throughout the WCF codebase extensively so I'm not going to change it here, but I will open an issue to refactor the entire codebase that way.

---
In reply to: [274014150](https://github.com/dotnet/wcf/pull/3538#discussion_r274014150) [](ancestors = 274014150)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/X509AsymmetricSecurityKey.cs
mconnew: There isn't any on .NET Framework so I didn't add it. I'll add it here.

---
In reply to: [274016549](https://github.com/dotnet/wcf/pull/3538#discussion_r274016549) [](ancestors = 274016549)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/X509AsymmetricSecurityKey.cs
mconnew: You are right. This is a bug brought over from .NET Framework. I'll remove this method.

---
In reply to: [274025148](https://github.com/dotnet/wcf/pull/3538#discussion_r274025148) [](ancestors = 274025148)

src/System.Private.ServiceModel/src/System/IdentityModel/Tokens/X509SubjectKeyIdentifierClause.cs
mconnew: I'll add a comment for future generations. This issue is also there on .NET Framework so I think you are right that it will probably never happen otherwise it would have been fixed there.

---
In reply to: [274035034](https://github.com/dotnet/wcf/pull/3538#discussion_r274035034) [](ancestors = 274035034)

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/ReliableChannelBinder.cs
mconnew: We don't need to worry about it. Because WCF public async api's are composites of lots of other api's and we pretty much always need to go async because of actual real network calls, we took a different approach. At an early point in the public api call, we check if we are using the default sync context and default task scheduler, and if we're not, we force an artificial async continuation onto a threadpool thread using the default Task Scheduler. This means we never have the ambiguous situation of not knowing if we're in a situation where ConfigureAwait is needed.

---
In reply to: [274046096](https://github.com/dotnet/wcf/pull/3538#discussion_r274046096) [](ancestors = 274046096)

src/System.Private.ServiceModel/src/System/ServiceModel/Security/WSSecurityOneDotZeroSendSecurityHeader.cs
mconnew: The difference is here we are adding a reference where the digest has already been calculated. We don't pass in the source data and have SignedXml apply the transforms and run the digest algorithm. In the other usages, we supply a memory stream which contains the XML to be signed so the transform needs to be provided. A lot of this code can probably be optimized, I was going for correctness first and then once everything is working, improve performance.

---
In reply to: [274064455](https://github.com/dotnet/wcf/pull/3538#discussion_r274064455) [](ancestors = 274064455)

PR #3555, opened 2019-04-17
Adopt arcade

StephenBonikowsky:


StephenBonikowsky:
@ericstj 
I have not been able to construct the System.ServiceModel.Primitives package the way I think it needs to be.

[This change](https://github.com/dotnet/wcf/pull/3555/commits/f4caac4a41019005c0c1b2472d21fe4cf09669aa) specifically adds a property to an existing API in the S.ServiceModel.Primitives contract that is only supported when targeting netcoreapp2.1

I believe this means that it should only add a ref assembly for netcoreapp2.1 in the package.
We don't need a netcoreapp2.1 build of S.Private.SM so I also don't think I need to add a netcoreapp2.1 dependency to S.P.SM

If this makes sense then my current problem is that while I see that this builds a ref assembly for netcoreapp2.1 it isn't getting put into the package. I believe that is because of [this logic](https://github.com/dotnet/wcf/blob/AdoptArcade/eng/FacadeAssemblies.props#L38).

Eric, can we discuss, make sure this is the correct approach and figure out the best way to add the ref assembly to the package?

src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.Ref.csproj
```diff
@@ -1,8 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <AssemblyVersion>4.6.0.0</AssemblyVersion>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
     <AssemblyName>System.ServiceModel.Primitives</AssemblyName>
     <RootNamespace>System.ServiceModel</RootNamespace>
   </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" >
```
ericstj: I think this condition is wrong.

ericstj:
You need to add a netcoreapp2.1 configuration to the facade project.  That way it can be built using your new reference assembly, and have the additional type forwards. It's ok that S.P.SM doesn't change.

StephenBonikowsky:
I don't think there are any new type forwards since it is only adding a new property inside of an existing type already being forwarded. I tried this anyway and it added an empty dependency for netcoreapp2.1 which seemed strange because tfm netstandard2.0 listed a dependency for S.Private.SM.

ericstj:
Make sure you update the condition on the dependency: https://github.com/dotnet/wcf/blob/13be0b3595d4a724bdb1d3f7167cef54dc42a4a3/src/System.ServiceModel.Primitives/src/System.ServiceModel.Primitives.Facade.csproj#L10

StephenBonikowsky:
Already updated to `<ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'netstandard2.0'">`
Is that what you would expect?

ericstj:
That will fix your empt dependency group, did it?

You could decide to change the dependency to be some other property (like not harvesting or something).

src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.Ref.csproj
ericstj: Consider
```
  <ItemGroup>
    <Compile Include="System.ServiceModel.Primitives.Manual.cs" />
    <Compile Include="System.ServiceModel.Primitives.cs" />
  </ItemGroup>
  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'" >
    <Compile Include="System.ServiceModel.Primitives.Netcoreapp.cs" />
  </ItemGroup>

File: src/System.ServiceModel.Primitives/src/System.ServiceModel.Primitives.Facade.csproj [hunk 2]
```diff
@@ -1,13 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <AssemblyVersion>4.6.0.0</AssemblyVersion>
+    <!--<TargetFrameworks>MonoAndroid10;MonoTouch10;net45;net46;net461;netcore50;netcoreapp2.1;netstandard1.0;netstandard1.1;netstandard1.3;netstandard2.0;portable-net45+win8+wp8;win8;wp8;xamarinios10;xamarinmac20;xamarintvos10;xamarinwatchos10</TargetFrameworks>-->
     <TargetFrameworks>MonoAndroid10;MonoTouch10;net45;net46;net461;netcore50;netstandard1.0;netstandard1.1;netstandard1.3;netstandard2.0;portable-net45+win8+wp8;win8;wp8;xamarinios10;xamarinmac20;xamarintvos10;xamarinwatchos10</TargetFrameworks>
```
ericstj: You need to add a TargetFramework for `netcoreapp2.1`

StephenBonikowsky:
The empty dependency for netcoreapp2.1 is fixed.

Can you confirm that I have to add a lib assembly for netcoreapp2.1 even though the all type forwards are identical?
The type forward `[assembly: TypeForwardedTo(typeof (DataContractSerializerOperationBehavior))]` is all that is needed and already exists in the netstandard2.0 lib assembly.

I guess the question is, if someone targets netcoreapp2.1 using S.SM.Primitives, and the process doesn't find a lib assembly for netcoreapp2.1 won't the hierarchical process  end up using the netstandard2.0 lib assembly?

Or does "best-practices" indicate that it is better to have the specific lib assembly even if it is identical?

ericstj:
> end up using the netstandard2.0 lib assembly?

Yes, it would, but...

> Or does "best-practices" indicate that it is better to have the specific lib assembly even if it is identical

Exactly.  It's a special case that the only thing you've added is members.  You don't want to set up a system that folks will need to think about these special cases.  The expectation is that you establish patterns that will work in general moving forward.

In the future you'll want to re-enable API-compat, at which point it matters that you have a different ref for each of these.  You'll want API compat to run on both facades and use different references for each.

StephenBonikowsky:
Ok, thanks for the clarification.

src/System.ServiceModel.Primitives/src/System.ServiceModel.Primitives.Facade.csproj
```diff
@@ -1,13 +1,14 @@
...
     <AssemblyName>System.ServiceModel.Primitives</AssemblyName>
     <CLSCompliant>true</CLSCompliant>
     <IsPackable>true</IsPackable>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'netstandard2.0'">
```
ericstj: You may want to try something like:
```xml
<BuildFrameworks>netstandard;netcoreapp2.1;</BuildFrameworks>
<HarvestFrameworks>MonoAndroid10; ...;</HarvestFrameworks>
<TargetFrameworks>$(BuildFrameworks)$(HarvestFrameworks)</TargetFrameworks>
<IsBuilding Condition="$(BuildFrameworks.Contains('$(TargetFramework);'))">true</IsBuilding>
<IsHarvesting Condition="$(BuildFrameworks.Contains('$(TargetFramework);'))">true</IsHarvesting>
```

That could simplify conditions a little.  Up to you.

StephenBonikowsky: I like that, it provides built-in clarity as well.

PR #3592, opened 2019-05-01
Add setup steps for Linux and MacOS test runs.

StephenBonikowsky:
* We need to install certificates on test machines before tests are run on Linux and MacOS machines.

### DIFF
--- a/eng/TestProperties.props
+++ b/eng/TestProperties.props
@@ -12,4 +12,9 @@
     <WcfScenarioTestCommonProj>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\src\System.Private.ServiceModel\tests\Common\Scenarios\ScenarioTests.Common.Tests.csproj'))</WcfScenarioTestCommonProj>
     <WcfInfrastructureCommonProj>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\src\System.Private.ServiceModel\tests\Common\Infrastructure\Infrastructure.Common.csproj'))</WcfInfrastructureCommonProj>
   </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Directory that is zipped up and sent as a correlation payload to Helix -->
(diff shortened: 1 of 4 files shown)


StephenBonikowsky:
@jonfortescue the examples in the Arcade readme have "DefaultTargets="Test" as a property of the Project, because of where I am adding this to the build system not sure if that matters.

StephenBonikowsky:
By the way, tests aren't being run yet with the PR checks so that won't work to validate this, have to merge to know for sure.

eng/pipelines/macos.yml
```diff
@@ -79,6 +79,7 @@ jobs:
       buildExtraArguments: ''
 
       variables:
+        - _helixPreCommands: "chmod a+x $HELIX_CORRELATION_PAYLOAD/InstallRootCertificate.sh;sudo -E -n $HELIX_CORRELATION_PAYLOAD/InstallRootCertificate.sh --service-host $(ServiceUri) --cert-file /tmp/wcfrootca.crt"
```
jonfortescue: I still think that you should probably add the cert as a correlation payload a reference it that way because I have a feeling that there was something else that managed getting it on the machine and putting it in temp

eng/TestProperties.props
```diff
@@ -12,4 +12,9 @@
     <WcfScenarioTestCommonProj>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\src\System.Private.ServiceModel\tests\Common\Scenarios\ScenarioTests.Common.Tests.csproj'))</WcfScenarioTestCommonProj>
     <WcfInfrastructureCommonProj>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\src\System.Private.ServiceModel\tests\Common\Infrastructure\Infrastructure.Common.csproj'))</WcfInfrastructureCommonProj>
   </PropertyGroup>
+
+  <ItemGroup>
```
jonfortescue: this will need to be in a separate .proj file and you'll need to make sure you execute msbuild on that proj file. You'll also need to make sure you're including the workitems.

PR #3593, opened 2019-05-02
Add setup steps for Linux and MacOS test runs.

StephenBonikowsky:
* We need to install certificates on test machines before tests are run on Linux and MacOS machines.

### DIFF
--- a/eng/SendToHelix.proj
+++ b/eng/SendToHelix.proj
@@ -1,0 +1,25 @@
+<Project Sdk="Microsoft.DotNet.Helix.Sdk" DefaultTargets="Test">
+  <PropertyGroup>
+    <Language>msbuild</Language>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <HelixAccessToken Condition="'$(_isOfficialBuild)' == 'true'">$(_helixAccessToken)</HelixAccessToken>
+    <WaitForWorkItemCompletion Condition="'$(_isOfficialBuild)' == 'false'">true</WaitForWorkItemCompletion>
+    <Creator Condition="'$(_isOfficialBuild)' == 'false'">dotnet-wcf</Creator>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="!$(HelixTargetQueue.StartsWith('Windows'))">
+    <HelixPreCommands>"chmod a+x $HELIX_CORRELATION_PAYLOAD/InstallRootCertificate.sh;sudo -E -n $HELIX_CORRELATION_PAYLOAD/InstallRootCertificate.sh --service-host $(ServiceUri) --cert-file /tmp/wcfrootca.crt"</HelixPreCommands>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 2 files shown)


eng/SendToHelix.proj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.DotNet.Helix.Sdk" DefaultTargets="Test">
+  <PropertyGroup>
+    <Language>msbuild</Language>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <HelixAccessToken Condition="'$(_isOfficialBuild)' == 'true'">$(_helixAccessToken)</HelixAccessToken>
+    <WaitForWorkItemCompletion Condition="'$(_isOfficialBuild)' == 'false'">true</WaitForWorkItemCompletion>
+    <Creator Condition="'$(_isOfficialBuild)' == 'false'">dotnet-wcf</Creator>
```
jonfortescue: should be !$(_isOfficialBuild)

eng/SendToHelix.proj
```diff
@@ -0,0 +1,25 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup Condition="!$(HelixTargetQueue.StartsWith('Windows'))">
+    <HelixPreCommands>"chmod a+x $HELIX_CORRELATION_PAYLOAD/InstallRootCertificate.sh;sudo -E -n $HELIX_CORRELATION_PAYLOAD/InstallRootCertificate.sh --service-host $(ServiceUri) --cert-file /tmp/wcfrootca.crt"</HelixPreCommands>
```
jonfortescue: Likely you'll want to prepend these with `$(HelixPreCommands)` to avoid overwriting command that the SDK creates itself. It shouldn't happen, but better safe than sorry.
```suggestion
    <HelixPreCommands>"chmod a+x $HELIX_CORRELATION_PAYLOAD/InstallRootCertificate.sh;sudo -E -n $HELIX_CORRELATION_PAYLOAD/InstallRootCertificate.sh --service-host $(ServiceUri) --cert-file /tmp/wcfrootca.crt"</HelixPreCommands>
```
Also, I do think you should consider sending the cert up as a second correlation payload.

eng/SendToHelix.proj
```diff
@@ -0,0 +1,25 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Directory that is zipped up and sent as a correlation payload to Helix -->
+    <HelixCorrelationPayload Include="$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\src\System.Private.ServiceModel\tools\scripts'))" />
```
jonfortescue: You need to include the `PayloadDirectory` property on this item and set it to match the identity

eng/SendToHelix.proj
StephenBonikowsky: The certificate is actually on our server, it gets pulled down from there and installed by this script.

jonfortescue: ah perfect! That will be fine then

PR #3596, opened 2019-05-02
Stebon/master/produce net461 facades

StephenBonikowsky:
@ericstj This produces wcf facades for 4.6.1 that type-forward to desktop assemblies. Please review and let me know if I missed anything.

I haven't yet run the test scenario that caught this issue, but will in a little while.

### DIFF
--- a/eng/FacadeAssemblies.targets
+++ b/eng/FacadeAssemblies.targets
@@ -2,17 +2,18 @@
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <IsFacadeAssembly Condition="'$(IsFacadeAssembly)'=='' AND ($(MSBuildProjectName.EndsWith('.Facade')))">true</IsFacadeAssembly>
+    <IsPartialFacadeAssembly Condition="'$(IsPartialFacadeAssembly)'=='' AND ($(MSBuildProjectName.EndsWith('.Facade')))">true</IsPartialFacadeAssembly>
   </PropertyGroup>
   
-  <PropertyGroup Condition="'$(IsFacadeAssembly)'=='true'">
-    <IsPartialFacadeAssembly>true</IsPartialFacadeAssembly>
-    <GenFacadesIgnoreMissingTypes>true</GenFacadesIgnoreMissingTypes>
+  <PropertyGroup Condition="'$(IsPartialFacadeAssembly)'=='true'">
+    <!-- [todo:arcade] Set GenFacadesIgnoreMissingTypes to true only for the ServiceModel shim -->
+    <GenFacadesIgnoreMissingTypes Condition="'$(IsReferenceAssembly)' != 'true'">true</GenFacadesIgnoreMissingTypes>
     <!--Don't produce or publish symbols for Facade assemblies.-->
     <PublishWindowsPdb>false</PublishWindowsPdb>
     <IsPackable>true</IsPackable>
     <!-- GenFacades requires a PDB -->
     <DebugType>portable</DebugType>
+    <NoWarn Condition="'$(IsReferenceAssembly)' == 'true'">$(NoWarn);0436</NoWarn>
   </PropertyGroup>
 
   <ItemGroup Condition="'$(IsPartialFacadeAssembly)'=='true'">
@@ -20,7 +21,7 @@
   </ItemGroup>
 
   <!--This is where we tell GenFacades what Ref assembly to look for.-->
-  <Target Condition="'$(IsPartialFacadeAssembly)'=='true'" Name="ResolveMatchingContract" AfterTargets="ResolvePackageDependenciesForBuild">
+  <Target Condition="'$(IsPartialFacadeAssembly)'=='true' AND '$(IsReferenceAssembly)' != 'true'" Name="ResolveMatchingContract" AfterTargets="ResolvePackageDependenciesForBuild">
     <ItemGroup>
       <ProjectReference Include="..\ref\$(AssemblyName).Ref.csproj">
         <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
--- a/eng/FacadeAssemblies.targets
+++ b/eng/FacadeAssemblies.targets
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <PropertyGroup>
+  <!--<PropertyGroup>
     <IsPartialFacadeAssembly Condition="'$(IsPartialFacadeAssembly)'=='' AND ($(MSBuildProjectName.EndsWith('.Facade')))">true</IsPartialFacadeAssembly>
-  </PropertyGroup>
+  </PropertyGroup>-->
   
   <PropertyGroup Condition="'$(IsPartialFacadeAssembly)'=='true'">
     <!--Don't produce or publish symbols for Facade assemblies.-->
--- a/eng/FacadeAssemblies.targets
+++ b/eng/FacadeAssemblies.targets
@@ -1,9 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
-  <!--<PropertyGroup>
-    <IsPartialFacadeAssembly Condition="'$(IsPartialFacadeAssembly)'=='' AND ($(MSBuildProjectName.EndsWith('.Facade')))">true</IsPartialFacadeAssembly>
-  </PropertyGroup>-->
   
   <PropertyGroup Condition="'$(IsPartialFacadeAssembly)'=='true'">
     <!--Don't produce or publish symbols for Facade assemblies.-->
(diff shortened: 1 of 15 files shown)


StephenBonikowsky:
The manual test project works with these new packages.

src/System.ServiceModel.NetTcp/src/System.ServiceModel.NetTcp.Facade.csproj
```diff
@@ -32,11 +37,6 @@
     <Reference Include="System.IdentityModel" />
     <Reference Include="System.ServiceModel" />
   </ItemGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
-    <PackageReference Include="System.ServiceModel.Primitives" Version="4.5.3" />
```
Lxiamail: Why the package reference is removed? Same comment applies to all files. From the previous released facade package, it had package dependency on System.ServiceModel.Primitives.

StephenBonikowsky: The last released packages do have that dependency. If we keep them in these new packages it should be to the new 4.6.0 packages.
I will re-build these packages and re-run the manual test project and use a type in S.SM.Http that has a dependency to S.SM.Primitives.

StephenBonikowsky: @ericstj If it is correct that .NETFramework, Version=v4.6.1 have a dependency to S.SM.Primitives and version 4.6.0 then I'm not sure how we did that in the buildtools world. Since it isn't built yet and therefore not available on any feed.

src/System.ServiceModel.Duplex/src/System.ServiceModel.Duplex.Facade.csproj
```diff
@@ -5,18 +5,23 @@
...
     <IsHarvesting Condition="$(HarvestFrameworks.Contains('$(TargetFramework);'))">true</IsHarvesting>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(IsBuilding)' == 'true'">
+  <ItemGroup Condition="'$(IsBuilding)' == 'true' And '$(TargetFramework)' != 'net461'">
     <ProjectReference Include="..\..\System.Private.ServiceModel\src\System.Private.ServiceModel.csproj" />
     <ProjectReference Include="..\..\System.ServiceModel.Primitives\ref\System.ServiceModel.Primitives.Ref.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(IsBuilding)' == 'true' And '$(TargetFramework)' == 'net461'">
```
ericstj: IsBuilding is redundant here.

src/System.ServiceModel.NetTcp/src/System.ServiceModel.NetTcp.Facade.csproj
ericstj: The reference project should have a P2P reference to it.  We do that because a .NETStandardLibrary will need it, even if the facade doesn't actually refrence it (since it has no typedefs) we preserve the reference.  Outside the pkproj world you need to manually specify your project reference.

In general, be careful about conditioning refrences in the framework specific builds that are in the netstandard build: doing so could break the libraries who only build for netstandard but are consumed in netframework.

src/System.ServiceModel.Duplex/src/System.ServiceModel.Duplex.Facade.csproj
```diff
@@ -5,18 +5,23 @@
...
+    <BuildFrameworks>netstandard2.0;net461;</BuildFrameworks>
+    <HarvestFrameworks>MonoAndroid10;MonoTouch10;net45;netcore50;netstandard1.1;netstandard1.3;portable-net45+win8;win8;xamarinios10;xamarinmac20;xamarintvos10;xamarinwatchos10;</HarvestFrameworks>
     <TargetFrameworks>$(BuildFrameworks)$(HarvestFrameworks)</TargetFrameworks>
     <IsBuilding Condition="$(BuildFrameworks.Contains('$(TargetFramework);'))">true</IsBuilding>
     <IsHarvesting Condition="$(HarvestFrameworks.Contains('$(TargetFramework);'))">true</IsHarvesting>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(IsBuilding)' == 'true'">
+  <ItemGroup Condition="'$(IsBuilding)' == 'true' And '$(TargetFramework)' != 'net461'">
     <ProjectReference Include="..\..\System.Private.ServiceModel\src\System.Private.ServiceModel.csproj" />
```
ericstj: Only condition this projectReference for !net461.

StephenBonikowsky:
I believe this addresses your comments @ericstj 
But I'm concerned with the referenced assemblies in the ref/net461 facade dll.
For instance, in pkg 4.5.3 S.SM.Primitives the referenced assemblies are...
mscorlib
System
System.IdentityModel
System.Runtime.Serialization
System.ServiceModel
System.Xml

But in the package I'm building right now I'm only getting 'netstandard'

Is it correct that it be netstandard or should it be referencing the desktop libraries specifically as it does in our 4.5.3 packages?

ericstj:
I'll build your branch and have a look.

ericstj:
This is only making the src projects build for net4x, ref projects are not building for net4x.

StephenBonikowsky:
With this last commit everything is building and packages being created as expected.
Static analysis looks good.
Need to merge, create some official signed packages and begin additional testing.

eng/FacadeAssemblies.targets
```diff
@@ -1,26 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <PropertyGroup>
-    <IsFacadeAssembly Condition="'$(IsFacadeAssembly)'=='' AND ($(MSBuildProjectName.EndsWith('.Facade')))">true</IsFacadeAssembly>
-  </PropertyGroup>
+  <!--<PropertyGroup>
```
ericstj: Delete?

StephenBonikowsky: Ah, yes

eng/ReferenceAssemblies.props
```diff
@@ -1,6 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <PropertyGroup>
```
ericstj: nit its a little unusual to put this in the referenceassemblies.props file.

maybe a "conventions.props" or just put them in root directory.build.props?

EG:
```
<PropertyGroup>
  <IsPartialFacadeAssembly Condition="'$(IsPartialFacadeAssembly)'=='' AND ($(MSBuildProjectName.EndsWith('.Facade')))">true</IsPartialFacadeAssembly>
  <IsReferenceAssembly Condition="'$(IsReferenceAssembly)'=='' AND ($(MSBuildProjectName.EndsWith('.Ref')))">true</IsReferenceAssembly>
</PropertyGroup>
```

src/System.ServiceModel.Shim/System.ServiceModel.Shim.csproj
```diff
@@ -7,7 +7,6 @@
     <GenFacadesIgnoreMissingTypes>true</GenFacadesIgnoreMissingTypes>
     <!--Don't produce or publish symbols for Facade assemblies.-->
     <PublishWindowsPdb>false</PublishWindowsPdb>
-    <IsPartialFacadeAssembly>true</IsPartialFacadeAssembly>
```
ericstj: Why did you remove this?  I'm missing where this project sets this property now.

eng/ReferenceAssemblies.props
StephenBonikowsky: I agree, I have some other props files that also need re-visiting, was going to move it somewhere more appropriate then, but the root directory.build.props sounds like a better choice for now.

src/System.ServiceModel.Shim/System.ServiceModel.Shim.csproj
StephenBonikowsky: Does it actually need that property?
For this project we don't want to use any of the logic in FacadeAssemblies.targets right?
GenFacades and ResolvedMatchingContract happens right there in the same project.

By leaving IsPartialFacadeAssembly set to true it was also running the ResolveMatchingContract target in FacadeAssemblies.targets and trying to find a related Ref.csproj.

StephenBonikowsky: My above comments are not completely correct.
By removing  `<IsPartialFacadeAssembly>true</IsPartialFacadeAssembly>` I was also not triggering the AddReferenceAssemblyToPackage target in FacadeAssemblies.targets
Made changes to call that target and added a condition to skip the logic in the ResolveMatchingContract target.

@ericstj I do see one diff from previous packages that I'm not sure about. Maybe we can follow-up on this next week, the shim shipped in 4.5.3 also had a reference to System.Runtime that I am not getting right now. Not sure if it is needed.

ericstj: I believe that GenFacades requires the property in order to run the facade generation target.  Did you confirm it still has type forwards after this change?

StephenBonikowsky: My last update before merging added that property back and made a couple additional modifications to make sure we were doing the right thing regarding the targets in FacadeAssemblies.targets over the weekend our vendors did some package testing including specifically the shim and everything looked good.

Issue #3610, opened 2019-05-08
WebSockets incorrectly posting continuations to application SynchronizationContext.


StephenBonikowsky:
The WCF WebSocket implementation when calling receive on a WebSocket is posting continuations to the application SynchronizationContext which we expect to have dropped by that point but is not.

The fix is to drop the sync context, investigation is needed to identify the correct place to insert this fix.

Test cases failing due to this issue...
...Scenarios/Client/ChannelLayer/RequestReplyChannelShapeTests.4.0.0.cs 
- IRequestChannel_Async_Http_BasicHttpBinding
- IRequestChannel_Async_Http_CustomBinding

...Scenarios/Client/ChannelLayer/DuplexChannelShapeTests.4.0.0.cs
- IDuplexSessionChannel_Async_Tcp_NetTcpBinding

...Scenarios\Binding\WS\TransportWithMessageCredentialSecurity\WSHttpTransportWithMessageCredentialSecurityTests.cs
- Https_SecModeTransWithMessCred_CertClientCredential_Succeeds

...Scenarios\Binding\WS\TransportWithMessageCredentialSecurity\BasicHttpTransportWithMessageCredentialSecurityTests.cs
- Https_SecModeTransWithMessCred_CertClientCredential_Succeeds

StephenBonikowsky:
Verify tests are fixed now.

StephenBonikowsky:
None of these tests are being skipped now for this issue and they are passing on all runs.

PR #3635, opened 2019-05-13
Fix ChannelLayer tests

mconnew:
Fix SynchronizationContext issue when calling IDuplexSessionChannel.Receive directly.
Fix tests which aren't correctly closing session.


StephenBonikowsky:
I believe this is the fix for issue #3610 correct?

StephenBonikowsky:
The CI failures are unrelated to this PR, they are related to two test failures, one due to an infrastructure issue that is being investigated and the other is a bug in CoreFx that has already been fixed for which we have to pick-up the latest packages for test.

PR #3640, opened 2019-05-15
Fix some threading issues when using a non-default sync context

mconnew:
We have a policy that if a non-default sync context is being used, to avoid potential deadlock (e.g. with a single threaded sync context such as used by winforms and WPF) or thread starvation problems, we forcefully asynchronously continue execution on the worker thread pool.
We had a few code paths where this didn't happen. This fixes those code paths.


StephenBonikowsky:
The failing test cases are unrelated.

PR #3642, opened 2019-05-16
Make test projects support both netcoreapp3.0 and netcoreapp2.1

StephenBonikowsky:
* For sending to Helix use netcoreapp2.1 as the primary scenario.
* Will add netcoreapp3.0 as a secondary test run in Helix as part of the SendToHelix project.

### DIFF
--- a/eng/pipelines/wcf-base.yml
+++ b/eng/pipelines/wcf-base.yml
@@ -106,11 +106,13 @@ jobs:
         # Send to Helix variables
         - ${{ if eq(job.submitToHelix, 'true') }}:
           - _helixType: build/test/
-          - _xUnitPublishTargetFramework: netcoreapp3.0
+          - _xUnitPublishTargetFramework: netcoreapp2.1
           - _xUnitRuntimeTargetFramework: netcoreapp2.0
           - _xUnitRunnerVersion: 2.4.1
-          - _dotNetCliPackageType: sdk
-          - _dotNetCliVersion: 2.1.11 #the version of dotnet for running the tests as listed in the global.json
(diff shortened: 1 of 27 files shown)


StephenBonikowsky:
@jonfortescue could you see if I'm missing something here. This is failing the checks right now...
https://github.com/dotnet/arcade/blob/70b35c1bff68f11011e9f35818142a344f208961/src/Microsoft.DotNet.Helix/Sdk/tools/dotnet-cli/DotNetCli.targets#L5

I've been going through the Arcade documentation but haven't hit on the problem yet.
I just changed `DotNetCliPackageType `from "sdk" to "runtime" since I'm pulling in netcoreapp2.1 as an additional CLI runtime in the global.json, but that didn't make a difference.

I've made sure that `XUnitPublishTargetFramework `is set to netcoreapp2.1 and that DotNetCl`i`Version is also specifying the netcoreapp2.1 release version.

StephenBonikowsky:
Some notes for posterity...
- Build tests for more than one TFM by adding TFMs to `<TargetFrameworks>` in each test project.
- Download additional dotnet runtimes in global.json as needed for each TFM you add to test projects.
- The version of the additional runtimes should exist in [releases.json](https://raw.githubusercontent.com/dotnet/core/master/release-notes/releases.json)
- Setup XUnit in Helix to run the tests on the desired runtime by setting the following properties...
          `XUnitPublishTargetFramework:` [tfm] (such as netcoreapp2.1)
          `DotNetCliPackageType:` runtime
          `DotNetCliVersion:` [same runtime version specified in the global.json]

PR #3655, opened 2019-05-29
Fix WS HTTP Sync Context problem

mconnew:
Also includes a small test infrastructure fix when developing in VS.


StephenBonikowsky:
Looks great @mconnew 
Merging.

PR #3661, opened 2019-06-03
Build and package dotnet-svcutil tool.

StephenBonikowsky:
Update the project as necessary to build and package in the new Arcade infrastructure environment.
Verify test project also builds and tests run and pass.
Do initial package validation.
Assign to vendors for additional validation.


StephenBonikowsky:
First update was to get the test project added to the yaml for ADO to build and run it.
Second update was to fix a property needed for the 'Send to Helix' step but not needed when running locally which is why I missed it.

StephenBonikowsky:
Going to skip waiting on the CI checks so I can get the official build started.

PR #3664, opened 2019-06-04
Update scripts used on server for updating service changes.

StephenBonikowsky:
* Needed to be updated to account for new "artifacts" folder location.
* Also updating to use newer VS command prompt and tools.


StephenBonikowsky:
Going to push this PR through since the changes I made won't be validated by the normal test build and run.
Once it is in I can verify the changes on the server.

PR #3679, opened 2019-06-07
Fix component dependency vulnerabilities.

StephenBonikowsky:
* Update "NETStandard.Library" pkg dependencies to version 2.0.3 to get the latest serviced netstandard packages.
* Explicitly added "System.Net.Security" and "System.Net.WebSockets.Client" to use updated serviced versions not being picked-up implicitly by S.P.SM.

### DIFF
--- a/src/System.ServiceModel.Duplex/src/System.ServiceModel.Duplex.Facade.csproj
+++ b/src/System.ServiceModel.Duplex/src/System.ServiceModel.Duplex.Facade.csproj
@@ -21,6 +21,7 @@
   <PropertyGroup Condition="'$(IsHarvesting)' == 'true'">
     <IsPartialFacadeAssembly>false</IsPartialFacadeAssembly>
     <HarvestFromPackage>true</HarvestFromPackage>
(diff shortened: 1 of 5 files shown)


StephenBonikowsky:
@ericstj 

Updating the version of the NETStandard.Library to 2.0.3 fixed the problem of pulling in version 4.3.0 of System.Net.Http

Explicitly adding v4.3.1 of both "System.Net.Security" and "System.Net.WebSockets.Client" for TFMs 'netcore50' and 'netstandard1.3' does mean that I get those versions now BUT because I have a package reference in those same TFMs to "System.Private.ServiceModel" Version="4.5.3" the project.assets.json is still pulling in v4.3.0 of both of those packages as dependencies of S.P.SM.
I think this will still cause the component governance tool to throw an alert?

I've been trying to find a way to reference System.Private.ServiceModel without having it include its dependencies in the project.assets.json file but no luck so far.

ericstj:
> Explicitly adding v4.3.1 of both "System.Net.Security" and "System.Net.WebSockets.Client" for TFMs 'netcore50' and 'netstandard1.3' does mean that I get those versions now BUT because I have a package reference in those same TFMs to "System.Private.ServiceModel" Version="4.5.3" the project.assets.json is still pulling in v4.3.0 of both of those packages as dependencies of S.P.SM.
I think this will still cause the component governance tool to throw an alert?

It shouldn't.  If it does that is a bug in the tool.

PR #3706, opened 2019-06-26
Fix issue with WebSocket already being closed sometimes

mconnew:
Sometimes DuplexChannelShapeTests were throwing when running locally as the WebSocket was already closed. This fixes the it by not attempting to close and already closed WebSocket.


StephenBonikowsky:
LGTM, sometimes when there is a lot of demand for MacOS we will time out waiting for machines to be available. That's all this was, re-running them.

PR #3709, opened 2019-06-26
Fixed problem with HTTP streamed mode timeout

### LINKED ISSUE #3590: Why timeout helper not passed to GetStreamAsync()
Hi all,

I'm trying to understand why my application sometimes hangs when working with WCF (net core 2.2). I'm using HttpTransportBindingElement with TransferMode equal to StreamedResponse.
And the question is, why timeoutHelper is not used in GetStreamAsync() here ? 
https://github.com/dotnet/wcf/blob/f0b35d2f8d8352082d1e4f4f413db62b24e25ec5/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs#L299
Where is guarantee that we will not hang here because of a network error ?
Let me give some additional info. Here we say that we don't need timeout for HttpClient:
https://github.com/dotnet/wcf/blob/f0b35d2f8d8352082d1e4f4f413db62b24e25ec5/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs#L358
Then we say that only response headers should be read:
https://github.com/dotnet/wcf/blob/f0b35d2f8d8352082d1e4f4f413db62b24e25ec5/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs#L1024
So, response's body should be read manually by ReadAsync/PreReadBufferAsync in GetStreamAsync() method and we don't have any protection from network timeouts. I'm right?

mconnew:
We weren't passing an appropriate CancellationToken when reading from the stream.
Fixes #3590

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
@@ -39,7 +39,7 @@ namespace System.ServiceModel.Channels
         {
             ValidateAuthentication();
             ValidateResponseStatusCode();
-            bool hasContent = await ValidateContentTypeAsync();
+            bool hasContent = await ValidateContentTypeAsync(timeoutHelper);
             Message message = null;
 
             if (!hasContent)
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
@mconnew Should we add a scenario test for this?

mconnew:
Normally I'd say yes, but writing a scenario test to try and hit this would require not using WCF on the server side and taken more direct control of sending the response. As there's no logic flow change, simply propagating a timeout to calls which should have been using it, I think we can make do without scenario tests for this one.

PR #3737, opened 2019-07-12
Fixes for publishing to BAR.

StephenBonikowsky:


StephenBonikowsky:
@JohnTortugo This should reflect the changes you made yesterday that finally got a successful publish to BAR.

PR #3772, opened 2019-07-25
Fix script to require VS 2019.

StephenBonikowsky:
* Since converting to .NET 3.0 SDK projects we now need that SDK in order to build.


StephenBonikowsky:
Merging because other PRs are dependent on this fix.

PR #3821, opened 2019-08-14
Update the contract release notes.

StephenBonikowsky:
* For Preview8


StephenBonikowsky:
doc update only, no need to wait on checks.

PR #3836, opened 2019-08-22
Update the build pipeline to use the stages based approach.

StephenBonikowsky:
- Also simplified everything I could into the single root yaml.
- Generally cleaned-up a lot of things as well.


StephenBonikowsky:
@chcosta I had questions about the following things hoping you can help answer...

[SendToHelix.proj line 3](https://github.com/dotnet/wcf/blob/stagesbasedbuilds/eng/SendToHelix.proj#L3): Do we need to differentiate between CI and PR runs?

[SendToHelix.proj line 7](https://github.com/dotnet/wcf/blob/stagesbasedbuilds/eng/SendToHelix.proj#L7): This is a default build number, keep this?

[SendToHelix.proj line 46](https://github.com/dotnet/wcf/blob/stagesbasedbuilds/eng/SendToHelix.proj#L46): This is needed?

[SendToHelix.proj line 56](https://github.com/dotnet/wcf/blob/stagesbasedbuilds/eng/SendToHelix.proj#L56): This project needs to be packed?

[SendToHelix.proj line 66](https://github.com/dotnet/wcf/blob/stagesbasedbuilds/eng/SendToHelix.proj#L66): Is this still needed?

chcosta:
> Do we need to differentiate between CI and PR runs?

The way I interpret that is that it's just poorly named (considering our dnceng conventions but back in the pipebuild days it made more sense) and that pr == public, official == internal.  @alexperovich would know for sure.

> This is a default build number, keep this?

I don't think the value really matters.  In Azure DevOps builds it should get set by the previous line and local builds don't (currently) send to helix unless you do some acrobatics

> This is needed?

You appear to be specifying that parameter in your yaml, but I don't think it's really needed.  you should be running as public if you're in the public project and dont' define a HelixAccessToken.  Again @alexperovich may know better.

> This project needs to be packed?

I don't think so, this looks like it's overriding the "Pack" target and basically saying don't pack.

> Is this still needed?

Yes, probably.  Otherwise MSBuild won't know if it's a C# or VB or other project and will try to do things like produce binaries and such for a project that is just executing MSBuild code (not really a build project).

StephenBonikowsky:
> I don't think so, this looks like it's overriding the "Pack" target and basically saying don't pack.

Could I just use `<IsPackable>false</IsPackable>` instead?

chcosta:
Not sure if "IsPackable" works with MSBuild projects.  I'd guess that, since MSBuild projects are indeed packable, it does.  Worth a try!

alexperovich:
The "Pack" target definition is needed because arcade unconditionally calls it and will fail if it doesn't exist. The Language property is also needed for a similar reason. Arcade SDK will fail the build if it can't figure out the Language for a project.

PR #3840, opened 2019-08-26
picking up an older no longer used authenticode cert

robertborr:
This is similar to : https://github.com/dotnet/wcf/pull/3828
where an older, not valid, cert is being requested.

@StephenBonikowsky 
@MattGal


robertborr:
Its under release/2.1.0

PR #3872, opened 2019-09-04
WCF Issue#3853 - support System.IM.Selectors.SecurityTokenProvider

HongGit:
…Provider


HongGit:
@mconnew we are exposing only the class and the constructor, the rest of the APIs in this class still need to be implemented?  Can you you please review?

mconnew:
We need to expose everything on this class as it can be overriden.

mconnew:
@HongGit, you can close this PR as this is part of the iceberg of API's I mentioned before so I'm going to tackle it.

Issue #3857, opened 2019-08-30
[P2] Support System.IdentityModel.Tokens.SecurityToken


StephenBonikowsky:
| APIs | Port/Implement | Public Contract | Testing |
| :------------ | :-------------: | :------------: | :------------: |
`abstract class S.IM.Tokens.SecurityToken` | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Added</li></ol> | <ol><li>[x] Scenario</li></ol> |
`ctor internal SecurityToken()` | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Not Applicable</li></ol> | <ol><li>[x] Unit N/A <li>[x] Scenario</li></li></ol> |
`virtual method T CreateKeyIdentifierClause<T>() where T : SecurityKeyIdentifierClause`  | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit N/A <li>[x] Scenario</li></li></ul> |
`property abstract string Id { get; }` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
`property abstract ReadOnlyCollection<SecurityKey> SecurityKeys { get; }` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
`property abstract DateTime ValidFrom { get; }` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
***
**API Notes**
May only need to expose the class with no properties or methods.
***
**Implementation Notes**
Already ported.
***
**Public Contract Notes**
The default constructor cannot be made internal because then our other contracts such as System.ServiceModel.Security cannot access it. So we will have to allow this class to be derivable (is that a word?).

The method `CreateKeyIdentifierClause` does not need to be made public, it was being called by the Exchange implementation of WS Trust Client which is going away, in the context of this project it is only called by other WCF code so we do not need to add it to the public project.

- Public Contract: System.ServiceModel.Primitives
***
**Testing Notes**
- Unit tests are required.
- Scenario tests needed: ???

***
**PR Links**
Added to the PC: #3873 
Unit tests added with PR #3902
Scenario tests tracked by: #4022

Issue #3856, opened 2019-08-30
[P2] Support System.IdentityModel.Tokens.SecurityKey


StephenBonikowsky:
| APIs | Port/Implement | Public Contract | Testing |
| :------------ | :-------------: | :------------: | :------------: |
`abstract class S.IM.Tokens.SecurityKey` | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Added</li></ol> | <ol><li>[ ] Unit <li>[x] Scenario</li></li></ol> |
`method byte[] EncryptKey(string algorithm, byte[] keyData)` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
`property int KeySize { get; }` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
***
**API Notes**

The two properties probably do not need to be exposed, just the class.
**After further investigation we will also need to add KeySize.**
***
**Implementation Notes**
Already ported.
***
**Public Contract Notes**
- Public Contract: System.ServiceModel.Primitives
***
**Testing Notes**
- Unit tests are required.
- Scenario tests needed: ???

***
**PR Links**
Added 'SecurityKey' constructor to the PC: #3873 
Added 'KeySize' to the PC with: #3905
Unit tests added with: #3937
Scenario tests tracked by: #4022

Issue #3846, opened 2019-08-28
[P0] Support System.ServiceModel.Security.Tokens.BinarySecretSecurityToken


StephenBonikowsky:
| APIs | Port/Implement | Public Contract | Testing |
| :------------ | :-------------: | :------------: | :------------: |
class S.SM.S.T.BinarySecretSecurityToken | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Added</li></ol> | <ol><li>[x] Unit <li>[x] Scenario</li></li></ol> |
***
**API Notes**
This entire class should be supported.
***
**Implementation Notes**
Already ported.
***
**Public Contract Notes**

- Public Contract: System.ServiceModel.Security

***
**Testing Notes**
- Unit tests are required.
- Scenario tests needed: ???

***
**PR Links**
Added to the PC: #3873
Unit tests added with: #3902
Scenario tests tracked by: #4022

PR #3873, opened 2019-09-04
Adding S.SM.Security.Tokens.BinarySecretSecurityToken to the public c…

StephenBonikowsky:
…ontract.

* This is tracked by issue #3846
* This also required adding S.IM.Tokens.SecurityToken to the public contract.(tracked with #3857)
* S.IM.Tokens.SecurityToken required adding S.IM.Tokens.SecurityKey. (tracked with #3856)


StephenBonikowsky:
I added only the minimal set for...
System.IdentityModel.Tokens.SecurityKey
System.IdentityModel.Tokens.SecurityToken
Based on the required types added for BinarySecretSecurityToken

It may be that further down the road some other type will require more from those classes.

Issue #3759, opened 2019-07-24
[P1] Support System.ServiceModel.Security.SecurityAlgorithmSuite


soyang2828:
| APIs | Port/Implement | Public Contract | Testing |
| :------------ | :-------------: | :------------: | :------------: |
`class SecurityAlgorithmSuite` | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Added</li></ol> | <ol><li>[x] Unit <li>[x] Scenario</li></li></ol> |
`SecurityAlgorithmSuite.TripleDes` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
`S.SM.Channels.SecurityBindingElement. DefaultAlgorithmSuite` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
***
**API Notes**
- class SecurityAlgorithmSuite: No instance members needed
- SecurityAlgorithmSuite.TripleDes: Static property
- Also need to expose S.SM.Channels.SecurityBindingElement.DefaultAlgorithmSuite because that is how they set SecurityAlgorithmSuite.TripleDes on the binding.
***
**Implementation Notes**

***
**Public Contract Notes**

- Public Contract for SecurityAlgorithmSuite: System.ServiceModel.Security
- Public Contract for S.SM.Channels.SecurityBindingElement.DefaultAlgorithmSuite: System.ServiceModel.Security
***
**Testing Notes**
- Unit tests are required.
- Scenario tests needed: Yes
      Take existing TransportWithMessageCredential test and duplicate. In the new test, modify the DefaultAlgorithmSuite to use TripleDes. This will need a new service endpoint which uses TripleDes too.

***
**PR Links**
Porting and Adding to the PC: #3875
Unit tests added via: #3915
Scenario tests tracked by: #4022

mconnew:
Can you please list the properties that you need available? We have this implemented but not added to the contract but we are currently only using (so only have implementation for) `SecurityAlgorithmSuite.Basic256`. Do you only need the static properties to select a security algorithm suite or do you need to access the instance properties on a concrete implementation such as the instance property DefaultDigestAlgorithm?

soyang2828:
This is from our internal scan. I take it as the instance property on SecurityAlgorithmSuite

[...]

PR #3875, opened 2019-09-05
Port System.ServiceModel.Security.SecurityAlgorithmSuite.TripleDes

StephenBonikowsky:
* Also including adding the APIs to the public contract.
* SecurityAlgorithmSuite was already added to the S.SM.Security contract, added the static property TripleDes to this contract.
* Also adding SecurityBindingElement.DefaultAlgorithmSuite to the public surface area since that is how SecurityAlgorithmSuite.TripleDes is being set in this scenario.

This PR handles porting (from full fx) and adding types to the public contract for WSFederation project work item #3759


StephenBonikowsky:
@mconnew could you do a quick review, I'd like to get it in today if possible.

PR #3885, opened 2019-09-06
Update pkg & assembly versions for 3.1

StephenBonikowsky:
* Pkg version is now 4.7.0
* Assembly versions are now 4.7.0.0


StephenBonikowsky:
Using this version number based on discussion with @HongGit 
Not waiting for checks, need to verify with internal build.

Issue #3859, opened 2019-08-30
[P2] Support System.ServiceModel.Channels.BinaryMessageEncodingBindingElement


StephenBonikowsky:
| APIs | Port/Implement | Public Contract | Testing |
| :------------ | :-------------: | :------------: | :------------: |
`class S.SM.Channels. BinaryMessageEncodingBindingElement` | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Added</li></ol> | <ol><li>[x] Unit <li>[x] Scenario</li></li></ol> |
`property int MaxReadPoolSize` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
`property int MaxWritePoolSize` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
***
**API Notes**
***
**Implementation Notes**
Already ported.
***
**Public Contract Notes**
- Public Contract: System.ServiceModel.Security
***
**Testing Notes**
- Unit tests are required.
- Scenario tests needed: ???

***
**PR Links**
Added to PC in: #3886
Unit tests added via: #3924
Scenario tests tracked by: #4022

PR #3886, opened 2019-09-07
[P2] Support System.ServiceModel.Channels.BinaryMessageEncodingBindin…

HongGit:
…gElement #3859


src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
```diff
@@ -844,6 +844,8 @@ public sealed partial class BinaryMessageEncodingBindingElement : System.Service
         public System.ServiceModel.Channels.CompressionFormat CompressionFormat { get { return default(System.ServiceModel.Channels.CompressionFormat); } set { } }
         [System.ComponentModel.DefaultValueAttribute(2048)]
         public int MaxSessionSize { get { return default(int); } set { } }
+        public int MaxReadPoolSize { get; set; }
```
StephenBonikowsky: I think both of these need to declare the default return type like this...

`public int MaxReadPoolSize { get { retrun default(int); } set { } }`

PR #3901, opened 2019-09-11
Add S.IM.T.SymmetricSecurityKey to the public contract.

StephenBonikowsky:
* Also had to make the class public in the implementation.


StephenBonikowsky:
@mconnew The class had been made private in the implementation for some reason, it dates back to the initial port from full fx.

mconnew:
LGTM

PR #3902, opened 2019-09-11
Unit test for BinarySecretSecurityToken class

imcarolwang:
#3846

### DIFF
--- a/src/System.ServiceModel.Security/tests/ServiceModel/BinarySecretSecurityTokenTest.cs
+++ b/src/System.ServiceModel.Security/tests/ServiceModel/BinarySecretSecurityTokenTest.cs
@@ -1,0 +1,40 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
+using System;
+using System.IdentityModel.Tokens;
+using System.ServiceModel.Security.Tokens;
+using Infrastructure.Common;
+using Xunit;
+
+public static class BinarySecretSecurityTokenTest
+{
+    [WcfFact]
+    public static void Ctor_Default_Properties()
+    {
+        string id = "test id";
+        byte[] keyBytes = new byte[128];
+        Random rnd = new Random();
+        rnd.NextBytes(keyBytes);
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
@imcarolwang I ran out of time today but I will review this tomorrow.
Thanks!

src/System.ServiceModel.Security/tests/ServiceModel/BinarySecretSecurityTokenTest.cs
```diff
@@ -0,0 +1,40 @@
...
+    public static void Ctor_Default_Properties()
+    {
+        string id = "test id";
+        byte[] keyBytes = new byte[128];
+        Random rnd = new Random();
+        rnd.NextBytes(keyBytes);
+
+        BinarySecretSecurityToken bsst = new BinarySecretSecurityToken(keyBytes);
+        Assert.NotNull(bsst);
+        Assert.NotNull(bsst.Id);
```
mconnew: When using this constructor, the id is auto-generated and is supposed to be unique. Can you add a test to validate that a second instance of BinarySecretSecurityToken using this constructor has a different Id?

src/System.ServiceModel.Security/tests/ServiceModel/BinarySecretSecurityTokenTest.cs
```diff
@@ -0,0 +1,40 @@
...
+        Assert.NotNull(bsst.SecurityKeys);
```
mconnew: Can you add a test that SecurityKeys only contains a single key.

src/System.ServiceModel.Security/tests/ServiceModel/BinarySecretSecurityTokenTest.cs
imcarolwang: Thanks for review! Added a new test method for this.

src/System.ServiceModel.Security/tests/ServiceModel/BinarySecretSecurityTokenTest.cs
imcarolwang: Added single key element validation.

Issue #3847, opened 2019-08-28
[P0] Support System.IdentityModel.Tokens.GenericXmlSecurityToken


StephenBonikowsky:
| APIs | Port/Implement | Public Contract | Testing |
| :------------ | :-------------: | :------------: | :------------: |
`class S.IM.T.GenericXmlSecurityToken` | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Added</li></ol> | <ol><li>[x] Unit <li>[x] Scenario</li></li></ol> |
`ctor GenericXmlSecurityToken(non-default)` | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Added</li></ol> | <ol><li>[x] Unit <li>[x] Scenario</li></li></ol> |
***
**API Notes**
This entire class needs to be supported.
***
**Implementation Notes**
Already ported.
***
**Public Contract Notes**

- Public Contract: System.ServiceModel.Security

***
**Testing Notes**
- Unit tests are required.
- Scenario tests needed: ???

***
**PR Links**
Added to PC with: #3884
Unit tests added with: #3913
Scenario tests tracked by: #4022

PR #3913, opened 2019-09-16
Issue3847

imcarolwang:
For #3847.

I am not sure if it's appropriate to put the file in new test folder "IdentityModel" ?
@StephenBonikowsky @mconnew @HongGit


mconnew:
This test should be in a new folder. I'm not a fan of spending lots of time debating naming conventions so my preference would be to leave it in the folder name you created and if later we decide we don't like it, we can change it easily.

PR #3909, opened 2019-09-13
Add IssuedSecurityTokenParameters class and SecurityBindingElement.CreateIssuedTokenOverTransportBindingElement

mconnew:
For #3760 and #3809

PR #3915, opened 2019-09-17
Add unit tests for S.SM.Security.SecurityAlgorithmSuite

StephenBonikowsky:
* Also removed several properties from the S.SM.Security public contract added recently via pr #3909
* We shouldn't add items to the public contract unless explicitly needed for this feature.
* Also changed the name of the dotnet-svcutil-xmlserializer test project since it includes scenario tests that call a service endpoint therefore the tests will fail because the naming convention was that of a unit test project and no service is setup when running unit tests only.


src/System.ServiceModel.Security/tests/ServiceModel/SecurityAlgorithmSuiteTest.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Reflection;
+using System.ServiceModel;
+using System.ServiceModel.Security;
+using Xunit;
+
+public static class SecurityAlgorithmSuiteTest
+{
+    [WcfTheory]
+    [InlineData(nameof(System.ServiceModel.Security.TripleDesSecurityAlgorithmSuite.DefaultCanonicalizationAlgorithm), "http://www.w3.org/2001/10/xml-exc-c14n#")]
+    [InlineData(nameof(System.ServiceModel.Security.TripleDesSecurityAlgorithmSuite.DefaultDigestAlgorithm), "http://www.w3.org/2000/09/xmldsig#sha1")]
```
mconnew: Can you simplify the naming to remove the S.SM.Security namespace as you are using that namespace already?

src/System.ServiceModel.Security/tests/ServiceModel/SecurityAlgorithmSuiteTest.cs
```diff
@@ -0,0 +1,52 @@
...
+    [InlineData(nameof(System.ServiceModel.Security.TripleDesSecurityAlgorithmSuite.IsAsymmetricKeyLengthSupported), true, 1024)]
+    [InlineData(nameof(System.ServiceModel.Security.TripleDesSecurityAlgorithmSuite.IsAsymmetricKeyLengthSupported), false, 4097)]
+    public static void Method_SecurityAlgorithmSuite_TripleDes(string propertyName, bool expectedValue, int argVal)
+    {
+        // Use reflection to call the method with the specified arguments based on the inline data.
+        // This test verifies the called methods return true or false based on the values passed in.
+        Type objType = typeof(TripleDesSecurityAlgorithmSuite);
+        Type tripleDesSecurityAlgorithmSuite = Type.GetType(objType.AssemblyQualifiedName);
+        ConstructorInfo constructor = tripleDesSecurityAlgorithmSuite.GetConstructor(Type.EmptyTypes);
+        object desClassObject = constructor.Invoke(new object[] { });
```
mconnew: Why are you constructing an instance using reflection? Developers won't be able to do that, they will only be able to use the type returned from SecurityAlgorithmSuite.TripleDes so you just get that instance and test it.

src/System.ServiceModel.Security/tests/ServiceModel/SecutityBindingElementTest.cs
```diff
@@ -27,4 +28,11 @@ public static void Method_CreateUserNameOverTransportBindingElement()
         Assert.True(element != null, "SecurityBindingElement added to binding elements should not be null.");
         Assert.Equal(element, securityBindingElement);
     }
+
+    [WcfFact]
+    public static void Property_DefaultAlgorithmSuite()
+    {
+        TransportSecurityBindingElement securityBindingElement = new TransportSecurityBindingElement();
+        Assert.Equal(securityBindingElement.DefaultAlgorithmSuite, SecurityAlgorithmSuite.Basic256);
+    }
```
mconnew: Did we add Basic256 to the contract?

StephenBonikowsky: No, but we did just add SecurityBindingElement.DefaultAlgorithmSuite with #3909 so it was convenient to add a quick of the default value which is Basic256.

src/System.ServiceModel.Security/tests/ServiceModel/SecurityAlgorithmSuiteTest.cs
StephenBonikowsky: Fixed.

src/System.ServiceModel.Security/tests/ServiceModel/SecurityAlgorithmSuiteTest.cs
StephenBonikowsky: Done

Issue #3860, opened 2019-08-30
[P2] Support System.ServiceModel.Channels.SecurityBindingElement


StephenBonikowsky:
| APIs | Port/Implement | Public Contract | Testing |
| :------------ | :-------------: | :------------: | :------------: |
`property SecurityAlgorithmSuite DefaultAlgorithmSuite` | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Added</li></ol> | <ol><li>[x] Unit <li>[x] Scenario</li></li></ol> |
`property SecurityKeyEntropyMode KeyEntropyMode` | <ul><li>[x] Merged</li></ul> | <ul><li>[x] Added</li></ul> | <ul><li>[x] Unit <li>[x] Scenario</li></li></ul> |
***
**API Notes**
***
**Implementation Notes**
Already ported.
***
**Public Contract Notes**
- Public Contract: System.ServiceModel.Security
***
**Testing Notes**
- Unit tests are required : PR #3915, #3916.
- Scenario tests needed: ???

***
**PR Links**
DefaultAlgorithmSuite was added to PC via: #3875
KeyEntropyMode was added to PC via: #3900
Scenario tests tracked by: #4022

krispenner:
Hi @StephenBonikowsky,

Will this also add support for the derived class [AsymmetricSecurityBindingElement](https://docs.microsoft.com/en-us/dotnet/api/system.servicemodel.channels.asymmetricsecuritybindingelement?view=netframework-4.8)?

For example, below is effectively the one and only code block that is preventing us from targeting .NET Standard and running on .NET Core. I would be forever grateful if this would be supported with your upcoming release in some way.

```
var messageSecurity = new AsymmetricSecurityBindingElement()
{
    MessageSecurityVersion = MessageSecurityVersion.WSSecurity11WSTrustFebruary2005WSSecureConversationFebruary2005WSSecurityPolicy11,
    InitiatorTokenParameters = new X509SecurityTokenParameters { InclusionMode = SecurityTokenInclusionMode.Never },
    RecipientTokenParameters = new X509SecurityTokenParameters { InclusionMode = SecurityTokenInclusionMode.Never },
    MessageProtectionOrder = MessageProtectionOrder.SignBeforeEncrypt,
    EnableUnsecuredResponse = true,
    IncludeTimestamp = false,
    DefaultAlgorithmSuite = SecurityAlgorithmSuite.Basic128Rsa15,
};
[...]

PR #3916, opened 2019-09-18
Unit test for SecurityBindingElement.KeyEntropyMode.

imcarolwang:
For #3860. 

@StephenBonikowsky , @mconnew , @HongGit


StephenBonikowsky:
Looks good to me, the default value is the same as on full framework.

imcarolwang:
@StephenBonikowsky  this was not merged and I happen to found I am able to do the merge operation, so I merged it ...

StephenBonikowsky:
In this case that's fine, but for future PRs I want to get Matt's signoff as well. So if you have a PR that both he and I have approved and the checks are green then go ahead and merge.

You also need to do a couple other things once you merge which it looks like you did in this case...
   - move the related work item to the 'Scenario Tests' pipeline.
   - edit the description and check off the unit test field.
   - un-assign the work item

Thanks!

imcarolwang:
Ok :)

Issue #3903, opened 2019-09-11
[P1] Support System.IdentityModel.Tokens.SecurityKeyType


StephenBonikowsky:
| APIs | Port/Implement | Public Contract | Testing |
| :------------ | :-------------: | :------------: | :------------: |
enum S.IM.Tokens.SecurityKeyType | <ol><li>[x] Merged</li></ol> | <ol><li>[x] Added</li></ol> | <ol><li>[x] Unit <li>[x] Scenario</li></li></ol> |
***
**API Notes**
***
**Implementation Notes**
Already implemented.
***
**Public Contract Notes**
- Public Contract: System.ServiceModel.Primitives

***
**Testing Notes**
- Unit tests are required.
- Scenario tests needed: ???

***
**PR Links**
Added to the PC with: #3907
Unit test added via: #3917
Scenario tests tracked by: #4022

PR #3917, opened 2019-09-18
Enum test

v-fangfc:
Add Unit Test for issue #3903

### DIFF
--- a/src/System.ServiceModel.Security/tests/ServiceModel/SecurityKeyTypeTest.cs
+++ b/src/System.ServiceModel.Security/tests/ServiceModel/SecurityKeyTypeTest.cs
@@ -12,33 +12,22 @@ public static class SecurityKeyTypeTest
     [WcfFact]
     public static void Get_EnumMembers_Test()
     {
-        SecurityKeyType skt0 = SecurityKeyType.SymmetricKey;
-        Assert.Equal(SecurityKeyType.SymmetricKey, skt0);
+        SecurityKeyType sk = SecurityKeyType.SymmetricKey;
+        Assert.Equal(SecurityKeyType.SymmetricKey, sk);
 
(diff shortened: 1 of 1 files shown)


src/System.ServiceModel.Security/tests/ServiceModel/SecurityKeyTypeTest.cs
```diff
@@ -0,0 +1,33 @@
...
+        Assert.Equal(SecurityKeyType.SymmetricKey, sk);
+
+        SecurityKeyType ak = SecurityKeyType.AsymmetricKey;
+        Assert.Equal(SecurityKeyType.AsymmetricKey, ak);
+
+        SecurityKeyType bk = SecurityKeyType.BearerKey;
+        Assert.Equal(SecurityKeyType.BearerKey, bk);
+    }
+
+    [Theory]
```
StephenBonikowsky: This needs to be [WcfTheory]

StephenBonikowsky:
@mconnew Do we really need a unit test for an enum?

src/System.ServiceModel.Security/tests/ServiceModel/SecurityKeyTypeTest.cs
```diff
@@ -0,0 +1,33 @@
...
+
+
+using System.IdentityModel.Tokens;
+using Infrastructure.Common;
+using Xunit;
+
+public static class SecurityKeyTypeTest
+{
+    [WcfFact]
+    public static void Get_EnumMembers_Test()
```
StephenBonikowsky: Our convention is to not use the word "test" as part of the name of our test methods. The reason for this is that in reporting on test passes the namespace is included as part of the full test name, so this test would read as:  `SecurityKeyTypeTest.Get_EnumMembers_Test`
Instead we want it to be `SecurityKeyTypeTest.Get_EnumMembers`

PR #3924, opened 2019-09-20
Add Unit Test For MaxReadPoolSize and MaxWritePoolSize

v-fangfc:
Add Unit Test For #3859 

### DIFF
--- a/src/System.ServiceModel.Primitives/tests/Channels/BinaryMessageEncodingBindingElementTest.cs
+++ b/src/System.ServiceModel.Primitives/tests/Channels/BinaryMessageEncodingBindingElementTest.cs
@@ -84,10 +84,9 @@ public static class BinaryMessageEncodingBindingElementTest
 
     [WcfTheory]
     [InlineData(64)]
-    public static void MaxReadPoolSize_DefaultValue(int value)
+    public static void MaxReadPoolSize_DefaultValueIsSameAsNetFrameWork(int value)
     {
         BinaryMessageEncodingBindingElement bindingElement = new BinaryMessageEncodingBindingElement();
-        bindingElement.MaxReadPoolSize = value;
         Assert.Equal<int>(value, bindingElement.MaxReadPoolSize);
     }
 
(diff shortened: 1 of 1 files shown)


src/System.ServiceModel.Primitives/tests/Channels/BinaryMessageEncodingBindingElementTest.cs
```diff
@@ -81,4 +81,60 @@ public static void MessageVersion_Property_Set_Null_Value_Throws()
         BinaryMessageEncodingBindingElement bindingElement = new BinaryMessageEncodingBindingElement();
         Assert.Throws<ArgumentNullException>(() => bindingElement.MessageVersion = null);
     }
+
+    [WcfTheory]
+    [InlineData(64)]
+    public static void MaxReadPoolSize_DefaultValue(int value)
+    {
+        BinaryMessageEncodingBindingElement bindingElement = new BinaryMessageEncodingBindingElement();
+        bindingElement.MaxReadPoolSize = value;
```
StephenBonikowsky: You should not be setting MaxReadPoolSize, the default constructor already sets it to a default value, which in this case is 64, so you should remove this line entirely.
The Assert will then verify that the default value assigned to the property by the constructor is or is not the inlinedata value you provided.
You should always check the default value assigned on the full framework first, since that is what we expect the default value to be in Core as well.

src/System.ServiceModel.Primitives/tests/Channels/BinaryMessageEncodingBindingElementTest.cs
```diff
@@ -81,4 +81,60 @@ public static void MessageVersion_Property_Set_Null_Value_Throws()
...
+        BinaryMessageEncodingBindingElement bindingElement = new BinaryMessageEncodingBindingElement();
+        Assert.Throws<ArgumentOutOfRangeException>(() => bindingElement.MaxReadPoolSize = value);
+    }
+
+    [WcfTheory]
+    [InlineData(16)]
+    public static void MaxWritePoolSize_DefaultValue(int value)
+    {
+        BinaryMessageEncodingBindingElement bindingElement = new BinaryMessageEncodingBindingElement();
+        bindingElement.MaxWritePoolSize = value;
```
StephenBonikowsky: Same as comment for MaxReadPoolSize

PR #3927, opened 2019-09-24
Unit test for SecurityToken class.

imcarolwang:
For issue #3857, regarding unit tests for APIs listed in table:

1. I didn't find ctor internal SecurityToken() but find ctor internal SecurityKey() in the public contract, and I didn't find implementation for ctor SecurityKey() in "System.Private.ServiceModel" project. I have no idea how to add unit test for this.

2. The properties validation was covered in test added for BinarySecretSecurityToken which derived from SecurityToken. No extra test added for this.

3. The only test not covered is for "virtual method T CreateKeyIdentifierClause<T>() where T : SecurityKeyIdentifierClause" , and it's not added as public contract. I updated that and added unit test for it via mocking the abstract SecurityToken class. 

@StephenBonikowsky , @mconnew , @HongGit 
I am not sure if I missed something above, please help confirm. Thanks!

### DIFF
--- a/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
+++ b/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
@@ -199,7 +199,7 @@ namespace System.IdentityModel.Tokens
         public abstract DateTime ValidFrom { get; }
         public abstract DateTime ValidTo { get; }
         //public virtual bool CanCreateKeyIdentifierClause<T>() where T : System.IdentityModel.Tokens.SecurityKeyIdentifierClause { return default(bool); }
-        //public virtual T CreateKeyIdentifierClause<T>() where T : System.IdentityModel.Tokens.SecurityKeyIdentifierClause { return default(T); }
+        public virtual T CreateKeyIdentifierClause<T>() where T : System.IdentityModel.Tokens.SecurityKeyIdentifierClause { return default(T); }
         //public virtual bool MatchesKeyIdentifierClause(System.IdentityModel.Tokens.SecurityKeyIdentifierClause keyIdentifierClause) { return default(bool); }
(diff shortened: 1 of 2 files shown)


StephenBonikowsky:
@imcarolwang in C# if no constructor is declared a public default constructor is implied. In the case of the SecurityKey class by adding it to the public contract in order to make certain members in it publicly available we don't want to allow the class itself to be derivable, so we explicitly add the default constructor to the contract but make it internal. So there is nothing for you to test.

src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
```diff
@@ -199,7 +199,7 @@ public abstract partial class SecurityToken
         public abstract DateTime ValidFrom { get; }
         public abstract DateTime ValidTo { get; }
         //public virtual bool CanCreateKeyIdentifierClause<T>() where T : System.IdentityModel.Tokens.SecurityKeyIdentifierClause { return default(bool); }
-        //public virtual T CreateKeyIdentifierClause<T>() where T : System.IdentityModel.Tokens.SecurityKeyIdentifierClause { return default(T); }
```
StephenBonikowsky: @mconnew I either just missed adding this method or there was some reason for not adding it, but I didn't see any notes and I can't specifically remember if there was something discussed. Any reason you can think of for not having added this?

StephenBonikowsky:
@imcarolwang I just spoke with @mconnew and this method was not added to the public contract by design, I had just not updated the tracking issue with the information. I did now update #3857 with the reason.

So my apologies but that means this PR is not needed. Having a mock implementation of the abstract class was the correct way to go about testing it though.

imcarolwang:
@StephenBonikowsky  thank you for the confirm! Close this one since it's no longer needed.

Issue #3967, opened 2019-10-14
The DataContractSerializerOperationBehavior.SerializationSurrogateProvider property is not available in a library netstandard 2.0


FrancescoCrimi:
The "DataContractSerializerOperationBehavior.SerializationSurrogateProvider" property is available in library netcore but not available in library netstandard.

This https://github.com/FrancescoCrimi/DCSOB-Error is a simple solution with two same simple project one library netcore and one library netstandard, just clone with VisualStudio and open the file ServiceClient in both project to see the error in the library netstandard and not in library netcore.

I expected  to use DataContractSerializerOperationBehavior.SerializationSurrogateProvider method in the netstandard library same as netcore library. Moreover, it's no possible referece a netcore library in an UWP project but it's possible for netstandard library.

This https://github.com/FrancescoCrimi/wcf is my wcf fork with a little modify at the file System.ServiceModel.Primitives.Ref.csproj to resolve a problem, see #3968 pull request

Thanks you

StephenBonikowsky:
@FrancescoCrimi This is expected behavior. Our latest WCF Core releases continue to support .NET Standard2.0 which is also supported by .NET Framework 4.8. This type does not exist in net4.8 so we couldn't add it to our public contract targeting .NET Standard2.0 so instead it targets netcoreapp2.1.

If I understand you correctly you want to be able to use that API from a UWP project, we could look into adding this API to our UWP public contract in an up-coming release.

Could you give us any more details about your scenario, are you migrating an application to .NET Core or writing a new application?

FrancescoCrimi:
@StephenBonikowsky  Thanks for the reply.

I wanted to be able to use this API in a "UWP" project, in an attempt to migrate an app from "Net Framework" to "Net Core" and at the same time add a ui "UWP".

In "Net Framework" I used "DataContractSerializerOperationBehavior.DataContractSurrogate" to be able to use a custom implementation of "IList <T>" in my "Domain Models" but this is not available in "netstandard" and "netcoreapp".

StephenBonikowsky:
@mconnew your thoughts?

FrancescoCrimi:
Hello, i close this issue because  it is expected behavior @StephenBonikowsky and I open another one where I ask why this api does not exist in the UWP platform, see #4020

Thanks you

mconnew:
[...]

PR #3968, opened 2019-10-14
DataContractSerializerOperationBehavior.SerializationSurrogateProvider fix for netstandard library

FrancescoCrimi:
Simple fix to have DataContractSerializerOperationBehavior.SerializationSurrogateProvider in netstandard library, see the Issue #3967 .
See https://github.com/FrancescoCrimi/DCSOB-Error for the error

Thanks you


StephenBonikowsky:
@FrancescoCrimi I believe we can help you get this API but we can't just add it to the NET Standard 2.0 contract as that would break compatibility with .NET Framework 4.8. Please see my comments on #3967 and we can work on finding a solution for you but this PR won't work so am going to close.

PR #3974, opened 2019-10-17
Stebon/3.1.0/infraupdates

StephenBonikowsky:


StephenBonikowsky:
Created #3975 so that it is easy to produce packages for dotnet-svcutil.xmlserializer when needed.

PR #3989, opened 2019-10-23
Fixing branch name logic.

StephenBonikowsky:
* The existing logic didn't account for a branch name being "release/3.1.0"
* We leveraged the azure devops variable 'Build.SourceBranchName' which returned the "last path segment in the ref string".


StephenBonikowsky:
Need to merge in order to trigger the CI build to validate these changes.

PR #3994, opened 2019-10-24
Port FullFx fix for SocketConnection sync/async timeout ignore problem.

StephenBonikowsky:
* [Changeset 1615113](http://vstfdevdiv.redmond.corp.microsoft.com:8080/DevDiv2/DevDiv/_versionControl/changeset/1615113?_a=compare&path=%24%2FDevDiv%2FFX%2FFeature%2FNetFXDev1%2FNDP%2Fcdf%2Fsrc%2FWCF%2FServiceModel%2FSystem%2FServiceModel%2FChannels%2FSocketConnection.cs) in VSTS.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
@@ -73,7 +73,7 @@ namespace System.ServiceModel.Channels
             _closeState = CloseState.Open;
             _readBuffer = connectionBufferPool.Take();
             _asyncReadBufferSize = _readBuffer.Length;
-            socket.SendBufferSize = socket.ReceiveBufferSize = _asyncReadBufferSize;
+            _socket.SendBufferSize = _socket.ReceiveBufferSize = _asyncReadBufferSize;
             _asyncSendTimeout = _asyncReceiveTimeout = TimeSpan.MaxValue;
             _socketSyncSendTimeout = _socketSyncReceiveTimeout = TimeSpan.MaxValue;
 
@@ -81,14 +81,14 @@ namespace System.ServiceModel.Channels
 
             if (autoBindToCompletionPort)
             {
-                socket.UseOnlyOverlappedIO = false;
+                _socket.UseOnlyOverlappedIO = false;
             }
 
             // In SMSvcHost, sockets must be duplicated to the target process. Binding a handle to a completion port
             // prevents any duplicated handle from ever binding to a completion port. The target process is where we
             // want to use completion ports for performance. This means that in SMSvcHost, socket.UseOnlyOverlappedIO
             // must be set to true to prevent completion port use.
-            if (socket.UseOnlyOverlappedIO)
+            if (_socket.UseOnlyOverlappedIO)
             {
                 // Init BeginRead state
                 if (s_onReceiveCompleted == null)
@@ -135,7 +135,7 @@ namespace System.ServiceModel.Channels
                     catch (ObjectDisposedException objectDisposedException)
                     {
                         Exception exceptionToThrow = ConvertObjectDisposedException(objectDisposedException, TransferOperation.Undefined);
-                        if (object.ReferenceEquals(exceptionToThrow, objectDisposedException))
+                        if (ReferenceEquals(exceptionToThrow, objectDisposedException))
                         {
                             throw;
                         }
@@ -340,6 +340,9 @@ namespace System.ServiceModel.Channels
 
             try
             {
+                // A FIN (shutdown) packet has already been sent to the remote host and we're waiting for the remote
+                // host to send a FIN back. A pending read on a socket will complete returning zero bytes when a FIN
+                // packet is received.
                 if (BeginReadCore(0, 1, _readFinTimeout, s_onWaitForFinComplete, this) == AsyncCompletionResult.Queued)
                 {
                     return;
@@ -505,7 +508,7 @@ namespace System.ServiceModel.Channels
             catch (ObjectDisposedException objectDisposedException)
             {
                 Exception exceptionToThrow = ConvertObjectDisposedException(objectDisposedException, TransferOperation.Undefined);
-                if (object.ReferenceEquals(exceptionToThrow, objectDisposedException))
+                if (ReferenceEquals(exceptionToThrow, objectDisposedException))
                 {
                     throw;
                 }
@@ -522,7 +525,7 @@ namespace System.ServiceModel.Channels
             {
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(
                     ConvertObjectDisposedException(new ObjectDisposedException(
-                    GetType().ToString(), SR.Format(SR.SocketConnectionDisposed)), TransferOperation.Undefined));
+                    GetType().ToString(), SR.SocketConnectionDisposed), TransferOperation.Undefined));
             }
         }
 
@@ -532,7 +535,7 @@ namespace System.ServiceModel.Channels
             {
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(
                     ConvertObjectDisposedException(new ObjectDisposedException(
-                    GetType().ToString(), SR.Format(SR.SocketConnectionDisposed)), TransferOperation.Undefined));
+                    GetType().ToString(), SR.SocketConnectionDisposed), TransferOperation.Undefined));
             }
         }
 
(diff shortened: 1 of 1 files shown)


StephenBonikowsky:
Going into master first then will cherry-pick to release/3.1.0.

@dasetser There is a fair amount of difference between the desktop and core versions of SocketConnection.cs so I had to port the changes manually and not all code paths were present.
Please review.

mconnew:
@StephenBonikowsky, the reason they are different is because I had to modify the implementation because of two reasons.  
1. Socket.Close(int timeout) wasn't originally available in .NET Core so I had to rewrite the close code to use socket lingering capabilities.
2. Neither System.Net.Sockets or SslStream wasn't available for UWP so I had to rewrite everything on top of the UWP native classes for sockets which didn't have a 1:1 mapping of functionality so I refactored everything to a base class and the concrete implementation.

Neither of this problems exist today so my preference would be to replace the entire SocketConnection class with the implementation from .NET Framework as there are some compromises with the current implementation, e.g. close timeout only works with a granularity of 1 second.

dasetser:
@StephenBonikowsky It looks like you ported the fix correctly to the current implementation. However, I agree with Matt that it would be better if we could port the entire file over again if we can get rid of the compromises in the current implementation.

StephenBonikowsky:
Sounds good, I'll find Matt's commits and see if reverting them makes sense and then porting the SocketConnection class as it is in the full framework.

StephenBonikowsky:
This was not a straight forward port, no copy and paste and just fix formatting issues. There are quite a few things not present in Core that I had to account for. Particularly related to tracing.

Please review this carefully.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
```diff
@@ -316,33 +340,68 @@ private void CloseAsyncAndLinger()
 
             try
             {
-                // A FIN (shutdown) packet has already been sent to the remote host and we're waiting for the remote
-                // host to send a FIN back. A pending read on a socket will complete returning zero bytes when a FIN
```
mconnew: This comment is worth keeping as it describes what the code is doing here.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
```diff
@@ -486,7 +522,7 @@ private void ThrowIfNotOpen()
             {
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(
                     ConvertObjectDisposedException(new ObjectDisposedException(
-                    GetType().ToString(), SR.SocketConnectionDisposed), TransferOperation.Undefined));
+                    GetType().ToString(), SR.Format(SR.SocketConnectionDisposed)), TransferOperation.Undefined));
             }
```
mconnew: SR.Format isn't needed when there are no parameters for the string resource

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
```diff
@@ -496,37 +532,77 @@ private void ThrowIfClosed()
...
             if (_timeoutErrorString != null)
             {
-                return ConvertTimeoutErrorException(originalException, _timeoutErrorString, _timeoutErrorTransferOperation);
+                return ConvertTimeoutErrorException(originalException, transferOperation, _timeoutErrorString, _timeoutErrorTransferOperation);
             }
             else if (_aborted)
             {
-                return new CommunicationObjectAbortedException(SR.SocketConnectionDisposed, originalException);
+                return new CommunicationObjectAbortedException(SR.Format(SR.SocketConnectionDisposed), originalException);
             }
```
mconnew: Remove SR.Format as unnecessary, just use the resource string directly

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
```diff
@@ -535,42 +611,42 @@ private Exception ConvertObjectDisposedException(ObjectDisposedException origina
...
-                (int)socketException.SocketErrorCode == UnsafeNativeMethods.WSAECONNABORTED ||
-                (int)socketException.SocketErrorCode == UnsafeNativeMethods.WSAECONNRESET)
+            if (socketException.ErrorCode == UnsafeNativeMethods.WSAENETRESET ||
+                socketException.ErrorCode == UnsafeNativeMethods.WSAECONNABORTED ||
+                socketException.ErrorCode == UnsafeNativeMethods.WSAECONNRESET)
             {
                 if (aborted)
                 {
-                    return new CommunicationObjectAbortedException(SR.TcpLocalConnectionAborted, originalException);
+                    return new CommunicationObjectAbortedException(SR.Format(SR.TcpLocalConnectionAborted), originalException);
```
mconnew: Remove SR.Format

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
```diff
@@ -662,12 +724,12 @@ private void TraceWriteStart(int size, bool async)
...
                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(
-                    ConvertSendException(socketException, TimeSpan.MaxValue));
+                    ConvertSendException(socketException, TimeSpan.MaxValue, _asyncSendTimeout));
             }
             catch (ObjectDisposedException objectDisposedException)
             {
                 Exception exceptionToThrow = ConvertObjectDisposedException(objectDisposedException, TransferOperation.Write);
-                if (ReferenceEquals(exceptionToThrow, objectDisposedException))
+                if (object.ReferenceEquals(exceptionToThrow, objectDisposedException))
                 {
```
mconnew: I think there should be a VS Roslyn suggested fix here to simplify the code to remove object. as it's unnecessary. If there is, choose to apply the fix to the entire document.

mconnew:
Just a couple of small issues around string resources, otherwise LGTM. Fix those SR.Format uses and we can call it good!

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
StephenBonikowsky: 👍

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
StephenBonikowsky: 👍

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
StephenBonikowsky: 👍

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
StephenBonikowsky: 👍

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
StephenBonikowsky: 👍

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SocketConnection.cs
```diff
@@ -46,37 +56,100 @@ internal class SocketConnection : IConnection
...
-            AsyncReadBuffer = _connectionBufferPool.Take();
-            AsyncReadBufferSize = AsyncReadBuffer.Length;
-            _sendTimeout = _receiveTimeout = TimeSpan.MaxValue;
-            _closeState = CloseState.Open;
-            _socket.SendBufferSize = _socket.ReceiveBufferSize = AsyncReadBufferSize;
-            _sendTimeout = _receiveTimeout = TimeSpan.MaxValue;
-        }
+            _readBuffer = connectionBufferPool.Take();
+            _asyncReadBufferSize = _readBuffer.Length;
+            socket.SendBufferSize = socket.ReceiveBufferSize = _asyncReadBufferSize;
```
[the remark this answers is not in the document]
StephenBonikowsky: 👍

StephenBonikowsky:
The new tests added with #4000 are failing because the test server hasn't been updated. Will update it manually and then re-run the checks.

mconnew:
:shipit:

Issue #3940, opened 2019-09-27
Out params not being stripped out of array size logic.


StephenBonikowsky:
PR merged to...
- [x] master (pr #4000)
- [x] release/3.1.0 (pr #4021)

Error Message:
`Message=There was a mismatch between the number of supplied arguments and the number of expected arguments.  Specifically, the argument 'parts' has '6' elements while the argument 'parameters' has '9' elements. (Parameter 'parameters')`

Call Stack:
```
System.ArgumentException
  HResult=0x80070057
  Message=There was a mismatch between the number of supplied arguments and the number of expected arguments.  Specifically, the argument 'parts' has '6' elements while the argument 'parameters' has '9' elements. (Parameter 'parameters')
  Source=System.Private.ServiceModel
  StackTrace:
   at System.ServiceModel.Dispatcher.PrimitiveOperationFormatter.SerializeParameters(XmlDictionaryWriter writer, PartInfo[] parts, Object[] parameters)
   at System.ServiceModel.Dispatcher.PrimitiveOperationFormatter.SerializeRequest(XmlDictionaryWriter writer, Object[] parameters)
   at System.ServiceModel.Dispatcher.PrimitiveOperationFormatter.PrimitiveRequestBodyWriter.OnWriteBodyContents(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.BodyWriter.WriteBodyContents(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.BodyWriterMessage.OnWriteBodyContents(XmlDictionaryWriter writer)
   at System.ServiceModel.Security.SecurityAppliedMessage.OnWriteMessage(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.Message.WriteMessage(XmlDictionaryWriter writer)
   at System.ServiceModel.Channels.BufferedMessageWriter.WriteMessage(Message message, BufferManager bufferManager, Int32 initialOffset, Int32 maxSizeQuota)
   at System.ServiceModel.Channels.BinaryMessageEncoderFactory.BinaryMessageEncoder.WriteMessage(Message message, Int32 maxMessageSize, BufferManager bufferManager, Int32 messageOffset)
   at System.ServiceModel.Channels.MessageEncoder.WriteMessage(Message message, Int32 maxMessageSize, BufferManager bufferManager)
   at System.ServiceModel.Channels.BufferedMessageContent.EnsureMessageEncoded()
[...]

PR #4000, opened 2019-10-28
Fix contract methods which use out parameters

mconnew:
Addresses issue #3940


StephenBonikowsky:
plus cleanup and tests!
LGTM

PR #3891, opened 2019-09-09
Fix xml serializer encoding

maheshnlrb:
For #3870

PR #4001, opened 2019-10-29
Add "encodingStyle" attribute to generated SOAP message.

StephenBonikowsky:
* This fix was initially produced by @maheshnlrb with PR #3891 but the commit history got sufficiently scrambled that is was easier to reproduce his changes in a clean up-to-date branch.


mconnew:
LGTM. As mentioned F2F, we need a scenario test for this.

Issue #4006, opened 2019-10-30
Fix xUnit warnings


StephenBonikowsky:
- [x] master (no need to go into release/3.1.0)

Fix all of the xUnit warnings we are currently suppressing.
Add to this list below if I missed any.

- [x] xUnit1003 (master:#4011,#4042)
- [x] xUnit1006 (master:#4011)
- [x] xUnit1013 (master:#4011, #4043)
- [x] xUnit1014 (master:#4011, #4046)
- [X] xUnit1019 (master:#4049)
- [X] xUnit1024 (master:#4032)
- [x] xUnit2000 (insert PR #4019, #4053)
- [x] xUnit2006 (master:#4011,#4055)
- [x] xUnit2009 (insert PR #4024,#4060)
- [x] xUnit2010 (master:#4011,#4065)
- [X] xUnit2013 (master:#4031)

StephenBonikowsky:
Added master and release/3.1 check boxes.
After it has been merged into master, move this issue into the "Cherry-pick to release/3.1.0" pipeline of the 'WCF Core 3.1' project.

StephenBonikowsky:
Removed from the 3.1 project, no need to cherry-pick to 3.1 branch.

imcarolwang:
Hi @StephenBonikowsky, I just found the PRs I submitted earlier only handled xUnit warnings in test projects under System.Private.ServiceModel, and there're more test folders/projects not checked.

I'm planning to create PRs for each xUnit warning fixing for the remaining folders. Please help review. Sorry for the extra effort/inconvenience brought.

StephenBonikowsky:
Ok, will do.

imcarolwang:
Close as all suppressed xUnit warnings have been removed from test projects.

StephenBonikowsky:
Terrific, thanks @imcarolwang !

Issue #1024, opened 2016-04-15
Remove or refactor workarounds used during infrastructure branch transition


roncain:
This issue is used as a checklist of temporary workarounds that were applied during the switch to the new test infrastructure.  Workarounds in the code reference this issue for easier tracking.

roncain:
Closing this issue now because the temporary workarounds will disappear when we resolve #1007 and replace the workaround code with actual detection logic.

PR #4032, opened 2019-11-07
Fix xUnit warning #1024 : Test method cannot have overloads.

imcarolwang:
For #4006 .

Fix notes: renamed the private/static overloads method name by adding name suffix "_TestImpl".


StephenBonikowsky:
Approved.
Sorry for the delay.

PR #4038, opened 2019-11-15
Update PreReleaseVersionLabel to RTM

wtgodbe:
Non-shipping packages should be labeled RTM for GA. We need to take this fix for 3.1 GA.

After merging this, I need to queue a manual build with `DotNetFinalVersionKind` = `release`

CC @StephenBonikowsky


StephenBonikowsky:
Thanks @wtgodbe

wtgodbe:
Here's the build: https://dnceng.visualstudio.com/internal/_build/results?buildId=428262

PR #4039, opened 2019-11-15
Change release lablel for non-shipping package to rtm.

StephenBonikowsky:
Changing the pre-release label in Versions.props only affects the WCF packages.
To affect the dotnet-svcutil.xmlserializer package it has to be done in the project file.
This package ships on an entirely different cadence which is why it is handled separately like this.


StephenBonikowsky:
@wtgodbe pushing this through and queuing another stable build.

PR #4048, opened 2019-11-21
Simplified test service startup pattern

mconnew:
* Created attribute to drive service creation so no longer needed to add to web.config and Program.cs
* Removed ServiceHostFactory instances
* Created ConfigurationBuilder implementation to populate serviceActivations in web.config based on attributed classes
* Created service starting helper for Program which starts service based on attributed classes
* Cleaned up unused service classes
* Cleaned up usings
* Moved SelfHostedService forward to 4.7.1 to get access to ConfigurationBuilder


StephenBonikowsky:
@imcarolwang Please note that this PR significantly changes how we add new endpoints for scenario tests.

PR #4053, opened 2019-11-25
Fix xUnitWarning #2000 in test projects: Constants and literals should…

imcarolwang:
… be the expected argument

For #4006 .


StephenBonikowsky:
:shipit:

Issue #2006, opened 2017-06-28
Clean up SUPPORTS_WINDOWSIDENTITY in S.P.SM.csproj


zhenlan:
We should remove section below in [S.P.SM.csproj](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/src/System.Private.ServiceModel.csproj)
```
  <PropertyGroup Condition="'$(TargetGroup)' != 'netcore50' AND '$(TargetsWindows)' == 'true'">
    <DefineConstants>$(DefineConstants);SUPPORTS_WINDOWSIDENTITY</DefineConstants>
  </PropertyGroup>
```

SUPPORTS_WINDOWSIDENTITY should be true for both Windows and Linux except UWP. We already have it defined correctly in another section in the same file, but above section is causing confusion.

cc: @mconnew

shmao:
This was fixed by https://github.com/dotnet/wcf/pull/2043.

PR #4055, opened 2019-11-26
Fix xUnit warning #2006 for test projects.

imcarolwang:
Warning detail:  the generic overloads of Assert.Equal or Assert.StrictEqual are used with string. There is an optimized overload of both Assert.Equal and Assert.StrictEqual for string arguments.

For #4006 .


StephenBonikowsky:
:shipit:

PR #2009, opened 2017-06-28
Fix property issue in netcigroovy

StephenBonikowsky:

PR #4060, opened 2019-11-27
Fix xUnit warning #2009 in test projects.

imcarolwang:
[Detail](http://xunit.net/xunit.analyzers/rules/xUnit2009.html): Check substrings with Assert string methods like string.Contains, string.StartsWith and string.EndsWith, instread of Assert.True or Assert.False.

For #4006 .


imcarolwang:
Pinging... @StephenBonikowsky

StephenBonikowsky:
:shipit:

PR #4065, opened 2019-12-03
Fix xUnit warning #2010 in test projects.

imcarolwang:
For #4006 .

**Warning detail:** Do not use boolean check to check for string equality. Assert.Equal should be used because it give more detailed information upon failure.


StephenBonikowsky:
:shipit:

PR #4077, opened 2019-12-10
Add projects that are solution dependencies; dotnet core 3.1.100

gojimmypi:


global.json
```diff
@@ -8,7 +8,7 @@
     }
   },
   "sdk": {
-    "version": "3.0.101"
+    "version": "3.1.100"
```
StephenBonikowsky: We typically update these versions through the auto update system we have as part of our dependency on the Arcade repo.
However, we should move the release/3.1 branch to using version "3.1.100" since it is released and that is the matching version of the SDK. It also matches what other .NET Core repos have been doing such as [coreclr](https://github.com/dotnet/coreclr/commit/5558d18aafc84ecac3630ca6a31f52699cde0b66
)
For our master branch which is intended for .NET 5.0 we can leave it as is for now.

StephenBonikowsky:
If you remove the update to the SDK version I'm fine with adding those projects to the root solution.
I'll create a PR for updating the SDK version in the 3.1 branch.

gojimmypi:
@StephenBonikowsky I've reverted the SDK to the prior value 3.0.101

StephenBonikowsky:
Checks look good.

PR #4151, opened 2020-02-05
Add build scripts copied from dotnet/runtime

mconnew:
This should make building a lot simpler


StephenBonikowsky:
That looks great, I'll start using it right away.
Thanks Matt.

Issue #3750, opened 2019-07-19
Old SOAP service (web reference) - throwing 'Message {nameOfFunction} must not have headers to be used in RPC encoded style.'


hurrk:
This might not be the right place for this - it's not necessarily an error in dotnet/wcf. But I'm hoping you guys can either answer or point me to the right place. 

I'm consuming a SOAP service that uses the old .NET Framework Web Services 2.0 (I believe, there is no .svc, but there is a WSDL).

I can consume this service in .NET Framework by going to 'Add service reference -> Advanced -> Add web reference". Works no problem.

In .NET Core, when I add it as a Connected Service, firstly the code is somewhat different. But I also get an error when I call any endpoint, that goes as follows:

 'Message {nameOfFunction} must not have headers to be used in RPC encoded style.'

And well, yes, it did generate the code so that the functions are marked with RPC and encoded. 

The error happens in System.ServiceModel.Description.XmlSerializerOperationBehavior.Reflector.OperationReflector.LoadHeadersMapping(MessageDescription message, String mappingKey)  when checking that MessageDescription message.Headers.Count > 0 (and IsEncoded is true).
Do you think it's worthwhile chasing this error down? Is it likely that this kind of API  is  just not supported in .NET Core?

hurrk:
P.s. I can publish the wsdl if anyone's interested.

StephenBonikowsky:
@dasetser Tooling related, could you provide @hurrk some guidance or more info about the error he got?

dasetser:
@hurrk Old services like this are something we make a best effort to support with the connected service, but we've seen people running into issues with them. We can try to help you track down what's causing it, but putting a fix into the tool for this would be lower priority. 

@mconnew Do you have any ideas why we would be getting this error? I know we have problems with arrays with XmlSerializer and Soap encoding, but this seems like an issue before they even get to that point. I'm not sure what we might be generating incorrectly.

mconnew:
@hurrk, on a .NET Framework application, can you try using "Add **Service Reference**" and see if you get the same error on .NET Framework? The problem is that your service contract has been generated with a `MessageHeaderAttribute` on one of the parameters and that's not supported when using RPC encoded style contracts.  
Can you provide your WSDL?

mottykohn:
I have exact problem with https://aodws.attendanceondemand.com:8192/cc1.aew/wsdl/IAeXMLBridge
Works with Web Reference and exact RPC message as above when I try in core

[...]

PR #4159, opened 2020-02-11
Allow message headers with Rpc document and Encoded use

mconnew:
Added a test for scenario.
This change is for #3750


FredRoeCassNet:
Did this issue ever get resolved?  I have the same issue using the same service, and I get the same errors as reported above when I add/user a service reference in .Net Core.

PR #4168, opened 2020-02-21
Remove UAP from test attribute skipping logic.

StephenBonikowsky:
* Uap was removed from TargetFrameworkMonikers in Arcade.
* Since we no longer run tests on this TFm we can just remove these conditional skips.


StephenBonikowsky:
This will resolve the failures seen in #4168

Issue #4110, opened 2020-01-04
Support WS2007FederationHttpBinding


CumpsD:
As per @mconnew's request in https://github.com/dotnet/wcf/issues/8#issuecomment-570717036 this issue is used to track `WS2007FederationHttpBinding` support.

> @CumpsD, we're in the process of adding support for WS-Federation. We're waiting on another team to provide an implementation of a WS-Trust client which can get the federation token to then communicate with the server. That's then used by WSHttpBinding to communicate with the server. 
>
> I took a look at WS2007FederationHttpBinding to see how it's different than WSFederationHttpBinding and the difference is that it uses WS2007HttpBinding which we don't have as a specific binding. I checked the protocols used by WS2007HttpBinding and it looks like we probably have most if not all the implementation required to implement it.
>
> Can you open an issue specifically for WS2007FederationHttpBinding?

StephenBonikowsky:
Update - Work is progressing with the WSFederationHttpBinding, can't commit to a date but it's not too far away and then we can look into what is needed for the WS2007 version of it.

mconnew:
So a little bit of an update. It turns out that we need to implement WS2007HttpBinding for the goal of targeting support for ADFS (it requires BearerKey token type which WSFederationHttpBinding doesn't support, you need WS2007FederationHttpBinding). So support for the functionality of this binding will be available at the same time as WSFederationHttpBinding. The actual binding type itself will be different, but it should support the same functionality as the NetFx one. We aren't aiming to support every protocol variant that can be used, we're targeting the main common scenarios which should work for 99% of people. We are going to produce a single new binding which supports the features of WS2007FederationHttpBinding as well as WSFederationHttpBinding.

crodeheaver:
I'd be happy to help test out the new binding once it becomes available!

coder-5943123:
Any progress resolving this issue?

StephenBonikowsky:
There was some unexpected work needed in underlying dependencies, specifically IdentityModel APIs. We have been working with that team to get the issue resolved, they are actively working on it. This is a priority for us to get done and it is progressing.

m-straub:
Is there any progress or workaround? This is the main issue holding my team back from fully switching to dotnet core because we are dependent on a third party wcf interface.

mconnew:
We are working on some build issues, but once they are resolved we should be releasing a preview package with this functionality very soon.

mconnew:
[...]

PR #4169, opened 2020-02-21
Add support for WS2007HttpBinding plus some cleanup

mconnew:
This is the WS2007HttpBinding implementation which is a pre-requisite for issue #4110


src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs
```diff
@@ -22,6 +22,7 @@ public abstract partial class MessageSecurityVersion
         public static System.ServiceModel.MessageSecurityVersion WSSecurity10WSTrustFebruary2005WSSecureConversationFebruary2005WSSecurityPolicy11BasicSecurityProfile10 { get { return default; } }
         public static System.ServiceModel.MessageSecurityVersion WSSecurity11WSTrustFebruary2005WSSecureConversationFebruary2005WSSecurityPolicy11 { get { return default; } }
         public static System.ServiceModel.MessageSecurityVersion WSSecurity11WSTrustFebruary2005WSSecureConversationFebruary2005WSSecurityPolicy11BasicSecurityProfile10 { get { return default; } }
+        public static System.ServiceModel.MessageSecurityVersion WSSecurity11WSTrust13WSSecureConversation13WSSecurityPolicy12BasicSecurityProfile10 { get { return default; } }
```
brentschmaltz: what is the 'default' here?

mconnew: This is just the reference assembly. Using `default` produces the smallest IL for the reference assembly. The implementation was already there in S.Private.SM

src/System.Private.ServiceModel/src/System/ServiceModel/Security/TrustDriver.cs
```diff
@@ -140,8 +140,5 @@ public virtual string RequestTypeRenew
 
         // RSTR Collection method
         public abstract void WriteRequestSecurityTokenResponseCollection(RequestSecurityTokenResponseCollection rstrCollection, XmlWriter writer);
-
-        // Federation proxy creation
-        public abstract IChannelFactory<IRequestChannel> CreateFederationProxy(EndpointAddress address, Binding binding, KeyedByTypeCollection<IEndpointBehavior> channelBehaviors);
```
HongGit: >public abstract IChannelFactory [](start = 8, length = 31)

Why is this interface removed?  Is there any risk associated with this change?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/Tokens/ISecurityContextSecurityTokenCache.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace System.ServiceModel.Security.Tokens
 {
-    public interface ISecurityContextSecurityTokenCache
+    internal interface ISecurityContextSecurityTokenCache
```
HongGit: >internal [](start = 4, length = 8)

is there any risk with this change?  Is this a breaking change for existing customers?

src/System.Private.ServiceModel/src/System/ServiceModel/Security/Tokens/SecurityContextSecurityToken.cs
```diff
@@ -12,40 +11,17 @@
 
 namespace System.ServiceModel.Security.Tokens
 {
-    public class SecurityContextSecurityToken : SecurityToken, TimeBoundedCache.IExpirableItem, IDisposable
+    internal class SecurityContextSecurityToken : SecurityToken, TimeBoundedCache.IExpirableItem, IDisposable
```
HongGit: similar question here

src/System.Private.ServiceModel/src/System/ServiceModel/Security/TrustDriver.cs
mconnew: This was the stub for the previous implementation of the WSTrust piece. This was the method which returned a client which could be used to talk to the STS (e.g. an ADFS server). This is now going to be done by the new package that's coming which uses the work that Brent is doing. It was incorrectly pulled over previously and as I was adding another implementation of TrustDriver, I did some cleanup to avoid bringing over another non-used stub.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/Tokens/SecurityContextSecurityToken.cs
mconnew: This type isn't in our contract. We have no internal usage of the public constructors so I removed them. As I was able to also remove most of the unused code in the class, it became clear this class won't be needed to be exposed in our contracts so I moved the class to internal. This is what prompted changing the interface to internal too.

src/System.Private.ServiceModel/src/System/ServiceModel/Security/Tokens/ISecurityContextSecurityTokenCache.cs
mconnew: This is not in our contract. This was changed from public to internal as one of the types in one of the defined methods was moved to internal which a public interface isn't allowed to reference.

brentschmaltz:
@mconnew @HongGit Thanks for pushing this forward.

PR #4177, opened 2020-02-29
Add version 1 porting playbook.

StephenBonikowsky:


HongGit:
@StephenBonikowsky  can we add the POP3 case study and lessons learned?  We could merge after that.  Then continue to improve on this doc.

Thanks!

StephenBonikowsky:
I've added the case study.
Also opened #4228 to track the supporting doc that needs to be created and linked in this playbook.

Issue #3984, opened 2019-10-22
FaultException is not marked Serializable


wouterroos:
**Describe the bug**
When attempting to port code running the .NETFramework code that uses (and serializes) FaultException I ran into the issue that FaultException is not marked as Serializable in System.ServiceModel.Primitives:

https://github.com/dotnet/wcf/blob/f0b3b4935eb97807dbed7b73f25fbf160a523eac/src/System.Private.ServiceModel/src/System/ServiceModel/FaultException.cs#L18

Is this by design? If I look at the System.ServiceModel reference source the attribute is present (https://github.com/microsoft/referencesource/blob/master/System.ServiceModel/System/ServiceModel/FaultException.cs#L21) and netstandard/core does seem to support it just fine (https://docs.microsoft.com/en-us/dotnet/api/system.serializableattribute?view=netcore-3.0)

Thanks in advance,

Regards,

Wouter

StephenBonikowsky:
@wouterroos This is by design in .NET Core. There were quite a few changes around exceptions between the full framework and core this being one of them. A FaultException should not need to be serialized. Are you doing some custom work where you are doing something like that? If so could you describe your scenario?

wouterroos:
@StephenBonikowsky Our repro-case is using WCF including FaultException in Service Fabric and propagate FaultExceptions between services using remoting (which uses binary serialization) and some of these services being ported to NetCore. This works fine when using the full .NET System.ServiceModel but obviously won't using System.ServiceModel.Primitives due to the missing [Serializable] attribute. I don't know if this is a valid use-case from your perspective since there is indeed some 'custom work' involved, but it might highlight a type of 'breaking change' that is introduced by leaving out the attribute.

StephenBonikowsky:
@wouterroos Thanks for the additional info. We'll discuss/investigate this and respond back.

StephenBonikowsky:
@wouterroos There's nothing wrong with your scenario and we are ok with taking this change in our next release, this could be in a servicing update of release/3.1 or in 5.0.

.NET Core as a whole has adopted the position of not making exceptions Serializable, I believe this is because they want to be able to make changes to exceptions as they wish without breaking anyone. Since this is a WCF specific exception we don't have any issue making this change.

wouterroos:
@StephenBonikowsky Great, thanks. Do you have a timeline on the releases you mentioned so we can plan accordingly?

StephenBonikowsky:
[...]

PR #4205, opened 2020-03-16
Mark FaultException as Serializable

imcarolwang:
For issue #3984 .

@HongGit ,@mconnew , @StephenBonikowsky
I am not sure if mark the FaultException class as Serializable is enough to make the scenario function work, let me know your concerns.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/FaultException.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/FaultException.cs
@@ -11,6 +11,7 @@ using System.ServiceModel.Dispatcher;
 
 namespace System.ServiceModel
 {
+    [Serializable]
     [KnownType(typeof(FaultCodeData))]
     [KnownType(typeof(FaultCodeData[]))]
     [KnownType(typeof(FaultReasonData))]
@@ -325,6 +326,7 @@ namespace System.ServiceModel
         }
     }
 
+    [Serializable]
     public class FaultException<TDetail> : FaultException
     {
         public FaultException(TDetail detail)
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/FaultException.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/FaultException.cs
@@ -147,6 +147,16 @@ namespace System.ServiceModel
             get { return _fault; }
         }
 
+        internal void AddFaultCodeObjectData(SerializationInfo info, string key, FaultCode code)
+        {
+            info.AddValue(key, FaultCodeData.GetObjectData(code));
+        }
+
+        internal void AddFaultReasonObjectData(SerializationInfo info, string key, FaultReason reason)
+        {
+            info.AddValue(key, FaultReasonData.GetObjectData(reason));
+        }
+
         private static FaultCode CreateCode(string code)
         {
             return (code != null) ? new FaultCode(code) : DefaultCode;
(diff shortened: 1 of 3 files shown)


imcarolwang:
> This should also be added to the reference assembly. Can you add a simple serialization test to the unit tests for S.SM.Primitives to ensure the serialized payload is the same between NetFx and Core.

Now I've added the attribute to the ref project file System.ServiceModel.Primitives.cs.

And I tried to serialize a FaultException instance with XmlSerializer, it doesn't work in .Net Framework nor .Net Core due to error about the base class "Exception" implements IDictionary.

I can use BinaryFormater to do serialization for FaultException type but serialized payload comparison with NetFramework is different. Any suggestion? Btw, deserialization from the serialized MemoryStream won't work as there seems has other dependencies not implemented, shall we support that as well?

mconnew:
XmlSerializer can't serialize IDictionary, but DataContractSerializer can. I couldn't find an instance of MemoryStream on FaultException or Exception. Where are you seeing this?

imcarolwang:
> XmlSerializer can't serialize IDictionary, but DataContractSerializer can. I couldn't find an instance of MemoryStream on FaultException or Exception. Where are you seeing this?

I forgot to mention I tried DataContractSerializer, it threw exception with message:

_System.Runtime.Serialization.InvalidDataContractException : Type 'System.ServiceModel.FaultException+FaultCodeData' cannot be serialized. Consider marking it with the DataContractAttribute attribute, and marking all of its members you want serialized with the DataMemberAttribute attribute. Alternatively, you can ensure that the type is public and has a parameterless constructor - all public members of the type will then be serialized, and no attributes will be required._

Regarding the MemoryStream thing, I am sorry it's not clear said, I meant after serialize FaultException instance to a memory stream, I was trying to deserialize it back, but failed to do so.

mconnew:
The FaultCodeData class also needs the Serializable attribute applied. If you compare the NetFx source with the .NET Core source, you can see that the attribute has been applied in NetFx.

imcarolwang:
Thanks, @mconnew . 

I've updated the product and test codes. Serializaing FaultException instance with DataContractSerializer would get equal result as  .NetFramework baseline after making following changes:
1. mark class `FaultCodeData `and `FaultReasonData `with `[Serializable]` attribute
2. add method `public override void GetObjectData(SerializationInfo info, StreamingContext context)` and its dependencies to FaultException.cs ([.net source ref](https://github.com/microsoft/referencesource/blob/a7bd3242bd7732dec4aebb21fbc0f6de61c2545e/System.ServiceModel/System/ServiceModel/FaultException.cs#L221
))
3. rename the private variable names in class FaultCodeData and FaultReasonData.
4. update the original .NetFramework baseline string section 
 `<HResult i:type=""x:int"" xmlns="""">-2146233087</HResult>`  to
 `<HResult i:type=""x:int"" xmlns="""">-2146233088</HResult>`

Does the changes in step1,2,3 make sense to you? Is result in step4 expected?

mconnew:
Your step 4 is interesting. On Framework, CommunicationException derives from SystemException which sets the default HResult to COR_E_SYSTEM (-2146233087). SystemException didn't exist in Core until netstandard2.0 so we had to derive from Exception instead, which defaults HResult to COR_E_EXCEPTION (-2146233088). This is where you are seeing the change. I'll open an issue around whether we should switch to SystemException as our base type. For now we'll include that change for the tests as it's caused by the base class difference.

mconnew:
Tests are failing because the Windows machines are running with the Spanish culture "es-ES" and the test expects "en-US".

mconnew:
@imcarolwang, can you modify the `FaultExceptionTest.Serializable_Default` test to capture the current culture before instantiating the FaultException, set the current culture to `en-US`, instantiate the FaultException object and then restore the current culture before continuing with the test? Some test machines purposefully have a different culture to catch product issues which don't show up with `en-US` so we need to be tolerant of running on a different culture.

mconnew:
@imcarolwang, don't worry about my last comment. I've fixed the build to use the English Windows image.

PR #4218, opened 2020-03-25
Blazor web assembly fixes

vidir-svanberg:
Fixes to support calling Soap Web Services from Blazor apps running in browser

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
@@ -27,7 +27,9 @@ namespace System.ServiceModel.Channels
         public HttpResponseMessageHelper(HttpResponseMessage httpResponseMessage, HttpChannelFactory<IRequestChannel> factory)
         {
             Contract.Assert(httpResponseMessage != null);
-            Contract.Assert(httpResponseMessage.RequestMessage != null);
+            // httpResponseMessage.RequestMessage is null when running in browser ( WebAssembly )
+            if (!Fx.IsWasm)
+              Contract.Assert( httpResponseMessage.RequestMessage != null );
             Contract.Assert(factory != null);
             _httpResponseMessage = httpResponseMessage;
             _httpRequestMessage = httpResponseMessage.RequestMessage;
@@ -344,12 +346,12 @@ namespace System.ServiceModel.Channels
             {
                 if (_httpResponseMessage.StatusCode == HttpStatusCode.NotFound)
                 {
-                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new EndpointNotFoundException(SR.Format(SR.EndpointNotFound, _httpRequestMessage.RequestUri.AbsoluteUri)));
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new EndpointNotFoundException(SR.Format(SR.EndpointNotFound, _httpRequestMessage?.RequestUri.AbsoluteUri)));
                 }
 
                 if (_httpResponseMessage.StatusCode == HttpStatusCode.ServiceUnavailable)
                 {
-                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ServerTooBusyException(SR.Format(SR.HttpServerTooBusy, _httpRequestMessage.RequestUri.AbsoluteUri)));
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ServerTooBusyException(SR.Format(SR.HttpServerTooBusy, _httpRequestMessage?.RequestUri.AbsoluteUri)));
                 }
 
                 if (_httpResponseMessage.StatusCode == HttpStatusCode.UnsupportedMediaType)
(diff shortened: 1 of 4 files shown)


mconnew:
The correctness of this code depends on the intention of behavior when using Blazor Web Assembly. These changes will significantly change the behavior of existing code, specifically around credentials. Is it better to throw an exception with a configuration which won't work or is it better to silently do something different. The specific code I'm concerned about is setting credentials on the client handler. We have this existing code:
```c#
                clientHandler.UseDefaultCredentials = false;
                if (credential == CredentialCache.DefaultCredentials || credential == null)
                {
                    clientHandler.UseDefaultCredentials = true;
                }
                else
                {
                    if (Fx.IsUap)
                    {
                        clientHandler.Credentials = credential;
                    }
                    else
                    {
                        CredentialCache credentials = new CredentialCache();
                        credentials.Add(GetCredentialCacheUriPrefix(via),
                            AuthenticationSchemesHelper.ToString(AuthenticationScheme), credential);
                        clientHandler.Credentials = credentials;
                    }
                }
```
The proposed change is to just not run this code when running in Web Assembly. An alternative approach would be:
```c#
    bool defaultCreds = (credential == CredentialCache.DefaultCredentials || credential == null);
    if (Fx.IsWasm & !defaultCreds)
    {
        // This would need to be a string resource, just a constant here for demonstration
        throw FxTrace.Exception.AsError(new InvalidOperationException("Explicit credentials not supported on Web Assembly");
    }
    if (defaultCreds)
    {
        clientHandler.UseDefaultCredentials = true;
    }
    else
    {
        clientHandler.UseDefaultCredentials = false;
        if (Fx.IsUap)
        {
            clientHandler.Credentials = credential;
        }
        else
        {
            CredentialCache credentials = new CredentialCache();
            credentials.Add(GetCredentialCacheUriPrefix(via),
                  AuthenticationSchemesHelper.ToString(AuthenticationScheme), credential);
            clientHandler.Credentials = credentials;
        }
    }
```

@vidir-svanberg, what's the general ideology when it comes to capability gaps with Web Assembly? Is it to change behavior or is it to throw when not supported?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -27,7 +27,9 @@ internal class HttpResponseMessageHelper
         public HttpResponseMessageHelper(HttpResponseMessage httpResponseMessage, HttpChannelFactory<IRequestChannel> factory)
         {
             Contract.Assert(httpResponseMessage != null);
-            Contract.Assert(httpResponseMessage.RequestMessage != null);
+            // httpResponseMessage.RequestMessage is null when running in browser ( WebAssembly )
+            if (!Fx.IsWasm)
```
mconnew: This is a bug in WebAssembly. I would prefer that the WebAssembly implementation is fixed here.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -359,15 +361,15 @@ private void ValidateResponseStatusCode()
...
                     }
-                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ProtocolException(SR.Format(SR.FramingContentTypeMismatch, _httpRequestMessage.Content.Headers.ContentType.ToString(), _httpRequestMessage.RequestUri)));
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ProtocolException(SR.Format(SR.FramingContentTypeMismatch, _httpRequestMessage?.Content.Headers.ContentType.ToString(), _httpRequestMessage?.RequestUri)));
                 }
 
                 if (_httpResponseMessage.StatusCode == HttpStatusCode.GatewayTimeout)
                 {
-                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new TimeoutException(_httpResponseMessage.StatusCode + " " + _httpResponseMessage.ReasonPhrase));
+                    throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new TimeoutException(_httpResponseMessage.StatusCode + " " + _httpResponseMessage?.ReasonPhrase));
                 }
```
mconnew: Did you get carried away here and not notice this is the response message and not the request message? I would expect the response message to not be null.

src/System.Private.ServiceModel/src/System/ServiceModel/Description/DataContractSerializerOperationBehavior.cs
```diff
@@ -82,7 +82,8 @@ public virtual XmlObjectSerializer CreateSerializer(Type type, string name, stri
             settings.MaxItemsInObjectGraph = MaxItemsInObjectGraph;
             settings.DataContractResolver = DataContractResolver;
             DataContractSerializer dcs = new DataContractSerializer(type, settings);
-            dcs.SetSerializationSurrogateProvider(SerializationSurrogateProvider);
+             if (Runtime.Fx.IsWasm && SerializationSurrogateProvider != null)
+              dcs.SetSerializationSurrogateProvider(SerializationSurrogateProvider);
```
mconnew: Import System.Runtime and refer to Fx.IsWasm instead of semi-qualifying the namespace. Also, SetSerializationSurrogateProvider has been supported since .NET Standard 1.6, is it really missing from Web Assembly? It was in an extension package in earlier versions.

vidir-svanberg:
> The correctness of this code depends on the intention of behavior when using Blazor Web Assembly. These changes will significantly change the behavior of existing code, specifically around credentials. Is it better to throw an exception with a configuration which won't work or is it better to silently do something different. The specific code I'm concerned about is setting credentials on the client handler. We have this existing code:
> 
> ```cs
>                 clientHandler.UseDefaultCredentials = false;
>                 if (credential == CredentialCache.DefaultCredentials || credential == null)
>                 {
>                     clientHandler.UseDefaultCredentials = true;
>                 }
>                 else
>                 {
>                     if (Fx.IsUap)
>                     {
>                         clientHandler.Credentials = credential;
>                     }
>                     else
>                     {
>                         CredentialCache credentials = new CredentialCache();
>                         credentials.Add(GetCredentialCacheUriPrefix(via),
>                             AuthenticationSchemesHelper.ToString(AuthenticationScheme), credential);
>                         clientHandler.Credentials = credentials;
>                     }
>                 }
> ```
> 
> The proposed change is to just not run this code when running in Web Assembly. An alternative approach would be:
> 
> ```cs
>     bool defaultCreds = (credential == CredentialCache.DefaultCredentials || credential == null);
>     if (Fx.IsWasm & !defaultCreds)
>     {
>         // This would need to be a string resource, just a constant here for demonstration
>         throw FxTrace.Exception.AsError(new InvalidOperationException("Explicit credentials not supported on Web Assembly");
>     }
>     if (defaultCreds)
>     {
>         clientHandler.UseDefaultCredentials = true;
>     }
>     else
>     {
>         clientHandler.UseDefaultCredentials = false;
>         if (Fx.IsUap)
>         {
>             clientHandler.Credentials = credential;
>         }
>         else
>         {
>             CredentialCache credentials = new CredentialCache();
>             credentials.Add(GetCredentialCacheUriPrefix(via),
>                   AuthenticationSchemesHelper.ToString(AuthenticationScheme), credential);
>             clientHandler.Credentials = credentials;
>         }
>     }
> ```
> 
> @vidir-svanberg, what's the general ideology when it comes to capability gaps with Web Assembly? Is it to change behavior or is it to throw when not supported?

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
vidir-svanberg: I agree it would be better if it was fixed in mono. 
This is just a quick fix which worked for me.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
vidir-svanberg: You are right.

src/System.Private.ServiceModel/src/System/ServiceModel/Description/DataContractSerializerOperationBehavior.cs
vidir-svanberg: No, it is not missing but it triggers a NullReferenceException in mono when the parameter is null.
You could say that this is a bug in mono.

vidir-svanberg:
I agree that throwing an exception is better but there is a problem with your fix.
Any references to either Credentials or UseDefaultCredentials will result in  PlatformNotSupportedException being throwned. 
I think you can view source for HttpClientHandler in the WebAssemby verision of mono on https://github.com/mono/mono/blob/master/mcs/class/System.Net.Http/System.Net.Http/HttpClientHandler.wasm.cs

PR #4221, opened 2020-03-30
Create central assembly version property for WCF assemblies.

StephenBonikowsky:
* Since all WCF assemblies share the same assembly version number no need to update more than one location.


StephenBonikowsky:
Consider cherry-picking to release/2.1.0 so both branches remain the same for releasing packages purposes.

HongGit (approved):

PR #4247, opened 2020-04-29
Exposing GenericXmlSecurityKeyIdentifierClause and virtual CloneCore().

brentschmaltz:
Exposing GenericXmlSecurityKeyIdentitfierClause is needed for WsFederationSecurityBinding effort as WsTrustChannelSecurityTokenProvider needs to set this clause type.

Exposing CloneCore() is needed so the derived WsTrustClientCredentails.CloneCore() will be called.

### DIFF
--- a/src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs
+++ b/src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs
@@ -5,8 +5,6 @@
 // ------------------------------------------------------------------------------
 
 
-using System.Xml;
-
 namespace System.ServiceModel
 {
     public partial class DnsEndpointIdentity : System.ServiceModel.EndpointIdentity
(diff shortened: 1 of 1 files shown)


src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs
```diff
@@ -5,6 +5,8 @@
 // ------------------------------------------------------------------------------
 
 
+using System.Xml;
+
```
mconnew, line 9: We don't add using statements, we fully qualify everything.

src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs
```diff
@@ -227,4 +234,16 @@ public enum SecurityKeyType
         AsymmetricKey,
         BearerKey
     }
+    public partial class GenericXmlSecurityKeyIdentifierClause : SecurityKeyIdentifierClause
+    {
```
mconnew, line 238: You need to declare that the Matches method has been overridden otherwise someone deriving from this class might get wrong behavior if calling base.Matches(). You should probably add the ReferenceXml property too otherwise you can't get the XmlElement back out if you want to.

brentschmaltz: updated

brentschmaltz:
@mconnew thanks !

PR #4290, opened 2020-05-29
Fixing The project format is incorrect

michal-ciechan:
Fixing `The project format is incorrect` exception when package identity name is `Update="Some.Package"` instead of `Include="Some.Package"`.

Skipping Package if `Include` attribute is not found.

I have reverted the tests I created, as I saw there is an outstanding PR #4271 which adds the Tests .csproj correctly!

I may leave a comment there to add tests for these cases?


src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
```diff
@@ -394,12 +398,26 @@ private static string GetItemValue(XElement reference, string itemName, bool thr
             return null;
         }
 
+        public static bool TryGetItemIdentity(XElement itemName, out string itemIdentity)
```
michal-ciechan, line 401: Wasn't sure if we also want to use this method when processing "Reference" nodes (line 234).

PR #4295, opened 2020-06-07
Fix Read() with count=0 exhausting MessageBodyStream

### LINKED ISSUE #4281: MessageBodyStream.Read() with count=0 unexpectedly terminates Stream
**Describe the bug**
If [MessageBodyStream.Read()](https://github.com/dotnet/wcf/blob/42b5572be18cb6d0d60718f24143a541b5dea755/src/dotnet-svcutil/lib/src/FrameworkFork/System.ServiceModel/System/ServiceModel/Dispatcher/StreamFormatter.cs#L239) is called with count=0, then _reader.ReadContentAsBase64() will return 0 bytes read, and so this code will execute which terminates the Stream:

[if (bytesRead == 0)](https://github.com/dotnet/wcf/blob/42b5572be18cb6d0d60718f24143a541b5dea755/src/dotnet-svcutil/lib/src/FrameworkFork/System.ServiceModel/System/ServiceModel/Dispatcher/StreamFormatter.cs#L278)
{
    Exhaust(_reader);
}

This 'Exhaust' case is presumably intended for when the bytesRead=0 result is due to no more bytes being able to be read from the XmlDictionaryReader, not for when bytesRead=0 simply is the trivial result of a count=0 argument.

I ran into this bug on .NET Desktop Framework, but the code appears the same here, so I'm filing an issue here, though I technically haven't reproduced this issue with this repo's .NET Core Nuget package.

**To Reproduce**
I think this bug is obvious from looking at the code, and my current projects in which I am reproducing this are too large to post here. So, I'm not posting a repro here, but let me know if you need one.

**Expected behavior**
I expect that calling MessageBodyStream.Read() with count=0 should have no side-effects on the state of the MessageBodyStream (i.e. should not advance Position and should not affect the result of any subsequent call to MessageBodyStream.Read()), since that is the behavior for any other Stream in .NET as far as I have observed.

Granted, the [Stream.Read() documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.read) doesn't seem to explicitly call out how to handle count=0, and, actually, strictly interpreting this sentence in the documentation would actually support the current behavior: "Read returns 0 only when there is no more data in the stream and no more is expected". However, that seems to me to be a documentation bug since that is not the behavior of passing count=0 to any other .NET Stream that I've worked with, such as FileStream or MemoryStream.

**Additional context**
In my case, and, perhaps, in most cases, it is simple enough to adjust the consuming code to avoid calling MessageBodyStream.Read() in the first place if count is going to be passed as 0, which is probably ideally how code should be written anyway. However, in order to know to work around it, first we have to notice the behavior is not what we expected based on how our code works with all other .NET Streams. Also, the code making the call to MessageBodyStream.Read() may not always be under our control to change (e.g. in third-party library code). So, I think this should be fixed in MessageBodyStream itself.

mriehm:
Fixes #4281

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
@@ -299,6 +299,11 @@ namespace System.ServiceModel.Dispatcher
 
             public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
             {
+                if (count == 0)
+                {
+                    return 0;
+                }
+
                 if (_reader == null)
                 {
                     if (_message.Properties.ContainsKey(BufferedReadStream.BufferedReadStreamPropertyName))
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/StreamFormatter.cs
@@ -344,6 +344,11 @@ namespace System.ServiceModel.Dispatcher
                     throw TraceUtility.ThrowHelperError(new ArgumentException(SR.Format(SR.SFxInvalidStreamOffsetLength, offset + count)), _message);
                 }
 
+                if (count == 0)
+                {
+                    return 0;
+                }
+
                 try
                 {
                     if (_reader == null)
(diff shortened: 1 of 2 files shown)


mconnew:
@mriehm, thank you for this. Can you add a similar check in the ReadAsync method too?

mriehm:
Sure. Done.

ReadAsync() appears to be just a wrapper for Read(), so I don't know if adding the check to ReadAsync() itself is behaviorally required, but it may be a performance optimization.

mconnew:
@mriehm, the async read does an async pre-read before synchronously reading. This is because the read goes through a serializer which is sync only. The pre-read is to make and effort to fill a buffer hoping it's enough and can be done asynchronously. It's a best effort attempt to try and get things to actually happen async.
I'll merge this as soon as the tests pass. The failure was nothing to do with your change.

mconnew:
Test which is failing is due to needing an updated version of the testserver being deployed and this PR is based off an older commit. This should be safe to merge so merging despite test failures.

PR #4297, opened 2020-06-15
System.ServiceModel.Federation package

mconnew:
This is the initial code drop of System.ServiceModel.Federation


mconnew:
This is still WIP. I need to clean up the logging to use the WCF error handling mechanisms. There's also at least one bug, turning off EstablishSecurityContext doesn't work. And there's at least one api oddity that needs to be resolved. Also we need the Microsoft.IdentityModel.Protocols.WsTrust package to be available in nuget and it isn't available there yet.

PR #4298, opened 2020-06-17
System.ServiceModel.Federation package

mconnew:
This is still WIP. I need to clean up the logging to use the WCF error handling mechanisms. There's also at least one bug, turning off EstablishSecurityContext doesn't work. And there's at least one api oddity that needs to be resolved. Also we need the Microsoft.IdentityModel.Protocols.WsTrust package to be available in nuget and it isn't available there yet.
I also need to update this to use an official WSTrust package from nuget once it's available.


mconnew:
This is still pulling the IdentityModel packages from a separate feed so can't be merged yet. Still waiting on a resolution to having the dependent packages available on nuget.

PR #4309, opened 2020-07-16
Remove IsTest=true from some utility projects

mconnew:
This is blocking updating of versions pulled in by Aracade

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Infrastructure/Infrastructure.Common.csproj
+++ b/src/System.Private.ServiceModel/tests/Common/Infrastructure/Infrastructure.Common.csproj
@@ -3,7 +3,7 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <OutputType>Library</OutputType>
     <CLSCompliant>false</CLSCompliant>
-    <IsTestProject>true</IsTestProject>
+    <IsPackable>false</IsPackable>
   </PropertyGroup>
   
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


HongGit:
@mconnew why remove one, the other just change value from true to false?  why not remove both?

src/System.Private.ServiceModel/tests/Common/Infrastructure/Infrastructure.Common.csproj
```diff
@@ -3,7 +3,7 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <OutputType>Library</OutputType>
     <CLSCompliant>false</CLSCompliant>
-    <IsTestProject>true</IsTestProject>
+    <IsPackable>false</IsPackable>
```
HongGit, line 6: Why not remove this line as well for consistency?

mconnew:
I think you might have misread the diff. I did remove it from both. I also added IsPackable=false to the Infrastructure project as it was in the unit test project. I wanted to make sure that removing the IsTest property didn't cause the project to attempt to be packaged. Rather than spend time experimenting, as it's already in the Unit project, I just added it to the infrastructure one too.

src/System.Private.ServiceModel/tests/Common/Infrastructure/Infrastructure.Common.csproj
HongGit: Huh, IsTest is removed, but add IsPackable

PR #4322, opened 2020-07-31
Fix RST generation issues in WsFederation

mjrousos:
This allows RSTs to be generated with null token types (to match NetFx behavior) and includes a requested ComputedKeyAlgorithm in combined entropy scenarios.

### DIFF
--- a/src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WsTrustChannelSecurityTokenManager.cs
+++ b/src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WsTrustChannelSecurityTokenManager.cs
@@ -41,7 +41,8 @@ namespace System.ServiceModel.Federation
             // TODO - we should check the value of the IssuedTokenType on WsTrustTokenParameters
             if (Saml2Constants.OasisWssSaml2TokenProfile11.Equals(tokenRequirement.TokenType) ||
                 Saml2Constants.Saml2TokenProfile11.Equals(tokenRequirement.TokenType) ||
-                SamlConstants.OasisWssSamlTokenProfile11.Equals(tokenRequirement.TokenType))
+                SamlConstants.OasisWssSamlTokenProfile11.Equals(tokenRequirement.TokenType) ||
+                tokenRequirement.TokenType is null) // Treat unspecified token types as being SAML
             {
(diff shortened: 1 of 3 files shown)


src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WsTrustChannelSecurityTokenManager.cs
```diff
@@ -38,10 +40,10 @@ public override SecurityTokenProvider CreateSecurityTokenProvider(SecurityTokenR
             if (tokenRequirement == null)
                 throw LogHelper.LogArgumentNullException(nameof(tokenRequirement));
 
-            // TODO - we should check the value of the IssuedTokenType on WsTrustTokenParameters
-            if (Saml2Constants.OasisWssSaml2TokenProfile11.Equals(tokenRequirement.TokenType) ||
-                Saml2Constants.Saml2TokenProfile11.Equals(tokenRequirement.TokenType) ||
-                SamlConstants.OasisWssSamlTokenProfile11.Equals(tokenRequirement.TokenType))
+            // If the token requirement includes an issued security token parameter of type
+            // WsTrustTokenParameters, then tokens should be provided by a WsTrustChannelSecurityTokenProvider.
+            if (tokenRequirement.TryGetProperty(IssuedSecurityTokenParametersProperty, out SecurityTokenParameters issuedSecurityTokenParameters) &&
```
mjrousos, line 45: Will the value of the IssuedSecurityTokenParametersProperty always be of type `SecurityTokenParameters`? In my testing it always is, but the `TryGetProperty` call will fail if we ever have a different type here. Should I make the out parameter an object to be safe or do we expect this property to always be a `SecurityTokenParameters` (if it exists)?

mconnew: We have code [here](https://github.com/dotnet/wcf/blob/5e5fd9e58cc410f48f4f9feed34801a0a6a65def/src/System.Private.ServiceModel/src/System/ServiceModel/Security/ClientCredentialsSecurityTokenManager.cs#L208) which makes this presumption about the type so if it ever isn't that type, we will break in that code too.

PR #4337, opened 2020-08-13
Use NET5 version with Http auth bug fixed

mconnew:
The version of .NET 5 we were using had a bug which caused Windows auth with default credentials over https to fail.


mconnew:
@HongGit, @StephenMolloy can one of you CR this?

PR #4353, opened 2020-09-03
Add support for AuthenticationSchemes.IntegratedWindowsAuthentication

### LINKED ISSUE #3694: The HTTP request is unauthorized with client authentication scheme 'Ntlm'. The authentication header received from the server was 'Negotiate, NTLM'.
Hi,

We have a bunch of WCF Services (.NET Framework 4.6.1) that need to be consumed by an ASP.NET Core (2.2 in IIS) client over an endpoint configured to use a custom binding with the Ntlm authorization scheme. Both WCF Service and the ASP.NET Core client are configured to use Windows Authentication with Negociate and NTLM as providers. Every time we try to consume the service, we have this error message: “The HTTP request is unauthorized with client authentication scheme 'Ntlm'. The authentication header received from the server was 'Negotiate, NTLM'.” I have to say that I’ve tried almost every voodoo thing’s I’ve found on the Internet to solve this without success. Changing provider order on both sides didn’t have any effect. Even If I remove Negotiate in the client providers, it still sends Negociate in the message header (see attached screenshots).  I don’t understand why… The only way I can make this thing work is to remove the Negociate provider from the WCF Service. We need Negociate and NTLM so this is not a solution for us. I’m wondering if this is a bug or is there something I’m missing?

Thank you!

fail - 1 - client NTLM only
fail - 2 - service Negociate, NTLM
fail - 3 - Negociate in client header even if its not in its provider
succes - 1 - client with negociate and ntlm
succes - 2 - server NTLM only
succes - 3 - client negociate, ntlm and service ntlm only

[BasicWcfFromCore.zip](https://github.com/dotnet/wcf/files/3306339/BasicWcfFromCore.zip)

mconnew:
When the server responds with both Negotiate and NTLM as supported authentication types, HttpClient will look in the credential cache for Negotiate credentials but not NTLM. This change allows specifying AuthenticationSchemes.IntegratedWindowsAuthentication on the HttpTransportBindingElement.AuthenticationScheme and WCF will add the credentials to the CredentialCache for Negotiate and NTLM.
Fixes #3694

### DIFF
--- a/src/System.Private.ServiceModel/src/Resources/Strings.resx
+++ b/src/System.Private.ServiceModel/src/Resources/Strings.resx
@@ -862,10 +862,10 @@
     <value>Cannot resolve the host name of URI "{0}" using DNS.</value>
   </data>
   <data name="HttpRequiresSingleAuthScheme" xml:space="preserve">
-    <value>The '{0}' authentication scheme has been specified on the HTTP factory. However, the factory only supports specification of exactly one authentication scheme. Valid authentication schemes are Digest, Negotiate, NTLM, Basic, or Anonymous.</value>
+    <value>The '{0}' authentication scheme has been specified on the HTTP factory. However, the factory only supports specification of exactly one authentication scheme. Valid authentication schemes are Digest, Negotiate, NTLM, Basic, IntegratedWindowsAuthentication, or Anonymous.</value>
   </data>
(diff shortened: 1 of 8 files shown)


StephenMolloy (approved):
lgtm

HongGit (approved):

PR #4356, opened 2020-09-04
Add support for X509CertificateIdentity with https

### LINKED ISSUE #3881: Supporting client endpoint identity certificate
I tried to create programmatically in .NET Core Preview 9 something similar to to the following .NET Framework code, since xml configurations are not supported.

```
var binding = new WSHttpBinding("WSHttpBinding_ICustomerService");
var cert = CreateCertificate();
var endpointAddress = new EndpointAddress(
    new Uri("url"),
    EndpointIdentity.CreateX509CertificateIdentity(cert));
var client = new CustomerService.CustomerServiceClient(binding, endpointAddress);
```

This is instead of the following client configuration:

```
<endpoint address="url" binding="wsHttpBinding" 
	  bindingConfiguration="WSHttpBinding_ICustomerService" 
          contract="CustomerServiceDev.CustomerService" name="WSHttpBinding_ICustomerService">
    <identity>
        <certificate encodedValue="cert"/>
    </identity>
</endpoint>
```

It seems that .NET Core Preview 9 is missing the API to create the identity certificate with `EndpointIdentity.CreateX509CertificateIdentity` or with `X509CertificateEndpointIdentity` class, though the class is in [this repository](https://github.com/dotnet/wcf/blob/f0b3b4935eb97807dbed7b73f25fbf160a523eac/src/System.Private.ServiceModel/src/System/ServiceModel/X509CertificateEndpointIdentity.cs) and it is public.

Is there a way in .NET Core 3 to achieve the same effect as what is done in the previous configuration? Maybe `X509CertificateValidator`? I am interested in any code for client to do service certificate validation.

I see that `DnsEndpointIdentity` made its way in Preview 9, maybe I can propose that one to my team to do the validation instead. Will it be officially supported in final release?

mconnew:
Fixes #3881

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportSecurityHelpers.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportSecurityHelpers.cs
@@ -4,6 +4,7 @@
 
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IdentityModel.Selectors;
 using System.IdentityModel.Tokens;
 using System.Net;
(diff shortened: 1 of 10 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportSecurityHelpers.cs
```diff
@@ -336,89 +337,61 @@ public static void RemoveIdentityMapping(Uri via, EndpointAddress target, bool v
...
                     "HTTPS server certificate identity contains multiple certificates");
-                AddServerCertMapping(request, remoteCertificateIdentity.Certificates[0].Thumbprint);
-            }
-        }
+                var thumbprint = remoteCertificateIdentity.Certificates[0].Thumbprint;
+                bool identityValidator(HttpRequestMessage requestMessage, X509Certificate2 cert, X509Chain chain, SslPolicyErrors policyErrors)
+                {
+                    try
+                    {
+                        ValidateServerCertificate(cert, thumbprint);
```
StephenMolloy, line 353: thumbprint isn't being checked for null anymore. I couldn't tell if it was checked before because we were retrieving something that may or may not have existed from s_serverCertMap... or because it might actually be null in some cases.

StephenMolloy (approved):
Looks fine to me. I had one question about the potential for a null thumbprint. I'd guess it's fine, but don't know much about this space.

HongGit (approved):

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/TransportSecurityHelpers.cs
mconnew: I verified with @bartonjs that Thumprint can't ever be null, and in the process of the conversation came to the conclusion that this isn't the best implementation. I'm going to switch it to comparing the RawData (which also can't be null) of the certificate as it will be both faster and avoids hash collisions.

mconnew: @StephenMolloy. I've modified the approach can you re-review.

Issue #4377, opened 2020-09-25
Consider changing calls from GetNonZeroBytes to GetBytes


GrabYourPitchforks:
The file https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/src/System/ServiceModel/Security/WSTrust.cs contains two calls to `RandomNumberGenerator.GetNonZeroBytes`. Is it intentional that this code use `GetNonZeroBytes` over the preferred API `RandomNumberGenerator.GetBytes`? Using `GetNonZeroBytes` reduces entropy somewhat, so it should only be used when the caller absolutely needs to prevent any zero bytes from being included in the buffer.

mconnew:
This is actually dead code, this method was pulled in by mistake and is only used in NetFx by server side code. So the action here is to remove the the entire ProcessRstAndIssueKey method.

HongGit:
@imcarolwang could try to submit a PR for this issue?

mconnew:
This is now fixed

PR #4397, opened 2020-10-16
Remove dead code.

imcarolwang:
For issue #4377. Have run tests locally with the change applied, see no regression.

@HongGit , @mconnew .


mconnew:
Fixes #4377

PR #4436, opened 2020-11-20
Change default WSTrustTokenParameters.KeyType from Bearer to Symmetric

### LINKED ISSUE #4425: CreateWS2007FederationTokenParameters should default key type to symmetric
**Describe the bug**

WSFederation and WS2007Federation token parameters should default KeyType to Symmetric. Currently, CreateWSFederationTokenParameters does this but [CreateWS2007FederationTokenParameters uses bearer](https://github.com/dotnet/wcf/blob/master/src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WsTrustTokenParameters.cs#L148).

As a workaround, WS2007Federation users can set the token parameters' key type to Symmetric explicitly, but we should update the default.

ssa3512:
Fixes #4425


mconnew:
Sorry, I missed this. I'm going to close this one in preference of #4440 as that includes necessary fixes to the tests.

PR #4438, opened 2020-11-20
Make SecureConversation inner channel factory share the same HttpClient as outer factory

mconnew:
When using SecureConversation, an inner channel factory is used to obtain the Secure Conversation token. When using a load balancer, this can result in the session establishment and service request being made to different backend hosts resulting in the service call failing.

### DIFF
--- a/src/System.Private.ServiceModel/src/Internals/System/Runtime/MruCache.cs
+++ b/src/System.Private.ServiceModel/src/Internals/System/Runtime/MruCache.cs
@@ -2,10 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-
 using System.Collections.Generic;
 using System.ServiceModel;
-using System.Threading;
 
 namespace System.Runtime
 {
@@ -19,15 +17,26 @@ namespace System.Runtime
         private readonly int _highWatermark;
         private CacheEntry _mruEntry;
         private int _refCount = 1;
+        private object _mutex = new object();
 
         public MruCache(int watermark)
             : this(watermark * 4 / 5, watermark)
         {
         }
 
-        public void AddRef()
+        // Returns whether the cache can be used by the caller after calling AddRef
+        public bool AddRef()
         {
-            Interlocked.Increment(ref _refCount);
+            lock(_mutex)
+            {
+                if (_refCount == 0)
+                {
+                    return false;
+                }
+
+                _refCount++;
+                return true;
+            }
         }
 
         //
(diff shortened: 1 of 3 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpTransportBindingElement.cs
```diff
@@ -406,6 +413,20 @@ public override T GetProperty<T>(BindingContext context)
             }
         }
 
+        private MruCache<string, HttpClient> EnsureHttpClientCache()
```
StephenMolloy, line 416: Maybe I'm missing something up the stack, but I don't think this is thread safe. Does it need to be?

src/System.Private.ServiceModel/src/Internals/System/Runtime/MruCache.cs
```diff
@@ -202,7 +209,12 @@ public bool TryGetValue(TKey key, out TValue value)
 
         public void Dispose()
         {
-            Dispose(true);
+            int refCount = Interlocked.Decrement(ref _refCount);
```
StephenMolloy, line 212: Isn't there a race with this ref counting? Thread A can begin disposing and decrement, Thread B can AddRef and increment, and then Thread B can continue to dispose, even though Thread B thinks they have a valid reference.

StephenMolloy (changes requested):
I'm not familiar with the code upstream that gets into these changes, but I have some thread safety questions, which is always a concern with a ref-counting approach.

src/System.Private.ServiceModel/src/Internals/System/Runtime/MruCache.cs
mconnew: Although it won't be used in that way as this is only possible to hit when a single instance of HttpTransportBindingElement is reused and it will be reused sequentially, I agree that having this have potential race conditions if that usage changes isn't the best idea. The only way I can see to make this bullet proof is with a lock. Do the ref counting in the lock and change AddRef to return a bool if it can't obtain increment it because it is disposed.

src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpTransportBindingElement.cs
mconnew: Doesn't need to be thread safe. When this class is first passed to ChannelFactory or any other of our classes, we clone it. This means the call path which gets here will be on an instance not accessible outside of our own code. In our code, we won't call the code path which gets to here on multiple threads simultaneously. It will happen sequentially.

mconnew:
@StephenMolloy, can you review the latest changes?

StephenMolloy (approved):
lgtm

PR #4451, opened 2020-11-27
Fix IOTS non windows condition in Cleanup

### LINKED ISSUE #4449: net5.0 on non-windows System.ArgumentNullException at System.Threading.Overlapped.Free
**Describe the bug**
After upgrading our application to .net5.0 we experience regular exceptions on linux docker containers with the stack below

```
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'nativeOverlappedPtr')
   at System.Threading.Overlapped.Free(NativeOverlapped* nativeOverlappedPtr)
   at System.Runtime.IOThreadScheduler.ScheduledOverlapped.Cleanup()
   at System.Runtime.IOThreadScheduler.Cleanup()
   at System.Runtime.IOThreadScheduler.Finalize()
Aborted (core dumped)
``` 

**To Reproduce**
.net5.0 application calling an external resource through WCF generated client code wsdl [here](https://s7sps3apissl.scene7.com/scene7/IpsApi-2014-04-03.wsdl)
running in linux docker container 

**Additional context**
We highly suspect something related to #4360 . At first glance the comments and changes seems to match our current issue. Strangely the change made in `IOThreadScheduler.cs` file seems to be the opposite of the original purpose of the PR. 

Any help would be greatly appreciated
Thanks in advance
Morgan

richamo:
Fix #4449

### DIFF
--- a/src/System.Private.ServiceModel/src/Internals/System/Runtime/IOThreadScheduler.cs
+++ b/src/System.Private.ServiceModel/src/Internals/System/Runtime/IOThreadScheduler.cs
@@ -498,7 +498,7 @@ namespace System.Runtime
                     throw Fx.AssertAndThrowFatal("Cleanup called on an overlapped that is in-flight.");
                 }
 				
-                if (!s_isWindows)
+                if (s_isWindows)
                 {
                     Overlapped.Free(_nativeOverlapped);
                 }


mconnew:
I believe these tests are failing because this change is based off of an earlier point in master and the Federation tests were modified since then making the client tests from the source branch incompatible with the current head server side. This change won't have any effect on the failing tests. The same failure was seen in a different PR. When I rebased that PR, the same tests started passing. Once the current run completes, I'll merge.

Issue #4359, opened 2020-09-08
--Update option does not seem to work


Bitfiddler:
**Describe the bug**
Used the following to create a service reference
dotnet svcutil http://localhost/MyApp/MyServices/SomeService.asmx?wsdl -syn -pf "C:\projects\MySolution\MyProject\MyProject.csproj" -d "C:\projectsprojects\MySolution\MyProject\MyService" 

No combination of options seem to work to update this reference.

Please see:
https://stackoverflow.com/questions/63800395/svcutil-update-error-output-not-making-sense

**To Reproduce**
1. Create a service reference as above
2. Try to update it using the command above.

You get an error:
Warning: The specified --projectFile option is not expected in the current operational context of the tool!
Error: The --update option requires a project to be specified, please use the --projectFile option to specify the project to use.

On top of the fact that this shouldn't be an error, this combination of warning/error makes no sense.
First it tells you that --projectFile is not expected in this context, then it tells you it is an error NOT to have --projectFile.

**Expected behavior**
I expect something like:
dotnet svcutil -u --projectFile "C:\projects\MySolution\MyProject\MyProject.csproj"

To work without error (as per the documentation and error messages)

imcarolwang:
`-projectFile` option is not supported in the context to `--update` a service. The recommended usage of the feature is to go to directory which contains the project file and run the update command. For example, in your case:

```
cd C:\projects\MySolution\MyProject
dotnet-svcutil --update <web service reference>
```

`<web service reference>` is folder name which contains the generated proxy code, it's optional if you have only one service to be updated.

Sorry for the confusing. As a fix, we will add an example in the Command Usage Helper to make it clear.

oguzhankahyaoglu:
This does not work definitely. No matter what I have tried, it did not work.
I had a project structure (.netcore3.1) like this:
-Clients.SatSasDetailService\
----Connected Services\
--------ZMM_SAT_SAS_DETAILService\
-------------Reference.cs
-------------ConnectedService.json
----Clients.SatSasDetailService.csproj

Could you give me some ides how can I update this exisiting wsdl service reference?

imcarolwang:
@oguzhankahyaoglu  looks like you're using WCF Connected Service tool from Visual Studio to generate the service reference, in this case you should be able to update the service via right clicking service name "ZMM_SAT_SAS_DETAILService" from the project tree and select `"Update Microsoft WCF Web Service Provider..." `from the menu list to update the service reference.

[...]

PR #4453, opened 2020-11-30
dotnet svcutil - add example for update option usage

imcarolwang:
@HongGit , @mconnew,  

Fixing issue #4359,  to throw meaningful error when running dotnet-svcutil to update service reference in global tool context but from unexpected working directory (projectFile not found under the dir).


imcarolwang:
@HongGit  the fix is updated base on our discussion, could you please help review?

PR #4466, opened 2020-12-17
Make all WCF exceptions serializable

### LINKED ISSUE #4448: FaultException implements Serializable but does not use serialization constructors
**Describe the bug**
As part of #3984 FaultException was marked Serializable which was released in System.ServiceModel.Primitives 4.8.0
However I am still not able to use the use case as described by Wouter. Using code that serializes FaultExceptions using remoting. After updating my dependency on System.ServiceModel.Primitives I still receive the stack trace

```
Failed to deserialize and get remote exception  System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.ServiceModel.CommunicationException..ctor(SerializationInfo info, StreamingContext context)
   at System.ServiceModel.FaultException..ctor(SerializationInfo info, StreamingContext context)
   at System.ServiceModel.FaultException`1..ctor(SerializationInfo info, StreamingContext context)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeConstructorInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters)
   at System.Runtime.Serialization.ObjectManager.CompleteISerializableObject(Object obj, SerializationInfo info, StreamingContext context)
   at System.Runtime.Serialization.ObjectManager.FixupSpecialObject(ObjectHolder holder)
   at System.Runtime.Serialization.ObjectManager.DoFixups()
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.Deserialize(BinaryParser serParser, Boolean fCheck)
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Deserialize(Stream serializationStream, Boolean check)
   at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Deserialize(Stream serializationStream)
   at Microsoft.ServiceFabric.Services.Remoting.V2.RemoteException.TryDeserializeException(Stream data, Exception& result)
```

As far as I can tell that is because `protected FaultException(SerializationInfo info, StreamingContext context)` throws a `PlatformNotSupportedException`

In reference source that method is implemented https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/System.ServiceModel/System/ServiceModel/FaultException.cs#L120 

The scenario is the same as described by Wouter in the referenced issue. 
Is this something that also needs to be taken into account?

Thanks in advance,
Regards,
Max

mconnew:
Fixes #4448
Also improves the use of DataContractSerializer with FaultException\<TDetail\> so you don't need to provide typeof(TDetail) to DataContractSerializer.

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/ActionNotSupportedException.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/ActionNotSupportedException.cs
@@ -2,18 +2,18 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-
 using System.Diagnostics.Contracts;
 using System.Runtime.Serialization;
 using System.ServiceModel.Channels;
 
 namespace System.ServiceModel
 {
+    [Serializable]
     public class ActionNotSupportedException : CommunicationException
     {
         public ActionNotSupportedException() { }
         public ActionNotSupportedException(string message) : base(message) { }
-        public ActionNotSupportedException(string message, Exception innerException) : base(message, innerException) { throw new PlatformNotSupportedException(); }
+        public ActionNotSupportedException(string message, Exception innerException) : base(message, innerException) { }
         protected ActionNotSupportedException(SerializationInfo info, StreamingContext context) : base(info, context) { }
 
         internal Message ProvideFault(MessageVersion messageVersion)
@@ -21,8 +21,7 @@ namespace System.ServiceModel
             Contract.Assert(messageVersion.Addressing != AddressingVersion.None);
             FaultCode code = FaultCode.CreateSenderFaultCode(AddressingStrings.ActionNotSupported, messageVersion.Addressing.Namespace);
             string reason = Message;
-            return System.ServiceModel.Channels.Message.CreateMessage(
-               messageVersion, code, reason, messageVersion.Addressing.FaultAction);
+            return Channels.Message.CreateMessage(messageVersion, code, reason, messageVersion.Addressing.FaultAction);
         }
     }
 }
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/FaultException.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/FaultException.cs
@@ -19,90 +19,89 @@ namespace System.ServiceModel
     public class FaultException : CommunicationException
     {
         internal const string Namespace = "http://schemas.xmlsoap.org/Microsoft/WindowsCommunicationFoundation/2005/08/Faults/";
-        private MessageFault _fault;
 
         public FaultException()
             : base(SR.SFxFaultReason)
         {
-            Code = FaultException.DefaultCode;
-            Reason = FaultException.DefaultReason;
+            Code = DefaultCode;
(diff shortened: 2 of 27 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/ActionNotSupportedException.cs
```diff
@@ -2,27 +2,26 @@
...
 using System.ServiceModel.Channels;
 
 namespace System.ServiceModel
 {
+    [Serializable]
     public class ActionNotSupportedException : CommunicationException
     {
         public ActionNotSupportedException() { }
         public ActionNotSupportedException(string message) : base(message) { }
-        public ActionNotSupportedException(string message, Exception innerException) : base(message, innerException) { throw new PlatformNotSupportedException(); }
```
StephenMolloy, line 16: Presumably this was intentional and the base constructor does everything necessary? Just double checking since the repetitive pattern in this PR is removing PNSEx from serialization constructors, which this is not.

StephenMolloy (approved):
One quick question to add to Hong's. Otherwise looks good to me.

src/System.Private.ServiceModel/src/System/ServiceModel/ActionNotSupportedException.cs
mconnew: I checked the history of the original change and this looks like it was a mistake where the PNSE was thrown from the wrong method, and nobody noticed. How it is now is correct. WCF has lots of exceptions which provide no additional functionality and are there purely to provide strong typing on your catch clauses to handle different failure modes differently. So most of them just call the base constructor.

src/System.Private.ServiceModel/src/System/ServiceModel/InvalidMessageContractException.cs
```diff
@@ -2,33 +2,16 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-
 using System.Runtime.Serialization;
 
 namespace System.ServiceModel
 {
     public class InvalidMessageContractException : Exception
     {
```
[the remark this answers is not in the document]
mconnew: Yep, that was a miss. Fixed it now.

src/System.Private.ServiceModel/src/System/ServiceModel/FaultException.cs
```diff
@@ -19,90 +19,89 @@ namespace System.ServiceModel
     public class FaultException : CommunicationException
```
HongGit, line 19: @mconnew what about this class?  should this be marked as well?

HongGit: I missed the line above.

PR #4470, opened 2020-12-18
Make CallbackBehaviorAttribute's ConcurrencyMode property publicly accessible

imcarolwang:
@mconnew Turns out I was missing some change in this PR #4348, could you please help confirm that and review?

### DIFF
--- a/src/System.ServiceModel.Duplex/ref/System.ServiceModel.Duplex.cs
+++ b/src/System.ServiceModel.Duplex/ref/System.ServiceModel.Duplex.cs
@@ -13,10 +13,17 @@ namespace System.ServiceModel
         public CallbackBehaviorAttribute() { }
         public bool AutomaticSessionShutdown { get { return default(bool); } set { } }
         public bool UseSynchronizationContext { get { return default(bool); } set { } }
+        public System.ServiceModel.ConcurrencyMode ConcurrencyMode { get { return default(System.ServiceModel.ConcurrencyMode); } set { } }
         void System.ServiceModel.Description.IEndpointBehavior.AddBindingParameters(System.ServiceModel.Description.ServiceEndpoint serviceEndpoint, System.ServiceModel.Channels.BindingParameterCollection parameters) { }
         void System.ServiceModel.Description.IEndpointBehavior.ApplyClientBehavior(System.ServiceModel.Description.ServiceEndpoint serviceEndpoint, System.ServiceModel.Dispatcher.ClientRuntime clientRuntime) { }
         void System.ServiceModel.Description.IEndpointBehavior.ApplyDispatchBehavior(System.ServiceModel.Description.ServiceEndpoint serviceEndpoint, System.ServiceModel.Dispatcher.EndpointDispatcher endpointDispatcher) { }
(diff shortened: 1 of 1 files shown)


mconnew:
Before merging this, I want to have a discussion about deprecating the Reentrant concurrency mode. It causes problems with being able to enable OperationContext flowing asynchronously.

DrabanL:
> Before merging this, I want to have a discussion about deprecating the Reentrant concurrency mode. It causes problems with being able to enable OperationContext flowing asynchronously.

any chance of at least enabling Multiple mode meanwhile?

mconnew:
@imcarolwang, can you add the Obsolete attribute to ConcurrencyMode.Reentrant in the reference assembly only?

mconnew:
@DrabanL, we should be releasing this as part of a preview build very soon.

imcarolwang:
> @imcarolwang, can you add the Obsolete attribute to ConcurrencyMode.Reentrant in the reference assembly only?

Updated.

mconnew:
It was set to obsolete because enabling this conflicted with the ability to flow OperationContext.Current through await calls in an async method. It was an unsolvable problem to make the two work together.  

Using the Reentrant mode is equivalent to doing this:
```csharp
private object _reentrantLock = new object();

public string CallbackSomething(string data)
{
    lock(_reentrantLock)
    {
        /* Multiple lines of code
           involved in doing something */
    }
    client.ServerCall(42);
    lock(_reentrantLock)
    {
        /* More lines of code involved
           in doing something */
    }
}
```
The reason we chose to favor enabling flowing the OperationContext.Current through an await over supporting reentrant is that in over a decade of working with customers using WCF, every time someone is using reentrant,  they didn't actually need reentrant and they didn't actually understand what it was doing. Not even once. I would be very surprised if you actually need it, and I don't mean that you can just use the workaround I mentioned. Generally they just needed multiple.  

Can you explain your use case for reentrant? Even if you do actually need it, there might be another way to solve what you need.

8lue5ky:
> It was set to obsolete because enabling this conflicted with the ability to flow OperationContext.Current through await calls in an async method. It was an unsolvable problem to make the two work together.
> 
> Using the Reentrant mode is equivalent to doing this:
> 
> ```cs
> private object _reentrantLock = new object();
> 
> public string CallbackSomething(string data)
> {
>     lock(_reentrantLock)
>     {
>         /* Multiple lines of code
>            involved in doing something */
>     }
>     client.ServerCall(42);
>     lock(_reentrantLock)
>     {
>         /* More lines of code involved
>            in doing something */
>     }
> }
> ```
> 
> The reason we chose to favor enabling flowing the OperationContext.Current through an await over supporting reentrant is that in over a decade of working with customers using WCF, every time someone is using reentrant, they didn't actually need reentrant and they didn't actually understand what it was doing. Not even once. I would be very surprised if you actually need it, and I don't mean that you can just use the workaround I mentioned. Generally they just needed multiple.
> 
> Can you explain your use case for reentrant? Even if you do actually need it, there might be another way to solve what you need.

Thank you for the reply.

We have a service where we can open a new wcf connection with a callback contract where we receive events over that callback.

When we receive a specific "finished" event, we close the connection from the client side (from the callback class we call another WCF service method to unsubscribe and close this specific connection).

PR #4481, opened 2021-01-05
Make federation implementation async and fix memory leak

### LINKED ISSUE #4461: WSTrustChannelSecurityTokenProvider doesn't implement an async version of GetTokenCore
**Describe the bug**
SecurityTokenProvider has async versions of GetTokenCore to provide an async implementation. WSTrustChannelSecurityTokenProvider should implement this and make the request to the STS using an async request.

### LINKED ISSUE #4462: Memory leak in WSTrustChannelSecurityTokenProvider.GetTokenCore
**Describe the bug**
WSTrustChannelSecurityTokenProvider.GetTokenCore creates an instance of a channel and makes a request. The channel doesn't get closed and no persistent reference is kept by WSTrustChannelSecurityTokenProvider. The ChannelFactory maintains a reference to all opened channels and is the channel hasn't been closed, the ChannelFactory continues to root the object. This will result in a new channel being created and held on to each time a new token is requested.  

The request rate of Trust tokens is usually low but this could still add up in a middleware scenario.

mconnew:
Implemented BeginGetTokenCore and EndGetTokenCore to provide async implementation and fixed memory leak around ChannelFactory and channel lifetime.
Fixes #4462
Fixes #4461

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecuritySessionClientSettings.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecuritySessionClientSettings.cs
@@ -487,7 +487,7 @@ namespace System.ServiceModel.Security
             {
                 TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
                 SetupSessionTokenProvider();
-                SecurityUtils.OpenTokenProviderIfRequired(_sessionTokenProvider, timeoutHelper.RemainingTime());
+                await SecurityUtils.OpenTokenProviderIfRequiredAsync(_sessionTokenProvider, timeoutHelper.RemainingTime());
                 using (ServiceModelActivity activity = DiagnosticUtility.ShouldUseActivity ?
                     ServiceModelActivity.CreateBoundedActivity() : null)
                 {
--- a/src/System.ServiceModel.Federation/src/System/IdentityModel/Security/WrapperSecurityCommunicationObject.cs
+++ b/src/System.ServiceModel.Federation/src/System/IdentityModel/Security/WrapperSecurityCommunicationObject.cs
@@ -1,0 +1,146 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using System.ServiceModel.Federation.System.Runtime;
+using System.ServiceModel.Security;
+
+namespace System.IdentityModel.Security
+{
+    internal class WrapperSecurityCommunicationObject : CommunicationObject
+    {
+        ISecurityCommunicationObject _innerCommunicationObject;
+
+        public WrapperSecurityCommunicationObject(ISecurityCommunicationObject innerCommunicationObject) : base()
+        {
+            _innerCommunicationObject = innerCommunicationObject ?? throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(innerCommunicationObject));
+        }
+
+        protected override Type GetCommunicationObjectType()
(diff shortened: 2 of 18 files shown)


mconnew:
@brentschmaltz, can you take a look at this. I've tidied some things up in the process of fixing a couple of issues.

src/System.ServiceModel.Federation/src/System/IdentityModel/Security/WrapperSecurityCommunicationObject.cs
```diff
@@ -0,0 +1,146 @@
...
+
+using System.Runtime;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+using System.ServiceModel.Federation.System.Runtime;
+using System.ServiceModel.Security;
+
+namespace System.IdentityModel.Security
+{
+    class WrapperSecurityCommunicationObject : CommunicationObject
```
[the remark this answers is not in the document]
mconnew: You're right, I missed adding that when porting from the framework. I've fixed it now.

HongGit (approved):

src/System.ServiceModel.Federation/src/System/ServiceModel/Security/ISecurityCommunicationObject.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading.Tasks;
+
+namespace System.ServiceModel.Security
+{
+    interface ISecurityCommunicationObject
```
mconnew, line 9: Missing accessor

src/System.Private.ServiceModel/src/System/ServiceModel/Security/SecuritySessionClientSettings.cs
```diff
@@ -487,7 +487,7 @@ protected internal override async Task OnOpenAsync(TimeSpan timeout)
             {
                 TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
                 SetupSessionTokenProvider();
-                SecurityUtils.OpenTokenProviderIfRequired(_sessionTokenProvider, timeoutHelper.RemainingTime());
+                await SecurityUtils.OpenTokenProviderIfRequiredAsync(_sessionTokenProvider, timeoutHelper.RemainingTime());
```
brentschmaltz, line 490: should you have .ConfigureAwait(false) on these?

mconnew: It's not needed. I hated the idea of littering my code everywhere with ConfigureAwait because it's messy, and it only takes missing one to create a difficult to find bug. So instead I have some code which is called when calling WCF using an async contract which checks if you are using a non-default sync context or a non-default task scheduler, and if you are, forces an artificial continuation into the worker thread pool. So this code is guaranteed to not need ConfigureAwait. ConfigureAwait is needed if you _might_ go async. If you are guaranteed to always go async (and WCF always does as it always does network IO which requires a response), then you can just forcibly go async from the start.

brentschmaltz: OK, cool just checking.
If you have a controlled call graph, you are probably OK,
We created a Roslyn Code Analyzer to check for our other libraries we have many entry points.

@mconnew What was the memory leak?

mconnew: The memory leak was not closing the channel. The channel factory will hold on to the channel in an internal list. Each call would create a new instance which would accumulate instances. The fix is to close the channel after usage. Also we now close the channel factory when the parent channel factory is closed as that closes the token provider.

PR #4485, opened 2021-01-08
Fix async lock

### LINKED ISSUE #4435: Continuously degrading performance in repeated one-way calls (S.SM 4.8, netcore3.1)
**Describe the bug**
When calling a one way service with reliable sessions repeatedly, the time to make the call continually increases on the client side.

**To Reproduce**
https://github.com/ssa3512/wcf-reliable-sessions/tree/perf-repro
1) Run the wcfasynctestservice project
2) From wcfasynctestclient directory run `dotnet run -f netcoreapp3.1`
Observe the call time gradually increase (starts at ~5ms and after a few minutes it is over 500ms)
3) From wcfasynctestclient directory run `dotnet run -f net472`
Observe the call time is consistently <10ms over any time frame

**Expected behavior**
Subsequent calls to a service are as performant as the first call

**Screenshots**
image

**Additional context**
I set up an azure function that is calling a legacy one way WCF service on a timer (every three minutes). This is designed to replace a windows service that is calling the service on a timer. I noticed over the span of 24 hours that the function execution time increased from ~10ms to about 2.5 seconds. Stopping and starting the app service immediately reset the execution time back to the expected 10ms.

Some notes from console app repro tests:
- Using a ClientBase implementation or ChannelFactory does not seem to make a difference
- Async vs sync does not seem to make a difference
- Using separate binding/contract/endpoint address objects does not seem to make a difference
- Reusing the proxy/leaving the channel open seems to alleviate the issue, but that isn't an option in my scenario as the function execution does not allow reuse of the channel from one execution to the next.
- Edit: This occurs with both NetTcpBinding and WSHttpBinding

mconnew:
Fixed a SocketConnection memory leak where we were pinning the SocketConnection and buffer in memory after the socket is closed due to not cancelling timers. NetFx doesn't actually support synchronous completion of Socket.SendAsync or Socket.ReceiveAsync so this bug wasn't discovered on NetFx. There was also a bug with not disposing the System.Threading.Timer instances at close. Switched back to IOThreadTimer that NetFx uses which is cheap to cancel.
Fixed IOThreadTimer issue where the first WaitableTimer wasn't being reset so could have caused what would effectively be an asynchronous while(true) loop.
Fixed AsyncLock to correctly use a static AsyncLocal instead of an instance. This was causing a memory leak. Adjusted logic to account for multiple AsyncLock's on the same ExecutionContext. This was the cause of #4435.
Fixes #4435
Other issues were found while investigating #4435


mconnew:
@stephentoub, could you look at my AsyncLock implementation and check I haven't missed anything else obvious?

HongGit (approved):

src/System.Private.ServiceModel/src/Internals/System/Runtime/AsyncLock.cs
```diff
@@ -18,101 +19,130 @@ internal class AsyncLock
...
-            _lockTaken = new AsyncLocal<bool>(LockTakenValueChanged);
-            _lockTaken.Value = false;
         }
 
-        private void LockTakenValueChanged(AsyncLocalValueChangedArgs<bool> obj)
+        private static void LockTakenValueChanged(AsyncLocalValueChangedArgs<object> obj)
         {
             // Without this fixup, when completing the call to await TakeLockAsync there is
-            // a switch of Context and _localTaken will be reset to false. This is because
+            // a switch of Context and _heldLocks will be reset to null. This is because
```
stephentoub, line 33:
```suggestion
            // a switch of Context and s_heldLocks will be reset to null. This is because
```

src/System.Private.ServiceModel/src/Internals/System/Runtime/AsyncLock.cs
```diff
@@ -18,101 +19,130 @@ internal class AsyncLock
...
             // of leaving the task.
 
             if (obj.ThreadContextChanged)
             {
-                _lockTaken.Value = obj.PreviousValue;
+                s_heldLocks.Value = obj.PreviousValue;
             }
```
stephentoub, line 39: I'm surprised this handler doesn't lead to other problems, like a lock showing as being held even when it's not.  If I'm understanding this handler correctly, it's basically undoing the undo?  That would suggest if you had code like:
```C#
public async Foo()
{
    await TakeLock();
    await Something();
    ReleaseLock();
}
...
Task t = Foo();
// is lock held here?
```
that if TakeLock completed synchronously and then Something completed asynchronously, the code after `Task t = Foo()` would see the lock as being held?

src/System.Private.ServiceModel/src/Internals/System/Runtime/AsyncLock.cs
```diff
@@ -135,7 +165,23 @@ public void Dispose()
...
+                {
+                    s_heldLocks.Value = null;
+                }
+                else if (s_heldLocks.Value is List<AsyncLock> listOfLocks)
+                {
+                    Fx.Assert(listOfLocks.Contains(_asyncLock), "The list of AsyncLock's didn't contain the expected lock");
+                    // As locks are expected to be released in the order they are taken and they are always appended to the end,
+                    // removal should be O(n) simply to look for the lock and removal should be constant time. If this becomes
+                    // a significant overhead, then manual search in reverse will fix it. Keeping simple for now.
+                    listOfLocks.Remove(_asyncLock);
```
stephentoub, line 178: What makes this thread-safe?  If I start multiple async operations from the same context, won't they all end up potentially operating on the same list, and potentially trying to remove from the same list in parallel?

mconnew:
This PR also fixes a couple of minor issues discovered while debugging the AsyncLock where I found the ExecutionContext was being held on to by the Timer based IOTimer (which IOThreadTimer doesn't do).

mconnew:
@stephentoub, can you take a look at my new version?

src/System.Private.ServiceModel/src/Internals/System/Runtime/AsyncLock.cs
```diff
@@ -4,139 +4,146 @@
...
-#if DEBUG
-            _lockTakenCallStack = new StackTrace();
-            _lockTakenCallStackString = _lockTakenCallStack.ToString();
-#endif
-            return _semaphoreRelease;
+            if (_isDisposed)
+                return;
+
+            _isDisposed = true;
+            // Ensure the lock isn't held. If it is, wait for it to be release
```
stephentoub, line 64: Nit:
```suggestion
            // Ensure the lock isn't held. If it is, wait for it to be released
```

src/System.Private.ServiceModel/src/Internals/System/Runtime/AsyncLock.cs
```diff
@@ -4,139 +4,146 @@
...
 using System.Threading.Tasks;
-#if DEBUG
-using System.Diagnostics;
-#endif
+using Microsoft.Extensions.ObjectPool;
 
 namespace System.Runtime
 {
-    internal class AsyncLock
+    internal class AsyncLock : IAsyncDisposable
```
stephentoub, line 11: It feels like this deserves an extensive set of tests, and I don't see any in this PR.  What you have here looks strictly more correct than what you previously had, though I'd feel better about it if there were tests validating all the scenarios required by its internal usage.

stephentoub (approved):

src/System.Private.ServiceModel/src/Internals/System/Runtime/AsyncLock.cs
mconnew: The scenario required by it's internal usage is reliable sessions, and we have pretty extensive tests around that functionality going so far as to intercept the underlying HTTP requests to mess around with the timing and/or delivery of messages to trigger all the behaviors that ReliableSessions provides. The original implementation worked from a lock standpoint, but leaked memory.

PR #4536, opened 2021-02-11
Documentation: Add line for WebHttpBinding

yowl:
Its not supported so even the red xs are informative.


HongGit:
Thanks @yowl  for catching this!

PR #4546, opened 2021-02-24
Update WCF-Web-Service-Reference-notes.md

PatrickMcDonald:
WCF Web Service Reference connected services provider is also part of Visual Studio 2019


mconnew:
Thank you.

Joebeazelman:
The link is broken yet it says it's labeled "verified."  WCF Web Services doesn't appear in the connected service screen with all the other services. It can only be added via right-clicking on the project file and selecting "Add Service Reference".  It's not fixed!

PR #4567, opened 2021-03-17
Support targetframewrok net5.0-windows.

angewa:
Support targetframewrok net5.0-windows.

### DIFF
--- a/src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
+++ b/src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
@@ -159,7 +159,7 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
                     var targetFramework = targetFrameworkElements.Last().Value.Trim();
                     if (!string.IsNullOrWhiteSpace(targetFramework))
                     {
-                        if ("net5.0-windows" == targetFramework)
+                        if (targetFramework.StartsWith("net5.0-"))
                         {
                             targetFramework = "net5.0";
                         }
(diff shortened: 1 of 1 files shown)


angewa:
@HongGit Please review this PR.

src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
```diff
@@ -159,6 +160,13 @@ public static async Task<MSBuildProj> ParseAsync(string projectText, string proj
                     var targetFramework = targetFrameworkElements.Last().Value.Trim();
                     if (!string.IsNullOrWhiteSpace(targetFramework))
                     {
+                        Regex re = new Regex(@"\w+\d+\.\d+");
```
HongGit, line 163: What is the reason of using regex here?  What about the approach with just comparing with "net5.0-windows*" approach?

HongGit (commented):
Regex approach seems to be too broad, not narrow enough, thoughts?

src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
angewa: The possible target framework format examples are: netcoreapp3.1, net5.0, net5.0-windows, the regex here is to truncate the extra characters after version number, so for both net5.0 and net5.0-windows, or other target framework format like "net5.0***" appear in the furture, we will get net5.0 for the internal starter project to generate service proxy code.

angewa: > What is the reason of using regex here? What about the approach with just comparing with "net5.0-windows*" approach?

I have removed regex code and use "net5.0-windows" approach, could you please help review the update?

angewa: I've modified the code according to the meeting, could you please help review the update?

HongGit (approved):

angewa:
@HongGit I have rebased the latest main code, the check now all passed.

Issue #4589, opened 2021-04-15
CreateCertificateOverTransportBindingElement


jason-russell:
**Describe the bug**
`CreateCertificateOverTransportBindingElement` is not available for `SecurityBindingElement` in `CustomBinding`

.NET Framework code is using this to create a Custom Binding, and I'm trying to port to .NET Core 3.1
I'm trying to connect to a third-party service that requires a certificate binding

**To Reproduce**

This is the .NET Framework code. I'm trying to recreate the custom binding

    private CustomBinding CreateCustomBinding()
	{
		var dvsBinding = new CustomBinding();

		var securityElement = SecurityBindingElement.CreateCertificateOverTransportBindingElement();
		securityElement.MessageSecurityVersion = MessageSecurityVersion.WSSecurity11WSTrustFebruary2005WSSecureConversationFebruary2005WSSecurityPolicy11BasicSecurityProfile10;
		securityElement.SecurityHeaderLayout = SecurityHeaderLayout.Lax;
		securityElement.SetKeyDerivation(true);                     // require derived keys
		securityElement.LocalClientSettings.DetectReplays = false;
		securityElement.LocalServiceSettings.DetectReplays = false;
		securityElement.DefaultAlgorithmSuite = SecurityAlgorithmSuite.Default;

		var textElement = new TextMessageEncodingBindingElement(MessageVersion.Soap12, Encoding.UTF8); // TODO: Verify Message Version
		var httpsElement = new HttpsTransportBindingElement();

		dvsBinding.Elements.AddRange(securityElement, textElement, httpsElement); // Transport Binding Element must occur last (in order)

		return dvsBinding;
	}

This custom binding was derived from the following app.config

        <binding name="Https2">
          <security defaultAlgorithmSuite="Basic256Sha256Rsa15" authenticationMode="CertificateOverTransport"
                requireDerivedKeys="true" securityHeaderLayout="Lax" includeTimestamp="true"
                messageSecurityVersion="WSSecurity10WSTrustFebruary2005WSSecureConversationFebruary2005WSSecurityPolicy11BasicSecurityProfile10">
            <localClientSettings detectReplays="false" />
            <localServiceSettings detectReplays="false" />
          </security>
          <textMessageEncoding />
          <httpsTransport />
        </binding>

**Expected behavior**
To be able to create a Certificate Over Transport binding

**Screenshots**

**Additional context**

HongGit:
Thanks for reporting this, @jason-russell !

[...]

PR #4617, opened 2021-05-06
Add api to SecurityAlgorithmSuite.

imcarolwang:
For #4589 .


mconnew:
You didn't add the implementation of SecurityAlgorithmSuite.Basic256Sha256 in System.Private.ServiceModel. It also needs to be public and not internal in the contract interface.

tapizquent:
Are these changes available in version 4.9.0-preview1.21220.1?

HongGit:
@tapizquent no, it's not.  It will be the next update.

Issue #4237, opened 2020-04-14
Add net6.0 TFM to packages


mconnew:
Today our packages suppress our implementation for xamarin based applications. The next version of the xamarin platform will be based on .NET 5. The previous version of xamarin was based on mono and carried it's own implementation of WCF which is why we suppressed our assets. Based on the design [here](https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md#precedences) we need to add a net5.0 TFM to our packages so that newer xamarin apps can use our implementation.

HongGit:
push out to .NET 6.0, move this to .NET 6.0.

akoeplinger:
@HongGit just checking whether this is still planned for .NET6? We need it for Xamarin scenarios, I'm happy to send a PR if it's just a matter of adding the net6.0 TFM to the projects.

mconnew:
@akoeplinger, the delay has been in finding bandwidth to test it. Adding net6.0 is simple to do, but when the reason is to enable Xamarin usage, we should test the generated packages with a Xamarin app to make sure it is picked up and used at runtime. If you can do that validation, then I'm happy to accept a PR adding net6.0. I'm just not comfortable adding something blindly without testing it in the intended scenario it's addressing.

akoeplinger:
@mconnew ok, I think we can do the testing/validation on our side.

akoeplinger:
I forced using the netcoreapp2.1 assets in the System.ServiceModel.Primitives 4.8.1 package from nuget into a sample .NET 6 Xamarin Android app and was able to call a SOAP service with a WCF client generated from the WSDL by `dotnet-svcutil` and it worked so we know at least the basics are working :)

@mconnew I'll look at adding net6.0 to the .csproj's and test with that too.

PR #4640, opened 2021-06-23
Add .NET 6.0 TFM to packages

akoeplinger:
Fixes https://github.com/dotnet/wcf/issues/4237

See https://github.com/dotnet/designs/pull/222#issuecomment-861617052 for context.

@mconnew I verified these packages restore using the `net6.0` assets in the HelloAndroid sample from https://github.com/dotnet/maui-samples and at least a basic call to a SOAP webservice works.

### DIFF
--- a/src/System.ServiceModel.Duplex/ref/System.ServiceModel.Duplex.Ref.csproj
+++ b/src/System.ServiceModel.Duplex/ref/System.ServiceModel.Duplex.Ref.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <AssemblyVersion>$(WcfAssemblyVersion)</AssemblyVersion>
     <StrongNameKeyId>Microsoft</StrongNameKeyId>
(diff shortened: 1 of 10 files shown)


mconnew:
@ericstj, can you comment on whether we need to add net6.0 to the Ref projects too? This work is to make the latest Xamarin platform which is based on .NET 6.0 (as opposed to mono) to use the WCF runtime assets from our packages instead of the previous Xamarin in-box mono based implementation.

akoeplinger:
Yes we do need them in the Ref projects too since the packages contain `ref/MonoAndroid10/_._` right now as well, so we need the same explicit net6.0 asset as for `lib`.

akoeplinger:
@mconnew this should be good to go now, Eric gave his thumbs up :)

ericstj:
Yes, to state explicitly: please make parallel changes to TFM additions in ref/src.  You need both here.

mconnew (approved):

PR #4688, opened 2021-08-06
Update the version of  packages

ZhaodongTian:
Update the version of packages

### DIFF
--- a/src/dotnet-svcutil/lib/src/DebugLogger.cs
+++ b/src/dotnet-svcutil/lib/src/DebugLogger.cs
@@ -35,6 +35,30 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
             }
         }
 
+        private void LogMessage(string message, NuGet.Common.LogLevel level, bool logToUI = false)
+        {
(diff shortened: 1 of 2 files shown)


src/dotnet-svcutil/lib/src/DebugLogger.cs
```diff
@@ -101,10 +125,31 @@ void NuGet.Common.ILogger.LogInformationSummary(string data)
             LogMessage(data, LogTag.LogMessage);
         }
 
+        void NuGet.Common.ILogger.Log(NuGet.Common.ILogMessage message)
```
HongGit, line 128: Why is this change needed for updating the Newtonsoft packages?

ZhaodongTian: > Why is this change needed for updating the Newtonsoft packages?

The interface ILogger in Higher version of NuGet.Common package has defined more methods, we should  have the implementation in the inherited class

HongGit (approved):

PR #4691, opened 2021-08-12
Allow dotnet-svcutil referenced package name end with dll.

### LINKED ISSUE #4126: Attempting to generate a svc client fails if the project references a nuget package with a name ending in .dll
**Describe the bug**
Attempting to generate a SOAP client via `svcutil` fails if the project references a nuget package with a name ending in `.dll`

**To Reproduce**

Install a package like `DocuSign.eSign.dll` and then attempt to generate a client for this project. You will receive the following error:

`Error: The specified parameter value 'DocuSign.eSign.dll' is not valid. Parameter name: 'packageName'.`

**Expected behavior**

Generate a svc client.

**Screenshots**

image

imcarolwang:
Fixes #4126.

### DIFF
--- a/src/dotnet-svcutil/lib/src/Shared/ProjectDependency.cs
+++ b/src/dotnet-svcutil/lib/src/Shared/ProjectDependency.cs
@@ -112,12 +112,6 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
                 // even when it is a Pacakge type, it may contain a valid binary, let's check.
                 fileHasKnownExtension = string.IsNullOrWhiteSpace(filePath) ? false :
                      (s_binaryExtensions.Any((ext) => String.Compare(Path.GetExtension(filePath), ext, StringComparison.OrdinalIgnoreCase) == 0));
-
-                // check that package name is not confused with assembly name.
-                if (s_binaryExtensions.Any((ext) => String.Compare(Path.GetExtension(packageName), ext, StringComparison.OrdinalIgnoreCase) == 0))
-                {
-                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Shared.Resources.ErrorInvalidDependencyValue, packageName, nameof(packageName)));
-                }
             }
             else // Binary or Project.
             {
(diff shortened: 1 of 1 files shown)


HongGit:
@imcarolwang will this change allow extensions outside of .exe or .dll?  What happens if there is no extension or extension like .txt?  Should those scenarios be allowed at all?

imcarolwang:
> @imcarolwang will this change allow extensions outside of .exe or .dll? What happens if there is no extension or extension like .txt? Should those scenarios be allowed at all?

Not allowing package name end with ".exe" or ".dll" are explicitly specified before the change.
This change remove that part of code so it won't throw if package name end with .exe or .dll.

As for the other extensions,  the original code doesn't call out it's forbidden so it should be allowed if package name end with .txt or others.

src/dotnet-svcutil/lib/src/Shared/ProjectDependency.cs
```diff
@@ -28,6 +28,7 @@ internal class ProjectDependency : IComparable<ProjectDependency>
 
         private static readonly IReadOnlyList<string> s_binaryExtensions = new List<string> { ".exe", ".dll" }.AsReadOnly();
```
HongGit, line 29: is this private member still needed?  or is it possible to modify this with .exe only?

src/dotnet-svcutil/lib/src/Shared/ProjectDependency.cs
```diff
@@ -28,6 +28,7 @@ internal class ProjectDependency : IComparable<ProjectDependency>
...
         private static readonly IReadOnlyList<string> s_projectExtensions = new List<string> { ".csproj" }.AsReadOnly();
+        private static readonly IReadOnlyList<string> s_exeExtensions = new List<string> { ".exe" }.AsReadOnly();
 
         /// <summary>
         /// The dependency name.  This can be a package or assembly name.
```
HongGit, line 34: similar to comments above, is this new variable really needed?

src/dotnet-svcutil/lib/src/Shared/ProjectDependency.cs
imcarolwang: Yes, it's being used in other code path, e.g. line115, L133, L315.

src/dotnet-svcutil/lib/src/Shared/ProjectDependency.cs
imcarolwang: In our case, this new variable **s_exeExtensions** defines EXE extension names we want to disallow as a package name ending, base on our discussion, **.exe** may have some side-effect so we still disallow it, and we exclude **.dll** extension which would be allowed for a package name end with it to address issue #4126. I think the new variable is necessary to differentiate with **s_binaryExtensions** which include both **.exe**  and **.dll.**

HongGit (approved):

Issue #4186, opened 2020-03-03
Channel Factory caching


pmiddleton:
I'm trying to figure out what the best approach is for caching with the new wcf library now that ClientBase no longer caches.

I have a middleware asp.net core web service which in turn needs to call out to a 3rd party soap service using wcf.  The 3rd party service is at a single fixed url https endpoint.

My understanding is that everything in the wcf layer is threadsafe - is this correct?  If so am I better off caching the generated clientBase derived object, a channel factory, or a channel proxy?

Calls to the 3rd party service will be sporadic so will I be facing any timeout issues with any of the 3 options?  Is it recommended to cache local to where I need to make the calls, or use DI on the channel factor or proxy?

Due to the size of the 3rd party wsdl generating a new clientBase each call is not an option.  The creating time is on the order of 500ms.

StingyJack:
Your title captures the answer - cache the factory.  Just be aware that there isnt a reliable Dispose() on the channels that ChannelFactory<T> creates and it could be leaky. IMO not really worth it, sporadic (subjective) or not.

Some of the word choices are giving me the impression that you are generating code from wsdl while the program is executing and then executing that. Is that what you mean, or are you using "generate" in the same way as you would use "construct an object" or "create a new instance of " and svcutil or some other tool already generated the code thats compiled and that the runtime is using to make new instances ?

pmiddleton:
sorry bad use of the word generating in that last sentence, should had said instantiating.  

The derived clientbase is part of the code generated by svcutil.  The generated reference.cs file, that svcutil produces, is on the order of ~700,000 lines of code.  Instantiating a new clientbase on each call was spending ~600-800ms creating a new channel factory.

I need to cache something because that cost is too high for our use case.

StingyJack:
That sounds like a web service that has too many functions. I would probably break up the code generated by the wsdl using a script or similar, starting with removing parts you don't use.

mconnew:
[...]

PR #4696, opened 2021-08-20
Add support for ClientBase<T>.CacheSetting

mconnew:
This will help with #4186. We should do some follow-up work to have dotnet-svcutil emit the code to turn on caching as the only constructor on ClientBase\<TChannel\> that has caching turned on by default is the configuration based constructors which we don't support.

### DIFF
--- a/src/System.Private.ServiceModel/src/Internals/WcfEventSource.cs
+++ b/src/System.Private.ServiceModel/src/Internals/WcfEventSource.cs
@@ -9,6 +9,10 @@ namespace System.Runtime
 {
     internal sealed partial class WcfEventSource : EventSource
     {
+        // All events from NetFx can be found here:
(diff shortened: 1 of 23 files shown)


mconnew:
I've added some additional caching so that there's a benefit even when you don't use the ClientBase\<TChannel\>.CacheSettings property. I added caching with a weak reference in TypeLoader so that repeated calls to `ContractDescription.GetContract(Type contractType)` will return the same instance. This means repeatedly instantiating a ChannelFactory for the same interface contract will return the same ContractDescription instance. There is a potential issue where if someone modifies the cached ContractDescription, they get the same modified instance returned instead of a fresh instance. To mitigate this, I made it so that only ChannelFactory uses the cached mechanism. ContractDescription.GetContract will continue to return a new instance each time. If someone is going to modify the ContractDescription, this is the way I would expect them to do so.

HongGit (approved):

Issue #4412, opened 2020-10-22
Error: The method or operation is not implemented


diegonc:
I'm getting the error in the subject when importing a wsdl file. No code is generated and the wizard ends with "no .net core compatible endpoint found" (sorry, the error is in Spanish and I don't know the exact message)

I have no idea what in the wsdl file might be wrong. May I have someone give a look at it?

[Autenticacion.zip](https://github.com/dotnet/wcf/files/5424761/Autenticacion.zip)

HongGit:
@imcarolwang can you please see if you could repro this?

imcarolwang:
Hi @diegonc , I tried running dotnet-svcutil against the attached wsdl file and I cannot reproduce the Error as you posted in the issue title "The method or operation is not implemented", I got following error instead:

```
Warning: Cannot import wsdl:portType
Detail: An exception was thrown while running a WSDL import extension: System.ServiceModel.Description.DataContractSerializerMessageContractImporter
Error: Schema with target namespace 'http://tempuri.org/' could not be found.
```

It's due to the referenced xsd files (`Autenticacion.xsd` and `Autenticacion1.xsd`) in the wsdl file are relative local paths which is unavailable on my repro environment, could you provide the dependency xsd files as well?

diegonc:
Sorry for the late response, work drifted away of this project. Here are the XSDs files imported by the WSDL. Thanks

[xsds.zip](https://github.com/dotnet/wcf/files/6086833/xsds.zip)

HongGit:
Close this issue as this is resolved.  Please open a new issue if there are more problems.

Issue #4498, opened 2021-01-14
Unsure How to Proceed With Error: The method or operation is not implemented.


MrTovson:
I've spent a few days researching why I can't work with a vendor's service in .NET Core 3.1.  The service works fine in .NET 4.5.2.

In the end, only 1 of the 2 bindings are generated: WS binding.  In 4.5.2 we are successfully using Basic, which is not generated.

I suppose my question is two-fold:
1. What isn't supported by this relatively simple SOAP service
2. Are there any known workarounds?

I have tried many, have build endpoint+binding dynamically as many ways as I can and am stuck.  

Detail:

PM> dotnet-svcutil https://servicesv4.qa.provexchnet.com/DSIRating.svc -d C:\Temp
Microsoft (R) WCF Service Model Proxy Generation Tool for .Net Core platform
[Microsoft.Tools.ServiceModel.Svcutil, Version 2.0.2]
...
Attempting to download metadata from 'https://servicesv4.qa.provexchnet.com/DSIRating.svc' using WS-Metadata Exchange and HttpGet.
Warning: Cannot import wsdl:binding
Detail: An exception was thrown while running a WSDL import extension: System.ServiceModel.Channels.StandardBindingImporter
Error: The method or operation is not implemented.
XPath to Error Source: //wsdl:definitions[@targetNamespace='https://www.provexchnet.com/dsi']/wsdl:binding[@name='BasicHttpBinding_IDSIRating']
Warning: Cannot import wsdl:port
Detail: 
XPath to Error Source: //wsdl:definitions[@targetNamespace='https://www.provexchnet.com/dsi']/wsdl:service[@name='DSIRating']/wsdl:port[@name='BasicHttpBinding_IDSIRating']
Generating files...
C:\Temp\Reference.cs

MrTovson:
'No connection could be made because the target machine actively refused it.'

Quite tired of seeing this message, regardless of anything I do.  I'm going back to .NET Framework until this service is supported, the service is upgraded by the vendor, or someone who knows what they're doing can help.

imcarolwang:
We tried to repro the issue with the given service https://servicesv4.qa.provexchnet.com/DSIRating.svc  mentioned in issue, we got accessed denied as we don't have a valid certificate.  @MrTovson could you please provide a simplified repro project for us to investigate this further?

MrTovson:
You've got to be kidding.  There is nothing simpler than this.

[...]

Issue #3969, opened 2019-10-16
Support TransportWithMessageCredential in dotnet-svcutil


johnpenmanweb:
Not sure if this is related to [#8](https://github.com/dotnet/wcf/issues/8) or not but I am getting an issue when trying to add wsdl's from SAP.  Wsdl works in .Net Framework 4.6 but fails in .net core 2.2. 
 Below is the verbose of dotnet-svcutil:

Warning: The optional WSDL extension element 'Policy' from namespace 'http://schemas.xmlsoap.org/ws/2004/09/policy' was not handled.
XPath: //wsdl:definitions[@targetNamespace='urn:sap-com:document:sap:soap:functions:mc-style']/wsdl:portType[@name='Z_RFC_VALIDATE_ADDRESS']
Warning: The optional WSDL extension element 'Policy' from namespace 'http://schemas.xmlsoap.org/ws/2004/09/policy' was not handled.
XPath: //wsdl:definitions[@targetNamespace='urn:sap-com:document:sap:soap:functions:mc-style']/wsdl:portType[@name='Z_RFC_VALIDATE_ADDRESS']/wsdl:operation[@name='ZRfcValidateAddress']
Warning: Cannot import wsdl:binding
Detail: An exception was thrown in a call to a policy import extension.
Extension: System.ServiceModel.Channels.SecurityBindingElementImporter
Error: The method or operation is not implemented.
XPath to Error Source: //wsdl:definitions[@targetNamespace='urn:sap-com:document:sap:soap:functions:mc-style']/wsdl:binding[@name='Z_RFC_VALIDATE_ADDRESS']
Warning: Cannot import wsdl:port
Detail:
XPath to Error Source: //wsdl:definitions[@targetNamespace='urn:sap-com:document:sap:soap:functions:mc-style']/wsdl:service[@name='Z_RFC_VALIDATE_ADDRESS']/wsdl:port[@name='Z_RFC_VALIDATE_ADDRESS']
Warning: No endpoints compatible with .Net Core apps were found.
[28032.222] Time elapsed: 00:00:21.1325156
[28032.355] Processing Code DOM ...
Generating files...

StephenBonikowsky:
@johnpenmanweb We believe this is a known issue, @dasetser has the details, Daniel could you provide the information regarding this issue?

dasetser:
[...]

PR #4698, opened 2021-08-26
Support Transport and TransportWithMessageCredential security mode

ZhaodongTian:
Support TransportWithMessageCredential security mode : #3969 
Fixed issues: #4412 and #4498


HongGit:
@ZhaodongTian can you please describe this change so we understand why these changes are needed for dotnet-svcutil to support TransportWithMessageCredential feature.

src/dotnet-svcutil/lib/src/FrameworkFork/System.ServiceModel/System/ServiceModel/FederatedMessageSecurityOverHttp.cs
```diff
@@ -139,7 +139,93 @@ public Collection<XmlElement> TokenRequestParameters
 
         internal static bool TryCreate(SecurityBindingElement sbe, bool isSecureTransportMode, bool isReliableSession, MessageSecurityVersion version, out FederatedMessageSecurityOverHttp messageSecurity)
         {
-            throw new NotImplementedException();
+            Fx.Assert(null != sbe, string.Empty);
```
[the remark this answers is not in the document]
ZhaodongTian: Any bindings with Security mode will have to call [WSFederationHttpBinding.TryCreate](https://github.com/dotnet/wcf/blob/9bf1eea452c8bfad9a4f833097ac3d4c093d987e/src/dotnet-svcutil/lib/src/FrameworkFork/System.ServiceModel/System/ServiceModel/WSHttpBindingBase.cs#L336) method . The implementation for this is forked from framework. Do you think we can update code here to remove Federation related?

ZhaodongTian:
> Does anything need to be done to support WSHttpBinding and WS2007HttpBinding? I'm not seeing tests added for these scenarios either. I believe the existing tests download the WSDL from some of the WCF test endpoints to validate.

I have updated code to support them, i will summit the changes together with test case

HongGit:
@ZhaodongTian or @imcarolwang could you please rebase this PR?

src/dotnet-svcutil/lib/src/FrameworkFork/System.ServiceModel/System/ServiceModel/FederatedMessageSecurityOverHttp.cs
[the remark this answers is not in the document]
mconnew: Yes, remove the call to WSFederationHttpBinding.TryCreate

mconnew:
Other than needing to remove call to the federation binding to create that from the WSDL,

mconnew:
@HongGit, I realized there's another problem caused by the recent security tightening of our test services. The tests for svcutil all hit wcfcoresvr5 which means they can't be run from a remote dev machine any more. We should modify these tests to use localhost unless there's a ServiceUri set, like the WCF tests do.

src/dotnet-svcutil/lib/src/CodeDomFixup/EndpointSelector.cs
```diff
@@ -62,6 +62,10 @@ private static bool IsBindingSupported(Binding binding)
                     {
                         s_bindingValidationErrors.Add(string.Format(SR.BindingMessageEncodingNotSupportedFormat, wsHttpBinding.MessageEncoding, WSMessageEncoding.Text));
                     }
+                    if (wsHttpBinding.Security.Mode == SecurityMode.Message)
+                    {
```
mconnew, line 66: We don't have a second for it, but we also need to do this check for BasicHttpBinding, although there the mode type is BasicHttpSecurityMode. You need to remove the check for MessageEncoding just above this code as we now support that.

src/dotnet-svcutil/lib/src/Shared/TargetFrameworkHelper.cs
```diff
@@ -63,6 +63,7 @@ internal class TargetFrameworkHelper
                 ProjectDependency.FromPackage("System.ServiceModel.Http", "4.8.*"    ),
                 ProjectDependency.FromPackage("System.ServiceModel.NetTcp", "4.8.*"  ),
                 ProjectDependency.FromPackage("System.ServiceModel.Security", "4.8.*"),
+                ProjectDependency.FromPackage("System.ServiceModel.Federation", "4.8.*"),
             } }
```
mconnew, line 67: We don't support Federation with svcutil yet so no need to add that package.

src/dotnet-svcutil/lib/src/CodeDomFixup/EndpointSelector.cs
HongGit: @ZhaodongTian  chatted with Matt, it's okay we are not supporting ReliableSession, and remove ling 61 to 64; and add check for BasicHttpBinding security mode as Matt suggested.

Thanks!

src/dotnet-svcutil/lib/src/Shared/TargetFrameworkHelper.cs
ZhaodongTian: I have removed that package

src/dotnet-svcutil/lib/src/CodeDomFixup/EndpointSelector.cs
ZhaodongTian: Done, I have add the code for checking BasicHttpBinding security mode

HongGit (approved):

PR #4700, opened 2021-08-27
Test case will fail on Linux with SelfHost service mode

ZhaodongTian:
The value for ServiceUri_PropertyName with IIS host  contains serivce name

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.4.1.0.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.4.1.0.cs
@@ -74,24 +74,40 @@ public class NegotiateStream_Http_Tests : ConditionalWcfTest
             BasicHttpBinding binding = new BasicHttpBinding(BasicHttpSecurityMode.Transport);
             binding.Security.Transport.ClientCredentialType = HttpClientCredentialType.Windows;
             string spn = GetSPN().ToLowerInvariant().Replace("host", "HTTP");
-
-#if NETCOREAPP3_1
             factory = new ChannelFactory<IWcfService>(
                 binding,
                 new EndpointAddress(Endpoints.Https_WindowsAuth_Address));
-#else
-            factory = new ChannelFactory<IWcfService>(
-                binding,
-                new EndpointAddress(new Uri(Endpoints.Https_WindowsAuth_Address), new SpnEndpointIdentity(spn)));
-#endif
-            serviceProxy = factory.CreateChannel();
 
+            serviceProxy = factory.CreateChannel();
+#if NET5_0
+            if (TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName).Contains("/")) //iis-host service uri
+            {
+                // *** EXECUTE *** \\
+                string result = serviceProxy.Echo(testString);
+
+                // *** VALIDATE *** \\
+                Assert.Equal(testString, result);
+            }
+            else //selfhost uri
+            {
+                try
+                {
+                    // *** EXECUTE *** \\
+                    string result = serviceProxy.Echo(testString);
+                    Assert.Null(result); //known issue: on net5.0 with non-standard port will throw excepiton
+                }
+                catch (System.ServiceModel.ProtocolException ex)
+                {
+                    Assert.Contains("(400) Bad Request", ex.Message);
+                }
+            }
+#else
             // *** EXECUTE *** \\
             string result = serviceProxy.Echo(testString);
 
             // *** VALIDATE *** \\
             Assert.Equal(testString, result);
-
+#endif
             // *** CLEANUP *** \\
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.4.1.0.cs
```diff
@@ -112,11 +119,11 @@ public static void NegotiateStream_Http_AmbientCredentialsForNet50()
             // *** SETUP *** \\
             BasicHttpBinding binding = new BasicHttpBinding(BasicHttpSecurityMode.Transport);
             binding.Security.Transport.ClientCredentialType = HttpClientCredentialType.Windows;
-            string host = TestProperties.GetProperty(TestProperties.ServiceUri_PropertyName);
+            string spn = GetSPN().ToLowerInvariant().Replace("host", "HTTP");
 
             factory = new ChannelFactory<IWcfService>(
                 binding,
-                new EndpointAddress(new Uri(Endpoints.Https_WindowsAuth_Address), new SpnEndpointIdentity($"HTTP/{host}")));
+                new EndpointAddress(new Uri(Endpoints.Https_WindowsAuth_Address), new SpnEndpointIdentity(spn)));
```
HongGit, line 126: Does this part of code always execute for net5.0?  Why not if 3.1 or else 5.0?

ZhaodongTian: > Does this part of code always execute for net5.0? Why not if 3.1 or else 5.0?

Under 3.1, test case used the default SPN,  5.0 and 6.0 used the new created SPN

afifi-ins: Could we use the if condition for targeting .NET 5.0 only? The reason is .NET 6.0 will be following .NET 3.1 behavior, which doesn't use the port number by default. I validated the test against a .NET 6.0 nightly build and it works fine (similar to .NET 3.1). 
Also, there will be back porting to .NET 5.0 as a servicing fix. Here’s the link for reference: https://github.com/dotnet/runtime/issues/51701#issuecomment-902636271

afifi-ins: Another thing just FYI, .NET 6.0 introduced a new environment variable switch to control the SPN port behavior, System.Net.Http.UsePortInSpn or DOTNET_SYSTEM_NET_HTTP_USEPORTINSPN, which would allow flexible switch between .NET 5.0 and .NET 6.0 behavior. We might consider adding a test for this switch as well.

ZhaodongTian: The sdk version in [global.json](https://github.com/dotnet/wcf/blob/8e18e27cd563d0bf9f084c671f8e8ce7374eb30d/global.json#L3) is lower , if we changed it follow afifi's suggestion, it will be failed , @HongGit any plan to update global.json?

HongGit: I agree with @afifi-ins to add a case for the config switch.  
Per my understanding, global.json  is used at build time, the fix is for runtime, correct?

HongGit: I think the other benefit to fix this as suggested will make it easier to read for the future.  Please add more comments with explanation regarding the intend for maintainability!

HongGit (changes requested):

src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.4.1.0.cs
```diff
@@ -73,10 +73,17 @@ public static void NegotiateStream_Http_AmbientCredentials()
...
 
+#if NETCOREAPP3_1
             factory = new ChannelFactory<IWcfService>(
                 binding,
                 new EndpointAddress(Endpoints.Https_WindowsAuth_Address));
+#else
+            factory = new ChannelFactory<IWcfService>(
+                binding,
+                new EndpointAddress(new Uri(Endpoints.Https_WindowsAuth_Address), new SpnEndpointIdentity(spn)));
+#endif
```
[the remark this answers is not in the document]
mconnew: It's only an issue if running on a non-standard port.

HongGit:
@ZhaodongTian this PR probably also need to be rebased.  Please rebase the PR, and re-submit appropriately!

PR #4703, opened 2021-08-30
Add more apis for generic xml security token

mconnew:
Adding api's which are missing from the contract for GenericXmlSecurityToken and GenericXmlSecurityKeyIdentifierClause. This is needed to support the WSTrustChannelFactory usage case where customers will need to access the returned security token's TokenXml property in some cases.

### DIFF
--- a/src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs
+++ b/src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs
@@ -234,10 +234,19 @@ namespace System.IdentityModel.Tokens
             SecurityKeyIdentifierClause internalTokenReference,
             SecurityKeyIdentifierClause externalTokenReference,
             System.Collections.ObjectModel.ReadOnlyCollection<System.IdentityModel.Policy.IAuthorizationPolicy> authorizationPolicies) {}
-        public override string Id { get; }
-        public override DateTime ValidFrom { get; }
-        public override DateTime ValidTo { get; }
-        public override System.Collections.ObjectModel.ReadOnlyCollection<System.IdentityModel.Tokens.SecurityKey> SecurityKeys { get; }
+        public override string Id => default;
+        public override DateTime ValidFrom => default;
+        public override DateTime ValidTo => default;
+        public SecurityKeyIdentifierClause InternalTokenReference => default;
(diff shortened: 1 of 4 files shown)


src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs
```diff
@@ -234,10 +234,19 @@ public partial class GenericXmlSecurityToken : System.IdentityModel.Tokens.Secur
...
+        public override DateTime ValidTo => default;
+        public SecurityKeyIdentifierClause InternalTokenReference => default;
+        public SecurityKeyIdentifierClause ExternalTokenReference => default;
+        public System.Xml.XmlElement TokenXml => default;
+        public SecurityToken ProofToken => default;
+        public System.Collections.ObjectModel.ReadOnlyCollection<System.IdentityModel.Policy.IAuthorizationPolicy> AuthorizationPolicies => default;
+        public override System.Collections.ObjectModel.ReadOnlyCollection<System.IdentityModel.Tokens.SecurityKey> SecurityKeys => default;
+        public override bool CanCreateKeyIdentifierClause<T>() => default;
+        public override T CreateKeyIdentifierClause<T>() => default;
+        public override bool MatchesKeyIdentifierClause(SecurityKeyIdentifierClause keyIdentifierClause) => default;
```
StephenMolloy, line 248: Aren't these overrides just re-defining the same results inherited from SecurityToken?

StephenMolloy (approved):
One question about overriding with the same results... but looks fine otherwise.

src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs
mconnew: Yes, but I think it might do something with regards to what happens when a derived class calls base.MatchesKeyIdentifierClause. When calling a base method, it can't just look up in the vtable for the abstract type and find the method, it's got to look up the method in the actual base class implementation. At least semantically there's a difference. If you didn't know this class also overrode it, you might expect it to call the implementation in the parent class of this. I suspect it would resolve correctly at runtime regardless, but semantically this is more correct.

HongGit (approved):

Issue #4431, opened 2020-11-17
Visual Studio 16.8 breaks SvcUtil build targets


pmiddleton:
We are using the dotnet-avcutil.xmlserilizer package in our solution to build xml serializes for a 3rd party wcf service we consume.  We just upgraded our version of Visual Studio to 16.8 and are now receiving the following two build errors.

Error	MSB4063	The "ResolvePackageAssets" task could not be initialized with its input parameters.

Error	MSB4064	The "TargetFrameworkMoniker" parameter is not supported by the "ResolvePackageAssets" task. Verify the parameter exists on the task, and it is a settable public instance property.	

We verified with that the code builds fine on a machine with 16.7.x and then fails on the same machine once the 16.8.1 update is applied.

Our code is in a .Net Standard 2.0 class library.

The errors is coming from the ResolvePackageAssets task in the dotnet-svcutil.xmlserializer.targets file.

yalin78:
Any luck with this? I am also facing similar issue my project is .net core 2.1

pmiddleton:
Not much luck.  

It turns out the build team renamed the TargetFrameworkMoniker to just TargetFramework.  I tried updating the targets file and that gets me past those two errors.  However now a new error appears.

Error NETSDK1005	Assets file 'c:\......\obj\project.assets.json' doesn't have a target for '.NETStandard,Version=v2.0'. Ensure that restore has run and that you have included '.NETStandard,Version=v2.0' in the TargetFrameworks for your project.		C:\Users\...\.nuget\packages\dotnet-svcutil.xmlserializer\1.2.0\build\dotnet-svcutil.xmlserializer.targets	16	

That's where I got stuck.  I am targeting .Net Standard 2.0 in the project.  The project.assets.json file does have a .netstandard 2 target so I'm not sure what the issue is.
 
The start of that file looks like this.

{
  "version": 3,
  "targets": {
    ".NETStandard,Version=v2.0": {
      "dotnet-svcutil.xmlserializer/1.2.0": {
        "type": "package",
        "build": {
          "build/dotnet-svcutil.xmlserializer.props": {},
          "build/dotnet-svcutil.xmlserializer.targets": {}
        }
      },

I'm not sure what else to try at this point.

yalin78:
Thanks for your help. But I edited target framework as below for xml serializer target as per your suggestion and it started working.

TargetFramewok="netcoreapp2.1" 

and it started working.

ajarrott:
[...]

PR #4707, opened 2021-09-03
update targets file

nckroush:
The properties defined in ResolvePackageAssets task changed between different .NETCore SDK releases #4431


nckroush:
Reviewing...

mconnew:
@nckroush, there's nothing wrong with this PR, there's a test infrastructure problem currently blocking it. I tested it manually and it worked there.

nckroush:
@mconnew I appreciate the feedback. Let me know if there is anything I can do to assist?

HongGit:
@nckroush could you please rebase your PR so it could pick up the latest change, in order for this PR to pass the checkin gate?

Thank you so much for your contribution!

HongGit:
@nckroush when we test PR on machine with .NET Core SDK smaller than .NET 5.0, it seems that the previous behavior is desired.  Could you please look into update this PR with `<MSBuildRuntimeVersion>$([System.Runtime.InteropServices.RuntimeInformation]:: FrameworkDescription.ToString())</MSBuildRuntimeVersion>` property to toggle the behavior between lower SDK and recent SDK version?

nckroush:
That would work much better. I think I can handle that. Thanks @HongGit

src/svcutilcore/files/dotnet-svcutil.xmlserializer.targets
```diff
@@ -1,6 +1,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
- <PropertyGroup>
-   <MyProperty>$(IntermediateOutputPath)</MyProperty>
+  <PropertyGroup>
+    <_IsNet5Compatable>$([MSBuild]::IsTargetFrameworkCompatible('net5.0-windows', 'net5.0'))</_IsNet5Compatable>
```
mconnew, line 3: >IsTargetFrameworkCompatible [](http://example.com/codeflow?start=36&length=27)

This looks like this would always return true as net5.0 is compatible with net5.0-windows. Would I be right to presume it can't answer this question if sdk 5.0 hasn't been installed so will return false in that case?

mconnew:
It looks like it's correct, but I'm going to have someone verify before merging. I'll add an update if there's a problem, otherwise this will be merged once verified.

HongGit:
@nckroush  we will work on a couple of customer reported issues with this tool and see if we could release a new version later this year or early next year.

Thanks for your contribution!

betmix-matt:
There was never a version of this tool released after this fix. Can we get a release please? @HongGit

PR #4753, opened 2021-10-21
Fix non standard port test issue on net5 on Linux OS

imcarolwang:
Continue work of pr #4700.


src/System.Private.ServiceModel/tests/Scenarios/Security/TransportSecurity/Negotiate/NegotiateStream_Http_Tests.4.1.0.cs
```diff
@@ -78,13 +78,27 @@ public static void NegotiateStream_Http_AmbientCredentials()
                 binding,
                 new EndpointAddress(Endpoints.Https_WindowsAuth_Address));
             serviceProxy = factory.CreateChannel();
-
+#if NET5_0
+            // on NET5.0, expect non-standard port (selfhost uri) throw exception on non-windows platform
```
mconnew, line 82: I'm not a big fan of conditional compilation. You can check which version of .NET the test is running on by checking Environment.Version and getting the major version number. Then you can add that check to the if block and not duplicate the .NET 6 test code.

imcarolwang: Thank you @mconnew for the review. As advised test code is updated.

PR #4757, opened 2021-11-01
Adding support for WSTrustChannelFactory and WSTrustChannel

brentschmaltz:
This is similar to what existed in .net framework with some small changes.
WSTrustChannel ctor takes a ChannelFactory rather than a WSTrustChannelFactory.
The new Async model has replaced the Begin / End model found in the original WCF release.
Common methods that were used in WSTrustSecurityTokenProvider and WSTrustChannel were moved to WSTrustUtilities.
An additional constructor was added that takes a separate IRequestChannel.

### DIFF
--- a/src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/IWSTrustChannelContract.cs
+++ b/src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/IWSTrustChannelContract.cs
@@ -1,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IdentityModel.Tokens;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Protocols.WsTrust;
+
+namespace System.ServiceModel.Federation
+{
+    /// <summary>
+    /// A service contract that defines the Issue request.
+    /// </summary>
+    [ServiceContract]
+    public interface IWSTrustChannelContract
+    {
+        /// <summary>
+        /// Sends a WS-Trust Issue request to a STS.
+        /// </summary>
+        /// <param name="request">The <see cref="WsTrustRequest" /> to send to the STS.</param>
+        /// <returns>A <see cref="SecurityToken" /> issued by the STS.</returns>
+        /// <summary>
+        [OperationContract(Name = "Issue", Action = "*", ReplyAction = "*")]
+        Task<SecurityToken> IssueAsync(WsTrustRequest request);
+    }
+}
--- a/src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
+++ b/src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
@@ -28,6 +28,9 @@ namespace System.ServiceModel.Federation
         public WSTrustChannel(IRequestChannel requestChannel)
         {
             RequestChannel = requestChannel ?? throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(requestChannel));
+            if (requestChannel.State != CommunicationState.Created)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(LogHelper.FormatInvariant(SR.GetResourceString(SR.IRequestChannelMustBeCreated), requestChannel.State)));
+
             MessageVersion = RequestChannel.GetProperty<MessageVersion>();
             if (MessageVersion == null || MessageVersion == MessageVersion.None)
                 MessageVersion = MessageVersion.Default;
--- a/src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
+++ b/src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
@@ -1,0 +1,374 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.ServiceModel.Channels;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Protocols;
+using Microsoft.IdentityModel.Protocols.WsTrust;
+using Microsoft.IdentityModel.Tokens;
+using WCFSecurityToken = System.IdentityModel.Tokens.SecurityToken;
+
+namespace System.ServiceModel.Federation
+{
+    /// <summary>
+    /// A channel that is used to send WS-Trust messages to an STS.
+    /// </summary>
+    public class WSTrustChannel : IWSTrustChannelContract, IChannel, ICommunicationObject
+    {
+        private const int FaultMaxBufferSize = 20 * 1024;
+        private WsSerializationContext _wsSerializationContextTrustFeb2005;
+        private WsSerializationContext _wsSerializationContextTrust1_3;
+        private WsSerializationContext _wsSerializationContextTrust1_4;
+
+        /// <summary>
+        /// Constructs a <see cref="WSTrustChannel" />.
+        /// </summary>
+        /// <param name="requestChannel">The <see cref="IRequestChannel" /> this channel will be used to send a <see cref="WsTrustRequest" /> to the STS.</param>
+        public WSTrustChannel(IRequestChannel requestChannel)
+        {
+            RequestChannel = requestChannel ?? throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(requestChannel));
+            MessageVersion = RequestChannel.GetProperty<MessageVersion>();
+            if (MessageVersion == null || MessageVersion == MessageVersion.None)
+                MessageVersion = MessageVersion.Default;
+
+            EndpointAddress endpointAddress = RequestChannel.GetProperty<EndpointAddress>();
+            if (endpointAddress != null)
+                Address = endpointAddress.Uri?.AbsoluteUri;
+        }
+
+        private string Address { get; }
+
+        /// <summary>
+        /// The <see cref="IRequestChannel" /> this class uses for sending and receiving <see cref="Message" /> objects.
+        /// </summary>
+        private IRequestChannel RequestChannel { get; }
+
+        /// <summary>
+        /// Gets the version of WS-Trust this channel will use for serializing <see cref="Message" /> objects.
+        /// </summary>
+        private MessageVersion MessageVersion { get; }
+
+        /// <summary>
+        /// Gets the version of WS-Trust this channel will use for serializing <see cref="Message" /> objects.
+        /// </summary>
+        //private WsSerializationContext SerializationContext
+        //{
+        //    get; set;
+        //}
+
+        ///// <summary>
+        ///// Gets the <see cref="WsTrustSerializer" /> this channel will used for serializing WS-Trust request messages.
+        ///// </summary>
+        private WsTrustSerializer TrustSerializer { get; } = new WsTrustSerializer();
+
+        /// <summary>
+        /// Creates a <see cref="Message"/> that represents a the <see cref="WsTrustRequest"/>.
+        /// </summary>
+        /// <param name="trustRequest">The <see cref="WsTrustRequest"/> to serialize into the message.</param>
+        /// <returns>The <see cref="Message" /> that represents the <see cref="WsTrustRequest"/>.</returns>
+        protected virtual Message CreateRequest(WsTrustRequest trustRequest)
+        {
+            _ = trustRequest ?? throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(trustRequest));
+            return Message.CreateMessage(MessageVersion,
+                                         GetRequestAction(trustRequest),
+                                         new WSTrustRequestBodyWriter(trustRequest, TrustSerializer));
+        }
+
+        /// <summary>
+        /// Gets the WS-Addressing SOAP action that corresponds to the <see cref="WsTrustRequest"/>.RequestType and <see cref="WsTrustRequest"/>.WsTrustVersion.
+        /// </summary>
+        /// <param name="trustRequest">The <see cref="WsTrustRequest"/> to generate the WS-Addressing action.
+        /// <returns>The WS-Addressing action to use.</returns>
+        public static string GetRequestAction(WsTrustRequest trustRequest)
+        {
+            _ = trustRequest ?? throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(trustRequest));
+
+            WsTrustActions wsTrustActions;
+            if (trustRequest.WsTrustVersion == WsTrustVersion.Trust13)
+                wsTrustActions = WsTrustActions.Trust13;
+            else if (trustRequest.WsTrustVersion == WsTrustVersion.TrustFeb2005)
+                wsTrustActions = WsTrustActions.TrustFeb2005;
+            else if (trustRequest.WsTrustVersion == WsTrustVersion.Trust14)
+                wsTrustActions = WsTrustActions.Trust14;
+            else
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new NotSupportedException(SR.GetResourceString(SR.WsTrustVersionNotSupported, trustRequest.WsTrustVersion.ToString())));
+
+            if (trustRequest.RequestType.Equals(wsTrustActions.Issue))
+                return wsTrustActions.IssueRequest;
+            else if (trustRequest.RequestType.Equals(wsTrustActions.Cancel))
+                return wsTrustActions.CancelRequest;
+            else if (trustRequest.RequestType.Equals(wsTrustActions.Renew))
+                return wsTrustActions.RenewRequest;
+            else if (trustRequest.RequestType.Equals(wsTrustActions.Validate))
+                return wsTrustActions.ValidateRequest;
+            else
+               throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new NotSupportedException(SR.GetResourceString(SR.RequestTypeNotSupported, trustRequest.RequestType)));
+        }
+
+        /// <summary>
+        /// Gets the <see cref="WsSerializationContext"/> to use when serializing the <see cref="WsTrustRequest"/>.
+        /// </summary>
+        /// <param name="trustRequest">The <see cref="WsTrustRequest"/> that will be serialized.
+        /// <returns>The <see cref="WsSerializationContext"/> for the <see cref="WsTrustRequest"/>.</returns>
+        private WsSerializationContext GetSerializationContext(WsTrustRequest trustRequest)
+        {
+            _ = trustRequest ?? throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(trustRequest));
+            if (trustRequest.WsTrustVersion == WsTrustVersion.TrustFeb2005)
+            {
+                if (_wsSerializationContextTrustFeb2005 == null)
+                    _wsSerializationContextTrustFeb2005 = new WsSerializationContext(trustRequest.WsTrustVersion);
+
+                return _wsSerializationContextTrustFeb2005;
+            }
+            else if (trustRequest.WsTrustVersion == WsTrustVersion.Trust13)
+            {
+                if (_wsSerializationContextTrust1_3 == null)
+                    _wsSerializationContextTrust1_3 = new WsSerializationContext(trustRequest.WsTrustVersion);
+
+                return _wsSerializationContextTrust1_3;
+            }
+            else if (trustRequest.WsTrustVersion == WsTrustVersion.Trust14)
+            {
+                if (_wsSerializationContextTrust1_4 == null)
+                    _wsSerializationContextTrust1_4 = new WsSerializationContext(trustRequest.WsTrustVersion);
+
+                return _wsSerializationContextTrust1_4;
+            }
+
+            throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new NotSupportedException(SR.GetResourceString(SR.WsTrustVersionNotSupported, trustRequest.WsTrustVersion.ToString())));
+        }
+
+        #region IChannel Members
+        /// <summary>
+        /// Returns a typed object requested, if present, from the appropriate layer in the channel stack.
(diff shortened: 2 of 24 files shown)


src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/IWSTrustChannelContract.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.IdentityModel.Protocols.WsTrust;
+
+namespace System.ServiceModel.Federation
+{
+    /// <summary>
+    /// A service contract that defines the Issue request.
+    /// </summary>
+    [ServiceContract]
+    [ComVisible(false)]
+    public interface IWSTrustChannelContract : IWSTrustContract
```
mconnew, line 17: We don't have this on any of our other interfaces. Is this here for a specific reason or just because of existing coding patterns?

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/IWSTrustContract.cs
```diff
@@ -0,0 +1,48 @@
...
+    public interface IWSTrustContract
+    {
+        /// <summary>
+        /// Method for Cancel binding for WS-Trust.
+        /// </summary>
+        /// <param name="message">The WS-Trust request to send to an STS.</param>
+        /// <returns>A message containing a WS-Trust response.</returns>
+        [OperationContract(Name = "Cancel", Action = "*", ReplyAction = "*")]
+        Task<Message> CancelAsync(Message message);
+
```
mconnew, line 23: All of these methods are interchangeable as they take and return a Message and they all have the Action and ReplyAction to be "*". I need to spend a little bit of time trying some things with this code, but I'm pretty certain that the Name property is only used to provide the Action and ReplyAction if they aren't explicitly specified. This whole interface seems like it could be eliminated and replaced with a Task based equivalent of IRequestChannel. IWSTrustChannelContract has the strongly typed operations on it. I'll update once I know more.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
```diff
@@ -0,0 +1,427 @@
...
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.GetResourceString(SR.ChannelFactoryMustSupportIRequestChannel)));
+
+            Initialize(channelFactory, requestChannel, trustVersion);
+        }
+
+        /// <summary>
+        /// Constructs a <see cref="WSTrustChannel" />.
+        /// </summary>
+        /// <param name="channelfactory">The <see cref="Federation.ChannelFactory" /> that is creating this object.</param>
+        /// <param name="requestChannel">The <see cref="IRequestChannel" /> this object will be used to send and receive <see cref="Message" /> objects.</param>
```
mconnew, line 40: Should the cref say Federation.ChannelFactory? I thought this was a regular S.SM.ChannelFactory?

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
```diff
@@ -0,0 +1,427 @@
...
+        /// <summary>
+        /// Constructs a <see cref="WSTrustChannel" />.
+        /// </summary>
+        /// <param name="channelfactory">The <see cref="ChannelFactory" /> that is creating this object.</param>
+        /// <param name="trustVersion">The version of WS-Trust this channel will use for serializing <see cref="Message" /> objects.</param>
+        public WSTrustChannel(ChannelFactory channelFactory,
+                              WsTrustVersion trustVersion)
+        {
+            if (!(channelFactory is IRequestChannel requestChannel))
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(SR.GetResourceString(SR.ChannelFactoryMustSupportIRequestChannel)));
```
mconnew, line 31: I'm a little confused by this. Should the passed in ChannelFactory implement IRequestChannel, or should it produce clients which implement IRequestChannel? If the latter, there's no need to do any check as in Core, all channels implement IRequestChannel, but this code doesn't check for that. ChannelFactory itself doesn't implement IRequestChannel so this would always throw.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
```diff
@@ -0,0 +1,427 @@
...
+        /// <summary>
+        /// Causes a communication object to transition from the created state into the opened state within a specified interval of time.
+        /// </summary>
+        /// <param name="timeout">
+        /// The <see cref="TimeSpan" /> that specifies how long the open operation has to complete before timing out.
+        /// </param>
+        void ICommunicationObject.Open(TimeSpan timeout)
+        {
+            RequestChannel.Open(timeout);
+        }
```
mconnew, line 336: If the passed in IRequestChannel is already open, this will throw. Either need a state check here and only call Open if in state Created, or add validation that the passed in channel should be in the Created state.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
```diff
@@ -0,0 +1,427 @@
...
+        #endregion
+
+        #region IWSTrustContract  Members
+        /// <summary>
+        /// Sends a WS-Trust Cancel message to an endpoint.
+        /// </summary>
+        /// <param name="message">The <see cref="Message" /> that contains the instructions for the request to the STS.</param>
+        /// <returns>The <see cref="Message" /> returned from the STS.</returns>
+        public async Task<Message> CancelAsync(Message message)
+        {
```
mconnew, line 392: I don't understand the purpose of these methods. They could be sending anything to the server.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+    [ComVisible(false)]
+    public class WSTrustChannelFactory : ChannelFactory<IWSTrustChannelContract>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WSTrustChannelFactory" /> class with a specified endpoint.
+        /// configuration name.
+        /// </summary>
+        /// <param name="endpointConfigurationName">The configuration name used for the endpoint.</param>
+        public WSTrustChannelFactory(string endpointConfigurationName)
+            : base(endpointConfigurationName)
```
mconnew, line 24: We don't support endpoint configuration name as we don't support app.config. Remove this constructor as the base class will just throw.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WSTrustChannelFactory" /> class associated with a specified
+        /// name for the endpoint configuration and remote address.
+        /// </summary>
+        /// <param name="endpointConfigurationName">The configuration name used for the endpoint.</param>
+        /// <param name="remoteAddress">The <see cref="EndpointAddress" /> that provides the location of the service.</param>
+        public WSTrustChannelFactory(string endpointConfigurationName, EndpointAddress remoteAddress)
+            : base(endpointConfigurationName, remoteAddress)
```
mconnew, line 44: Same for this one, it will just throw so remove.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// <summary>
+        /// Creates a <see cref="WSTrustChannel" /> using parameters that reflect the configuration of
+        /// this factory.
+        /// </summary>
+        /// <param name="requestChannel">The <see cref="IRequestChannel"/> that will use used to send and receive messages.</param>
+        /// <returns></returns>
+        protected virtual WSTrustChannel CreateTrustChannel(IRequestChannel requestChannel)
+        {
+            return new WSTrustChannel(this, requestChannel, TrustVersion);
+        }
```
mconnew, line 88: As WSTrustChannel has such an odd constructor where it is passed this class basically just to get the MessageVersion from the binding and the EndpointAddress uri, what are your thoughts of making the constructors on WSTrustChannel internal? They aren't really there for public consumption really, the expected use case is to get an instance of WSTrustChannel from WSTrustChannelFactory.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustUtilities.cs
```diff
@@ -0,0 +1,163 @@
...
+using SecurityToken = System.IdentityModel.Tokens.SecurityToken;
+
+namespace System.ServiceModel.Federation
+{
+    /// <summary>
+    /// <see cref="WSTrustChannelSecurityTokenProvider"/> has been designed to work with the <see cref="WSFederationHttpBinding"/> to send a WsTrust message to obtain a SecurityToken from an STS. The SecurityToken is
+    /// added as an IssuedToken on the outbound WCF message.
+    /// </summary>
+    public static class WSTrustUtilities
+    {
```
mconnew, line 21: All members of this class are internal. I think you intended for the class to be internal too?

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.ServiceModel.Description;
+using Microsoft.IdentityModel.Protocols.WsTrust;
+
+namespace System.ServiceModel.Federation
+{
+    /// <summary>
+    /// A <see cref="WSTrustChannelFactory" /> that produces <see cref="WSTrustChannel" /> objects used to communicate with a WS-Trust endpoint.
+    /// </summary>
+    [ComVisible(false)]
+    public class WSTrustChannelFactory : ChannelFactory<IWSTrustChannelContract>
```
mconnew, line 16: I think this can probably be removed. It looks like you added it because originally on NetFx this was there.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/IWSTrustChannelContract.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        /// Sends a WS-Trust Issue request to a STS.
+        /// </summary>
+        /// <param name="request">The <see cref="WsTrustRequest" /> to send to the STS.</param>
+        /// <returns>A <see cref="SecurityToken" /> issued by the STS.</returns>
+        Task<SecurityToken> IssueAsync(WsTrustRequest request);
+    }
```
mconnew, line 25: You had originally planned to add a method which returns a WsTrustResponse instead of/as well as the fully baked SecurityToken. Is that not needed now?

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/IWSTrustChannelContract.cs
brentschmaltz: Existing patterns.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/IWSTrustChannelContract.cs
brentschmaltz: For returning a security token, this is all we need.
We could open up a simplified api if we exposed WSTrustResponseBodyReader.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
brentschmaltz: We could certainly do that that would limit the ability for any custom overrides of  IssueAsync.
We could wait until we are asked for the overrides, you are closer to consumers at this point in time, what do you think?

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
brentschmaltz: Yes the message is open, this was part of the original WSTrustChannel object.
This allows users to send messages for Cancel, Renew, Validte.
I have sample code to show how to use this.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
brentschmaltz: When would it be called if it was already open?

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/IWSTrustChannelContract.cs
brentschmaltz: removed [ComVisible(false)] in: ab9b3c1cb00f558dcd6ac6f3b02663ace5e25750

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
brentschmaltz: address in: ab9b3c1cb00f558dcd6ac6f3b02663ace5e25750

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
brentschmaltz: addressed in: ab9b3c1cb00f558dcd6ac6f3b02663ace5e25750

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
brentschmaltz: addressed in: ab9b3c1cb00f558dcd6ac6f3b02663ace5e25750

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
brentschmaltz: addressed in: ab9b3c1cb00f558dcd6ac6f3b02663ace5e25750

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustUtilities.cs
brentschmaltz: addressed in: ab9b3c1cb00f558dcd6ac6f3b02663ace5e25750

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
brentschmaltz: addressed in: ab9b3c1cb00f558dcd6ac6f3b02663ace5e25750

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/IWSTrustContract.cs
mconnew: I have done some local testing and confirmed what I believed. When an operation is defined like this, the is zero request shape that comes from the operation. Every method in this interface will generate the exact same on the wire message. I have no idea what the original intention was of the authors of the class in NetFx, but this interface make no sense at all to exist. We can remove this interface and just have IWSTrustChannelContract live on it's own.

brentschmaltz: It does make a difference, the responsibility is on the user to create the correct message.
To make it easier for users we probably would want to make WSTrustBodyWriter public, although the code is simple.

see code here: https://github.com/brentschmaltz/WCF/blob/9b084e846fea828fbb5d65bbe8ac70bdd26f9ded/WsTrustClientCore/Program.cs#L46

SelfHostSTS will identify the different messages and route them.
For Renew see: https://github.com/brentschmaltz/WCF/blob/9b084e846fea828fbb5d65bbe8ac70bdd26f9ded/SelfHostSTS/SelfHostSecurityTokenService.cs#L215

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
mconnew: Then make the constructor internal protected. That still stops direct instantiation but allows derivation.

mconnew: Or another option is to not do exactly the same as NetFx but instead pass the things which the channel factory is actually used for. We could change the constructor to take a MessageVersion and a Uri. This way you don't need multiple channel factory instances if your use case is to talk to multiple STS's with everything else the same. ChannelFactory.CreateChannel has an overload to create a channel for a specific endpoint so I'm going to see if there's a reliable way to extract the endpoint address from the requestChannel as that would be preferable.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
```diff
@@ -0,0 +1,406 @@
...
+            SerializationContext = new WsSerializationContext(trustVersion);
+            TrustSerializer = new WsTrustSerializer();
+
+            // If possible use the Binding's MessageVersion for creating our requests.
+            if (channelfactory?.Endpoint?.Binding?.MessageVersion != null)
+                MessageVersion = channelfactory.Endpoint.Binding.MessageVersion;
+
+            if (channelfactory?.Endpoint?.Address?.Uri?.AbsoluteUri != null)
+                Address = channelfactory.Endpoint.Address.Uri.AbsoluteUri;
+        }
```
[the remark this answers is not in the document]
mconnew: You can get the MessageVersion by calling `requestChannel.GetProperty<MessageVersion>()` so we can eliminate passing the ChannelFactory to this class.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannel.cs
mconnew: Using our code it won't be. But you've got to protect against people using the public API's outside of our usage. So someone could do this:
```
var factory = new ChannelFactory<IRequestChannel>(binding, endpoint);
var channel = factory.CreateChannel();
channel.Open();
var trustChannel = new WSTrustChannel(factory, channel, WsTrustVersion.Trust13);
((ICommunicationObject)trustChannel).Open();
```
If this code is a little more convoluted and the IRequestChannel gets passed in from somewhere else, someone could presume a newly created WSTrustChannel needs to be opened as that's the normal case. So thinking about it, I think we should throw in the constructor if the passed in IRequestChannel isn't in the Created state as that's really where the problem comes from.  
Also, while it's not part of the ICommunicationObject interface (which is explicitly implemented so you must cast to call open), do you think there's any benefit in adding an OpenAsync and CloseAsync method to this class? No cast would be needed, but this class isn't like a regular channel instance anyway so I think that's fine. Just use Task.Factory.FromAsync to wrap the Begin/End methods. Otherwise we're expecting people to either handle APM (I refer you to our teams chat about that) or use the synchronous versions.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
brentschmaltz: @mconnew i am all for that, i originally wanted to keep the API's as close as possible..
But, given that users will need to make some code changes, i agree there is a simpler and more natural model.
For example, the WSTrustChannelFactory shouldn't need to know what trust version is used.
The model in NetFx required a version specific WsTrustSerializer, however the current work doesn't require that.
I agree if we had the endpoint address, the api would be simpler.

ericstj:
cc @safern @joperezr 
@mconnew let me know that they're facing an issue with these tests using the reference assemblies.  

All the other tests in this repo don't actually use the System.ServiceModel.* contracts, instead they compile against System.Private.ServiceModel so they don't need the facades.  This test needs the facades, since it's testing the System.ServiceModel.Federation product assembly which compiles against the reference assemblies.

I think the solution might be to add references from the test project to the src projects for all the shims used by System.ServiceModel.Federation, and make sure that those don't conflict with the transitive references from the product assembly.

The way we solve a similar problem in dotnet/runtime is to flow the reference assemblies across along with the src project reference using `ReferenceAssembly` metadata.  This might be another option to allow System.ServiceModel.Federation to reference the src projects, yet compile against the reference assemblies.

src/System.ServiceModel.Federation/src/System/ServiceModel/Federation/WSTrustChannelFactory.cs
brentschmaltz: Simplified and removed methods

mconnew:
@ericstj I replaced the reference to S.P.SM project with a reference to System.ServiceModel.Primitives.Facade.csproj and it now compiles and the tests run. I'm seeing an odd problem though. When editing in VS, I get the error:

ericstj:
> I replaced the reference to S.P.SM project with a reference to System.ServiceModel.Primitives.Facade.csproj and it now compiles and the tests run. I'm seeing an odd problem though. When editing in VS, I get the error:
CS7069 Reference to type 'Binding' claims it is defined in 'System.ServiceModel.Primitives', but it could not be found System.ServiceModel.Federation.Tests (net5.0), System.ServiceModel.Federation.Tests (netcoreapp3.1)

Is System.Private.ServiceModel in the SLN?  Can you try just adding facade references (in addition to S.P.SM) rather than replacing S.P.SM?  That error indicates that the design time build is unable to follow the type-forward, which indicates it didn't have a reference to S.P.SM (or that the typeforward was missing from S.SM.P in design time).  I would expect the design time build should see both S.SM.P and S.P.SM with only the S.SM.P reference, but something about your build might be preventing that.  You can debug design time builds with information described here: https://github.com/dotnet/project-system/blob/main/docs/design-time-builds.md

Issue #4766, opened 2021-11-12
svcutil dosen't work with .NET6


zTGross:
After installing the new Visual Studio 2022 with its .NET 6 features the dotnet-svcutils tools doesen't work anymore.

What i did:

`dotnet-svcutil.exe https://myDomain.org/MyService?wsdl --outputFile SvcClient.cs --targetFramework netstandard2.0 --serializer XmlSerializer --namespace *,My.Service --enableDataBinding`

```
Output before updating the svc-util:

Microsoft (R) WCF Service Model Proxy Generation Tool for .Net Core platform
[Microsoft.Tools.ServiceModel.Svcutil, Version 2.0.1]
Copyright (c) Microsoft Corporation.  All rights reserved.

This tool collects information about how it is used in order to improve the tool. This functionality can be disabled by setting the environment variable "DOTNET_SVCUTIL_TELEMETRY_OPTOUT" to 1.

Resolving project references ...
Bootstrapping application ...
Error: Microsoft (R) Build Engine version 17.0.0+c9eb9dd64 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

C:\Program Files\dotnet\sdk\6.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: The target framework 'netcoreapp2.0' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy. [C:\Users\myself\AppData\Local\Temp\dotnet-svcutil-lib_Temp\2021_Nov_12_14_32_32\SvcutilBootstrapper\SvcutilBootstrapper.csproj]
CSC : error CS8630: Invalid 'nullable' value: 'Enable' for C# 7.3. Please use language version '8.0' or greater. [C:\Users\myself\AppData\Local\Temp\dotnet-svcutil-lib_Temp\2021_Nov_12_14_32_32\SvcutilBootstrapper\SvcutilBootstrapper.csproj]

Build FAILED.

[...]

PR #4772, opened 2021-12-09
Add net6.0 support for dotnet-svcutil and update tests.

imcarolwang:
For #4766 .

PR #4779, opened 2022-01-08
#4766 - Support netstandard2.0 when using net6.0 utility

TWhidden:
When targeting netstandard2.0, using 2.0.3 dotnet-svcutil under net6.0, the generation process will fail stating the following error:  

`SC : error CS8630: Invalid 'nullable' value: 'Enable' for C# 7.3. Please use language version '8.0' or greater`

Using `latest` tag to target what is latest available on the developers machine. 

This edit only modifies the csproj file used during the generation process

### DIFF
--- a/src/dotnet-svcutil/lib/src/Bootstrapper/SvcutilBootstrapper.cs
+++ b/src/dotnet-svcutil/lib/src/Bootstrapper/SvcutilBootstrapper.cs
@@ -151,6 +151,9 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
             // Don't treat warnings as errors so the bootstrapper will succeed as often as possible.
             this.MSBuildProj.ClearWarningsAsErrors();
 
+            // Nullable requires newer lang versions.
+            this.MSBuildProj.SetLangVersion();
+
             await this.MSBuildProj.SaveAsync(logger, cancellationToken).ConfigureAwait(false);
         }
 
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/wcf?pullRequest=4779) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/wcf?pullRequest=4779) before we can accept your contribution.<br/><br/>:x: TWhidden [sign now](https://cla.dotnetfoundation.org/dotnet/wcf?pullRequest=4779)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/wcf?pullRequest=4779) it.</sub>

mconnew:
@imcarolwang, can you check to see if this is still a problem with the latest sdk? If it is, can you validate this fixes things? Ideally we would have a test scenario to validate this, but I don't know how easy that would be to do with how we do our tests?

imcarolwang:
This is no longer an issue with the latest SDK. This PR was intended to resolve issue #4766, but the problem was actually addressed in PR #4772.

The issue occurred in the bootstrapper project, which was created to reuse referenced assembly types. The tool was updating the project's target framework from net6.0 to netstandard2.0 when the user's project targeted netstandard2.0. The fix now ensures that the bootstrapper project targets the latest .NET version available on the user's machine, and this solution has been effective so far.

We validated this scenario manually. There's an automation case testing type reuse scenario but targeting the latest .net version, not netstandard2.0. I'll try to add one later along with the current test suite fix (PR #5052).

Issue #4786, opened 2022-02-16
add HttpChannelFactory Authorization header support


Bencargs:
I'm having an issue when adding a valid AWS Signature 4 authorization header to a wcf channel's request.

As per AWS docs - https://docs.aws.amazon.com/general/latest/gr/sigv4-add-signature-to-request.html
a valid authorisation header should look like this:
`Authorization: AWS4-HMAC-SHA256 Credential=AKIDEXAMPLE/20150830/us-east-1/iam/aws4_request, SignedHeaders=content-type;host;x-amz-date, Signature=5d672d79c15b13162d9279b0855cfba6789a8edb4c82c400e06b5924a6f2b5d7`

However adding this to a wcf request header will result in an error about an invalid header format, eg:
`System.InvalidOperationException: Failed to copy the HTTP header 'Authorization' with value 'AWS4-HMAC-SHA256 `

`The format of value 'AWS4-HMAC-SHA256 Credential=xxx/ap-southeast-2/execute-api/aws4_request, SignedHeaders=host;x-amz-date;x-amz-security-token, Signature=xxx' is invalid.`

When adding this to a REST request normally the method `TryAddWithoutValidation` is called, see https://github.com/FantasticFiasco/aws-signature-version-4/blob/master/src/Private/Signer.cs#L52

I believe the fix would be to add something here - https://github.com/dotnet/wcf/blob/main/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs#L1274
```
...
else if (string.Compare(name, "authorization", StringComparison.OrdinalIgnoreCase) == 0)
{
    // allows adding authorisation that doesnt conform to header standards, eg. AWS Signiture 4 headers
    _httpRequestMessage.Headers.TryAddWithoutValidation(name, value);
}
...

HongGit:
@mconnew  could you please review and the PR as well?

mconnew:
I closed the PR as it was the wrong approach. You can either submit a new PR following the approach I suggested in your original PR, or you can use an intercepting http handler which adds the Authorization header outside of the WCF code using a technique similar to [this blog post](https://justsimplycode.com/2019/11/02/disable-header-100-continue-in-net-core-wcf-client/).  
Closing the issue as there's been no response for over a year.

PR #4787, opened 2022-02-16
addHttpChannelFactory Authorization header support

Bencargs:
for [#4786](https://github.com/dotnet/wcf/issues/4786)

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
@@ -1271,6 +1271,11 @@ namespace System.ServiceModel.Channels
                             {
                                 // specifying a range doesn't make sense in the context of WCF
                             }
+                            else if (string.Compare(name, "authorization", StringComparison.OrdinalIgnoreCase) == 0)
+                            {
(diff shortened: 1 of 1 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpChannelFactory.cs
```diff
@@ -1271,6 +1271,11 @@ private bool PrepareMessageHeaders(Message message)
                             {
                                 // specifying a range doesn't make sense in the context of WCF
                             }
+                            else if (string.Compare(name, "authorization", StringComparison.OrdinalIgnoreCase) == 0)
+                            {
+                                // allows adding authorisation that doesnt conform to header standards, eg. AWS Signiture 4 headers
+                                _httpRequestMessage.Headers.TryAddWithoutValidation(name, value);
```
mconnew, line 1277: On .NET Framework, we just blindly add any headers that we don't explicitly want to handle, which is equivalent to calling TryAddWithoutValidation. I think it you wish to add a header which doesn't conform to the HTTP spec (such as in this case), who are we to argue. So instead of this change, I think it would be better to change the code a few lines below to use TryAddWithoutValidation instead of Add.

mconnew:
Closing as this is the wrong approach to solve the problem.

PR #4814, opened 2022-04-21
Publishing IChannelListener and IChannelListener<TChannel> interfaces

### LINKED ISSUE #4812: IChannelListener interface not available
**Describe the bug**
The IChannelListener interface is not available when using the System.ServiceModel.Primitives package

**To Reproduce**
See the project in this project: https://github.com/bstordrup/WcfMissingInterface

The project will not compile as the IChannelListener is not visible in the System.ServiceModel.Channels namespace although the interface is marked public in the code base.

**Expected behavior**
IChannelListener interface should be visible.

**Additional context**
I think that the problem is caused by the IChannelListener interface not being part of the System.ServiceModel.Primitives.cs in the System.ServiceModel.Primitives.Ref project. However, I'm not able to test this as I'm not able to build the solution, but that's another story.

bstordrup:
The change will expose IChannelListener and IChannelListener<TChannel> interfaces thus making it possible to use these in projects consuming the System.ServiceModel.Primitives package.

Also a few code style changes came in (placement of opening brackets and some leading white spaces removed)

@mconnew 

Fix #4812

### DIFF
--- a/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
+++ b/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
@@ -8,44 +8,45 @@ using System.ServiceModel.Channels;
 using System.ServiceModel.Description;
 using System.ServiceModel.Dispatcher;
 
-namespace System.Collections.Generic {
-      public partial class SynchronizedCollection<T> : IList<T>, IList
-      {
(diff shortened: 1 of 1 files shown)


mconnew:
These api's aren't appropriate for the client packages.

bstordrup:
@mconnew, would then apply for going into the CoreWCF package instead, if they are server related?

Issue #4782, opened 2022-01-17
CallbackDebugBehavior and ChannelDispatcher.IncludeExceptionDetailInFaults are missing


Danielku15:
**Describe the bug**
This is a follow up on the extensions which were done already as part of https://github.com/dotnet/wcf/issues/4686. While some aspects were covered there, I am still missing `CallbackDebugBehavior` and `ChannelDispatcher.IncludeExceptionDetailInFaults` to properly migrate from .net Framework to .net core. 

**To Reproduce**
1. Try to activate the `CallbackDebugBehavior` on a duplex client channel and notice that it is missing: 
```cs
var factory = new DuplexChannelFactory<IMyContract>(...);
factory.Endpoint.EndpointBehaviors.Add(new CallbackDebugBehavior(true));
```
2. Try to make an own implementation of `CallbackDebugBehavior` which sets the `IncludeExceptionDetailInFaults` based on some custom logic and notice that quite some classes which were available in .net framework are still missing: 

```cs
public class CallbackDebugBehavior: IEndpointBehavior
{
    public CallbackDebugBehavior(bool includeExceptionDetailInFaults)
    {
        IncludeExceptionDetailInFaults = includeExceptionDetailInFaults;
    }

    public bool IncludeExceptionDetailInFaults { get; set; }
    
    void IEndpointBehavior.ApplyClientBehavior(ServiceEndpoint serviceEndpoint, ClientRuntime behavior)
    {
        var channelDispatcher = behavior.CallbackDispatchRuntime.ChannelDispatcher;
        if (channelDispatcher != null && IncludeExceptionDetailInFaults)
        {
            channelDispatcher.IncludeExceptionDetailInFaults = true;
        }
    }
    // ....
}
```

- `System.ServiceModel.Dispatcher.DispatchRuntime` only exposes `MessageInspectors` but no `ChannelDispatcher` anymore. 
-  ([.net framework](https://docs.microsoft.com/en-us/dotnet/api/system.servicemodel.dispatcher.dispatchruntime.channeldispatcher?view=netframework-4.8#system-servicemodel-dispatcher-dispatchruntime-channeldispatcher) vs [.net core](https://docs.microsoft.com/en-us/dotnet/api/system.servicemodel.dispatcher.dispatchruntime?view=dotnet-plat-ext-6.0))
- `ChannelDispatcher` is fully hidden. 

**Expected behavior**
[...]

PR #4834, opened 2022-05-25
Expose ChannelDispatcher and CallbackDebugBehavior.

imcarolwang:
For #4782 .

### DIFF
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Description/CallbackDebugBehavior.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Description/CallbackDebugBehavior.cs
@@ -10,34 +10,35 @@ namespace System.ServiceModel.Description
 {
     public class CallbackDebugBehavior : IEndpointBehavior
     {
-        bool includeExceptionDetailInFaults = false;
-
         public CallbackDebugBehavior(bool includeExceptionDetailInFaults)
         {
-            this.includeExceptionDetailInFaults = includeExceptionDetailInFaults;
+            IncludeExceptionDetailInFaults = includeExceptionDetailInFaults;
         }
 
         public bool IncludeExceptionDetailInFaults
         {
-            get { return this.includeExceptionDetailInFaults; }
-            set { this.includeExceptionDetailInFaults = value; }
+            get;
+            set;
         }
 
         void IEndpointBehavior.Validate(ServiceEndpoint serviceEndpoint)
         {
         }
+
         void IEndpointBehavior.AddBindingParameters(ServiceEndpoint serviceEndpoint, BindingParameterCollection bindingParameters)
         {
         }
+
         void IEndpointBehavior.ApplyDispatchBehavior(ServiceEndpoint serviceEndpoint, EndpointDispatcher endpointDispatcher)
         {
             throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new InvalidOperationException(
                 SR.Format(SR.SFXEndpointBehaviorUsedOnWrongSide, typeof(CallbackDebugBehavior).Name)));
         }
+
         void IEndpointBehavior.ApplyClientBehavior(ServiceEndpoint serviceEndpoint, ClientRuntime behavior)
         {
             ChannelDispatcher channelDispatcher = behavior.CallbackDispatchRuntime.ChannelDispatcher;
-            if (channelDispatcher != null && this.includeExceptionDetailInFaults)
+            if (channelDispatcher != null && IncludeExceptionDetailInFaults)
             {
                 channelDispatcher.IncludeExceptionDetailInFaults = true;
             }
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Description/CallbackDebugBehavior.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Description/CallbackDebugBehavior.cs
@@ -1,0 +1,46 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
+using System.ServiceModel.Channels;
+using System.ServiceModel.Dispatcher;
+
+namespace System.ServiceModel.Description
+{
+    public class CallbackDebugBehavior : IEndpointBehavior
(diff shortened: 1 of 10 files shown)


src/System.Private.ServiceModel/src/System/ServiceModel/Description/CallbackDebugBehavior.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System.ServiceModel.Channels;
+using System.ServiceModel.Dispatcher;
+
+namespace System.ServiceModel.Description
+{
+    public class CallbackDebugBehavior : IEndpointBehavior
+    {
+        bool includeExceptionDetailInFaults = false;
+
```
mconnew, line 14: This is missing accessibility modifier (private), and isn't using the correct naming convention. This field can go away though if you convert IncludeExceptionDetailsInFaults into an auto property.

src/System.Private.ServiceModel/src/System/ServiceModel/Description/CallbackDebugBehavior.cs
```diff
@@ -0,0 +1,46 @@
...
+// See the LICENSE file in the project root for more information.
+
+
+using System.ServiceModel.Channels;
+using System.ServiceModel.Dispatcher;
+
+namespace System.ServiceModel.Description
+{
+    public class CallbackDebugBehavior : IEndpointBehavior
+    {
```
mconnew, line 12: We should have a scenario test for this. You would test this by creating a service which uses a callback contract, throw an exception in a callback method and have the service check that the service side exception that was caught contained the client side exception details.

HongGit:
@imcarolwang can you please review Matt's comments?

imcarolwang:
> @imcarolwang can you please review Matt's comments?

@HongGit , yes I am working on adding test scenario as Matt suggests.

src/System.Private.ServiceModel/src/System/ServiceModel/Description/CallbackDebugBehavior.cs
imcarolwang: Hi @mconnew , when validating the feature in test scenario, after applying **CallBackDebugBehavior.IncludeExceptionDetailInFaults=true**, the **exception.Message** property value of exception caught from service is `"Exception has been thrown by the target of an invocation."` although the exception.ToString() contains the exception detail. This is different than behavior in Desktop where exception.Message has the exact string initiated by exception thrown from the callback method. Do you have any idea here?

src/System.Private.ServiceModel/tests/Scenarios/Client/ExpectedExceptions/CallbackDebugBehaviorTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        
+        // *** EXECUTE *** \\
+        try
+        {
+            string response = serviceProxy.Hello(greeting, includeExceptionDetailInFaults);
+            Assert.Null(response);
+        }
+        catch
+        {
+            string result = Environment.GetEnvironmentVariable(envVar, EnvironmentVariableTarget.Machine);
```
imcarolwang, line 60: I just realize the environment variable set at service side is not available to client machine. Working on way to validate the test result...

src/System.Private.ServiceModel/src/System/ServiceModel/Description/CallbackDebugBehavior.cs
mconnew: Can you provide the exact text of what you see on .NET Framework and in Core as I'm not quite understanding what the difference is that you are seeing. This is likely a bug where we aren't unwrapping a TargetInvocationException when getting the exception details.

imcarolwang: Ok. If we` throw new Exception("SpeficiedErrorMessage")` in the client call back method, and in service side, the exception is caught via `catch (Exception ex)`, then check the` ex.Message` property , with Desktop client, the value is exact the string "**SpeficiedErrorMessage**", for .NETCore client, it is "**Exception has been thrown by the target of an invocation.**" , the `ex.ToString()` is like below for Core client:

```
System.ServiceModel.FaultException`1[System.ServiceModel.ExceptionDetail]: Exception has been thrown by the target of an invocation. (Fault Detail is equal to An ExceptionDetail, likely created by IncludeExceptionDetailInFaults=true, whose value is:
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ----> System.Exception: SpeficiedErrorMessage
```

mconnew: Yep, I was right. This is a failure to unwrap the TargetInvocationException. I'll see if I can work out where the 1 line of code needs to go to fix it and let you know.

mconnew: The issue is caused by calling MethodInfo.Invoke on Core whereas on Framework we use RefEmit to output IL. MethodInfo.Invoke wraps the exception in a TargetInvocationException whereas the RefEmit solution doesn't. The problematic code is at:
https://github.com/dotnet/wcf/blob/72598df55f58b575415ae458f1fd4af332ee0310/src/System.Private.ServiceModel/src/System/ServiceModel/Dispatcher/InvokerUtil.cs#L61-L68

Replace it with the following:
```c#
try
{
    if (returnsValue)
    {
        result = method.[Invoke](https://source.dot.net/System.Private.CoreLib/A.html#c8277ed81262a367)(target, inputsLocal);
    }
    else
    {
        method.[Invoke](https://source.dot.net/System.Private.CoreLib/A.html#c8277ed81262a367)(target, inputsLocal);
    }
}
catch(TargetInvocationException tie)
{
    var actualException = tie.InnerException;
    ExceptionDispatchInfo.Capture(actualException).Throw(); // Keep original call stack
}

imcarolwang: Thank you @mconnew ! 
I have replaced the code and test shows the behavior on Core is the same as on Framework now.

PR #4839, opened 2022-06-09
wcfcs tool add service failed with 0 endpoints found error if type-reuse enabled

imcarolwang:
This is to fix below error thrown from WCF Connected Service tool when add some service with type reuse feature enabled:
```
Importing web service metadata ...
Scaffolding service reference code ...
Number of service endpoints found: 0
ErrorNo code was generated.
If you were trying to generate a client, this could be because the metadata documents did not contain any valid contracts or services, or because all contracts/services were discovered to exist in --reference assemblies. Verify that you passed all the metadata documents to the tool.
Done.
ErrorFailed adding service reference(s). Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
```

The error message is not very specific, root cause is that the referenced package has dependency file failed to be loaded in the WCFCS execution context, for example System.Runtime version 4.2.2.0 can't be loaded when reusing type SqlClientFactory in Microsoft.Data.SqlClient package version 4.1.0.

### DIFF
--- a/src/dotnet-svcutil/lib/src/FrameworkFork/System.Runtime.Serialization/System/Runtime/Serialization/DataContract.cs
+++ b/src/dotnet-svcutil/lib/src/FrameworkFork/System.Runtime.Serialization/System/Runtime/Serialization/DataContract.cs
@@ -1710,18 +1710,10 @@ namespace System.Runtime.Serialization
             string clrNs = type.Namespace;
             if (clrNs == null)
                 clrNs = String.Empty;
-            string ns = null;
-            try
-            {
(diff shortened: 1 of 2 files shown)


imcarolwang:
Fix is included in PR #4858. Close as dup.

imcarolwang:
When resolving referenced assembly/types (for reuse), we observed ReflectionTypeLoadException exception thrown, one of the reasons is package dependency assembly version conflict. The fix here is: once the ReflectionTypeLoadException happened, we caught the exception, logged the error/warnings and emptied the types instead of returning the "problem" types.

Issue #3812, opened 2019-08-11
Configure Web Service Reference - Reuse types in referenced assemblies - Still broken


PhilPJL:
Create a web service reference with a shared assembly.  If the DTO has a DataContract that specifies a namespace the type isn't reused.

This [bug ](https://github.com/dotnet/wcf/issues/3253) was supposed to have been fixed.  It appears to be fixed if the DataContract doesn't have a namespace.

Say we have the shared type
```
[DataContract]
public class CompositeType
```
This is reused.

But if we have

```
[DataContract(Namespace = "nowhere.com")]
public class CompositeType2

```

it isn't.

See the attached sample app. [WCFTest.zip](https://github.com/dotnet/wcf/files/3490383/WCFTest.zip)

**Expected behavior**
I expect shared types to be reused whether they have an explicit namespace or not.

**Screenshots**
![image](https://user-images.githubusercontent.com/271656/62838896-3273e980-bc7a-11e9-8119-f330d7594ac2.png)

**Additional context**
VS 16.3.0 Preview 1.0

StephenBonikowsky:
Thanks for reporting this @PhilPJL 
@dasetser can you take a look when you have time please.

StephenBonikowsky:
Needs to be fixed but lower priority.

PhilPJL:
@StephenBonikowsky This isn't low priority for us.  We have 20 WCF services which all use shared types across our code base.  If we want to move to .NET Core (which we do because we're now developing for Xamarin/Android) this is a big problem.  Obviously we can get rid of WCF and switch to WebApi or Grpc but really shouldn't have to.

If you do get around to fixing this don't forget it's possible for a WCF service to consume two (or more classes) with the same name disambiguated by namespace, e.g

namespace Company.Project1.Types
{
  [DataContract(namespace="namespace1)]
  public class AThing {}
}

namespace Company.Project2.Types
{
  [DataContract(namespace="namespace2)]
  public class AThing {}
}

StephenBonikowsky:
Thanks for the additional context @PhilPJL we will discuss this further.

PhilPJL:
@StephenBonikowsky: When if ever will this be fixed?

StephenBonikowsky:
@dasetser We need to add this issue into our planning for 3.1 Servicing/5.0 release. Going to add it to 3.1.1 for right now to make sure we don't miss it.

Thraka:
[...]

PR #4858, opened 2022-07-06
dotnet svcutil issue fix

imcarolwang:
1. #3812
2. suppress exception caused by reused assembly/package loading version confliction 
3. extend ServiceModelPacakges list with more WCF package dependencies, so if the client project only references WCF packages, we avoid unnecessary attempts to reuse types from project's referenced assemblies.


imcarolwang:
1. Issue #3812 is already fixed by PR #4868 (merged)
2. Reused assembly loading issue is supposed to be addressed in PR #4839
3.  The remaining code in this PR: exclude more WCF dependency packages when resolving type-reused assemblies.

PR #4868, opened 2022-08-02
Fix for DataContractAttribute evaluation

BitKFu:
In my tests the DataContractAttribute could not be found, because Globals.TypeOfDataContractAttribute is a DataContractAttribute with ref. to the Framework Fork.
Because this Framework Fork won't be used in "real live" projects, the DataContractAttribute could not be evaluated.
To fix this, the code now searches for the same FullName and read the properties via reflection in order to build the DataContractAttribute used by the Framework Fork.

This fix will probably fix, because namespace will now be evaluated correctly
[#3812](https://github.com/dotnet/wcf/issues/3812)

### DIFF
--- a/src/dotnet-svcutil/lib/src/FrameworkFork/System.Runtime.Serialization/System/Runtime/Serialization/DataContract.cs
+++ b/src/dotnet-svcutil/lib/src/FrameworkFork/System.Runtime.Serialization/System/Runtime/Serialization/DataContract.cs
@@ -1523,9 +1523,9 @@ namespace System.Runtime.Serialization
 
             // Try to find an attribute with matching name
             // type compare does not work, because of the different types used in the FrameworkFork compared to various .NET implementations 6.0/7.0
-            object[] attributes = type.GetTypeInfo()
-                .GetCustomAttributes(false)
-                .Where(a => a.GetType().FullName == Globals.TypeOfDataContractAttribute.FullName)
+            CustomAttributeData[] attributes = type.GetTypeInfo()
+                .GetCustomAttributesData()
+                .Where(a => a.AttributeType.FullName == Globals.TypeOfDataContractAttribute.FullName)
                 .ToArray();
 
 #if DEBUG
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/wcf?pullRequest=4868) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/wcf?pullRequest=4868) before we can accept your contribution.<br/><br/>:x: BitKFu [sign now](https://cla.dotnetfoundation.org/dotnet/wcf?pullRequest=4868)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/wcf?pullRequest=4868) it.</sub>

mconnew:
I think you might be able to get a cleaner implementation using GetCustomAttributesData so you don't need to use reflection.

BitKFu:
@mconnew You're right, that's slightly nicer. 
Even if I think that internally the GetCustomAttributesData also uses reflection ;)

mconnew:
It doesn't in quite the same way. There's a class called MetadataLoadContext which enables you to load assemblies in a non-executable way. You can even load assemblies of different versions than are currently loaded and being used in the process. GetCustomAttributes causes the attribute to be instantiated so doesn't work with MetadataLoadContext, but GetCustomAttributesData works fine as it doesn't instantiate an instance. The avoiding of reflection is because you can refer to properties and constructor arguments by name and avoid the reflection there. One thing to keep in mind is you have to check the constructor arguments as well as the properties as it doesn't run the constructor which would normally result in the property getting set.

BitKFu:
The DataContractAttribute has no constructor arguments. 
Please have a look at the code changes where I use the adviced GetCustomAttributesData method.
It fixes the bug #3812 and I have no clue how the original implementation ever worked?!

HongGit:
@imcarolwang can you please test this PR?

HongGit (approved):

Issue #4877, opened 2022-08-11
Improper Restriction of XML External Entity Reference [System.Security.Cryptography.Xml]


Havunen:
Overview
Affected versions of this package are vulnerable to Improper Restriction of XML External Entity Reference during the XML signature verification process.

Remediation
Upgrade System.Security.Cryptography.Xml to version 4.7.1, 6.0.1 or higher.

-> System.ServiceModel.Primitives@4.9.0
|-> System.Private.ServiceModel@4.9.0
|--->System.Security.Cryptography.Xml@5.0.0
|------> Fixed in: System.Security.Cryptography.Xml 4.7.1,6.0.1

Could this package be updated to version 6.0.1 https://github.com/dotnet/wcf/blob/main/eng/Versions.props#L37 ?

mconnew:
We will update this, but FYI WCF client isn't vulnerable to this issue. I examined the problem the fix addresses and the WCF client doesn't have the vulnerable scenario. Basically, we never use SignedXml to verify a signature on a received message, we only use it to sign a security header on an outgoing message and the entire payload for that is generated by WCF. The only purpose for WCF to updating our dependency and release a new package (which we will do) is to satisfy any systems which scan for vulnerable dependencies. There is no actual threat to using WCF with an older version.

Havunen:
Good to hear, thanks! :)

mconnew:
FYI there will be a small delay on this as we missed updating all our dependency versions in our last release so I want to evaluate and update everything to the latest version at once.

PR #4878, opened 2022-08-11
Updated netcore System.Security.Cryptography.Xml to 6.0.1

Havunen:
Fixes https://github.com/dotnet/wcf/issues/4877


andreicristianpetcu:
Is there any update on this? @mconnew @Havunen

mconnew:
@andreicristianpetcu, take a look at the referenced issue for more details. Basically, you don't need to worry about this as the vulnerable scenario isn't one that the WCF Core Client uses (validating an externally provided XML signature). We only use it to sign an internally created XML fragment so there's no untrusted XML being processed which could trigger the vulnerability. If you need to do something to quieten some automated tooling which is flagging this, you can add a direct dependency to your own project to the later version as top level project dependent versions override what packages ask for.

andreicristianpetcu:
I'm not sure I feel comfortable with swapping libraries like that. It might cause unexpected behavior. I think that if this is such an easy fix, it should be upgrade here, right? @mconnew

Havunen:
Yeah updating this package in the application level is problematic as we have multiple solutions and hundreds of projects and this issue is coming from many places through dependencies

PR #4884, opened 2022-08-17
Add wcfcore perf testing and deploy service with corewcf

ZhaodongTian:
Add perf testing with crank and deploy new service with corewcf. @HongGit @mconnew


HongGit:
The WCFCorePerf and WCFCorePerfService is duplicate with the desktop version.  What about rename WCFCorePerfService to CoreWCFService, and WCFCorePerf to WCFCorePerfClientUsingCoreWCFService (you could omit Service)?

ZhaodongTian:
> The WCFCorePerf and WCFCorePerfService is duplicate with the desktop version. What about rename WCFCorePerfService to CoreWCFService, and WCFCorePerf to WCFCorePerfClientUsingCoreWCFService (you could omit Service)? 
 I have renamed the project name

mconnew:
This PR looks like it has a lot of other unrelated changes included in it that were already merged. It's making it really difficult to review. Can you rebase to main?

ZhaodongTian:
> This PR looks like it has a lot of other unrelated changes included in it that were already merged. It's making it really difficult to review. Can you rebase to main?

I have rebased main and do a force push

src/System.Private.ServiceModel/tests/Benchmarks/WCFCorePerf/benchmarks.wcfcore.Perf.yml
```diff
@@ -2,7 +2,7 @@ jobs:
   WCFCorePerf:
     source:
       repository: https://github.com/dotnet/wcf.git
-      branchOrCommit: master
+      branchOrCommit: main
```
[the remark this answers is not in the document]
ZhaodongTian: > These two changes aren't related with using CoreWCF as server, correct?

You are right, that means the client code will pull from main branch

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/CoreWCFPerfService/CoreWCFPerfService.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>   
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="CoreWCF.Http" Version="1.1.0" />
+    <PackageReference Include="CoreWCF.NetTcp" Version="1.1.0" />
```
mconnew, line 10: Can you specify `1.*` as the version so it will always use the latest CoreWCF version.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/CoreWCFPerfService/CoreWCFPerfService.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
```
mconnew, line 2: Use the Sdk `""Microsoft.NET.Sdk.Web"` and remove the package reference to Microsoft.AspNetCore

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/CoreWCFPerfService/MyCustomValidator.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace WCFCorePerfService
+{
+    public class MyCustomValidator : UserNamePasswordValidator
+    {
+        public override ValueTask ValidateAsync(string userName, string password)
+        {            
+            if (!string.IsNullOrEmpty(userName) && !string.IsNullOrEmpty(password))
+            {
+                return new ValueTask(Task.CompletedTask);
+            }
```
mconnew, line 14: You can just return `default` as the default ValueTask represents a completed operation.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/CoreWCFPerfService/Program.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Net;
+
+namespace WCFCorePerfService
+{
+    public class Program
+    {        
+        static void Main(string[] args)
+        {
+            string filePath = Path.Combine(Environment.CurrentDirectory, "WCFCorePerfService.exe");
+            Console.WriteLine(filePath);
```
mconnew, line 19: I think a better option is `Assembly.GetEntryAssembly().Location` as the exe doesn't need to be launched from the current directory.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/CoreWCFPerfService/Program.cs
```diff
@@ -0,0 +1,126 @@
...
+            string command = $" advfirewall firewall add rule name=\"WCFCorePerfService\" dir=in protocol=any action=allow program=\"{filePath}\" enable=yes";
+            ExecuteCommand(command, Environment.CurrentDirectory, TimeSpan.FromSeconds(20));
```
mconnew, line 21: A better option is to directly modify the firewall. You can find some old code which [does this here](https://github.com/dotnet/wcf/blob/v1.1.9/src/System.Private.ServiceModel/tools/test/Bridge/WcfTestBridgeCommon/PortManager.cs). It's basically a helper class which uses the COM component for programming the firewall and includes automatic cleanup of process exit. I believe this will work now as there's COM support in recent versions of .NET.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/CoreWCFPerfService/SayHello.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Threading.Tasks;
+
+namespace WCFCorePerfService
+{
+    public class SayHello : ISayHello
+    {
+        public Task<string> HelloAsync(string name)
+        {
+            return Task.Factory.StartNew(() => { return name; });
+        }
```
mconnew, line 12: Task.FromResult(name) would be better to return as it will execute faster and we're testing the client here.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/WCFCorePerfClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+                            default:
+                                break;
+                        }
+
+                        Console.WriteLine($"Testing TransferMode: {binding.TransferMode}");
+                        ChannelFactory<ISayHello> factory = new ChannelFactory<ISayHello>(binding, new EndpointAddress(test._paramServiceUrl));
+                        var stopwatchChannelOpen = new Stopwatch();
+                        stopwatchChannelOpen.Start();
+                        factory.Open();
+                        BenchmarksEventSource.Measure("wcfcoreperf/channelopen", stopwatchChannelOpen.ElapsedMilliseconds);
```
mconnew, line 72: We should test the synchronous Open and the async Open (`Task.Factory.FromAsync(factory.BeginOpen, factory.EndOpen, null)`) as a separate test variant.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/WCFCorePerfClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+
+                        var client = factory.CreateChannel();
+                        var stopwatchFirstReq = new Stopwatch();
+                        stopwatchFirstReq.Start();
```
mconnew, line 76: You can reuse the same stopwatch and just restart it.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/WCFCorePerfClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+                        var result = client.HelloAsync("helloworld").Result;
+                        BenchmarksEventSource.Measure("wcfcoreperf/firstrequest", stopwatchFirstReq.ElapsedMilliseconds);
```
mconnew, line 78: If we're going to be using synchronous code, we should have the contract have be a synchronous method. .NET allows you to make the Main method async now so this should probably be awaiting the async method. We say async is the way people should use it, but we should measure sync too.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/WCFCorePerfClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+
+                        while (DateTime.Now <= startTime.Add(test._paramPerfMeasurementDuration))
+                        {
```
mconnew, line 81: Cache the `startTime.Add` result so it's not recalculated on every loop.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/WCFCorePerfClient/Program.cs
```diff
@@ -0,0 +1,200 @@
...
+                            var rtnResult = client.HelloAsync("helloworld").Result;
+                            request++;
+                        }
+
+                        BenchmarksEventSource.Measure("wcfcoreperf/requests", request);
+                        BenchmarksEventSource.Measure("wcfcoreperf/rps/max", request / test._paramPerfMeasurementDuration.TotalSeconds);
+                        break;
```
mconnew, line 88: The value you are dividing by is the wrong one. The total time taken will always be more than the specified duration time by an average of half the average request time. You need to measure the actual duration taken to complete the while loop and divide by that.

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/WCFCorePerfClient/WCFCorePerfClient.csproj
```diff
@@ -0,0 +1,16 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net7.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Crank.EventSources" Version="0.2.0-alpha.22228.2" />
+    <PackageReference Include="System.ServiceModel.Duplex" Version="4.10.*" />
+    <PackageReference Include="System.ServiceModel.Http" Version="4.10.*" />
```
mconnew, line 11: This should specify 4.* so that as new versions are released, we get updated results for them.

mconnew:
>             string filePath = Path.Combine(Environment.CurrentDirectory, "WcfCorePerfCrankService.exe");

My understanding was we weren't going to be downloading anything any more and would always compile the perf crank service fresh each time.

---
Refers to: src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingNetfxWCFService/NetfxWCFPerfService/Program.cs:35 in c7d5bf1. [](commit_id = c7d5bf17a854e04627f2655e908f371f0e1a2a31, deletion_comment = False)

ZhaodongTian:
> > ```
> >         string filePath = Path.Combine(Environment.CurrentDirectory, "WcfCorePerfCrankService.exe");
> > ```
> 
> My understanding was we weren't going to be downloading anything any more and would always compile the perf crank service fresh each time.
> 
> Refers to: src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingNetfxWCFService/NetfxWCFPerfService/Program.cs:35 in [c7d5bf1](https://github.com/dotnet/wcf/commit/c7d5bf17a854e04627f2655e908f371f0e1a2a31). [](commit_id = [c7d5bf1](https://github.com/dotnet/wcf/commit/c7d5bf17a854e04627f2655e908f371f0e1a2a31), deletion_comment = False)

@mconnew Crank only support  .netcore and  cannot build projects which target with .netframework. so I use download way to start service.  do you know other way to host .netframewrok service?

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/CoreWCFPerfService/CoreWCFPerfService.csproj
ZhaodongTian: If change to "Microsoft.NET.Sdk.Web",  visualstudio will throw IndexOutOfRangeException exception

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/CoreWCFPerfService/Program.cs
ZhaodongTian: Assembly.GetEntryAssembly().Location will return the full path of coreWCFPerfService.dll.  but I need to get the full path of  exe and add into firewall

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/CoreWCFPerfService/Program.cs
ZhaodongTian: I have use COM to add rules in the firewall. but this only support in windows OS

src/System.Private.ServiceModel/tests/Benchmarks/PerfUsingCoreWCFService/WCFCorePerfClient/Program.cs
ZhaodongTian: I used EndTime-BeginTime , and found the value is equal to the specified duration. Maybe something wrong in the code

Issue #4867, opened 2022-08-02
New Service Reference with .net 6 defined in Directory.Build.props-File not working


mhoermann:
Hi,

when I want to add a new "WCF Web Service" reference in a .net 6 solution I get an error when the target framework is defined in the Directory.Build.props-file and not in the project file.
When I define the target framework in the project file, the wizard actually works.

Please, look at the attachments for further details.
![1](https://user-images.githubusercontent.com/18574569/182357290-499d5738-bc05-4b7b-bd76-0027695260be.png)
![2](https://user-images.githubusercontent.com/18574569/182357302-6f04da18-6083-4453-a415-dd7986651bab.png)
![3](https://user-images.githubusercontent.com/18574569/182357308-6d9b8fef-07a4-4f10-a549-cc912c40f386.png)

imcarolwang:
A fix of the issue has been released to both dotnet-svcutil tool and VS 2022 preview. You can get the fix by installing the most recent release of [dotnet-svcutil](https://www.nuget.org/packages/dotnet-svcutil) or [VS2022 Preview](https://visualstudio.microsoft.com/vs/preview/). Closing issue.

PR #4893, opened 2022-08-25
Fix dotnet-svcutil error when project targetframework is defined in D…

imcarolwang:
…irectory.Build.props

For #4867.

### DIFF
--- a/src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
+++ b/src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
@@ -166,12 +166,6 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
                     var targetFramework = targetFrameworkElements.Last().Value.Trim().ToLowerInvariant();
                     if (!string.IsNullOrWhiteSpace(targetFramework))
                     {
-                        var tfx = targetFramework.Split('-');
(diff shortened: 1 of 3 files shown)


src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
```diff
@@ -122,7 +122,14 @@ public static async Task<MSBuildProj> FromPathAsync(string filePath, ILogger log
             return project;
         }
 
-        public static async Task<MSBuildProj> ParseAsync(string projectText, string projectFullPath, ILogger logger, CancellationToken cancellationToken)
+        internal static async Task<MSBuildProj> FromPathAsync(string filePath, ILogger logger, string tfMonitor, CancellationToken cancellationToken)
```
HongGit, line 125: nit, I might rename this function to "ProjectFromPathAsync, as it returns project.  Just from reading the function name it will be clear what is intended.

HongGit (approved):
Looks good, except one recommendation.

src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
imcarolwang: Thanks for the suggestion, it's an overloaded function, so for compatibility purpose the name remains the same as the existing one (with different number of arguments).

PR #4933, opened 2022-10-10
Use latest docker containers

adiaaida:
This change moves all of the docker images to the latest tag as part of https://github.com/dotnet/arcade/issues/10377.


adiaaida:
@imcarolwang thanks for rerunning the CI legs. Let me know if I'm good to check this in

mconnew (approved):

PR #4955, opened 2022-11-16
Initial NetNamedPipe implementation

mconnew:
Things other than NetNamedPipe that are included in this PR:
* Refactor NetTcp out of System.Private.ServiceModel into its own existing package
* Deprecate the S.SM.Duplex and S.SM.Security packages as they ended up with circular references. 
  * Moved apis into Primitives or NetFramingBase
  * Added type forwarding facade to Primitives so upgrading to latest packages doesn't cause breaks
* Refactor Http out of System.Private.ServiceModel into its own existing package
* Created new common package System.ServiceModel.NetFramingBase which contains base implementation common to NetTcp, NetNamedPipe, and a soon to be created System.ServiceModel.NetUds
* Cleanup of dead code which was triggering "This code only runs on Windows" errors due to targeting net6.0 now
* Removed package harvesting

Still a little bit of work needed
* Removal of remaining code in Common folder
* Disable NamedPipe tests for CI until we have working CoreWCF service to run against
* Get ServiceModelShim building and included in Primitives
* Fix facade generation for NetNamedPipe
* Remove System.Private.ServiceModel package from source and move scenario tests and tools


ericstj:
Have a look at this
https://github.com/dotnet/wcf/blob/a2856848c4eff95e21d60039e2ab80ae30df8ab5/eng/FacadeAssemblies.targets#L45-L63

It's only included for facade assemblies
https://github.com/dotnet/wcf/blob/a2856848c4eff95e21d60039e2ab80ae30df8ab5/Directory.Build.targets#L4

So it's not set for your net6.0 or net6.0-windows TargetFrameworks.  This part of your FacadeAssemblies.targets is both finding the reference assembly to include in the package and setting it for `ResolvedMatchingContract` which is the item used to determine what API to generate for a not-supported assembly.  Since the package has only a net461 ref and not net6.0 that's the source of the nuget warning as error.

You have reference assemblies in other WCF packages because you have different reference identity than implementation due to System.Private.ServiceModel.  If you don't need that here you could delete the reference assemblies.  That would simplify this project quite a bit.  I think this was what @ViktorHofer's was reccomending.

Rather than using a platform not-supported assembly for net6.0 why not just build a normal assembly for net6.0 and then throw in APIs you can't support on non-windows and annotate them with [SupportedOSPlatformAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.versioning.supportedosplatformattribute?view=net-7.0).

src/System.ServiceModel.NetTcp/src/System.ServiceModel.NetTcp.csproj
```diff
@@ -0,0 +1,58 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <AssemblyVersion>$(WcfAssemblyVersion)</AssemblyVersion>
+    <StrongNameKeyId>Microsoft</StrongNameKeyId>
+    <AssemblyName>System.ServiceModel.NetTcp</AssemblyName>
+    <CLSCompliant>true</CLSCompliant>
+    <IsPackable>true</IsPackable>
+    <!-- Workaround for https://github.com/NuGet/NuGet.Client/pull/3016 -->
+    <NoWarn>$(NoWarn);NU5131</NoWarn>
+    <TargetFrameworks>net6.0;net461</TargetFrameworks>
```
ericstj, line 10: It looks to me like your `net461` implementation is a full facade.  Since you aren't exposing any types on that `TargetFramework` and you don't have a `netstandard2.0` assembly with types that you need to replace with forwards on .NETFramework , could you just delete your `net461` build?

src/System.ServiceModel.NetNamedPipe/src/System.ServiceModel.NetNamedPipe.csproj
```diff
@@ -0,0 +1,63 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <AssemblyVersion>$(WcfAssemblyVersion)</AssemblyVersion>
+    <StrongNameKeyId>Microsoft</StrongNameKeyId>
+    <AssemblyName>System.ServiceModel.NetNamedPipe</AssemblyName>
+    <CLSCompliant>true</CLSCompliant>
+    <IsPackable>true</IsPackable>
+    <IsShipping>$(Ship_WcfPackages)</IsShipping>
+	<!-- TODO: Make this net462 and netstandard2.0 after it's multi-targetting .NET Framework -->
+    <TargetFrameworks>net6.0;net6.0-windows;net461</TargetFrameworks>
```
ericstj, line 10: It looks to me like your `net461` implementation is a full facade.  Since you aren't exposing any types on that `TargetFramework` and you don't have a `netstandard2.0` assembly with types that you need to replace with forwards on .NETFramework , could you just delete your `net461` build?

mconnew: FYI, this PR removes any implementation living in System.Private.ServiceModel. We're getting rid of it.  
If I remove the Ref project and build this project for net6.0 and net6.0-windows only, won't that make this package unusable on .NET Framework? I know there are customers who multi target between NetFx and .Net. 

Previously our packages had reference assemblies and implementations for net461, netcoreapp3.1, net60 (needed for MAUI due to Xamarin suppression in our package having higher precedence that netcoreapp3.1), and netstandard2.0.  
With .NET Core 3.1 about to go out of support, there's no need for us to have netstandard2.0. net461 and net6.0 should cover us.

mconnew: @ericstj, are you suggesting the assembly shouldn't be supported on net461? The intention was that the public api's which exist on .NET Framework will type forward to System.ServiceModel on .NET Framework. The idea is that a consumer can multi target to .NET Framework and .NET 6 and use the same package.  

Is that a waste of time and it would be better to have the consumer do the conditional stuff in their project to reference System.ServiceModel when compiling for .NET Framework and only reference the nuget package on .NET 6?  

The reason for not targeting netstandard2.0 is that the implementation would only run on .NET 6+ (as 3.1 is about to go out of support and .NET Framework carries its own) and it frees me up to use newer apis. I'd be fine with a reference assembly targeting netstandard2.0 as I won't be exposing any newer types on the public api surface, just want to consume them in the implementation.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnectionInitiator.cs
```diff
@@ -0,0 +1,437 @@
...
+            _bufferSize = bufferSize;
+        }
+
+        private Exception CreateConnectFailedException(Uri remoteUri, PipeException innerException)
+        {
+            return new CommunicationException(
+                SR.Format(SR.PipeConnectFailed, remoteUri.AbsoluteUri), innerException);
+        }
+
+        public async ValueTask<IConnection> ConnectAsync(Uri remoteUri, TimeSpan timeout)
```
JamesNK, line 35: There are a lot of methods that pass around a TimeSpan timeout. Have you considered using a CancellationToken instead? That's the idiomatic way to do cancellation with Tasks.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnectionInitiator.cs
```diff
@@ -0,0 +1,437 @@
...
+            {
+                backoffTimeout = Ticks.ToTimeSpan((Ticks.FromMilliseconds(backoffBufferMilliseconds) / 2) + 1);
+            }
+
+            backoffHelper = new BackoffTimeoutHelper(backoffTimeout, TimeSpan.FromMinutes(5));
+        }
+
+        private async ValueTask<IConnection> TryConnectAsync(Uri remoteUri, string resolvedAddress, BackoffTimeoutHelper backoffHelper)
+        {
+            bool lastAttempt = backoffHelper.IsExpired();
```
JamesNK, line 168: It seems odd that this method still attempts to connect after timeout

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnectionInitiator.cs
```diff
@@ -0,0 +1,437 @@
...
+            // to local pipes, we pass "." as the serverName.
+            // PipeDirection means the pipe is opened READ/WRITE
+            // PipeOptions.Asynchronous sets the flag FILE_FLAG_OVERLAPPED and binds the handle to the threadpool IOCP
+            // TokenImpersonationLevel.Anonymous sets the SECURITY_QOS_PRESENT flag and disables impersonation 
+            // HandleInheritability.None sets the security attributes to null
+            NamedPipeClientStream namedPipeClient;
+            try
+            {
+                namedPipeClient = new NamedPipeClientStream(".", resolvedAddress, PipeDirection.InOut, PipeOptions.Asynchronous, TokenImpersonationLevel.Anonymous, HandleInheritability.None);
+                await namedPipeClient.ConnectAsync((int)backoffHelper.OriginalTimeout.TotalMilliseconds);
```
JamesNK, line 179: Shouldn't this be the remaining time? Maybe I don't understand how timeouts work here. I would expect that the method immediately throws as soon as a timeout is exceeded.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnectionInitiator.cs
```diff
@@ -0,0 +1,437 @@
...
+
+            }
+
+            return new PipeConnection(namedPipeClient, _bufferSize);
+        }
+    }
+
+    internal unsafe class PipeSharedMemory : IDisposable
+    {
+        internal const string PipeLocalPrefix = @"Local\";
```
JamesNK, line 238: I noticed that named pipes on my computer tend to capitalize `LOCAL\`. I'm guessing it doesn't matter.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
```diff
@@ -0,0 +1,576 @@
...
+            {
+                return new CommunicationObjectAbortedException(exceptionMessage, pipeException);
+            }
+            else
+            {
+                return new CommunicationException(exceptionMessage, pipeException);
+            }
+        }
+
+        public async ValueTask<int> ReadAsync(Memory<byte> buffer, TimeSpan timeout)
```
JamesNK, line 114: +1 to `CancellationToken` instead of `TimeSpan timeout`.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
```diff
@@ -0,0 +1,576 @@
...
+        {
+            ValidateBufferBounds(buffer);
+            TimeoutHelper timeoutHelper = new TimeoutHelper(timeout);
+            var cancellationToken = await timeoutHelper.GetCancellationTokenAsync();
```
JamesNK, line 118: Why is GetCancellationTokenAsync async? Seems like an odd place to do IO

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
```diff
@@ -0,0 +1,576 @@
...
+                    }
+                    catch (TimeoutException e)
+                    {
+                        throw DiagnosticUtility.ExceptionUtility.ThrowHelper(
+                            new TimeoutException(SR.PipeShutdownReadError, e), ExceptionEventType);
+                    }
+                }
+                else if (existingReadIsPending)
+                {
+                    if (!TimeoutHelper.Wait(_atEOFEvent, timeoutHelper.RemainingTime()))
```
JamesNK, line 289: Should EOFEvent ManualResetEventSlim be a TaskCompletionSource? It looks like this will block. A TCS would allow an await here.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
```diff
@@ -0,0 +1,576 @@
...
+                    writeValueTask = StartWriteZeroAsync(cancellationToken);
+
+                    // read an ack for eof
+                    readValueTask = StartReadZeroAsync(cancellationToken);
+
+                    // wait for write to complete/fail
+                    await WaitForWriteZero(writeValueTask, timeout, false);
+
+                    // wait for read to complete/fail
+                    await WaitForReadZero(readValueTask, timeout, false);
```
JamesNK, line 310: timeout doesn't look like it is used by these methods

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
```diff
@@ -0,0 +1,576 @@
...
+            {
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelper(
+                    ConvertPipeException(SR.PipeCloseFailed, e, TransferOperation.Undefined), ExceptionEventType);
+            }
+            finally
+            {
+                if (shouldCloseHandle)
+                {
+                    CloseHandle(false, null, TransferOperation.Undefined);
+                }
```
JamesNK, line 345: nit: Does handle come from old code where native pipe types were used directly? I'd change handle to pipe in names.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
```diff
@@ -0,0 +1,576 @@
...
+                if (DiagnosticUtility.ShouldTrace(traceEventType))
+                {
+                    TraceUtility.TraceEvent(traceEventType, TraceCode.PipeConnectionAbort, SR.TraceCodePipeConnectionAbort, this);
+                }
+            }
+        }
+
+        private void EnterReadingState()
+        {
+            _inReadingState = true;
```
JamesNK, line 392: nit: `Debug.Assert(Monitor.IsEntered(_readLock));` is a simple way to validate methods like these are always called in the correct lock

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
JamesNK: NVM, it is used in message.

src/System.ServiceModel.NetNamedPipe/src/System.ServiceModel.NetNamedPipe.csproj
ViktorHofer: > The reason for not targeting netstandard2.0 is that the implementation would only run on .NET 6+ (as 3.1 is about to go out of support and .NET Framework carries its own) and it frees me up to use newer apis. I'd be fine with a reference assembly targeting netstandard2.0 as I won't be exposing any newer types on the public api surface, just want to consume them in the implementation.

I know you were asking Eric for an opinion but let me quickly share mine. These days, with multi-targeting being a first-class concept I wouldn't include a .NETFramework facade assembly just for the sake of a customer not needing to apply a condition when targeting both modern .NET and .NET Framework and referencing this package. We have been doing this in corefx/runtime for a while, but only because we also offer .NETStandard assets. In this case, the package targets modern .NET only, and that's OK and future proof.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnectionInitiator.cs
mconnew: See my previous comment, the final delay time is truncated to hit the originally specified timeout exactly. IsExpired means we just hit the deadline provided to the backoff helper. This had 150ms subtracted from it in PrepareConnect so we're actually up to 150ms before the original timeout (there's some wiggle room because of system timer resolution, so likely between 135ms and 150ms).

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnectionInitiator.cs
mconnew: You're right that this is wrong. I looked at the implementation and it tries to connect to the named pipe, but that requires the service side to have called ConnectNamedPipe to have a pending connection. If that's not the case the connect fails and it calls WaitNamedPipe with the timeout. This synchronously blocks until ConnectNamedPipe is called by the service. If I pass zero, it uses the what was specified when the service created the pipe, which defaults to 50ms (and WCF doesn't set it to anything explicit). I'm going to pass 1 so that it doesn't hold a thread blocking waiting for the service to begin listening.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnectionInitiator.cs
mconnew: This was copied from WCF in NetFx so I think you're right. I suspect it's the whole case preserving case insensitive thing that Windows does for its file system by default.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
mconnew: We use IOThreads for running our own timer queue baseed on a Fibonacci priority queue as we had massive performance problems with the runtime's TimerManager (global lock around cancelling timers which CancellationToken does). Some of our perf tests were spending 30%+ CPU time registering and cancelling tiemrs. This was code copied from NetFx. I could rework that to create our own equivalent of a CancellationTokenSource which uses our own timer mechanism under the hood. We do have a problem of when a CancellationToken gets cancelled that we need to throw an exception with the original Timeout and that's not easily retrieved from a CancellationToken. I'll give this some thought and try to come up with a solution.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
mconnew: It's not doing IO, it's coalescing CancellationToken's based on deadline. This was a mitigation to reduce load on the TimerManager. As two threads could be requesting a CancellationToken for the same deadline (within a window), the second thread that tries to get an instance will have to wait for the first one to finish initializing it. Think of it like an asynchronous version of Lazy\<T\> where the loser of the race condition doesn't block waiting for the winner to finish.

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
mconnew: The caller won't know what's the appropriate exception message should be as we have different exception messages for socket etc. I could catch the OperationCanceledException, catch it and throw TimeoutException with the literal `{0}` where the timeout should be, have the caller catch TimeoutException, reformat the string with the timeout and then rethrow again. It seems wrong to throw 3 times to achieve this though. Do you have any thoughts about a more elegant way to solve this?

src/System.ServiceModel.NetNamedPipe/src/System.ServiceModel.NetNamedPipe.csproj
mconnew: @ericstj, @ViktorHofer, what are your thoughts on the ServiceModel shim? Its purpose was to allow you to use an assembly compiled for .NET Framework on .NET Core. I'm having problems getting the shim to be generated with the removal of netstandard. Should I put in the work to get that working or drop it? Is that still an important scenario now?

ViktorHofer: IMHO I would drop it as running .NET Framework compiled assemblies on modern .NET was never a "first class" supported scenario. When you do a "dotnet add reference" pointing to a .NET Framework project you will even get an error today:

```cmd
C:\temp\modernent>dotnet add reference ..\netfx\
Project `C:\temp\netfx\netfx.csproj` cannot be added due to incompatible targeted frameworks between the two projects. Review the project you are trying to add and verify that is compatible with the following targets:
    - net7.0
```

src/System.ServiceModel.NetNamedPipe/src/System/ServiceModel/Channels/PipeConnection.cs
```diff
@@ -0,0 +1,578 @@
...
+            if (namedPipeClient == null)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(namedPipeClient));
+            if (!namedPipeClient.IsConnected)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperArgumentNull(nameof(namedPipeClient));
+
+            _pipe = namedPipeClient;
+            _closeState = CloseState.Open;
+            _exceptionEventType = TraceEventType.Error;
+            _connectionBufferSize = connectionBufferSize;
+            _atEOFTask = new TaskCompletionSource();
```
JamesNK, line 51: You probably want this:
```suggestion
            _atEOFTask = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
```
https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/0c82fa823796c38d0436c0c08665c944ab12f91d/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously

Kind of related, does WCF care about the synchronization context? Should `ConfigureAwait(false)` be added to all awaits?

Issue #4948, opened 2022-11-01
System.ServiceModel.Http 4.10.0 consist of known vulnerability


quukske77:
My trivy scanner reported the following on System.ServiceModel.Http 4.10.0.
![image](https://user-images.githubusercontent.com/106823500/199280748-fcd744aa-7763-4317-93e8-2d2e5406dadf.png)

idkburkes:
I am seeing the same issue from System.ServiceModel.Primitives 4.10.0

HongGit:
@imcarolwang could you please investigate this?

fforjan:
@HongGit  there is the same with System.ServiceModel.NetTcp  which references System.Security.Cryptography.Xml and there is CVE-2022-34716 
Note sure what is the expectation ?

imcarolwang:
Thanks for raising the issue! The vulnerabilities actually stem from System.Private.ServiceModel referencing System.Security.Cryptography.Xml 5.0.0, we have merged PR #4970 to bump the version to address this. Closing issue.

PR #4970, opened 2022-12-14
Upgrade SystemSecurityCryptographyXmlPackageVersion

imcarolwang:
For #4948 .

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -34,7 +34,7 @@
     <MicrosoftPrivateCoreFxNETCoreAppPackageVersion>4.5.0-rtm</MicrosoftPrivateCoreFxNETCoreAppPackageVersion>
     <SystemReflectionDispatchProxyPackageVersion>4.7.1</SystemReflectionDispatchProxyPackageVersion>
     <SystemSecurityPrincipalWindowsPackageVersion>5.0.0</SystemSecurityPrincipalWindowsPackageVersion>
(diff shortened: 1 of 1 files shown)


HongGit (approved):

felipe-canaviri:
Greetings !
I noticed this commit is upgrading the dependency to version 6.0.1. However, there is a new important upgrade in SystemSecurityCryptographyXml package, please see the following link about it:

https://github.com/dotnet/runtime/issues/98827

Wondering if you can consider to move to a newest version of that dependency.
I will really appreciate if you can provide more information for your future releases regarding to that dependency.

Thanks so much in advance !

mconnew:
Of the 6.x releases of System.Security.Cryptography.Xml, 6.0.1 is the latest version. I don't see any mention of an important upgrade in the referenced issue. As was stated in the discussion in that issue:  
> Runtime packages (i.e. Cryptography.Xml) are only serviced when they themselves are vulnerable. In this case, Cryptography.Xml wasn't vulnerable therefore a new package version didn't get released.

That also applies to WCF. We update our dependencies to the latest patch level when a vulnerability exists that affects WCF. There are no known vulnerabilities in System.Security.Cryptography.Xml which is why they haven't released an updated version.  

If a new patch is released for System.Security.Cryptography.Xml, we will update our dependency and release a new version. As it stands, there are no direct dependencies for us to update.

Issue #4984, opened 2023-01-12
MTOM request is buffered with TransferMode.Streamed or TransferMode.StreameadRequest


MarcWils:
**Describe the bug**
We're migrating some of our WCF clients from .NET 4.7.2 to .NET 6. Some of these clients upload large files to a WCF streaming service (SOAP 1.1, MTOM, HTTPS). With .NET 4.7.2 this works quit well. Streaming works end-to-end from the client to the service. We're able to upload files up to 5GB. 

After migrating to .NET 6 (System.ServiceModel.Http 4.10.0) we notice the streaming concept is broken. The request is buffered in memory. Requests larger than 2.147.483.647 bytes are not allowed because of the maxSizeInBytes setting on the MTOM writer.

**To Reproduce**
1. Checkout this repository: https://github.com/MarcWils/WcfMtomStreaming. This contains both a full framework and .NET 6 sample project.

.NET 4.7.2 runs successfully sending a 4GB file to a WCF service. 

.NET 6 fails with:
System.Xml.XmlException: The 'maximum size in bytes' quota (2147483647) has been exceeded while writing MTOM data. This quota may be increased by changing the maxSizeInBytes setting used when creating the MTOM writer.

**Expected behavior**
WCF streaming works without buffering the request. Requests larger than 2GB should also work.

**Additional context**
I can see a difference in behavior with the XmlMtomWriter:
https://github.com/dotnet/wcf/blob/5830d45a94738f00c80361cc281e1249963d9650/src/System.ServiceModel.Primitives/src/Internals/System/Xml/XmlMtomWriter.cs#L291

Running the project in .NET 4.7.2 the WriterState is set to Content. The buffer is directly passed to the inner XmlDictionaryWriter.

Running the project in .NET 6 the WriterState is set to Element. Each block of 256 bytes is added to the _binaryDataChunks list. After reading 2147483647 from the request stream an exception is thrown:
https://github.com/dotnet/wcf/blob/5830d45a94738f00c80361cc281e1249963d9650/src/System.ServiceModel.Primitives/src/Internals/System/Xml/XmlMtomWriter.cs#L324

HongGit:
@afifi-ins can you please take a look at this and see if you can identify where the issue is?

HongGit:
This is fixed now.

PR #5012, opened 2023-02-16
Update XmlMtomWriter and StreamFormatter

afifi-ins:
Reference issue: https://github.com/dotnet/wcf/issues/4984

### DIFF
--- a/src/System.ServiceModel.Primitives/src/Internals/System/Xml/XmlMtomWriter.cs
+++ b/src/System.ServiceModel.Primitives/src/Internals/System/Xml/XmlMtomWriter.cs
@@ -286,6 +286,25 @@ namespace System.Xml
                 Writer.WriteValue(value);
         }
 
+        public override Task WriteValueAsync(IStreamProvider value)
+        {
+            if (value == null)
+                throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ArgumentNullException("value"));
+
+            if (Writer.WriteState == WriteState.Element)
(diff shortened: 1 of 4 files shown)


src/System.ServiceModel.Primitives/src/Internals/System/Xml/XmlMtomWriter.cs
```diff
@@ -286,6 +286,24 @@ public override void WriteValue(IStreamProvider value)
...
+                if (_binaryDataChunks == null)
+                {
+                    _binaryDataChunks = new List<MtomBinaryData>();
+                    _contentID = GenerateUriForMimePart((_mimeParts == null) ? 1 : _mimeParts.Count + 1);
+                }
+                _binaryDataChunks.Add(new MtomBinaryData(value));
+            }
+            else
+                await Writer.WriteValueAsync(value);
+        }
```
mconnew, line 305: Making a method async causes a state machine to be generated for it and has some extra overhead. When the only async call is the last thing in a method there's an optimization you can do. You can drop the async keyword from the method and return the Task result from `Writer.WriteValueAsync(value)` to the caller. In the other code path which doesn't do anything async, return `Task.CompletedTask`.

mconnew:
Just the one small optimization change and this is correct.

mconnew:
I fixed wrapping a Task in a Task in one of your changes and then I'll merge once the tests complete. Take a look and it should be obvious what the mistake was.

PR #5030, opened 2023-03-03
Suppress codeql warning and fix typo

imcarolwang:


HongGit:
@mconnew can you please look at the code change?

mconnew:
Looks good. Looks like Carol fixed two bugs too, the switch was going the wrong way, and had the wrong name. I'll need to add a test variation which hits this code path. It only gets used if the path is too long to use as is which is why it wasn't caught with the tests I wrote as the path was too short to trigger this.

mconnew (approved):

PR #5034, opened 2023-03-07
Add Test for Mtom Stream Upload 

afifi-ins:


mconnew:
Can you remove the commented out CoreWCF service and fix the global.json conflict

afifi-ins:
> Can you remove the commented out CoreWCF service and fix the global.json conflict

Thanks, Matt, for your suggestions. There is a problem when using the Mtom headers method, the client will not return the correct number of uploaded bytes, but it will always return zero (seems like the client callbacks are not called). Calling Assert() for test validation will always fail. 

```c#
long result = serviceProxy.UploadStream(new BloatedStream(uploadBytes));
Assert.Equal(uploadBytes, result); //result=0
```

mconnew:
In working on some other fixes, I actually discovered an async bug in the mtom message encoder when running on a concurrency constrained SynchronizationContext. That fix has been merged so if you rebase, I think the CoreWCF based service will work now.

src/System.Private.ServiceModel/tests/Scenarios/Binding/Http/BasicHttpBindingTests.4.0.0.cs
```diff
@@ -359,4 +362,43 @@ public static void DecompressionEnabled_Echo_RoundTrips_String(bool decompressio
...
+    {
+        ChannelFactory<MtomBindingTestHelper.IMtomStreamingService> factory = null;
+        MtomBindingTestHelper.IMtomStreamingService serviceProxy = null;
+        WebApplication app = null;
+        try
+        {
+            // *** SETUP *** \\
+            // CoreWCF Server
+            app = MtomBindingTestHelper.BuildCoreWCFService();
+            app.Start();
```
mconnew, line 382: The app needs to be stopped, and Dispose'd.

mconnew:
I think your branch might need to be rebased on main. I believe that's what's causing the failures.

PR #4909, opened 2022-09-19
Add new WCF package version support and not generate CloseAsync() in …

imcarolwang:
…verison above 4.10.

Related issue: #4891

Issue #4891, opened 2022-08-22
System.ServiceModel.Http v4.10 


hoerup:
**Describe the bug**
While trying to update to System.ServiceModel.Http  v4.10.0 we now get this new warning (and since our pipelines runs with warnaserror this breaks the pipeline)

error CS0108: '`OurServiceNameClient.CloseAsync()`' hides inherited member `'ClientBase<IOurServiceName>.CloseAsync()'`. Use the new keyword if hiding was intended. 

We have tried to regenerate the client client, even with latest VS2022, but doesn't help ? Any guide on how to proceed?

marhoi:
Same problem here but with System.ServiceModel.Primitives v4.10.0. Same underlying issue though - WCF 3.4.0 (https://github.com/dotnet/wcf/pull/4865 I guess). Manually adding "new" to a auto generated client class does not feel right.

HongGit:
The workaround is to delete the `OurServiceNameClient.CloseAsync()`, and the fix is to fix dotnet-svcutil not to generate '`OurServiceNameClient.CloseAsync()`'.

imcarolwang:
The latest release of dotnet-svcutil tool (2.1.0) fixed the problem in generated code. Closing issue.

PR #5042, opened 2023-03-16
dotnet-svcutil test update: remove unsupported .net version tests and…

imcarolwang:
… fixing baselines.

To make changes of this PR clear, split it into following three:
https://github.com/dotnet/wcf/pull/5050: remove CloseAsync() 
https://github.com/dotnet/wcf/pull/5051: remove unsupported framework targets
https://github.com/dotnet/wcf/pull/5052: fix ParamsFiles and typereuse cases which hard code to net6.0


src/dotnet-svcutil/lib/tests/Baselines/ParamsFiles_SDK_TFM/_edb/_edb.csproj
```diff
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>N.N</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
```
HongGit, line 6: should we still support coreapp2.0 when we are deleting netcoreapp3.1?

src/dotnet-svcutil/lib/tests/Baselines/ParamsFiles_SDK_TFM/_edb/_edb.params.json
```diff
@@ -1,18 +1,18 @@
 {
-  "providerId": "TFnet5_0",
+  "providerId": "_edb",
```
HongGit, line 2: why not use 6.0 to replace 5.0?  Is _edb defined somewhere, I might have missed it>

src/dotnet-svcutil/lib/tests/Baselines/TypeReuse/TypeReuse60/SvcutilBootstrapper/Program.cs
HongGit, line 1: is this part of this change?  From the name, sound like this is used for net6.0?  Is that not the case?

src/dotnet-svcutil/lib/tests/Baselines/TypeReuse/TypeReuseDefault/ServiceReference/Reference.cs
```diff
@@ -7,12 +7,12 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
-namespace TypeReuse60_NS
```
HongGit, line 10: I see that you are making this generic.  How does this impact combability?  

In addition, perhaps we should throw an erro message when working with version less than 5.0, what do you think?

HongGit: Does generic will make it work with all versions of target Framework?  We do need to think about this.

src/dotnet-svcutil/lib/tests/Baselines/WcfRuntimeNetHttpSvcs/NetHttps/Reference.cs
```diff
@@ -616,8 +616,8 @@ public interface IWcfService
         System.Threading.Tasks.Task TestFaultIntAsync(int faultCode);
         
         [System.ServiceModel.OperationContractAttribute(Action="http://tempuri.org/IWcfService/TestFaults", ReplyAction="http://tempuri.org/IWcfService/TestFaultsResponse")]
-        [System.ServiceModel.FaultContractAttribute(typeof(NetHttps_NS.FaultDetail), Action="http://tempuri.org/IWcfService/TestFaultFaultDetailFault2", Name="FaultDetail2", Namespace="http://www.contoso.com/wcfnamespace")]
         [System.ServiceModel.FaultContractAttribute(typeof(NetHttps_NS.FaultDetail), Action="http://tempuri.org/IWcfService/TestFaultFaultDetailFault", Name="FaultDetail", Namespace="http://www.contoso.com/wcfnamespace")]
+        [System.ServiceModel.FaultContractAttribute(typeof(NetHttps_NS.FaultDetail), Action="http://tempuri.org/IWcfService/TestFaultFaultDetailFault2", Name="FaultDetail2", Namespace="http://www.contoso.com/wcfnamespace")]
```
HongGit, line 620: sorry, I am not sure I know what this is for.

src/dotnet-svcutil/lib/tests/src/E2ETests.cs
```diff
@@ -204,12 +201,12 @@ public void NamespaceParam(string testCaseName, string options, bool expectSucce
 
         [Trait("Category", "BVT")]
         [Theory]
-        [InlineData("TypeReuse60", "net6.0")]
```
HongGit, line 207: rather than use a generic format, I do prefer to add net7.0 and net 8.0 here to make it explicitly clear.  However, I do see your point that we don't need to update this in the future.  Similar to another comment before, does this mean we also allow other earlier .net versions?  We do need to be careful here.

src/dotnet-svcutil/lib/tests/src/GlobalToolTests.cs
```diff
@@ -107,24 +105,24 @@ public void FullFramework(string testCaseName)
...
         [Trait("Category", "BVT")]
         [Theory]
-        [InlineData("net6.0", "-elm")]
-        public void ParamsFiles_SDK_TFM(string targetFramework, string extraOptions)
+        [InlineData("-edb")]
+        public void ParamsFiles_SDK_TFM( string extraOptions)
         {
             this_TestCaseName = "ParamsFiles_SDK_TFM";
             TestFixture();
-            var testCaseName = $"TF{targetFramework}".Replace(".", "_");
```
HongGit, line 115: same as above

HongGit (commented):
1. please documents all the changes in the description to make this PR clear
2. does it make sense to separate the CloseAsync in a separate PR?
3. I see making the version generic is more dangerous than the gain, particularly in the type reuse side, what do you think?  I would suggest to make this PR purely remove the old non-supported framework PR.  We could rethink about the future, whether to make it generic or not.
4. I also think we need to throw an error message when the targeted app is on non-supported target framework.

Directory.Build.props
```diff
@@ -25,8 +25,8 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <ScenarioTestTargetFrameworks>net6.0;net7.0</ScenarioTestTargetFrameworks>
-	<UnitTestTargetFrameworks>net6.0;net7.0</UnitTestTargetFrameworks>
+    <ScenarioTestTargetFrameworks>net6.0;net8.0</ScenarioTestTargetFrameworks>
```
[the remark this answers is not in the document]
imcarolwang: I meant to upgrade net7.0 to net8.0 to accommodate the .net sdk and runtime definitions in https://github.com/dotnet/wcf/blob/main/global.json

src/dotnet-svcutil/lib/tests/Baselines/Collections/array/Reference.cs
```diff
@@ -889,11 +889,6 @@ public virtual System.Threading.Tasks.Task OpenAsync()
             return System.Threading.Tasks.Task.Factory.FromAsync(((System.ServiceModel.ICommunicationObject)(this)).BeginOpen(null, null), new System.Action<System.IAsyncResult>(((System.ServiceModel.ICommunicationObject)(this)).EndOpen));
         }
         
-        public virtual System.Threading.Tasks.Task CloseAsync()
```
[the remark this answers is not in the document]
imcarolwang: The test baselines are updated for changes of PR https://github.com/dotnet/wcf/pull/4909 (issue https://github.com/dotnet/wcf/issues/4891), dotnet-svcutil no longer generate CloseAsync() when the referenced wcf package version equals or greater than 4.10.

The risk here is we missed testing the tool running in projects with different .net targets,  it is due to the limitation of .net defined in global.json under the repository root. We do cover this scenario in manual testing, specifically, verifying that when the test project references wcf packages older than 4.10.*, the CloseAysnc() still get generated.

src/dotnet-svcutil/lib/tests/Baselines/WcfRuntimeNetHttpSvcs/NetHttps/Reference.cs
imcarolwang: In generated code, the attributes' order is different than before, looks relate to CodeDom other than bug from the tool.

src/dotnet-svcutil/lib/tests/src/E2ETests.cs
imcarolwang: The format is made generic because the .net version is actually not used at all in the test, in the case context it always creates a test.csproj targets to the latest .net version (currently it's net8.0 as defined in global.json) to test type-reuse feature.

imcarolwang:
> 1. please documents all the changes in the description to make this PR clear
> 2. does it make sense to separate the CloseAsync in a separate PR?
> 3. I see making the version generic is more dangerous than the gain, particularly in the type reuse side, what do you think?  I would suggest to make this PR purely remove the old non-supported framework PR.  We could rethink about the future, whether to make it generic or not.
> 4. I also think we need to throw an error message when the targeted app is on non-supported target framework.

Thanks for your review!  
1. I think it makes sense to put the CloseAsync in a separate PR, I will keep this one open as reference.
2. I'll make this PR purely remove the old non-supported framework to make it clearer. 
3. As regarding to make versions generic for 2 cases which hard coded to .net6.0, they currently fail to run because the test code can only create project targeting to .net8.0, I will make the change in new PR and work on it from there.

Directory.Build.props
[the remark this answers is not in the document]
imcarolwang: I found this setting only affects WCF tests, not dotnet-svcutil tests. So it's not changed.

PR #5050, opened 2023-03-21
dotnet-svcutil test baseline update: remove CloseAsync()

imcarolwang:
These test baselines are updated for changes of PR https://github.com/dotnet/wcf/pull/4909 (issue https://github.com/dotnet/wcf/issues/4891), dotnet-svcutil no longer generate CloseAsync() when the referenced wcf package version equals or greater than 4.10.

The risk here is we missed testing the tool running in projects with different .net targets, it is due to the limitation of .net defined in global.json under the repository root. We do cover this scenario in manual testing, specifically, verifying that when the test project references wcf packages older than 4.10.*, the CloseAysnc() still get generated.


HongGit:
@imcarolwang can you please rebase this PR?

HongGit (approved):

PR #5052, opened 2023-03-21
dotnet-svcutil tests: use default TFM for TypeReuse/ParamsFiles and refresh baselines for product upgrade

imcarolwang:
## What & why

The dotnet-svcutil test cases pinned old target frameworks (net6.0, net5.0, netcoreapp3.1) and carried stale baselines, so they broke as the product moved to net8.0/net462-only assets. This updates the affected cases to use the default target framework and refreshes the baselines accordingly.

## Test code changes

- **E2ETests.cs** — `TypeReuse` test no longer pins net6.0/SDK version; it now runs with the default TFM (`[InlineData("TypeReuse60", "net6.0")]` → `[InlineData("TypeReuse")]`).
- **GlobalToolTests.cs**
  - `ParamsFiles_SDK_TFM` renamed to `ParamsFiles`; dropped the net6.0 pinning so it uses the default TFM. Bumped the test's Newtonsoft.Json reference 13.0.1 → 13.0.2.
  - Added a new BVT test `NetStandard20TypeReuse` that exercises type-reuse against a netstandard2.0 library referenced via project (`TypeReuseClient` + `TypesLib`).
- **TestInit.cs** — `CompareFiles` now validates `<Content CopyToOutputDirectory …>` asset lines (removed from the `exceptLines` skip list) so baseline drift in asset entries is caught.
- **FixupUtil.cs** — added two normalization rules that collapse the build-output TFM prefix (e.g. net10.0) before an inner asset TFM (net8.0/net462) or the `BinLib`/`TypesLib` type-reuse assemblies, preserving the path separator so Windows and Linux baselines stay identical.

## Test-case / baseline changes

- Replaced the net6.0 `TypeReuse60` case with a default-TFM `TypeReuse` case; added `Netstandard20TypeReuse` test-case projects (`TypeReuseClient`, `TypesLib`).
- Removed the `ParamsFiles_SDK_TFM` net5.0/netcoreapp3.1 cases in favor of the default-TFM `ParamsFiles` case.
- Refreshed `.csproj` and `Reference.cs` baselines for the product upgrade: removed stale net6.0 asset Content lines (now net8.0/net462 only), picked up ExtensionData generation, and applied the DOTNET_VERSION path normalization.


HongGit:
@imcarolwang can you please research to see if it's possible to make a global variable or somewhere so all the code path could benefit in the future?

HongGit:
I agree with the latest approach, is there a concern that using the default framework may not be include all the test framework needed to be tested?

imcarolwang:
> I agree with the latest approach, is there a concern that using the default framework may not be include all the test framework needed to be tested?

I think it's ok because before updating to use default framework, test hard coded to use the latest framework version, which is no different than use the default, and we have other tests to cover different framework option test.

src/dotnet-svcutil/lib/src/Shared/TargetFrameworkHelper.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
 {
     internal class TargetFrameworkHelper
     {
-        internal static readonly List<string> s_currentSupportedVersions = new List<string>() { "8.0", "9.0" };
+        internal static readonly List<string> s_currentSupportedVersions = new List<string>() { "8.0", "9.0", "10.0" };
```
imcarolwang, line 16: Notes:
This list will determine the version of wcf packages  added to the user project. If the detected target framework equals any item in the list, the latest WCF package version (8.*) will be added, otherwise, add WCF 4.10.* packages.

The change here might require a separate PR, but is added here because the current wcf.git code base uses 10.0 SDK for test compilation and execution which leads to unexpected wcf references (4.10) added and causing a bunch of unexpected baseline mismatch test failures.

mconnew:
@imcarolwang, this PR includes the assembly src/dotnet-svcutil/lib/tests/TestCases/Netstandard20TypeReuse/BinLib/BinLib.dll. It's not normal for us to commit compiled code into the repo. Was this a mistake or intentional?

imcarolwang:
> @imcarolwang, this PR includes the assembly src/dotnet-svcutil/lib/tests/TestCases/Netstandard20TypeReuse/BinLib/BinLib.dll. It's not normal for us to commit compiled code into the repo. Was this a mistake or intentional?

Hi @mconnew, It was intentional. However, since it was not normal, I updated the test client project to reference .csproj instead of the compiled dll to cover the code generation type-reuse scenario.

PR #5055, opened 2023-03-21
Remove accidental adding of net7.0 ref

### LINKED ISSUE #5053: Breaking change going from System.ServiceModel.* 4.10.0 to 4.10.1
**Describe the bug**
Code that compiled on `System.ServiceModel.Primitives` `4.10.0` does not compile on `4.10.1`

```c#
// This line compiled previously:
await this.wcfGenerateClient.CloseAsync();
```

> {our generated WCF service} does not contain a definition for 'CloseAsync' and no accessible extension method 'CloseAsync' accepting a first argument of type ...

Our code has not changed.

In `4.10.0` `System.ServiceModel.ClientBase` had `CloseAsync`
Today, in `4.10.1` it does not.

**Expected behavior**
No breaking changes in a minor update - this should be a major version number change to `5.*` or above.

**Workaround**
This can be worked around by forcing all dependencies that use the service model primitives to a specific version:

Replace in the `.csproj`:

```xml
    <PackageReference Include="System.ServiceModel.Http" Version="4.10.*" />
```

With
```xml
    <PackageReference Include="System.ServiceModel.Http" Version="4.10.0" />
```

mconnew:
Fixes #5053

### DIFF
--- a/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.Ref.csproj
+++ b/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.Ref.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <AssemblyVersion>$(WcfAssemblyVersion)</AssemblyVersion>
     <StrongNameKeyId>Microsoft</StrongNameKeyId>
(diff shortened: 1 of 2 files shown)


mconnew:
Test failed due to timing issue with ReliableSessions test which can be fragile on slow test machines with a high latency on the network.

HongGit (approved):

Issue #5101, opened 2023-04-17
WSMessageEncoding does not exist in SystemModel namespace


ZhaodongTian:
Repro steps:

1. Create Console app and target with 7.0
2. Install below WCF package with latest version: 6.0.0-rc.23205.4
      **PackageReference Include="System.ServiceModel.Duplex" Version="6.0.0-rc.23205.4" 
    PackageReference Include="System.ServiceModel.Federation" Version="6.0.0-rc.23205.4" 
    PackageReference Include="System.ServiceModel.Http" Version="6.0.0-rc.23205.4"
    PackageReference Include="System.ServiceModel.NetTcp" Version="6.0.0-rc.23205.4" 
    PackageReference Include="System.ServiceModel.Security" Version="6.0.0-rc.23205.4" **
3.    Create BasicHttpBinding object and set MessageEncoding 
       System.ServiceModel.BasicHttpBinding result = new System.ServiceModel.BasicHttpBinding();
       **result.MessageEncoding = System.ServiceModel.WSMessageEncoding.Text;**

Actual Result:

WSMessageEncoding cannot be found 

![image](https://user-images.githubusercontent.com/25213046/232403508-c01f7560-4424-4f96-8d36-9c0f248596bf.png)

HongGit:
This has been fixed in WCF Client 6.0.

PR #5103, opened 2023-04-17
Expose WSMessageEncoding in Primitives ref project

imcarolwang:
For issue #5101.


mconnew:
This enum isn't used anywhere other than in System.ServiceModel.Http so the type should be moved to the Http package instead of exposing it in the reference assembly of Primitives/

imcarolwang:
> This enum isn't used anywhere other than in System.ServiceModel.Http so the type should be moved to the Http package instead of exposing it in the reference assembly of Primitives/

Thanks for the review!  I've moved WSMessageEncoding to System.ServiceModel.Http and do sanity check offline.

mconnew (approved):

PR #5141, opened 2023-05-04
NetNamedPipeBinding  support in dotnet-svcutil tool

imcarolwang:


dotnet-svcutil.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28728.38
+# Visual Studio Version 17
+VisualStudioVersion = 17.5.33530.505
```
HongGit, line 4: Is this change necessary for this PR?  What if someone is still on 16.0.x?

src/dotnet-svcutil/lib/NamedPipeMetadataImporter/IMetadataExchange.cs
```diff
@@ -0,0 +1,31 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.ServiceModel;
+using System.ServiceModel.Channels;
+
+namespace Microsoft.Tools.ServiceModel.Svcutil
+{
+    [ServiceContract]
+    public interface IMetadataExchange
```
HongGit, line 12: How does svcutil tool handle this?  Is there a similar class in the svcutil tool?

src/dotnet-svcutil/lib/src/Metadata/ServiceDescriptor.cs
```diff
@@ -93,7 +97,53 @@ public async Task ImportMetadataAsync(Action<WsdlImporter> onWsdlImporterCreated
...
+            if (MetadataUrl != null && MetadataUrl.Scheme.Equals("net.pipe"))
+            {
+                string tfn;                
+                if(OperationalCtx == OperationalContext.Infrastructure)
+                {
+                    tfn = "net472";
+                }
+                else
+                {
+                    tfn = "net6.0";
```
HongGit, line 110: what about net7.0 and above?

dotnet-svcutil.sln
imcarolwang: This change is not necessary, I will revert.

src/dotnet-svcutil/lib/src/Metadata/ServiceDescriptor.cs
imcarolwang: The tfn value depends on target framework of NamedPipeMetadataImporter.csproj which we used to get metadata for namedpipe binding service, that project multi-targets to net6.0 and net472, so the value here is fixed.

src/dotnet-svcutil/lib/NamedPipeMetadataImporter/IMetadataExchange.cs
HongGit: This is okay to duplicate here.

src/dotnet-svcutil/lib/src/Metadata/ServiceDescriptor.cs
HongGit: Now you mention that this is fixed for both net6.0 and net472, I am not sure that I understand why this couldn't work on net462.

imcarolwang: net6.0 is the same as the current also the lowest .net version of NetNamedPipe package, net472 is what wcf connected service tool project targets to.

NamedPipeMetadataImporter.csproj targets to net6.0 and net472 so the tfn value here is net6.0 and net472.

The value of variable "tfn" is part of the directory path to locate the internal assets (NamedPipeMetadataImporter.dll) used by dotnet-svcutil-lib to generate proxy.

Hope this clarifies the confusion.

HongGit: Let's use net462 for .NET Framework.

HongGit (approved):
Let's use net462, see my comments.

src/dotnet-svcutil/lib/src/dotnet-svcutil-lib.csproj
```diff
@@ -19,7 +19,7 @@
     <IsPackable>true</IsPackable>
     <IsShipping>$(Ship_SvcUtilPackages)</IsShipping>
     <MajorVersion>2</MajorVersion>
-    <MinorVersion>1</MinorVersion>
+    <MinorVersion>2</MinorVersion>
```
imcarolwang, line 22: @HongGit, I would like to double confirm with you on the version here, I remember we talked about making it a preview release, right?

HongGit: Correct, we can make this 2.2.0preview1.

imcarolwang: Ok. Updated.

Issue #5132, opened 2023-04-27
Add OperationDescription.SyncMethod, BeginMethod, and EndMethod


t-palanis:
Describe the bug
Hi! We are migrating our legacy application in ,Net framework to .NET core and noticed that 
OperationDescription.SyncMethod, BeginMethod, and EndMethod are missing from the .NET Core API but are available in the .NET 

Similar request to https://github.com/dotnet/wcf/issues/4549. Would it be possible to expose these?

cc @mconnew

mconnew:
We're locked in on the api surface that will be released for the upcoming 6.0 release. We can certainly add these, but it will be a while before they would be available to use. Can you describe a bit more about how you are using these properties? With the timing, if there's an alternative way to achieve your goal it might be better to modify your code to unblock you sooner.

t-palanis:
@mconnew Thanks for the reply! From legacy code, we have existing code to read custom attributes in service synchronization method(Syncmethod) via ContractDescription.Operations, currently unable to do so in .NET 6/7. 

We don't have an immediate need and if this could be available in next few months, that would be helpful

HongGit:
@imcarolwang  can you please see if you could add these?  In addition, please add tests for these APIs as well.

PR #5165, opened 2023-05-22
Public OperationDescription.Begin/End/SyncMethod properties and add unit test

imcarolwang:
For #5132.

### DIFF
--- a/src/System.ServiceModel.Primitives/tests/Description/ContractDescriptionTest.cs
+++ b/src/System.ServiceModel.Primitives/tests/Description/ContractDescriptionTest.cs
@@ -212,6 +212,16 @@ public static class ContractDescriptionTest
         {
             Assert.NotNull(operation.BeginMethod);
             Assert.NotNull(operation.EndMethod);
+            if(operation.Name.Equals("Echo"))
+            {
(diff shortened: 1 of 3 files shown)


src/System.ServiceModel.Primitives/tests/Description/ContractDescriptionTest.cs
```diff
@@ -202,4 +202,23 @@ public static void ContractDescription_GetContract()
...
+
+    [WcfFact]
+    public static void OperationDescription_BeginEndSyncMethod_Property()
+    {
+        ContractDescription contractDescription = ContractDescription.GetContract(typeof(IDescriptionTestsServiceBeginEndGenerated));
+        Assert.Equal(2, contractDescription.Operations.Count);
+        foreach(OperationDescription operation in contractDescription.Operations)
+        {
+            Assert.NotNull(operation.BeginMethod);
+            Assert.NotNull(operation.EndMethod);
```
mconnew, line 214: Can you also do `typeof(IDescriptionTestsServiceBeginEndGenerated).GetMethod(nameof(IDescriptionTestsServiceBeginEndGenerated.BeginMessageRequestReply)` and do an equals check. Do the same for End method and the sync method.

imcarolwang: Thanks for review! Equal check added.

mconnew (approved):

Issue #5159, opened 2023-05-17
Expose more DuplexChannelFactory constructor variants


pentp:
We would like to use `public DuplexChannelFactory(Type callbackInstanceType, Binding binding)` constructor variant specifically, but probably all related constructor variants that take a `callbackInstanceType` should be exposed by the ref assembly at the same time.

Related discussion: https://github.com/dotnet/wcf/issues/5091#issuecomment-1506099453

HongGit:
@imcarolwang please remember to add unit tests for this.

PR #5167, opened 2023-05-24
Expose DuplexChannelFactory<TChannel> constructor with Type overloads

imcarolwang:
For #5159 .


HongGit:
@imcarolwang can you please add one E2E test to validate the callback instance is created correctly and functions?

HongGit (approved):

Issue #4817, opened 2022-04-28
AllowInsecureTransport & EnableUnsecuredResponse in SecurityBindingElement class


phspies:
I'm trying to construct a SecurityBindingElement but the AllowInsecureTransport and EnableUnsecuredResponse attributes are not available within this class.

When you look at the code [here](https://github.com/dotnet/wcf/blob/main/src/dotnet-svcutil/lib/src/FrameworkFork/System.ServiceModel/System/ServiceModel/Channels/SecurityBindingElement.cs), one can see that these attributes are listed.

Is this code old and no longer used?

HongGit:
@imcarolwang can you please see if you could add code and tests for this issue?

seth-hamill:
I also have this issue, did you get around it? @phspies

wzchua:
Hi this is because the ref assembly does not expose those properties
 https://github.com/dotnet/wcf/blob/7a261fa1ba61e6aa40021eecf8f6fd918d155981/src/System.ServiceModel.Primitives/ref/System.ServiceModel.Security.cs#L72-L92

Issue #3705, opened 2019-06-25
SecurityBindingElement does not expose EnableUnsecuredResponse


hector-quintero:
`BindingElementCollection elements = Client.Endpoint.Binding.CreateBindingElements();
            elements.Find<SecurityBindingElement>().EnableUnsecuredResponse = true;`

Line 2 above results in compile time error stating SecurityBindingElement does not contain a definition for EnableUnsecuredResponse. 
Our team is currently using a local build of master branch version: 4.6.0-ci.19325.1

I see the property here:
[System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityBindingElement.cs](https://github.com/dotnet/wcf/blob/master/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/SecurityBindingElement.cs#L118) 
But not here: 
[System.ServiceModel.Security/ref/System.ServiceModel.Security.cs](https://github.com/dotnet/wcf/blob/master/src/System.ServiceModel.Security/ref/System.ServiceModel.Security.cs#L54)

hector-quintero:
I see there is a current workaround [here](https://github.com/dotnet/wcf/issues/3653#issuecomment-503734676) that works for me.

mconnew:
The property is on the implementation class as it was copied across from .NET Framework, but it wasn't exposed in the contract as it's not hooked up to anything functional. I'm glad the workaround fixed this for you.
Closing as it looks like there's nothing more that's needed. Feel free to re-open if you still have problems.

echavez087:
@mconnew Is it possible that you can provide a specific example of how to set the EnableUnsecuredResponse property using your workaround? Thanks!

gustavogavancho:
I concur with @echavez087. @mconnew an example of how to set the EnableUnsecuredResponse property would be really helpful.

PR #5176, opened 2023-06-19
Expose EnableUnsecuredResponse in SecurityBindingElement and add unit test

imcarolwang:
Relates to #4817, #3705. 

**Note:**
#4817 mentioned to expose both AllowInsecureTransport & EnableUnsecuredResponse,  this PR only exposes EnableUnsecuredResponse since the impementation of property AllowInsecureTransport doesn't exist in the class on WCF core, it needs to be discussed whether to add it.


HongGit:
@imcarolwang can you please create a TWMC and send a invalid credentials?  Service will reply a fault, toggle the value of the property and you should get two different exceptions.

imcarolwang:
> @imcarolwang can you please create a TWMC and send a invalid credentials? Service will reply a fault, toggle the value of the property and you should get two different exceptions.

I don't quite understand what TWMC stands for, I added a scenario test base on my understanding of the comments, and for different settings of `EnableUnsecuredResponse` the test does get different exception when sending invalid client credential.

mconnew:
TWMC means SecurityMode.TransportWithMessageCredentials. The custom binding you used is equivalent to what you can get from using TWMC so your test matches what was asked.

mconnew (approved):

Issue #5182, opened 2023-06-28
dotnet-svcutil fails to run if passed -r parameter(s) with multiple assemblies with the same name


imcarolwang:
**Describe the bug**

Following command runs with two `TypesLib.dll` in `-r` parameters and it throws error for dotnet-svcutil but not for svcutil.exe.

`dotnet-svcutil  TypeReuseSvc.wsdl -r "..\TypesLib\TypesLib.dll" -r "..\TypesLib2\TypesLib.dll"`

> Error: The option 'reference' already contains the value 'E:\Apps\TypesLib2\TypesLib.dll'.

PR #5183, opened 2023-06-28
dotnet-svcutil: differentiate dependencies with full path instead of assembly name

imcarolwang:
For #5182.

### DIFF
--- a/src/dotnet-svcutil/lib/src/Shared/ProjectDependency.cs
+++ b/src/dotnet-svcutil/lib/src/Shared/ProjectDependency.cs
@@ -395,9 +395,16 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
         public int CompareTo(ProjectDependency other)
         {
             if (other == null) return 1;
-            return this.DependencyType == other.DependencyType ?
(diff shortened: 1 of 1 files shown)


HongGit:
Will this issue resolve the issue with two assemblies?  

Can you please run testing scenarios to make sure dotnet-svcutil.exe behaves the same as svcutil.exe.

imcarolwang:
> Will this issue resolve the issue with two assemblies?
> 
> Can you please run testing scenarios to make sure dotnet-svcutil.exe behaves the same as svcutil.exe.

This will unblock user from running the tool to generate code when two assemblies of the same name are referenced, but whether the types in the referenced assemblies are reused or not depend on which version of the assemblies get restored, that's because in dotnet-svcutil, it used bootstrapper project to deal with type reusing and the project can only restore one assembly if multiple assemblies with the same name are found in the svcutilBootstrapper.csproj.

imcarolwang:
@HongGit, update on the fix: I've confirmed with Afifi about the original scenario which caused this bug, in the scenario which works on svcutil.exe, the two referenced assemblies (with same file name and different full path) have the same namespace and class mappings, in this case, the behavior on dotnet-svcutil  behaves the same as svcutil on desktop. I have tested the original problem scenario now works, and no regression found for normal type reuse scenario.

Issue #5190, opened 2023-07-03
System.ServiceModel.Primitives has a vulnerable dependency


ondrokrc:
**Describe the bug**
The most recent version 6.0.0 System.ServiceModel.Primitives depends on System.Security.Cryptography.Xml >= 6.0.1, which depends on System.Security.Cryptography.Pkcs >= 6.0.1 (for .NET Standard 2.0). This package is vulnerable in this version: https://www.nuget.org/packages/System.Security.Cryptography.Pkcs/6.0.1.

**Expected behavior**
A new version of System.ServiceModel.Primitives without vulnerable dependencies.

Thank you.

Ismael-Pep:
There isn´t a patched stable version of System.Security.Cryptography.Xml that resolves this issue, so System.Security.Cryptography.Pkcs-6.0.3 should be reference directly in the csproj.

HongGit:
Thanks for raisin this issue with use!  Please refer to [this](https://github.com/dotnet/runtime/issues/68931#issuecomment-1125260283) as why WCF packages are not impacted.

PR #5193, opened 2023-07-07
Upgrade dependency pkg version to address security vulnerability.

imcarolwang:
For #5190. 
The vulnerabilities are also reported in ci pipeline. 

Cc: @mconnew , @HongGit

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -34,6 +34,7 @@
     <SystemReflectionDispatchProxyPackageVersion>4.7.1</SystemReflectionDispatchProxyPackageVersion>
     <SystemSecurityPrincipalWindowsPackageVersion>5.0.0</SystemSecurityPrincipalWindowsPackageVersion>
     <SystemSecurityCryptographyXmlPackageVersion>6.0.1</SystemSecurityCryptographyXmlPackageVersion>
+    <SystemSecurityCryptographyPkcsPackageVersion>6.0.4</SystemSecurityCryptographyPkcsPackageVersion>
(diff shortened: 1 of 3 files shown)


HongGit:
@imcarolwang  we will wait for a secure version of System.Security.Cryptography.Xml.

MoSelem:
> @imcarolwang we will wait for a secure version of System.Security.Cryptography.Xml.

is 7.0.1 not a secure version?

imcarolwang:
> > @imcarolwang we will wait for a secure version of System.Security.Cryptography.Xml.
> 
> is 7.0.1 not a secure version?

It appears that 7.0.1 version of `System.Security.Cryptography.Xml` relies on `System.Security.Cryptography.Pkcs` (>= 7.0.0) , and 7.0.0 version of `System.Security.Cryptography.Pkcs` has the same severity vulnerability. So we still need to wait.

MoSelem:
> 

I think the vulnerability exists in version 7.0.0 for only Net7,
WCF is releasing for Net6 only so far, and the vulnerability would be in Pks >=6.0.0 and <= 6.0.2, and the patched version is 6.0.3, 
so maybe version 7.0.0 for Net6 is fine?

MoSelem:
> > 
> 
> I think the vulnerability exists in version 7.0.0 for only Net7, WCF is releasing for Net6 only so far, and the vulnerability would be in Pks >=6.0.0 and <= 6.0.2, and the patched version is 6.0.3, so maybe version 7.0.0 for Net6 is fine?

@imcarolwang

PR #5194, opened 2023-07-07
Add missing quote to dotnet-svcutil.xmlserializer.IntegrationTests.csproj

tamasvajk:


mconnew:
I believe we can just remove the PropertyGroup as we don't build for target frameworks other than net6.0 any more. @imcarolwang, @HongGit, do you agree?

imcarolwang:
> I believe we can just remove the PropertyGroup as we don't build for target frameworks other than net6.0 any more. @imcarolwang, @HongGit, do you agree?

Yes as both the product and this test project fixed TargetFramework to net6.0.

HongGit (approved):

Issue #5202, opened 2023-07-18
Upgrade to 6.0.0 (from 4.10.2) is missing IBindingRuntimePreferences


amjidq:
I have a wcf CustomBinding that is using BasicHttpBinding.
After the upgrade to 6.0.0 I get the following error in VS 2022

Error	CS7069	Reference to type 'IBindingRuntimePreferences' claims it is defined in 'System.ServiceModel.Primitives', but it could not be found	

I have added the nuget package Primitives. As well as alot of others to test with no joy

Looking in github, the interface exists in the RTM tag but not the 6.0.0 tag
is this the problem???
Or am I doing something wrong??

https://github.com/dotnet/wcf/blob/**v6.0.0-rtm**/src/System.ServiceModel.Primitives/src/System/ServiceModel/Channels/IBindingRuntimePreferences.cs

https://github.com/dotnet/wcf/blob/**v6.0.0**/src/System.ServiceModel.Primitives/src/System/ServiceModel/Channels/IBindingRuntimePreferences.cs

I have replicated this on a new console app with a class library with a single class 
ConsoleApp1.zip](https://github.com/dotnet/wcf/files/12075724/ConsoleApp1.zip)

TestClass.cs
```
using System.ServiceModel;
using System.ServiceModel.Channels;

namespace TestNamespace
{

    public abstract class TestClass
    {
        private static CustomBinding BuildBinding()
        {
            //ERROR ON LINE BELOW
            var binding = new CustomBinding(new BasicHttpBinding
            {
                MaxBufferSize = int.MaxValue,
                ReaderQuotas = System.Xml.XmlDictionaryReaderQuotas.Max,
                MaxReceivedMessageSize = int.MaxValue,
                AllowCookies = true,
                Security =
                {
                    Mode = BasicHttpSecurityMode.TransportWithMessageCredential,
                }
            });

            return binding;

        }
    }
}
```
The ClassLibrary1.csproj file is as follows

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="System.ServiceModel.Duplex" Version="6.0.0" />
    <PackageReference Include="System.ServiceModel.Http" Version="6.0.0" />
[...]

PR #5203, opened 2023-07-18
Remove interface IBindingRuntimePreferences from HttpBindingBase

imcarolwang:
For #5202.

Cc @mconnew, @HongGit


imcarolwang:
> This isn't the correct solution. Instead can you remove the implementation from HttpBindingBase as it's a meaningless interface on client.

Ok, I will, thanks for review~

HongGit (approved):

PR #5233, opened 2023-08-04
Use SDK's version for restoring NetFx targeting pack

ViktorHofer:
The SDK exposes the current version of the .NET Framework targeting pack to restore: https://github.com/dotnet/sdk/blob/25c6d1202b73a07f7932c61262131ee08509ca0e/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props#L123C3-L123C3

Avoid the hardcoded version and use the SDK one instead.


mconnew:
We don't use the Shim package any more. With moving the implementation into the actual packages rather than it all being in System.Private.ServiceModel, we got a circular dependency problem when trying to build the Shim. Unless you have the time and will to make it work again, we should remove the Shim package project.

ViktorHofer:
I'll sync offline with you on that. Unsure I understood you correctly.

ViktorHofer:
@mconnew I now see what you mean and just deleted the project. PTAL

mconnew (approved):

PR #5253, opened 2023-08-18
Update xlf files for dotnet-svcutil resource change

imcarolwang:
Similar as PR #4917, we seem need this change as part of the localization task.


imcarolwang:
Closing as it should be done automatically by a dedicated team.

PR #5268, opened 2023-08-30
Fix some tests that would fail when running under a non en-US culture

0xced:
Before this commit, two tests would fail with this error (FaultExceptionTest.Serializable_TDetail and FaultExceptionTest.Serializable_Default):
```
Xunit.Sdk.EqualException
Assert.Equal() Failure
                                 ↓ (pos 1259)
Expected: ···:text><a:xmlLang>en-US</a:xmlLang></a:FaultException.FaultRea···
Actual:   ···:text><a:xmlLang>fr-CH</a:xmlLang></a:FaultException.FaultRea···
                                 ↑ (pos 1259)
```

Explicitly setting the expected culture with a `BeforeAfterTestAttribute` solves this issue.


mconnew:
I just rebased your branch so that the latest tests will run. Sorry it took so long to merge this

PR #5279, opened 2023-09-06
Update xlf files for dotnet-svcutil-lib.

imcarolwang:


HongGit (approved):

HongGit:
@imcarolwang I am not seeing the csproj update to add the package reference.  Is that not needed here?  What if more resource strings are added later?

imcarolwang:
> @imcarolwang I am not seeing the csproj update to add the package reference. Is that not needed here? What if more resource strings are added later?

We don't need the package reference here, the package reference plus a build command change (add /UpdateXlfOnBuild:true) will check  if we miss any xlf update during build and pull the changes. According to [ this document](https://github.com/dotnet/xliff-tasks/blob/998abd6f1ad4696e7555ddc6de5b837cf765976d/README.md#L39-L40:~:text=Many%20teams%20using%20XliffTasks%20default%20UpdateXlfOnBuild%20to%20true%20for%20local%20developer%20builds%2C%20but%20leave%20it%20off%20for%20CI%20builds.), package reference and change to the build command are recommended to be turned off on CI build and if we have resource strings need to sync up with xlf files we are supposed to do it locally.

PR #5283, opened 2023-09-10
Update MessageVersion ref

Rans4ckeR:
The MessageVersion properties Soap11WSAddressing10 & Soap12 were missing from the ref.
Also removed the non-existing file System.ServiceModel.Primitives.netframework.cs from the project file.


Rans4ckeR:
@dotnet-policy-service agree

mconnew:
FYI, until this ships you can work around this by using `MessageVersion.CreateVersion`

Rans4ckeR:
> FYI, until this ships you can work around this by using `MessageVersion.CreateVersion`

Indeed, that's what I do atm. Just didn't see a reason why it's not included in the ref.

PR #5298, opened 2023-09-26
Use Expressions to generate InvokeDelegate

g7ed6e:
Use Expressions to generate InvokeDelegate. This is a performance improvement which will make callback dispatch slight faster by removing MethodInfo.Invoke calls and using LINQ expressions to compile a delegate which is strongly typed.


mconnew (approved):

mconnew:
We should document the AppContext switch `System.ServiceModel.Dispatcher.UseLegacyInvokeDelegate` that it reverts to the old behavior in case there's an unforeseen compatibility issue.

PR #5300, opened 2023-09-28
Multi target type reuse

imcarolwang:
Allow muti-target framework project reuse referenced assembly types when adding wcf service reference, the fix use the best fit .NET target to load its referenced assemblies for type reuse.

Related issue from the community: https://developercommunity.visualstudio.com/t/Cannot-Reuse-types-in-referenced-assembl/10262818


src/dotnet-svcutil/lib/src/Bootstrapper/SvcutilBootstrapper.cs
```diff
@@ -123,22 +123,17 @@ internal async Task GenerateProjectAsync(bool keepBootstrapperDir, ILogger logge
...
-                //if (isSupportedTFM && frameworkInfo.IsKnownDnx)
-                //{
-                //    if (frameworkInfo.Name == FrameworkInfo.Netstandard)
-                //    {
-                //        targetFramework = FrameworkInfo.Netcoreapp + TargetFrameworkHelper.NetStandardToNetCoreVersionMap[frameworkInfo.Version];
-                //    }
-                //    this.MSBuildProj.TargetFramework = targetFramework;
-                //}
+                var targetFramework = frameworkInfo.FullName;
+                if (isSupportedTFM && frameworkInfo.Name != FrameworkInfo.Netstandard && frameworkInfo.Version.CompareTo(new Version(6, 0)) >= 0)
```
HongGit, line 127: Why rule out Net standard and 6.0 here?

imcarolwang: The bootstrapper project is an executable project, so NetStandard is excluded here.
.NET target lower than 6.0 are ruled out to be consistent with previous behavior, back to .net5.0 and netcoreapp3.1, bootstrapper project uses the default target framework from the generated project.

HongGit (approved):

PR #5337, opened 2023-10-28
Add OperationInvoker source generator

g7ed6e:


mconnew:
MacOS tests are failing due to line ending differences when diffing the expected generated source code:
```
  -}<CR><LF>
  +}<LF>
```

PR #5358, opened 2023-11-16
Remove configuration api.

imcarolwang:
Shall we remove the [ConfigurationName](https://github.com/dotnet/wcf/blob/f229469c67dc1d46978833f72332d3ef16b3a75a/src/System.ServiceModel.Primitives/src/System/ServiceModel/ServiceContractAttribute.cs#L21C6-L21C6) property from ServiceContractAttribute as well?


mconnew (approved):
:shipit:

PR #5367, opened 2023-12-01
Update ExtendedProtectionPolicy propagate test so it also runs on non-windows

imcarolwang:
Start from .NET8,  System.Security.Authentication.ExtendedProtection.OSSupportsExtendedProtection returns OperatingSystem.IsWindows() instead of always returning true.

The change affected case ExtendedProtectionPolicy_Propagates_To_TransportBindingElement. 

The test is trying to set an extended protection policy with a PolicyEnforcement value of Always. The default is Never, so the binding needs to be set to something different to Never to detect that the value propagated. Changing the test to use WhenSupported instead will fix this as we can still see that the value propagated as it will have changed from Never, and we won't throw as WhenSupported is compatible with an OS not supporting extended protection.


imcarolwang:
Change merged in PR #5369.

PR #5369, opened 2023-12-04
update to 8.0 TFM

HongGit:
In preparation for 8.0 release, update TFM to 8.0.

### DIFF
--- a/src/System.ServiceModel.NetNamedPipe/src/System/IO/PipeException.cs
+++ b/src/System.ServiceModel.NetNamedPipe/src/System/IO/PipeException.cs
@@ -28,6 +28,7 @@ namespace System.IO
         {
         }
 
+        [Obsolete(DiagnosticId = "SYSLIB0051")]
         protected PipeException(SerializationInfo info, StreamingContext context)
             : base(info, context)
         {
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -40,10 +40,10 @@
   </ItemGroup>
 
   <PropertyGroup>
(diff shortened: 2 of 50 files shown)


src/dotnet-svcutil/lib/tests/src/GlobalToolTests.cs
```diff
@@ -111,7 +111,7 @@ public void FullFramework(string testCaseName)
         [InlineData("netstd20")] //WCF package older than V4.10 get referened and CloseAsync() is generated
         [InlineData("net60")] //WCF package newer than V4.10 get referened and CloseAsync() is not generated
         [InlineData("net60net48")] //WCF package newer than V4.10 and System.ServiceModel.dll are referenced conditionally by target and CloseAsync() be generarted with conditional compilation mark
-        public void MultiTargetCloseAsyncGeneration(string testCaseName)
+        public async void MultiTargetCloseAsyncGeneration(string testCaseName)
         {
```
mconnew, line 115: Switch to `async Task` and not `async void` otherwise xunit doesn't see failures after the first async await

src/System.ServiceModel.NetNamedPipe/src/System/IO/PipeException.cs
```diff
@@ -28,6 +28,7 @@ public PipeException(string message, Exception inner)
         {
         }
 
+        [Obsolete(DiagnosticId = "SYSLIB0051")]
         protected PipeException(SerializationInfo info, StreamingContext context)
```
mconnew, line 32: This is changing api surface which will have a doc effect too. Can you use a pragma to ignore instead or is this a hard requirement to compile?

Issue #5362, opened 2023-11-22
svcutil doesn't work with .Net 8


KristijanZic:
I've installed the .net 8 and dotnet-svcutil but it doesn't work with .Net 8.

```sh
dotnet-svcutil --version
You must install or update .NET to run this application.

App: /Users/username/.dotnet/tools/dotnet-svcutil
Architecture: arm64
Framework: 'Microsoft.NETCore.App', version '6.0.0' (arm64)
.NET location: /usr/local/share/dotnet

The following frameworks were found:
  8.0.0 at [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]

Learn more:
https://aka.ms/dotnet/app-launch-failed

To install missing framework, download:
https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=6.0.0&arch=arm64&rid=osx-arm64&os=osx.14
```

imcarolwang:
Hi @KristijanZic, the source change for support running dotnet-svcutil on .Net8 is merged (PR [#5376](https://github.com/dotnet/wcf/pull/5376)), the change will take effect in the next coming release.
 
It's worth noting that the latest preview version of dotnet-svcutil supports .Net8 (you can install it by running `dotnet tool install --global dotnet-svcutil --version 2.2.0-preview1.23462.5`), that's because it was built using a preview version of .Net8 SDK which added support of `"rollForward": "Major"` config, but this config was then removed in later releases of .Net SDK ([more](https://github.com/dotnet/docs/issues/35267)) which building dotnet-svcutil relies on. So we still need the change of PR [#5376](https://github.com/dotnet/wcf/pull/5376) for the next release of dotnet-svcutil to work on machine environment where the lowest runtime available is .Net8.0.

maxmoore14:
Hi @imcarolwang, do you know when the next release is scheduled to be available?

stamminator:
We're a ways into .NET 8 being live now. Any news on when dotnet-svcutil will be compatible with it?

YounesCheikh:
Any news about this release ? .NET 9 is almost here :)

jangaehler:
Hi @imcarolwang, the preview version is now 9 months old, so the next release can only be days away, right?

Post-Impatica:
Does this mean that the visual studio creation of a wcf may have issues as well?  I'm asking because I'm having an issue creating a wcf service from the gui

Leonardo-Ferreira:
[...]

PR #5376, opened 2023-12-12
Support net8 SDK in dotnet-svcutil

betmix-matt:
Resolves Issue #5362


betmix-matt:
If we can get this merged before the 2.2 release, it can include both .NET 7 and .NET 8 support for dotnet-svcutil

imcarolwang (approved):
Looks good to me.

HongGit (approved):

0xced:
I think the `dotnet-svcutil` tool should be configured to roll forward to the latest major .NET version, just like the [dotnet ef tool did](https://github.com/dotnet/efcore/pull/27671).

Basically, that means adding the `<RollForward>Major</RollForward>` property in the project file.

So that when new runtimes are released the tool will automatically work with the new runtime. There's even a proposal for this behavior to be the default for .NET tools (https://github.com/dotnet/sdk/issues/30336) but it has not been implemented yet.

PR #5385, opened 2023-12-14
Make ServiceChannelProxy implement IAsyncDisposable

### LINKED ISSUE #5270: ServiceChannelProxy should implement both IAsyncCommunicationObject and IAsyncDisposable
It seems to me that the `ServiceChannelProxy` class should implement the `IAsyncCommunicationObject` and `IAsyncDisposable` interfaces. I was surprised to learn that it currently doesn't.

The implementation looks trivial to me since the underlying `ServiceChannel` object already implements both interfaces.

```csharp
async Task IAsyncCommunicationObject.CloseAsync(TimeSpan timeout)
{
    await ((IAsyncCommunicationObject)_serviceChannel).CloseAsync(timeout);
}

async Task IAsyncCommunicationObject.OpenAsync(TimeSpan timeout)
{
    await ((IAsyncCommunicationObject)_serviceChannel).OpenAsync(timeout);
}

async ValueTask IAsyncDisposable.DisposeAsync()
{
    await ((IAsyncDisposable)_serviceChannel).DisposeAsync();
}
```

Is there any risk in adding those two interfaces to `ServiceChannelProxy` ?

0xced:
So that channels created through a `ChannelFactory<T>` without inheriting from `ClientBase<T>` can also benefit from the asynchronous disposal improvements introduced in #4865.

Fixes #5270

### DIFF
--- a/src/System.ServiceModel.Primitives/src/System/ServiceModel/Channels/ServiceChannelProxy.cs
+++ b/src/System.ServiceModel.Primitives/src/System/ServiceModel/Channels/ServiceChannelProxy.cs
@@ -15,7 +15,7 @@ using System.Threading.Tasks;
 
 namespace System.ServiceModel.Channels
 {
-    public class ServiceChannelProxy : DispatchProxy, ICommunicationObject, IChannel, IClientChannel, IOutputChannel, IRequestChannel, IServiceChannel, IDuplexContextChannel
+    public class ServiceChannelProxy : DispatchProxy, ICommunicationObject, IChannel, IClientChannel, IOutputChannel, IRequestChannel, IServiceChannel, IDuplexContextChannel, IAsyncDisposable
     {
         private const String activityIdSlotName = "E2ETrace.ActivityID";
         private Type _proxiedType;
@@ -671,6 +671,11 @@ namespace System.ServiceModel.Channels
         void IDisposable.Dispose()
         {
             ((IClientChannel)_serviceChannel).Dispose();
+        }
+
+        ValueTask IAsyncDisposable.DisposeAsync()
+        {
+            return ((IAsyncDisposable)_serviceChannel).DisposeAsync();
         }
 
         bool IContextChannel.AllowOutputBatching


0xced:
Is there any chance for this to eventually land in a 8.x.x update of the `System.ServiceModel` NuGet packages?

0xced:
Here's why this pull request is important. 😁 https://github.com/0xced/Wcf.HttpClientFactory/blob/6254a6df13fddcd3143c0ece726ccd5778ff15c9/tests/B2BServiceTest.cs#L86-L90

There's a lot going on in my `Wcf.HttpClientFactory` so please don't hesitate to ask if you have questions.

mconnew:
This will go into our next 8.x.x release. We don't have a planned release right now, so this will go out in the next release that happens.

mconnew:
Btw, I think you can work around this by creating another type in DI which wraps the channel and implements IAsyncDisposable. In that, it does a non-throwing Dispose of the client. If I remember right, objects get disposed in a predictable order, I think it's construction order. But if it's the other way around, you can accommodate that too. You would something like this:
```c#
internal class WcfThrowingDisposeWrapper : IAsyncDisposable
{
    public CommunicationObject Disposable { get; set; }
    public ValueTask DisposeAsync()
    {
        if (Disposable == null) return;
        try
        {
            // Only want to close if it is in the Opened state
            if (Disposable.State == CommunicationState.Opened)
            {
                await Task.Factory.FromAsync(Disposable.BeginOpen, Disposable.EndOpen, null);
            }
            // Anything not closed by this point should be aborted
            if (Disposable.State != CommunicationState.Closed)
            {
                Disposable.Abort();
            }
        }
        catch (CommunicationException)
        {
            Disposable.Abort();
        }
        catch (TimeoutException)
        {
            Disposable.Abort();
        }
    }
}
```
In the place where you add a factory to create the WCF client, you also create an instance of WcfThrowingDisposeWrapper and set the WCF client on it. Something like this:
```c#
services.AddScoped<WcfThrowingDisposeWrapper>();
services.AddScoped<IService>(services =>
{
    var wrapper = services.GetRequiresService<WcfThrowingDisposeWrapper>();
    var channel = services.GetRequiresService<ChannelFactory<IService>>();
    wrapper.Disposable = channel;
    return channel;
});
```
When you dispose the scope, it should dispose the `WcfThrowingDisposeWrapper` first which should dispose the channel without throwing, then when the scope comes to dispose the channel, it will already have been done. The in-box DI service provider scope maintains a simple `List<object>` for all instantiated instances which are disposable, adding them as they are created. Then when the scope is disposed, it iterates the list and disposes them, which means WcfThrowingDisposeWrapper would get disposed first. You should still keep your catch block in your implementation as the order of disposal is an internal implementation detail and a 3rd party DI engine may not do it in the same order.

0xced:
.NET 9 is going to be released very soon. Are there plans to release either a 8.x.x update with this fix included, or maybe a 9.0.0 version?

0xced:
I see that [System.ServiceModel.Primitives 8.1.0](https://www.nuget.org/packages/System.ServiceModel.Primitives/8.1.0) has been released and includes this fix. 🥳

Issue #5425, opened 2024-02-16
Got the "An HTTP Content-Type header is required for SOAP messaging and none was found" error when the content-type header has multiple values


lucassantosTI:
**Describe the bug**
I got the "An HTTP Content-Type header is required for SOAP messaging and none was found" exception when I sent a third-party webservice request.

This errors occurs because the server send me a reply with a Content-Type header with multiple values.

**My request is like this:**
```
POST http://url/service.asmx HTTP/1.1
Content-Type: text/xml;charset=utf-8
SOAPAction: "http://url/serviceAction"
Host: url
Content-Length: 424
Expect: 100-continue
Accept-Encoding: gzip, deflate
Connection: Keep-Alive

<?xml version="1.0" encoding="utf-8"?><s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/"><s:Body xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema"><serviceAction xmlns="http://url"><ID_SATTRANS>9852365</ID_SATTRANS><mensagemLivre>Teste desenvolvimento</mensagemLivre></serviceAction></s:Body></s:Envelope>
```

**And I got this response from the server:**
```
HTTP/1.1 500 Internal Server Error
Cache-Control: private
Content-Type: text/xml; charset=utf-8,text/xml;charset=ISO-8859-1
Server: Microsoft-IIS/10.0
X-AspNet-Version: 4.0.30319
X-Powered-By: ASP.NET
Date: Thu, 15 Feb 2024 18:55:46 GMT
Connection: close
Content-Length: 422

<?xml version="1.0" encoding="utf-8"?><soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema"><soap:Body><soap:Fault><faultcode>soap:Server</faultcode><faultstring>Server was unable to process request. ---&gt; The validation business rule violated.</faultstring><detail /></soap:Fault></soap:Body></soap:Envelope>
```
In this response we have this on content-type header: `Content-Type: text/xml; charset=utf-8,text/xml;charset=ISO-8859-1` and it's results in an empty `ContentType` value on the `HttpResponseMessage`. 
[...]

PR #5426, opened 2024-02-16
Changes the ValidateContentTypeAsync to handle Content-Type header with multiple values to prevent a null content-type exception.

lucassantosTI:
In issue #5425 reported an error occurred when the content-type header had multiple values.

Because in our day-by-day we need to call third-party services when we don't have control over their process, we need to have the ability to handle messages with this content-type.


HongGit:
Close this PR as @mconnew suggested an alternative solution to this issue.

PR #5443, opened 2024-03-06
dotnet-svcutil: add reference to wcf package 8.0 for client project

imcarolwang:
1. update referenced wcf package to latest RTM version 8.0 for client project after code generated.
2. wcf 8.0 & 6.2 combines `System.ServiceModel.Security` & `System.ServiceModel.Duplex` into `System.ServiceModel.Primitives`, update to reflect that change, and add reference to all supported wcf pkgs including `System.Web.Services.Description` & `System.ServiceModel.UnixDomainSocket` for 6.0 and 8.0 targets.
3. update test baselines affected by the above changes.

### DIFF
--- a/src/dotnet-svcutil/lib/tests/Baselines/Collections/dictionary/dictionary.csproj
+++ b/src/dotnet-svcutil/lib/tests/Baselines/Collections/dictionary/dictionary.csproj
@@ -3,12 +3,14 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>N.N</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="dotnet-svcutil-lib" Version="99.99.99" />
+    <Content CopyToOutputDirectory="always" Include="$(NuGetPackageRoot)dotnet-svcutil-lib/99.99.99/content/internalAssets/**" Link="internalAssets/%(RecursiveDir)%(Filename)%(Extension)" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="System.ServiceModel.*", Version="N.N.N">
     <PackageReference Include="System.ServiceModel.*", Version="N.N.N">
     <PackageReference Include="System.ServiceModel.*", Version="N.N.N">
     <PackageReference Include="System.ServiceModel.*", Version="N.N.N">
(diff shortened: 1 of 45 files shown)


src/dotnet-svcutil/lib/tests/Baselines/Collections/dictionary/dictionary.csproj
```diff
@@ -3,15 +3,17 @@
...
   <ItemGroup>
     <PackageReference Include="dotnet-svcutil-lib" Version="99.99.99" />
+    <Content CopyToOutputDirectory="always" Include="$(NuGetPackageRoot)dotnet-svcutil-lib/99.99.99/content/internalAssets/**" Link="internalAssets/%(RecursiveDir)%(Filename)%(Extension)" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="System.ServiceModel.*", Version="N.N.N">
     <PackageReference Include="System.ServiceModel.*", Version="N.N.N">
     <PackageReference Include="System.ServiceModel.*", Version="N.N.N">
     <PackageReference Include="System.ServiceModel.*", Version="N.N.N">
-    <PackageReference Include="System.ServiceModel.*", Version="N.N.N">
```
HongGit, line 15: I guess this is cut is because System.security and Duplex are gone.  What about the new packages, like NetNamedPipe?

HongGit (commented):
I see the changes in tests repeated so many places.  Is it possible to create a helper method so only need to update once?

src/dotnet-svcutil/lib/tests/Baselines/Collections/dictionary/dictionary.csproj
imcarolwang: Yes, this change is because System.SerivceModel.Security and Duplex are gone and being replaced with System.ServiceModel.Primitives.  
System.ServiceModel.NetNamedPipe package is included since last release ([reference](https://github.com/dotnet/wcf/blob/98e3bd07ddc47b4f7e77874276e140406ff04f9e/src/dotnet-svcutil/lib/src/Shared/TargetFrameworkHelper.cs#L74)), so it doesn't reflect change for NetNamedPipe  here.

imcarolwang:
> I see the changes in tests repeated so many places. Is it possible to create a helper method so only need to update once?

The changes are baseline file changes, for each test we do test baseline file comparison in line basis, helper method doesn't seem work here.

HongGit (approved):

Issue #5495, opened 2024-04-08
dotnet-svcutil generates json which contains absolute path to the referenced assembly.


voroninp:
When I run `dotnet-svcutil`  the generated file `dotnet-svcutil.params.json` contains absolute path to the dll.
That's not what should go to source control system.

![image](https://github.com/dotnet/wcf/assets/9202028/c8325b83-00bf-4752-a4a1-07a8d91250a7)

Here's is a solution to reproduce the issue:
[TypesAreNotReused.zip](https://github.com/dotnet/wcf/files/14902229/TypesAreNotReused.zip)

1. cd to TypesAreNotReused project.
2. Run `dotnet-svcutil '.\AdAlliance (RTL) Service Definition - 12mar2024.wsdl' -n "*,MBS.Saleshouse.Webservices.Lib.RTL" -d RTL`
3. Look into generated `dotnet-svcutil.params.json`

I also modeled the situation of checking out on different machine where paths are obviously different. Just renamed the solution folder.
I tried to update the service `dotnet-svcutil -u` afterwards.
I see the path completely removed from the file, and types not reused. How come?

P.S. I'd expect the relative path to proj file, not to dll directly.

This is a continuation of [this discussion](https://developercommunity.visualstudio.com/t/WCF-Connected-Services---Compile-Error/10603813#TPIN-N10628932-N10632565).

imcarolwang:
Hi there, after checking out the project resource on different machine, the project needs to be compiled (if it wasn’t) for assembly type reuse to take effect, because the tool tried to load the project's dependency assembly and inspect for reusable types at runtime, if the project is not built the referenced assemblies are inaccessible to the tool, then the types won't be reused.
 
As for the local referenced assembly path being written to the parameters json file, indeed it could be an issue when type reuse mode is Specified. I am currently drafting a PR to address this.

voroninp:
> the project needs to be compiled (if it wasn’t) for assembly type reuse to take effect

This is the reason why the tool should support specifying the reference to the project instead of the assembly. Then it could force the compilation.
Or it least it should not ignore the absent references but return an error?

Montago:
I cant make Type-Reuse to work ... (using Visual Studio Professional 17.14.7) 

The configuration json gets messed up like you are experiencing with absolute paths to the referenced classes 

and regardless of what i do, the svcutil generates type-proxies / copies 🤷🏻‍♂️

I miss the old .NET Framework days where WCF just worked...

PR #5500, opened 2024-04-16
dotnet-svcutil: write generated json file with relative path for referenced assembly.

imcarolwang:
For issue #5495.

Write referenced assembly name instead of absolute path to the generated parameters json file, corresponding to the change, recalculate references base on project dependencies for update operation, update related test as well.

### DIFF
--- a/src/dotnet-svcutil/lib/src/Shared/Options/UpdateOptions.cs
+++ b/src/dotnet-svcutil/lib/src/Shared/Options/UpdateOptions.cs
@@ -103,6 +103,16 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
                     this.Inputs[idx] = new Uri(relPath, UriKind.Relative);
                 }
             }
+
+            // Update references
+            for (int idx = 0; idx < this.References.Count; idx++)
+            {
+                var reference = this.References[idx];
+                if (reference.DependencyType == ProjectDependencyType.Project || reference.DependencyType == ProjectDependencyType.Binary)
+                {
+                    this.References[idx].ReferenceIdentity = PathHelper.GetRelativePath(reference.FullPath, optionsFileDirectory, out relPath) ? relPath : reference.FullPath;
+                }
+            }
         }
 
         public void ResolveFullPathsFrom(DirectoryInfo optionsFileDirectory)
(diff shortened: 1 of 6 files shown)


voroninp:
Any estimations for the merge? ;-)

src/dotnet-svcutil/lib/src/Shared/Options/UpdateOptions.cs
```diff
@@ -97,6 +97,16 @@ public void MakePathsRelativeTo(DirectoryInfo optionsFileDirectory)
                     this.Inputs[idx] = new Uri(relPath, UriKind.Relative);
                 }
             }
+
```
imcarolwang, line 100: Only binary dependency type is affected, reconstruct the reference with reference.AssemblyName writes name instead of path to the .json file.

src/dotnet-svcutil/lib/src/Shared/Options/UpdateOptions.cs
```diff
@@ -97,6 +97,16 @@ public void MakePathsRelativeTo(DirectoryInfo optionsFileDirectory)
...
+            //update referenced assembly path
+            for (int idx = 0; idx < this.References.Count; idx++)
+            {
+                var reference = this.References[idx];
+                if (reference.DependencyType.Equals(ProjectDependencyType.Binary))
+                {
+                    this.References[idx] = ProjectDependency.FromAssembly(reference.AssemblyName);
```
imcarolwang, line 107: Only binary dependency type is affected, reconstruct the reference with reference.AssemblyName ensures name instead of path get written to the .json file.

src/dotnet-svcutil/lib/src/CommandProcessorOptions.cs
```diff
@@ -627,7 +627,12 @@ private async Task ProcessReferencesOptionAsync(CancellationToken cancellationTo
                     {
                         for (int idx = this.References.Count - 1; idx >= 0; idx--)
                         {
-                            if (!references.Contains(this.References[idx]))
```
imcarolwang, line 630: Clarification for the change:

The purpose of this code piece is to remove any reference that exist in References but not exist in references (as some reference could have been removed after service was added) when updating service.
`References `is the original reference list
`references `is the resolved reference list of the running project

With the fix of this issue applied, the reference "name" has changed from full local path to assembly name, so the old reference item in References collection might not exactly equals to the reference in references collection due to the path change. To resolve this discrepancy, 
1. the containment check previously done with `(if(references.Contains(this.References[idx])))` is now performed by comparing the reference item's Name property for equality.
2. since only reference items present in the references list can be used for type reuse (as they should be available or resolved from the running project), we need to update the existing Reference (which could be with a local full path) with the reference item from the resolved references list. The assignment `this.References[idx] = selected;` does not add a new item to the list but updates the existing item in References with the corresponding reference (having the same name) from the resolved project references.
3. traversing the collection using array index ensures the shape of the References collection is maintained during the removal and update operations.

Issue #5499, opened 2024-04-12
Issues migrating WCF Client from .NET Standard to MultiTarget


bwilliams1:
**Describe the bug**
Over the last couple years we've migrated all our SOAP clients to Connected Services WCF Client on .NET standard 2.0 referencing the System.ServiceModel.* 4.*.* NuGet packages for consumption in a .NET framework 4.8 solution with a long term goal of targeting the latest versions of .NET (8 and beyond). I recently stumbled upon the [drop of support of .NET standard 2.0](https://learn.microsoft.com/en-us/dotnet/core/compatibility/wcf-client/6.0/net-standard-2-support) for WCF Client and started to understand what that means for our team. The docs linked in the previous sentence lists multi targeting as the solution.  

I started noticing some issues with the migration to a multi target solution.

**To Reproduce**

1. Create a .net standard 2.0 library project.
2. Add a Connected Services WCF Service Client using the svc-util "wizard" within Visual Studio 2022 Version 17.8.7
3. This will Generate a Reference.cs file with an implementation for `CloseAsync()`
4. update `<TargetFramework>netstandard2.0</TargetFramework>` to `<TargetFrameworks>net48;net8.0</TargetFrameworks>`
5. replace all the System.ServiceModel PackageReference nodes with conditional statements as seen below:
```
	<ItemGroup Condition=" '$(TargetFramework)' == 'net48' ">
		<Reference Include="System.ServiceModel" />
	</ItemGroup>
	<ItemGroup Condition=" '$(TargetFramework)' == 'net8.0' ">
		<PackageReference Include="System.ServiceModel.Primitives" Version="8.0.0" />
		<PackageReference Include="System.ServiceModel.Http" Version="8.0.0" />
	</ItemGroup>
```
7. build the solution
8. observe CS1061 build failures `'XXXClient' does not contain a definition for 'CloseAsync' and no accessible extension method 'CloseAsync' accepting a first argument of type 'XXXClient' could be found (are you missing a using directive or an assembly reference?)`

**Expected behavior**
given the Microsoft documentation to simply "multi target" any .NET standard 2.0 projects, I expected a successful build. 

**Screenshots**
n/a

**Additional context**

I also noticed if I create a brand new .net8.0 class library and try to add a WCF Client, it's still using the 4.10.* version of the System.ServiceModel.* instead of the 8.0.0 versions. In this scenario, it does not generate the `CloseAsync` methods that failed the build in the multi target scenario.

[...]

PR #5519, opened 2024-04-30
dotnet-svcutil: optimize the referenced wcf package version and add test.

imcarolwang:
Addresses #5499. 

Description:
1. Correctly resolve netstandard2.0 if the TFM moniker string was passed from VS (WCF CS Tool)
2. Create clear table mapping between .NET target version and wcf package version, for the various wcf releases which support netstandard2.0, keep the latest version 4.10.* only; update CloseAsync() generation logic to generate it for TFM less than 6.0, and update test baselines
3. Add only the common wcf references for version 8.0 packages, update corresponding test baselines
4. Improve wcf assembly reference condition for .NET framework target (use $(TargetFramework.StartsWith('net4') instead of getting the first .NET target and check for equality, this will fix bug when there're multiple .NET framework 4.x targets)

### DIFF
--- a/src/dotnet-svcutil/lib/src/CodeDomFixup/CodeDomVisitors/AddAsyncOpenClose.cs
+++ b/src/dotnet-svcutil/lib/src/CodeDomFixup/CodeDomVisitors/AddAsyncOpenClose.cs
@@ -12,40 +12,6 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
 {
     internal class AddAsyncOpenClose : ClientClassVisitor
     {
-        //default behavior: won't generate CloseAsync()
-        private bool _generateCloseAsync = false;
-        private bool _addCondition = false;
-
-        public AddAsyncOpenClose(CommandProcessorOptions options)
-        {
-            if (options.Project != null)
-            {
-                Version lowestNetcoreVer = TargetFrameworkHelper.GetLowestNetCoreVersion(options.Project.TargetFrameworks);
-                bool containsNet6OrGreater = options.Project.TargetFrameworks.Any(t => TargetFrameworkHelper.IsSupportedFramework(t, out FrameworkInfo netfxInfo) && netfxInfo.IsDnx && netfxInfo.Version.Major >= 6);
-                bool containsNetFx = options.Project.TargetFrameworks.Any(t => TargetFrameworkHelper.IsSupportedFramework(t, out FrameworkInfo netfxInfo) && !netfxInfo.IsDnx);
-
-                // netfx or lowest.net core < 6.0
-                if (lowestNetcoreVer == null || (lowestNetcoreVer != null && lowestNetcoreVer.Major < 6))
-                {
-                    _generateCloseAsync = true;
-                    if (containsNet6OrGreater)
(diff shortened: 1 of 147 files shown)


src/dotnet-svcutil/lib/src/CodeDomFixup/CodeDomVisitors/AddAsyncOpenClose.cs
```diff
@@ -7,52 +7,43 @@
...
-                        if (vers.Length > 1)
-                        {
-                            Version v = new Version(int.Parse(vers[0]), int.TryParse(vers[1], out int minor) ? minor : 0);
-                            // For .NETCore targetframework found in the referenced table, generate CloseAsync() when WCF package version is less than 4.10
-                            if (v.CompareTo(new Version(4, 10)) < 0)
-                            {
-                                _generateCloseAsync = true;
-                            }
-                        }
+                        _addCondition = true;
```
HongGit, line 33: It sees that _addCondition and _generateCloseAsync are duplicate?  Could they be merged into one variable?

src/dotnet-svcutil/lib/src/Shared/TargetFrameworkHelper.cs
```diff
@@ -15,55 +15,32 @@ internal class TargetFrameworkHelper
...
+         });
+
+        public static ReadOnlyDictionary<Version, Version> NetCoreToWCFPackageReferenceVersionMap { get; } = new ReadOnlyDictionary<Version, Version>(new SortedDictionary<Version, Version>
+         {
+             {new Version("2.0"), new Version("2.0") },
+             {new Version("2.1"), new Version("2.0") },
+             {new Version("2.2"), new Version("2.0") },
+             {new Version("3.0"), new Version("2.0") },
+             {new Version("3.1"), new Version("2.0") },
+             {new Version("5.0"), new Version("2.0") },
```
HongGit, line 29: If I understand this mapping correctly, this is from Net Core to WCF Client, right?  If so, is the 3.1 and 5.0 line correct?

src/dotnet-svcutil/lib/src/CodeDomFixup/CodeDomVisitors/AddAsyncOpenClose.cs
imcarolwang: The variable _generateCloseAsync controls whether the CloseAsync() code should be generated, the variable _addCondition determines whether a condition should be added to the generated CloseAsync method in scenarios involving multiple target frameworks. These variables serve different purposes and are not duplicates.

src/dotnet-svcutil/lib/src/Shared/TargetFrameworkHelper.cs
imcarolwang: Yes, this mapping is from .NET Core to WCF Client, with the mapping value indicating the WCF Client version as specified in another table. For the 3.1 and 5.0 lines, the tag Version("2.0") corresponds to the 4.10 WCF package. Therefore, the data for these lines is correct.

PR #5521, opened 2024-04-30
[release/6.1] dotnet-svcutil: port 2 more PRs from main branch

imcarolwang:
Cc: @HongGit.


src/dotnet-svcutil/lib/src/Bootstrapper/SvcutilBootstrapper.cs
```diff
@@ -123,22 +123,17 @@ internal async Task GenerateProjectAsync(bool keepBootstrapperDir, ILogger logge
...
-                //if (isSupportedTFM && frameworkInfo.IsKnownDnx)
-                //{
-                //    if (frameworkInfo.Name == FrameworkInfo.Netstandard)
-                //    {
-                //        targetFramework = FrameworkInfo.Netcoreapp + TargetFrameworkHelper.NetStandardToNetCoreVersionMap[frameworkInfo.Version];
-                //    }
-                //    this.MSBuildProj.TargetFramework = targetFramework;
-                //}
+                var targetFramework = frameworkInfo.FullName;
+                if (isSupportedTFM && frameworkInfo.Name != FrameworkInfo.Netstandard && frameworkInfo.Version.CompareTo(new Version(6, 0)) >= 0)
```
HongGit, line 127: We will still support net standard.

imcarolwang: netstandard support remains both before and after the change. The bootstrapper project serves the purpose of type reuse and is required to be an executable target, hence it cannot be netstandard. With this update, the bootstrapper project now targets the default .NET Core version installed on the machine when the user project targets netstandard. Previously, the bootstrapper project inadvertently used the default .NET Core version installed on the machine regardless of the target of the user project, which was unintended.

PR #5543, opened 2024-05-13
dotnet-svcutil: update netnamedpipe pkg for metadata downloads and improve exception handling for WCFCSTools

imcarolwang:


src/dotnet-svcutil/lib/src/internalAssets/net6.0/Microsoft.Svcutil.NamedPipeMetadataImporter.dll
HongGit, line 1: Why are these needed?

imcarolwang: We utilize reflection to retrieve metadata for the netnamedpipe service; the DLLs are present to support this reflective approach.

src/dotnet-svcutil/lib/Microsoft.Svcutil.NamedPipeMetadataImporter/Microsoft.Svcutil.NamedPipeMetadataImporter.csproj
```diff
@@ -1,14 +1,40 @@
...
     <AssemblyName>$(MSBuildProjectName)</AssemblyName>
     <LangVersion>latest</LangVersion>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
 	</PropertyGroup>
 	<ItemGroup>
-		<PackageReference Include="System.ServiceModel.NetNamedPipe" Version="6.0.*" Condition="'$(TargetFramework)' == 'net6.0'">
-			<PrivateAssets>contentfiles;analyzers;build</PrivateAssets>
+		<PackageReference Include="System.ServiceModel.NetNamedPipe" Version="6.2.*" Condition="'$(TargetFramework)' == 'net6.0'">
+			<IncludeAssets>all</IncludeAssets>
 		</PackageReference>
```
HongGit, line 15: since this 8.0 version of dotnet-svcutil.exe, only need to reference 8.0 version of WCF Client library.

src/dotnet-svcutil/lib/Microsoft.Svcutil.NamedPipeMetadataImporter/Microsoft.Svcutil.NamedPipeMetadataImporter.csproj
```diff
@@ -1,14 +1,40 @@
...
+    <DotNetUseShippingVersions>true</DotNetUseShippingVersions>
+    <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
+    <IncludeSymbols>true</IncludeSymbols>
+    <IncludeSource>false</IncludeSource>
+  </PropertyGroup>
+  <Target Name="CopyPackageDlls" AfterTargets="Build">
+    <ItemGroup>
+      <PublishedDlls Include="$(OutputPath)\*.dll" />
+    </ItemGroup>
+    <Copy SourceFiles="@(PublishedDlls)" DestinationFolder="$(CustomOutputFolder)\internalAssets\net6.0" Condition="$(OutputPath.Contains('net6.0'))" />
```
HongGit, line 36: same reason as above.

src/dotnet-svcutil/lib/src/Metadata/ServiceDescriptor.cs
```diff
@@ -107,7 +107,7 @@ public async Task ImportMetadataAsync(Action<WsdlImporter> onWsdlImporterCreated
                 }
                 else
                 {
-                    tfn = "net6.0";
+                    tfn = Environment.Version.Major >= 8 ? "net8.0" : "net6.0";
```
HongGit, line 110: Is this only relying on .NET 8.0?  If so, throw some kind of exception?  Is this protected somewhere else?  If so, at least add a comment.

src/dotnet-svcutil/lib/src/Metadata/ServiceDescriptor.cs
```diff
@@ -116,27 +116,35 @@ public async Task ImportMetadataAsync(Action<WsdlImporter> onWsdlImporterCreated
                 Type type = assembly.GetType("Microsoft.Tools.ServiceModel.Svcutil.NamedPipeMetadataImporter");
                 if (type != null)
                 {
-                    object typeInstance = Activator.CreateInstance(type, null);
-                    MethodInfo methodInfo = type.GetMethod("GetMetadatadataAsync", BindingFlags.Public | BindingFlags.Instance);
-                    var xmlReader = await (Task<System.Xml.XmlReader>)methodInfo.Invoke(typeInstance, new object[] { MetadataUrl });
+                    try
+                    {
+                        object typeInstance = Activator.CreateInstance(type, null);
+                        MethodInfo methodInfo = type.GetMethod("GetMetadatadataAsync", BindingFlags.Public | BindingFlags.Instance);
```
HongGit, line 122: Is this necessary?  If necessary, at least add comment.

src/dotnet-svcutil/lib/Microsoft.Svcutil.NamedPipeMetadataImporter/Microsoft.Svcutil.NamedPipeMetadataImporter.csproj
imcarolwang: If only reference the 8.0 NetNamedPipe package, it can cause issues for user project targeting net6.0, particularly when adding net.pipe service reference with type reuse option enabled. In such case, the associated bootstrapper project will also target net6.0, causing problem situation of loading the 8.0 NetNamedPipe package assembly in a net6.0 runtime environment.

We have two potential solutions::
1. Use the 6.2 named pipe package for net6.0 target and 8.0 named pipe package for the net8.0 target.
2. Use the 6.2 named pipe package for both net6.0 and net8.0 targets.

The current implementation adopts the 1st option.

src/dotnet-svcutil/lib/src/Metadata/ServiceDescriptor.cs
imcarolwang: It is necessary. As per our original plan, we are utilizing a reflection approach to download metadata for the named pipe service. The method entry point for this is GetMetadataAsync, defined within Microsoft.Svcutil.NamedPipeMetadataImporter.csproj.

imcarolwang:
Closing. This will be addressed in separate PRs #5582 and #5604.

PR #5544, opened 2024-05-14
dotnet-svcutil: update metadataSourceFiles for URL input to correctly compute hash for caching ServiceDescriptor

imcarolwang:
This update addresses a bug in the WCF Connected Services Tool where switching between different service URL addresses would consistently cache the initial input, resulting in inaccurate display within the Services tree view.


HongGit (approved):

HongGit:
@imcarolwang can you please see if this change need to be ported to release/6.1 branch?

imcarolwang:
> @imcarolwang can you please see if this change need to be ported to release/6.1 branch?

This change affects WCF Connected Service tool only, so there's plan to insert release/6.1 release to VS, then it should be ported over, otherwise, I think it could be ignored.

PR #5563, opened 2024-06-04
Update Mac queue to osx.14

HongGit:


HongGit:
Close this as we are using Helix hosted pool, not Azure pipeline hosted pool agent.

Issue #5561, opened 2024-05-31
IErrorHandler missing from nuget version 8.0.0


Pixel452:
Create new .NET 8 class library project
Add System.ServiceModel.Primitives from nuget
IErrorHandler is missing from System.ServiceModel.Dispatcher namespace
Looking at the source code and the documentation, it should be there, but it's not, am I missing something?

mconnew:
Transferred issue to the dotnet/wcf repo. 
What you're missing is that we have a reference assembly which doesn't contain all the api's as some aren't suitable (e.g. service only usage) for the client. We have some extra code in the implementation assembly due to in the early days prioritizing speed of porting over features over stripping out code which wasn't appropriate to port. Some api implementations technically got ported, but as we didn't add tests to make sure they were functioning, and they were ported as a side effect of porting some other feature, we didn't expose them in the reference assembly.  
This api does seem like a reasonable api to expose. We should also expose `ClientRuntime.EnableFaults` at the same time.  
@imcarolwang, could you expose `IErrorHandler`, `ChannelDispatcher.ErrorHandlers`, and `ClientRuntime.EnableFaults` and add tests for them. This is almost certainly going to need a scenario test. You can likely use an existing service endpoint as the scenario is a callback contract method throws an exception. You might need to look at .NET Framework WCF tests to see all the facets that need testing with these api's. Let me know if you need some pointers beyond that.  

@Pixel452, what's the scenario that you are using an IErrorHandler for? It's going to take a little while before new api's are available in a public build, and there might be an alternative way to achieve the same thing using existing public api's.

Pixel452:
I'm porting a .NET Framework WPF app to .NET and there is an implementation of AsyncCommand that uses IErrorHandler, given that at the moment I'm still evaluating and testing if all the needed libraries and code are portable to .NET, I removed this feature, if everything else works, I'll go back to this. So, I can wait.

Thanks

imcarolwang:
> @imcarolwang, could you expose `IErrorHandler`, `ChannelDispatcher.ErrorHandlers`, and `ClientRuntime.EnableFaults` and add tests for them. This is almost certainly going to need a scenario test. You can likely use an existing service endpoint as the scenario is a callback contract method throws an exception. You might need to look at .NET Framework WCF tests to see all the facets that need testing with these api's. Let me know if you need some pointers beyond that.

[...]

PR #5593, opened 2024-06-27
Expose IErrorHandler and add test case.

imcarolwang:
For issue #5561.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
+++ b/src/System.Private.ServiceModel/tests/Common/Scenarios/Endpoints.cs
@@ -766,6 +766,11 @@ public static partial class Endpoints
     {
         get { return GetEndpointAddress("DuplexCallbackDebugBehavior.svc/tcp", protocol: "net.tcp"); }
     }
+
+    public static string DuplexCallbackErrorHandler_Address
+    {
+        get { return GetEndpointAddress("DuplexCallbackErrorHandler.svc/tcp", protocol: "net.tcp"); }
+    }
     #endregion net.tcp Addresses
 
     #region net.pipe Addresses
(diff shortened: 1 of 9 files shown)


src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
```diff
@@ -1834,6 +1834,7 @@ internal ClientRuntime() { }
         public System.Type ContractClientType { get { return default; } set { } }
         public string ContractName { get { return default; } }
         public string ContractNamespace { get { return default; } }
+        internal bool EnableFaults { get { return default; } set { } }
```
mconnew, line 1837: I'm trying to understand why you added this to the ref assembly as it's internal?

mconnew:
Everything looks good except the question around the internal EnableFaults. With the recent changes I made to support netstandard2.0 and netfx, we have multiple ref assemblies. When you rebase, only add to the netcoreapp ref assembly, we aren't updating the netstandard2.0 reference assembly beyond what it currently is as it's only there to support those using that capability from the 4.10.x packages.

src/System.ServiceModel.Primitives/ref/System.ServiceModel.Primitives.cs
imcarolwang: I think the request at https://github.com/dotnet/wcf/issues/5561#:~:text=time.%0A%40imcarolwang%2C-,could%20you%20expose%20IErrorHandler%2C%20ChannelDispatcher.ErrorHandlers%2C%20and%20ClientRuntime.EnableFaults,-and%20add%20tests%20for suggests adding `ClientRuntime.EnableFaults`. It is internal in the src implementation https://github.com/dotnet/wcf/blob/4e6e777613d380d9101969d15c8f08fc543a22fd/src/System.ServiceModel.Primitives/src/System/ServiceModel/Dispatcher/ClientRuntime.cs#L225, so I brought the same to the ref project.

Would you recommend that we update it?

imcarolwang: @mconnew Thank you for letting me know. I’ve updated the PR to remove the exposure of "EnableFaults." Following your suggestion, I’ve rebased the branch and made changes only to the netcoreapp reference assembly by modifying the System.ServiceModel.Primitives.Netcoreapp.cs file. Would you mind taking another look when you have a chance to confirm if everything looks good now?

mconnew (approved):

PR #5604, opened 2024-07-11
Fix for WCFCSTools issue - invalid namedPipe URL throws uncatchable exception in VS.

imcarolwang:
**Description:**

WCF CS tool invokes dotnet-svcutil-lib, which uses reflection to download metadata from a net.pipe URL. If an exception occurs during this reflection process and it isn't properly caught, Visual Studio may crash.

The fix is wrapping the reflection code's return type and any potential exceptions with `TaskCompletionSource<TheType> `instance. This approach ensures that exceptions are available to WCF CS tool, preventing VS crash and providing a more user-friendly exception message.


HongGit (approved):
Is it possible to add a testcase for this change?

imcarolwang:
> Is it possible to add a testcase for this change?

I found that the unhandled exception from dotnet-svcutil-lib will be output to the process running the command line tool, whether the fix is applied or not, and cmd.exe does not crash. I think to verify the fix, we will need to invoke the code from WCF CS Tool in VS. We will add this test scenario when testing the WCF CS Tool.

PR #5640, opened 2024-08-31
Fix sync hang issue after async task method call (#5634)

organich:
Fix #5634


mconnew:
I'm not sure what you're trying to do here. You've tried to merge in an old uwp branch into main, but have referenced a PR which already fixes the problem that your title mentions. I'm also not seeing any code changes in this PR from you. Whatever it is your are trying to do, this PR is not going to achieve it. Closing this PR.

PR #5653, opened 2024-09-26
dotnet-svcutil: refine TFX resolution and the referenced WCF versions, and update test baselines.

imcarolwang:
Split PR #5519 - Part 2

**Description**:

1. Correctly resolve the target framework moniker string passed from the WCF CS Tool when it is netstandard.
2. Add WCF references based on the user's project target frameworks, accounting for .NET Framework, .NET Core, and .NET Standard, including both single-target and multi-target scenarios.
3. Remove WCF references older than version 4.10; retain only the three common WCF packages for the 8.* version, update the test baselines accordingly.

### DIFF
--- a/src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
+++ b/src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
@@ -43,7 +43,9 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
         }
 
         private List<string> _targetFrameworks = new List<string>();
+        private List<string> _endOfLifeTargetFrameworks = new List<string>();
         public IEnumerable<string> TargetFrameworks { get { return _targetFrameworks; } }
+        public IEnumerable<string> EndOfLifeTargetFrameworks { get { return _endOfLifeTargetFrameworks; } }
 
         private string _runtimeIdentifier;
         public string RuntimeIdentifier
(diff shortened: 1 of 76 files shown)


src/dotnet-svcutil/lib/src/Shared/MSBuildProj.cs
```diff
@@ -43,7 +43,9 @@ public string TargetFramework
         }
 
         private List<string> _targetFrameworks = new List<string>();
+        private List<string> _endOfLifeTargetFrameworks = new List<string>();
         public IEnumerable<string> TargetFrameworks { get { return _targetFrameworks; } }
+        public IEnumerable<string> EndOfLifeTargetFrameworks { get { return _endOfLifeTargetFrameworks; } }
```
HongGit, line 48: Is this needed for dotnet-svcutil tool?

imcarolwang: It is used in dotnet-svcutil-lib. I have updated the visibility to internal.

src/dotnet-svcutil/lib/src/Shared/TargetFrameworkHelper.cs
```diff
@@ -13,84 +13,27 @@ namespace Microsoft.Tools.ServiceModel.Svcutil
...
-             {new Version("1.5"), new Version("1.0") },
-             {new Version("1.6"), new Version("1.0") },
-             {new Version("1.6.1"), new Version("1.1") },
-             {new Version("2.0"), new Version("2.0") }
-         });
-
-        internal static SortedDictionary<Version, List<ProjectDependency>> NetCoreVersionReferenceTable = new SortedDictionary<Version, List<ProjectDependency>>
+        private static readonly Version s_minEOFNetVersion = new Version("5.0");
+        private static readonly Version s_maxEOFNetVersion = new Version("7.0");
+        internal static readonly List<string> s_supportedVersions = new List<string>() { "8.0", "9.0", "10.0" };
```
HongGit, line 18: suggest to use currentSupportedVersions.

imcarolwang: Updated.

HongGit (approved):

PR #5655, opened 2024-09-30
Make _RunWithCoreWcfService a variable again

mconnew:
Due to Azure DevOps not trusting my GitHub account to make changes to the yml for the corewcf ci run, the PR triggered build will fail due to not being able to use the access token. I've kicked off a manual build for the same commit which can use the access token (because Azure DevOps trusts my Azure authenticated account). You can verify this is doing the right thing by seeing that the step "Update 
_updateService variable" is skipped. This is the step which populates the service url and sets a variable telling a later step to update the test servers to be in line with the PR. We still have tests failing due to still ramping up CoreWCF for tests, so looking at this step is the easiest way to check if the _RunWithCoreWcfService variable is working as expected.  
https://dev.azure.com/dnceng-public/public/_build/results?buildId=822804&view=results

### DIFF
--- a/azure-pipelines-arcade-PR.yml
+++ b/azure-pipelines-arcade-PR.yml
@@ -1,9 +1,3 @@
-parameters:
-- name: _RunWithCoreWcfService
-  displayName: Run tests using CoreWCF service
-  type: boolean
-  default: false
-
 # trigger ci builds for merged PRs into listed branches
 # Setting batch to true, triggers one build at a time.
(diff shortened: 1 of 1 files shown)


mconnew:
I pushed one more update since the manual pipeline run, but the only change was to update the commit comment as I originally tried to make the parameter default dynamic, but that didn't work. Instead I got it working with a variable. The changes in the commit are identical.

HongGit (approved):

PR #5663, opened 2024-10-12
Stop splitting the UPN credential into username and domain parts unless specified

imcarolwang:


mconnew:
LGTM

mconnew:
@imcarolwang, can you also check the behavior before explicitly setting the switch to validate the default too?

imcarolwang:
> @imcarolwang, can you also check the behavior before explicitly setting the switch to validate the default too?

Hi @mconnew, I’ve added the check for the default behavior before setting the switch value. Note: in the followed test variations, the switch is set using reflection since s_enableLegacyUpnUsernameFix is internal static and the AppContext.TryGetSwitch and assignment is invoked only once during the application's execution.

mconnew (approved):

Issue #2999, opened 2018-07-05
NRE when response has Content-Type but empty body


IGx89:
When a SOAP service returns no content but has a Content-Type set, WCF throws a NullReferenceException in HttpResponseMessageHelper. Leaving off Content-Type works fine.

**Example HTTP response that reproduces the issue**
```
HTTP/1.1 500
Content-Length: 0
Content-Type: text/xml; charset=UTF-8
Date: Thu, 05 Jul 2018 21:10:34 GMT
Connection: close
```

**Expected Result**
> System.ServiceModel.CommunicationException : The server did not provide a meaningful reply; this might be caused by a contract mismatch, a premature session shutdown or an internal server error."

**Actual Result**
> System.NullReferenceException : Object reference not set to an instance of an object.

The NRE happens due to inputStream being null in HttpResponseMessageHelper.DecodeBufferedMessageAsync when it attempts to dispose it in the finally block.

The root cause however looks to be in ValidateContentTypeAsync. It doesn't properly handle this situation. It returns true (last line in the method) if the content type has a value and content length is 0, when it should be returning false. Its return value is saved in a `hasContent` variable in ParseIncomingResponse, so it definitely looks like it should be returning false if there's no content.

StephenBonikowsky:
Thanks for reporting this @IGx89 we'll have someone take a look next week.

skyflyer:
Would it be possible to fix this bug anytime soon? Since it is a very trivial fix and we'd like to avoid using our own custom build. The patch for the fix is below.

```patch
diff --git a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
index 0c580539..f8407996 100644
--- a/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
+++ b/src/System.Private.ServiceModel/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
@@ -42,6 +42,7 @@ namespace System.ServiceModel.Channels
             bool hasContent = await ValidateContentTypeAsync();
             Message message = null;
 
+
             if (!hasContent)
             {
                 if (_encoder.MessageVersion == MessageVersion.None)
@@ -136,6 +137,11 @@ namespace System.ServiceModel.Channels
[...]

PR #5678, opened 2024-11-09
Handle null-ref during HttpResponseMessageHelper.DecodeBufferedMessageAsync()

wrall:
Handle current null-ref situation that can occur when Content-Type is present but Content-Length is 0.

Addresses #2999

### DIFF
--- a/src/System.ServiceModel.Http/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
+++ b/src/System.ServiceModel.Http/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
@@ -148,7 +148,7 @@ namespace System.ServiceModel.Channels
                 return false;
             }
 
-            if (_contentLength == 0)
+            if (_contentLength <= 0)
             {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


src/System.ServiceModel.Http/src/System/ServiceModel/Channels/HttpResponseMessageHelper.cs
```diff
@@ -144,25 +144,30 @@ private async Task<bool> ValidateContentTypeAsync(TimeoutHelper timeoutHelper)
                 {
                     throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(new ProtocolException(SR.HttpContentTypeHeaderRequired));
                 }
+
                 return false;
             }
-            else if (_contentLength != 0)
+
+            if (_contentLength <= 0)
```
mconnew, line 151: _contentLength will have a value of -1 if the content length header is missing. They could be using chunked encoding, or it could be an HTTP/1.0 response where it closes the connection after the response has been sent. So this isn't correct as it will falsely return that there is no content when that might not be the case. At this point, we know there's a content type provided, so we only want to return false if _contentLength is exactly 0. If it's -1, we don't have enough info. To do this correctly needs quite a bit of refactoring. In the case of a -1 value, we need to attempt to pre-read a single byte from the stream. When there's no actual content (HTTP/1.0 has closed the connection, or it's chunked encoded with just a 0 length chunk), we will get 0 bytes read and know there's no content.

PR #5691, opened 2024-11-19
Fix S.SM.Http ref assembly for net8.0

### LINKED ISSUE #5690: BasicHttpSecurity fails with 8.1
This worked with 8.0.0:

```csharp
 var binding = new BasicHttpsBinding()
 {
     MaxReceivedMessageSize = int.MaxValue,
     ReaderQuotas = System.Xml.XmlDictionaryReaderQuotas.Max,
     Security = new BasicHttpsSecurity()
     {
         Mode = BasicHttpsSecurityMode.Transport,
     },
 };
```

No longer works with 8.1.0 in .NET 8 library

`
'BasicHttpsSecurity' does not contain a constructor that takes 0 arguments
`

### LINKED ISSUE #5692: Description nuget package is missing ref assembly in 8.1
**Describe the bug**
System.Web.Services.Description nuget package now has a ref folder in 8.1. But its empty so essentially has no api surface area.

**To Reproduce**
1. Add the System.Web.Services.Description nuget package to a new project
2. The namespaces System.Web.Services.Description and System.Web.Services.Configuration can not be found

**Expected behavior**
Same as 8.0. Namespaces and classes are public

mconnew:
Http ref assembly had been generated from netstandard dummy implementation instead of real implementation.

Some properties/methods/types shouldn't have been public so cleaned up in implementation so ref is correct

Fixes #5690
Fixes #5692


ErikEJ:
@HongGit When is next release planned?

PR #5718, opened 2024-12-18
Upgrade the version of the dotnet-svcutil-lib dependencies.

imcarolwang:
NuGet.ProjectModel 6.11.1 has a transitive dependency on System.Formats.Asn1 6.0.0, which contains a critical security vulnerability.

CC: @HongGit , @mconnew


HongGit:
@imcarolwang please review all packages referenced are using the latest version!

imcarolwang:
> @imcarolwang please review all packages referenced are using the latest version!

Thanks for your review. I’ve checked the project, and the current dependencies are the most recent versions that are compatible with the project.

HongGit (approved):

Issue #4230, opened 2020-04-03
dotnet-svcutil --update should not reorder the class definitions


EricStG:
**Describe the bug**
When updating a service reference based off the same exact WSDL, the order of the classes in the `Reference.cs` file changes, which makes it less version control friendly. It also makes it harder to determine what changed when the WSDL did change.

**To Reproduce**
1. Create a new service reference from a WSDL
2. Make a copy of the `Reference.cs` file
3. Run `dotnet-svcutil --update` with your service reference
4. Compare the `Reference.cs` files generated in step 2 and 3.

**Expected behavior**
When running `dotnet-svcutil --update` on a WSDL that has not change, the order of the classes in `Reference.cs` should stay the same.

**Screenshots**
N/A

**Additional context**
using dotnet-svcutil 2.0.1

dasetser:
Thanks for the suggestion. This shouldn't impact functionality, but I can see why this would be nicer for version control and comparisons. We'll look into what we can do in a future version.

PR #5735, opened 2025-01-29
fix: import derived types order

kaqq:
fix for: https://github.com/dotnet/wcf/issues/4230


mconnew:
Long term this approach is going to be fragile. At the moment, dotnet-svcutil has an internal copy of a bunch of stuff that is in the runtime. This is because of historical reasons, lots of things were missing from the implementation in runtime. Over time they have added it back. Although not scheduled yet, we do have the plan to drop the internal framework fork implementation. Making these changes here will regress again when that happens.  
Instead of changing it inside of the schema importer, can something be done to make the schemas added to the schema set in a deterministic way? My guess as to why it's non deterministic is something is using a Dictionary to store things with a string key. Dictionary uses a special string comparer which doesn't use the same hash code values from one process to the next. If this is the cause, if we can identify the dictionary, then we could instantiate it with a deterministic string comparer.  
Whatever the cause, the sorting/change should happen outside of the schema importers and in the dotnet-svcutil implementation (or even forked framework WCF code).

PR #5741, opened 2025-02-11
Fixes scenario where library references packages and NetFx app consumes library

### LINKED ISSUE #5722: SecurityToken is not found by .NET Framework project after upgrade of System.ServiceModel.Primitives from 4.10.3 to 8.1.1 in netstandard2.0 project
**Describe the bug**
Version 8.1 of the System.ServiceModel.Primitives brings back netstandard and .NET Framework support to the System.ServiceModel.Primitives package, after it was removed in version 6.0.0. However, after updating the version from 4.10.3 to 8.1.1, when building a .NET Framework project consuming a .NET standard project, exposing the SecurityTopken type in a return parameter, the .NET framework project fails to build with 
> Error CS7069: Reference to type 'SecurityToken' claims it is defined in 'System.ServiceModel.Primitives', but it could not be found

**To Reproduce**
Build the attached solution: [SecurityTokenNotFound.zip](https://github.com/user-attachments/files/18298108/SecurityTokenNotFound.zip)

**Expected behavior**
The solution should build correctly, as it does when downgrading the System.ServiceModel.Primitives package reference to 4.10.3

mconnew:
Also fixes scenario where library references older 4.x retired packages (Duplex/Security) and NetFx app references the library and newer 8.x packages.
Fixes #5722


ericstj:
This PR did fix most of the compatibility problems found by APICompat - but there are still a few.  Have a look at https://github.com/dotnet/wcf/pull/5742/files?file-filters%5B%5D=.xml&show-viewed-files=true to see the remaining compatibility problems.

src/System.ServiceModel.Http/ref/NetFxTypeForwards.cs
```diff
@@ -0,0 +1,36 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
ericstj, line 1: I think you can delete these `ref` specific files now.

ericstj (approved):

mconnew:
Closing this PR as everything was included in #5742

PR #5742, opened 2025-02-12
Enable Package Validation - suppress all problems

ericstj:
This is with the 8.1.0 packages as a baseline.  Could be that other problems exist if you select an older baseline.

CC @ViktorHofer


src/System.ServiceModel.Federation/src/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- https://learn.microsoft.com/dotnet/fundamentals/package-validation/diagnostic-ids -->
+<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:System.ServiceModel.Federation.WSTrustTokenParameters.get_KeySize</Target>
+    <Left>lib/net8.0/System.ServiceModel.Federation.dll</Left>
```
mconnew, line 7: I think this is another bug in API compat tooling. WSTrustTokenParameters has the new keyword because when the class was first introduced, the base class didn't have a KeySize property. The base class did have a KeySize property exposed on NetFx and was brought back, so the definition in WSTrustTokenParameters needed the new keyword so that it didn't become binary incompatible. This hasn't changed single 6.1.0.

eng/Versions.props
```diff
@@ -6,6 +6,7 @@
     <MajorVersion>8</MajorVersion>
     <MinorVersion>1</MinorVersion>
     <PatchVersion>1</PatchVersion>
+    <PackageValidationBaselineVersion>$(MajorVersion).$(MinorVersion).$([MSBuild]::Subtract($(PatchVersion), 1))</PackageValidationBaselineVersion>
     <WcfAssemblyVersion>8.1.1.0</WcfAssemblyVersion>
```
mconnew, line 10: The next version will be 8.1.2. It looks like @HongGit doesn't update this version number until we're ready to build, so the comparison should be with 8.1.1 and not 8.1.0.

src/System.ServiceModel.Http/src/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,223 @@
...
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:System.ServiceModel.WSHttpBindingBase.set_EnvelopeVersion(System.ServiceModel.EnvelopeVersion)</Target>
+    <Left>ref/net8.0/System.ServiceModel.Http.dll</Left>
+    <Right>ref/net8.0/System.ServiceModel.Http.dll</Right>
+    <IsBaselineSuppression>true</IsBaselineSuppression>
+  </Suppression>
+  <Suppression>
+    <DiagnosticId>CP0002</DiagnosticId>
+    <Target>M:System.ServiceModel.Channels.HttpTransportBindingElement.get_Proxy</Target>
+    <Left>ref/netstandard2.0/System.ServiceModel.Http.dll</Left>
```
mconnew, line 196: This looks like a genuine problem, this api is listed in the netstandard reference assembly but isn't public on .NET Framework, so should only be in our net8 reference assembly.

src/System.Web.Services.Description/src/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- https://learn.microsoft.com/dotnet/fundamentals/package-validation/diagnostic-ids -->
+<Suppressions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+  <Suppression>
+    <DiagnosticId>CP0008</DiagnosticId>
+    <Target>T:System.Web.Services.Configuration.XmlFormatExtensionAttribute</Target>
+    <Left>ref/netstandard2.0/System.Web.Services.Description.dll</Left>
```
mconnew, line 7: @ericstj, do you know what's going on with these three? As far as I can tell, these are correct. The CP0008 error means a base interface was removed from the interface hierarchy from one of the compared sides. Is this caused by Attribute no longer implementing the interface `_Attribute`? Is this an API compat tool bug as the interface is defined on Attribute, which is external to these packages so isn't a difference between WCF Client packages.

ericstj: I'll have a closer look at this and the others.  The logged messages give a bit more context - @ViktorHofer has improved human readability of the suppression files in later 10.0 releases.

A lot of times you still need to look at the metadata to understand what's going on.  I'll have a review and add detail by mid-day tomorrow.

src/System.ServiceModel.Federation/src/CompatibilitySuppressions.xml
ericstj: Yeah, it does seem to be off:
```
CP0002: Member 'System.Nullable<int> System.ServiceModel.Federation.WSTrustTokenParameters.KeySize.get' exists on [Baseline] lib/net8.0/System.ServiceModel.Federation.dll but not on lib/netstandard2.0/System.ServiceModel.Federation.dll
```

This has to do with comparing the previous `net8.0` asset with the new `netstandard2.0` asset.  You'll only face this issue across this one baseline comparison.  Details here https://github.com/dotnet/sdk/issues/46834.

Issue #5736, opened 2025-01-30
Generate Sync only proxy code using dotnet-svcutil ?


Samba143:
While generating service reference using either IDE or command line using dotnet-svcutil, observed that generated code by default generates Async methods and we have option to include Sync calls.

How can I use the tool to generate only Sync calls? 
We would like to avoid /minimize WCF client code and just would like to migrate to .Net 8

Any suggestions plz

PR #5751, opened 2025-03-05
[dotnet-svcutil] Add option to support generating sync operations only.

imcarolwang:
For issue #5736.


src/dotnet-svcutil/lib/src/Shared/Options/UpdateOptions.cs
```diff
@@ -24,6 +24,7 @@ internal partial class UpdateOptions : ApplicationOptions
         public const string RuntimeIdentifierKey = "runtimeIdentifier";
         public const string SerializerModeKey = "serializer";
         public const string SyncKey = "sync";
+        public const string NoAsyncKey = "noAsync";
```
HongGit, line 27: Let's go with SyncOnly, same as svcutil.exe.

PR #5777, opened 2025-05-15
Fix pipeline build

mconnew:
Incorporates changes from #5772


akoeplinger (approved):

mconnew:
Build failure was due to a vendor modifying our test server to use individual app pools for each test site. This meant the test endpoints didn't have permission to use our test certificates. Should be fixed now.

akoeplinger:
there are a couple test failures but the arcade infrastructure seems to work again

mconnew:
The dotnet-wcf-with-corewcf--ci tests are known to fail, they are a work in progress. It's the non-corewcf tests that we care about which are all passing now.

Issue #5792, opened 2025-06-06
The svcutil tool fails when referencing an IXmlSerializable type that uses the XmlSchemaProviderAttribute


heatonmatthew:
**Describe the bug**
When referencing an assembly containing an `IXmlSerializable` type that needs to match a WSDL declared type in a custom namespace, use of the standard `System.Xml.Serialization.XmlSchemaProviderAttribute` doesn't enable the `dotnet-svcutil` tool to correctly match the type in the referenced assembly.  Instead, a new type is generated in the output code.

This prevents us from having a single implementation of types that are used across several web-services.

**Cause**
I've spent time debugging and found that the cause is in the `SchemaExporter.cs` file inside the `InvokeSchemaProviderMethod` where the implementation is searching for the `XmlSchemaProviderAttribute` type contained in the `FrameworkFork`.  That is, it's searching for `Microsoft.Xml.Serialization.XmlSchemaProviderAttribute` rather than `System.Xml.Serialization.XmlSchemaProviderAttribute`.

This is similar to the bug that @imcarolwang fixed in #5746 regarding the `IXmlSerializable` interface.  The fix is slightly more involved than the one in #5746 as additional types within the `InvokeSchemaProviderMethod` also need to be swapped to refer to the standard framework types rather than the FrameworkFork types.

I'm happy to provide a PR with my proposed fix.

**Expected behavior**
The `dotnet-svcutil` tool would find and use the `System.Xml.Serialization.XmlSchemaProviderAttribute` to resolve the custom namespace declaration and match the type in the referenced assembly to the WSDL type.

Issue #5638, opened 2024-08-28
The svcutil tool fails when one of the contracts implements IXmlSerializable


lipchev:
**Describe the bug**
The `"typeReuseMode": "All"` option no longer works when one of the `[DataMember]`s of a `[DataContract]` is itself not a `[DataContract]` but instead implements the `IXmlSerializable` interface. This isn't the case in `net472`  (using the "Update Service Reference").

**To Reproduce**
I've created a [simple repository](https://github.com/lipchev/CoreWCF-Service-Client-Tester) in order to demonstrate the issue (tested with all versions up to the latest build version: `8.0.0-dev.24427.1`).
The solution contains the following projects:
1. `CommonTypes` - this contians the "shared" data contracts (including the problematic `CustomSerializableType` contract). It is currently multi-targeting `net472` and `net8.0` but that's not a problem here.
2. `SerializationTests` - a simple test that demonstrates that the `DataContractSerializer` is capable of dealing with all of my `CommonTypes`
3. `CoreWCF-Service` - a .NET8 service hosting a vanilla IService returning the contracts from `CommonTypes`.
4. `NetFramework-Client` - a simple `net472` client referencing the `CommonTypes` which demonstrates that the scenario used to work before.
5. `CoreWCF-Service` - a simple `net8.0` client referencing the `CommonTypes` which demonstrates that the scenario doesn't work any more (tested using the latest build `svcutil tool`: `8.0.0-dev.24427.1`, as well as the _current_ version  `2.2.0-preview1.23462.5`)

The following json file
``` 
{
  "ExtendedData": {
    "inputs": [
      "https://localhost:7246/Service.svc"
    ],
    "collectionTypes": [
      "System.Array",
      "System.Collections.Generic.Dictionary`2"
    ],
    "namespaceMappings": [
      "*, ServiceReference1"
    ],
    "targetFramework": "net8.0",
    "typeReuseMode": "All"
  }
}

```
is passed to the svcutil using:
`dotnet svcutil -u "Connected Services/ServiceReference1" -v Debug`.

Here are the relevant parts of the WSDL:
```
<xs:schema elementFormDefault="qualified" targetNamespace="http://schemas.datacontract.org/2004/07/CommonTypes">
<xs:import schemaLocation="https://localhost:7246/Service.svc?xsd=xsd1" namespace="http://schemas.microsoft.com/2003/10/Serialization/"/>
<xs:complexType name="SomeSharedType">
<xs:annotation>
[...]

PR #5796, opened 2025-06-08
[dotnet-svcutil] Fully qualify XmlSchemaProvider to ensure correct resolution

heatonmatthew:
Addresses issue #5792: Fully qualify the use of the `XmlSchemaProvider` attribute with the correct namespace so that types which belong in non-default Xml namespaces can be correctly resolved from referenced libraries and projects. Without this fix, it resolves to the renamed XmlSchemaProvider type in the forked framework, causing the applied attribute to not be located and the type incorrect not matches as it's assumed to be in the default Xml namespace for a referenced assembly.  The effect is that the type is treated as non-reusable.

This is similar to @imcarolwang fix for #5638.


heatonmatthew:
@dotnet-policy-service agree company="Transmax"

heatonmatthew:
Note that I've extended the existing ReuseIXmlSerializableType unit test as it seemed the appropriate location. Previously that test used a pre-built binary for `CommonTypes.dll` that was checked into the repository. This made it difficult to either extend the existing test or write a parallel test using a similar referenced type assembly, so I substituted the pre-built binary for an equivalent `CommonTypes.csproj`.  That enabled me to build on the previous work by adding the new type required in the unit test case.

heatonmatthew:
Being unfamiliar with the test pipelines, I'm unsure whether the failures are known race-conditions or genuine failures.  Is anyone able to offer any advice or assist?  Thanks.  😄

mconnew:
@imcarolwang, can you look at this and see if it's something that has been addressed with any of the other changes made, or pending?

imcarolwang:
> @imcarolwang, can you look at this and see if it's something that has been addressed with any of the other changes made, or pending?

Hi @mconnew, I looked into this. It's still needed; none of the other pending changes address it.

The root cause is a type-identity mismatch: a reused user type carries the real [System.Xml.Serialization.XmlSchemaProviderAttribute], but the forked serialization code looks up the attribute against its own forked copy. Since same-name/different-assembly types aren't equal for [GetCustomAttributes], the schema-provider method is never found, dotnet-svcutil falls back to a default name, and type reuse silently fails. This PR fixes it by pinning the lookup to the real BCL attribute, and I've verified that works.

It's rarely reported because it needs a narrow combination: type reuse enabled, the reused type implements [IXmlSerializable], it declares a custom name via [[XmlSchemaProvider]], and it's from an externally compiled assembly. It also went unnoticed because svcutil's own generated types use the forked attribute (so self round-trips worked), and the failure is silent — it just regenerates instead of reusing.

My original thinking was to leave this open until the FrameworkFork folder was removed, which would make the fix unnecessary. That work is turning out to be longer-term and incremental, and this code path isn't trivial to remove , so this fix fills the gap in the meantime.

PR #5805, opened 2025-07-02
[dotnet-svcutil]  .NET 10.0 support

imcarolwang:
Description:

1. ~~Add net10.0 to the project's target frameworks to ensure the tool runs on machines where only .NET 10.0 SDK is installed (without .NET 8.0 or 9.0).~~
On machine with only net10.0 installed, installing the tool with the _--allow-roll-forward_ option ensures the support of net10.0.

2. Added .NET 10.0 to the supported target frameworks list during code generation, enabling the generation of WCF 8.x references for .NET 10.0 projects, while user projects targeting frameworks outside the list will still get WCF 4.10.x package references.


mconnew:
Can you also remove the net6.0 support?

PR #5809, opened 2025-07-14
Set OPENSSL_ENABLE_SHA1_SIGNATURES=1 in SendToHelix.proj

akoeplinger:
Instead of hardcoding this in the Helix image, see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1391

### DIFF
--- a/eng/SendToHelix.proj
+++ b/eng/SendToHelix.proj
@@ -78,6 +78,11 @@
     <HelixPreCommands>$(HelixPreCommands);sudo -E -n $HELIX_CORRELATION_PAYLOAD/InstallRootCertificate.sh --service-host $(ServiceHost) --cert-file /tmp/wcfrootca.crt</HelixPreCommands>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TestJob)' == 'Linux'" >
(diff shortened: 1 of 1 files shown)


mconnew:
Test failure looks to have just been a test infrastructure blip and nothing to do with the tests themselves.  

While I have no problem with this being in our pipeline setup, I am slightly concerned about there being zero need for this in the dotnet/runtime side. I fully understand that we shouldn't choose to use SHA1 for hashing, but we absolutely should be having tests to make sure validation of things like xml dsig signatures correctly validate (which is the consumption side of SHA1, not the generation side). Sometimes specs outside of our control require support for SHA1 and we should validate those scenarios (which is why WCF has a need for this). Also SHA1 isn't considered broken when using in an HMAC, only when used as a raw hashing function so there are still some legitimate (although pre-emptively discouraged) uses for SHA1.

akoeplinger:
Tests have been adjusted on dotnet/runtime in these environments, e.g. https://github.com/dotnet/runtime/pull/70821

mconnew:
A lot of defaults are using SHA1 with SOAP so it's not reasonable to disable those tests. I disagree with the runtime approach, but that's what the owners there chose to do, so this does seem like the best option for WCF.

mconnew (approved):

akoeplinger:
As far as I know it should only affect rsa+sha1, not sha1 in general.

PR #5810, opened 2025-07-15
[dotnet-svcutil] defer known types import to resolve IsReference inheritance issue

### LINKED ISSUE #5769: ReferencedTypeDoesNotMatch thrown on correct DataContract
**Describe the bug**
I have a WCF service that uses two DLLs with base types. Once we reference that base types, the data contract will throw the ReferencedTypeDoesNotMatch.

There must be a difference to older versions, where the Reference.cs has been written / even the referenced types are not used and therefore generated as duplicates.

The WSDL File is generated right before creating the service - so I assume that this is correct.

The strange thing this is, that this is only one service that is not working - other WSDL files in our project can be used without any hassle.

**To Reproduce**

1. I created a .NET Solution that produces the error with the current svc-util v8.0.0
2. You have to run the build.ps1 file
3. It crahes on CodeExporter.GetReferencedType 

When serializing the object it throws a null reference exception 

 ```
---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Runtime.Serialization.DataMember.CriticalHelper.get_MemberType() in C:\Github\wcf\src\dotnet-svcutil\lib\src\FrameworkFork\System.Runtime.Serialization\System\Runtime\Serialization\DataMember.cs:line 264
   at System.Runtime.Serialization.DataMember.get_MemberType() in C:\Github\wcf\src\dotnet-svcutil\lib\src\FrameworkFork\System.Runtime.Serialization\System\Runtime\Serialization\DataMember.cs:line 142

```
**Workaround**

The workaround is to change to comparison in CodeExporter.GetReferencedType (Line 563)

from:` if (referencedContract.Equals(dataContract))`
to: `if (referencedContract.StableName.Equals(dataContract.StableName))`

**Expected behavior**
That the Reference.cs will be created with respect to the referenced data types.

**Screenshots**

Image

[Error_WSDL_DataContract.zip](https://github.com/user-attachments/files/19906870/Error_WSDL_DataContract.zip)

imcarolwang:
Fixes #5769: The DataContract IsReference property was incorrectly set to false for some derived types when known types and inheritance were involved. This led to a mismatch between the imported DataContract type and the referenced type, which caused issues when trying to reuse the referenced types.

Root cause: ImportKnownTypes(typeName) was called immediately after importing each type, which could import derived types as known types before their base types were fully processed, causing inheritance-based IsReference propagation to fail.

Solution: Implement deferred known types import strategy:
- Add _deferredKnownTypesImport list to collect types requiring known types import
- Defer known types import until after all main types are imported
- Process deferred known types after inheritance chains are fully established
- Ensure IsReference property propagates correctly through inheritance hierarchy

This ensures that inheritance-based property propagation works correctly by processing known types only after the complete type hierarchy is established.

### DIFF
--- a/src/dotnet-svcutil/lib/src/FrameworkFork/System.Runtime.Serialization/System/Runtime/Serialization/SchemaImporter.cs
+++ b/src/dotnet-svcutil/lib/src/FrameworkFork/System.Runtime.Serialization/System/Runtime/Serialization/SchemaImporter.cs
@@ -28,6 +28,7 @@ namespace System.Runtime.Serialization
         private SchemaObjectDictionary _schemaObjects;
         private List<XmlSchemaRedefine> _redefineList;
         private bool _needToImportKnownTypesForObject;
+        private List<XmlQualifiedName> _deferredKnownTypesImport;
 
 
         // TODO: [Fx.Tag.SecurityNote(Critical = "Static field used to store serialization schema elements from future versions."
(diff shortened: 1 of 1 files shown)


imcarolwang:
**Update**
I was able to reproduce the issue on .NET Framework by rebuilding the referenced dll of the problem repro app to target .NET Standard 2.0, and confirmed the fix works on full framework as well.

Issue #5867, opened 2025-11-17
WCF test targets should be revisited


wfurt:
Looking at the SendToHelix.proj I see
```
  <ItemGroup Condition="'$(TestJob)' == 'Linux'" >
    <HelixTargetQueue Include="SLES.15.Amd64.Open" />
    <HelixTargetQueue Include="(Fedora.41.Amd64.Open)ubuntu.2204.amd64.open@mcr.microsoft.com/dotnet-buildtools/prereqs:fedora-41-helix" /> 
    <HelixTargetQueue Include="Ubuntu.2204.Amd64.Open" />
    <HelixTargetQueue Include="(Debian.12.Amd64.Open)Ubuntu.2204.Amd64.Open@mcr.microsoft.com/dotnet-buildtools/prereqs:debian-12-helix-amd64" />
    <HelixTargetQueue Include="(Mariner.2.0.Amd64.Open)Ubuntu.2204.Amd64.open@mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-helix-amd64" />
    <HelixTargetQueue Include="(openSUSE.15.6.Amd64.Open)Ubuntu.2204.Amd64.open@mcr.microsoft.com/dotnet-buildtools/prereqs:opensuse-15.6-helix-amd64" />
  </ItemGroup>
```

it seems like SLES and OpenSUSE are somewhat redundant - one with commercial support and one without. 
I'm curious if we could drop the SLES.15. and depend only on containers. That would allow us to retire that queue and associated infrastructure maintenance. 

On similar note, mariner-2.0 is out of support, replaced by AzureLinux.3 
For reference, here is what .NET runtime tests on:
https://github.com/dotnet/runtime/blob/main/eng/pipelines/libraries/helix-queues-setup.yml

Im happy to put up PR @mconnew if you want me to

mconnew:
That sounds good to me. Feel free to submit a PR. Our list was based on the runtime repo from a while ago.

PR #5871, opened 2025-12-03
update test queues

wfurt:
This brings test queues closer to dotnet/runtime
https://github.com/dotnet/runtime/blob/main/eng/pipelines/libraries/helix-queues-setup.yml

- runtime stop using SLES queue and uses OpenSuse docket image. It seems like WCF uses both now and that does not add much test coverage as the OpenSUSE is variant of commercially supported SLES. 
- Mariner 2 is EOL for a while and it was replaced by AzureLinux.3

runtime also moved from Ubuntu 22.04 as base queue to AzureLinux as the images are smaller, optimized for cloud and considered more secure. I did not want to make that change at the moment to decrease risk but perhaps something we can keep eyes when Ubuntu 22 gets closer to EOS. (April 2027) 

fixes https://github.com/dotnet/wcf/issues/5867


mconnew:
This PR will fail without some code changes to our tests. We have the ability to skip tests based on the Linux distro they are running, and out test infrastructure doesn't recognize the distro name "Microsoft Azure Linux 3.0"

PR #5889, opened 2026-01-21
[dotnet-svcutil] Remove forked framework codedom and use System.CodeDom package

imcarolwang:


src/dotnet-svcutil/lib/src/CodeDomFixup/CodeDomVisitors/AddAsyncOpenClose.cs
```diff
@@ -68,10 +68,10 @@ private CodeMemberMethod GenerateTaskBasedAsyncMethod(string methodName, NameSco
             if (methodName.Equals("Close"))
             {
                 string condition = _isVisualBasic ? "Not NET6_0_OR_GREATER" : "!NET6_0_OR_GREATER";
-                CodeIfDirective ifStart = new CodeIfDirective(CodeIfMode.Start, condition);
-                CodeIfDirective ifEnd = new CodeIfDirective(CodeIfMode.End, "");
-                implMethod.StartDirectives.Add(ifStart);
-                implMethod.EndDirectives.Add(ifEnd);
+                // System.CodeDom doesn't provide a CodeIfDirectiveStatement; we use a region directive as a stable marker
+                // and post-process the generated text to convert it into the desired conditional compilation directives.
```
mconnew: I asked copilot about this because it seems like there has to be an easier way than emitting this, and then doing a string replacement in CodeSerializer, and it looks like there might be. I think you can do this in the VisitClientClass method:
```c#
    type.Members.Add(new CodeSnippetTypeMember(
        "#if !NET6_0_OR_GREATER\r\n"
    ));
    type.Members.Add(GenerateTaskBasedAsyncMethod("Close", nameScope));
    type.Members.Add(new CodeSnippetTypeMember(
        "#endif\r\n"
    ));
```
And obviously switching it to `#If Not NET6_0_OR_GREATER Then\r\n` and `#End If\r\n` for Visual Basic.

src/dotnet-svcutil/lib/src/CodeDomFixup/CodeDomVisitors/RemoveExtensibleDataObjectImpl.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Microsoft.CodeDom;
+using System.CodeDom;
 using System.Runtime.Serialization;
```
mconnew: Out of scope for this PR, but do we still need this? I presume this is an artifact from the early days when IExtensibleDataObject wasn't in .NET Core 1.0 or 1.1, but they've been there since 2.0. I'm not finding equivalent code in .NET Framework to remove these. Can you open an issue and as a follow up task, work out if this is needed and if not, remove it.

src/dotnet-svcutil/lib/src/FrameworkFork/System.Web.Services/Services/Description/ServiceDescription.cs
```diff
@@ -11,7 +11,7 @@
 using System.IO;
 using System.Reflection;
 using System.ComponentModel;
-using Microsoft.CodeDom;
+using System.CodeDom;
 using System.Text;
```
mconnew: We shipped a WCF package called System.Web.Services.Description a few years ago. It doesn't have everything in there that is in this fork, but what's there is standalone, so you should be able to remove a lot of the code in this folder. This might be better done as a follow up PR to avoid adding too many changes to this PR and it taking a long time to merge.

src/dotnet-svcutil/lib/tests/Baselines/TFMBootstrap/tfmNetstd20/SvcutilBootstrapper/SvcutilBootstrapper.csproj
```diff
@@ -6,14 +6,17 @@
...
   <ItemGroup>
     <Reference Include="dotnet-svcutil-lib">
       <HintPath>$resultPath$/TestResults/TFMBootstrap/tfmNetstd20/bin/Debug/DOTNET_VERSION/dotnet-svcutil-lib.dll</HintPath>
     </Reference>
-    <Content CopyToOutputDirectory="always" Include="$resultPath$/TestResults/TFMBootstrap/tfmNetstd20/bin/Debug/net9.0/net6.0/**" Link="net6.0/%(RecursiveDir)%(Filename)%(Extension)" />
-    <Content CopyToOutputDirectory="always" Include="$resultPath$/TestResults/TFMBootstrap/tfmNetstd20/bin/Debug/net9.0/net8.0/**" Link="net8.0/%(RecursiveDir)%(Filename)%(Extension)" />
-    <Content CopyToOutputDirectory="always" Include="$resultPath$/TestResults/TFMBootstrap/tfmNetstd20/bin/Debug/net9.0/net462/**" Link="net462/%(RecursiveDir)%(Filename)%(Extension)" />
+    <Reference Include="System.CodeDom">
+      <HintPath>$resultPath$/TestResults/TFMBootstrap/tfmNetstd20/bin/Debug/net10.0/System.CodeDom.dll</HintPath>
+    </Reference>
```
mconnew: Same question about DOTNET_VERSION

src/dotnet-svcutil/lib/src/CodeDomFixup/CodeDomVisitors/AddAsyncOpenClose.cs
imcarolwang: Thank you for the suggestion. Using `CodeSnippetTypeMember `to emit `#if/#endif` around the generated `CloseAsync `method was considered. During validation, the CodeDOM provider did not always keep snippet members adjacent to the `CloseAsync `method, and the directives could be emitted at the start of the class instead of directly around the method.

To avoid that, the current change keeps the existing pattern: a unique `#region` marker is attached to the `CloseAsync `member via StartDirectives/EndDirectives, and during serialization only that marked region is rewritten into the corresponding conditional compilation directives (`#if/#endif` for C#, `#If/#End If` for VB). This keeps the rewrite small and focused on the intended member.

PR #5892, opened 2026-01-29
Context agnostic async methods continuation

11v1:
Apparently, the existing implementation does not account for applications with a synchronization context. Most methods lack ConfigureAwait calls, which causes continuations to capture the current synchronization context and resume execution on a context-specific thread.

This behavior is important for UI applications, where continuations may resume on the UI thread and, in certain scenarios (especially sync-over-async), can lead to deadlocks. For general-purpose library code, it is [recommended](https://devblogs.microsoft.com/dotnet/configureawait-faq/) to use ConfigureAwait(false) to avoid context capture and remain environment-agnostic.

This PR changes:
* ConfigureAwait(false) was added to all awaited asynchronous method calls, except in test projects.
* The solution was configured to treat missing ConfigureAwait usage as an error to prevent future omissions. This rule is enabled for all projects except tests.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -29,6 +29,10 @@ csharp_new_line_between_query_expression_clauses = true
 # IDE1006: Naming Styles
 dotnet_diagnostic.IDE1006.severity = error
 
+# CA2007: Consider calling ConfigureAwait on the awaited task
+[src/System.ServiceModel*/**/*.cs]
+dotnet_diagnostic.CA2007.severity = error
+
 # Indentation preferences
 csharp_indent_block_contents = true
 csharp_indent_braces = false
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -53,6 +53,8 @@
 
   <PropertyGroup Condition="'$(IsTestProject)' == 'true'">
     <EnableRunSettingsSupport Condition="'$(ContinuousIntegrationBuild)' != 'true'">true</EnableRunSettingsSupport>
+    <!-- Suppress CA2007 (ConfigureAwait) for test projects -->
+    <NoWarn>$(NoWarn);CA2007</NoWarn>
   </PropertyGroup>
 
   <ItemGroup Condition="'$(IsTestProject)' == 'true'">
--- a/src/System.ServiceModel.Http/src/System/ServiceModel/Channels/ClientWebSocketTransportDuplexSessionChannel.cs
+++ b/src/System.ServiceModel.Http/src/System/ServiceModel/Channels/ClientWebSocketTransportDuplexSessionChannel.cs
@@ -71,8 +71,8 @@ namespace System.ServiceModel.Channels
                 try
                 {
                     var clientWebSocket = new ClientWebSocket();
-                    await ConfigureClientWebSocketAsync(clientWebSocket, helper.RemainingTime());
-                    await clientWebSocket.ConnectAsync(Via, await helper.GetCancellationTokenAsync());
+                    await ConfigureClientWebSocketAsync(clientWebSocket, helper.RemainingTime()).ConfigureAwait(false);
+                    await clientWebSocket.ConnectAsync(Via, await helper.GetCancellationTokenAsync().ConfigureAwait(false)).ConfigureAwait(false);
                     ValidateWebSocketConnection(clientWebSocket);
                     WebSocket = clientWebSocket;
                 }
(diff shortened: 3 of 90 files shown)


11v1:
Probably I need help with build fix on CI
`##[error]src/dotnet-svcutil/lib/Microsoft.Svcutil.NamedPipeMetadataImporter/NamedPipeMetadataImporter.cs(58,27): error CA2007: (NETCORE_ENGINEERING_TELEMETRY=Build) Consider calling ConfigureAwait on the awaited task (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2007)`
I don't know where `NamedPipeMetadataImporter.cs` comes from, as it is not a part of this repository

mconnew:
Using `ConfigureAwait(false)` is the typical recommendation for async code. At the time WCF was being ported to .NET [Core], there weren't all the code analyzers available now to catch when you missed a ConfigureAwait call. In addition to that, if you have an async method with for example 5 calls to async methods that need to be awaited, the ConfigureAwait call is only needed the first time one of the async method calls doesn't complete synchronously. You need to "fix up" the continuation for the first time it goes async, but subsequent await's don't need the fixup. Overall this results in a lot of extra time being spent checking the result of the async call, seeing it completed synchronously, and ending up being a no-op, or seeing it's an asynchronous completion, but you're on the worker thread pool anyway so it being a no-op.  
It's a LOT of additional work which is only needed to be done if you are using a non-default synchronization context, AND it's the first async call to be awaited that actually completed asynchronously.  

Instead of the (at the time) risk of missing adding ConfigureAwait which could result in a regression breaking someone, WCF Client went a different path. WCF will always go async for any channel proxy call. Even if the outgoing request is able to be sent synchronously (e.g. the OS buffers the outgoing payload on the socket), we are going to be asynchronously waiting for the reply anyway. There's no point in trying to avoid an unnecessary thread hop as we're going to have that happen anyway at some point. What we do instead is at the earliest opportunity, we forcibly jump threads to a worked thread on the threadpool. We drop the execution context so we don't flow the SynchronizationContext. This means we don't need to ever call ConfigureAwait once we're past this point as it will always be a no-op due to already being on the default worker threadpool.  
We have tests to explicitly check that this happens. We use a single threaded SynchronizationContext that we execute some tests inside and make sure the call completes without blocking. The risk of not calling ConfigureAwait is that the continuation will be scheduled to the SynchronizationContext, and if it's single threaded, calling Wait() on the returned Task will cause a deadlock as you are holding the thread with the Wait() call, and the continuations can't progress. By forcing continuations to the worker threadpool before any actual async operations are possible, we guarantee we aren't going to deadlock on a single threaded SynchronizationContext. This means we never need to call ConfigureAwait. We avoid all the overhead of that call that's only needed in a very small percentage of the times it would get called, and don't have to worry about having missed calling it when code is changed.  

If you are having a problem which you think is caused by WCF not using ConfigureAwait, it's almost certainly not caused by that. If your problem goes away after applying these changes, it's more likely there's a race condition in your code that the extra overhead is changing the timing of and you are now avoiding. We have pretty robust tests around this. The only way you could have a problem is if you are using the WCF channel stack without a channel proxy (raw Message calls using a channel created from a raw ChannelFactory created directly from Binding.CreateChannelFactory). If this is your scenario, let me know and I can help.  

Basically, this isn't a problem, can you describe what symptoms you are seeing that you are trying to fix?

11v1:
@mconnew, thank you for the detailed answer. Your assumption about our usage of WCF without a channel proxy is correct - we create the channel directly via ChannelFactory.
The problem this change addresses is a System.TimeoutException that we encounter when using `NetNamedPipe`. After several successful sends, we start seeing:
`System.TimeoutException: Sending to via * timed out after 00:01:00. The time allotted to this operation may have been a portion of a longer timeout.`

Our investigation showed that `System.ServiceModel.Channels.TransportDuplexSessionChannel.OnSend` fails to acquire its semaphore. The semaphore is already held by another task, with the following stack trace:
```
>	[Async] System.ServiceModel.NetNamedPipe.dll!System.ServiceModel.Channels.PipeConnection.WriteAsync(System.ReadOnlyMemory<byte> buffer, bool immediate, System.TimeSpan timeout) Line 174	C#
 	[Async] System.ServiceModel.NetFramingBase.dll!System.ServiceModel.Channels.BufferedConnection.WriteNowAsync(System.ReadOnlyMemory<byte> buffer, System.TimeSpan timeout) Line 153	C#
 	[Async] System.Net.Security.dll!System.Net.Security.NegotiateStream.WriteAsync<System.Net.Security.AsyncReadWriteAdapter>(System.ReadOnlyMemory<byte> buffer, System.Threading.CancellationToken cancellationToken)	Unknown
 	[Async] System.ServiceModel.NetFramingBase.dll!System.ServiceModel.Channels.StreamConnection.WriteAsync(System.ReadOnlyMemory<byte> buffer, bool immediate, System.TimeSpan timeout) Line 311	C#
 	[Async] System.ServiceModel.NetFramingBase.dll!System.ServiceModel.Channels.TransportDuplexSessionChannel.OnSendAsync(System.ServiceModel.Channels.Message message, System.TimeSpan timeout) Line 436	C#
 	[Async] System.ServiceModel.Primitives.dll!System.Runtime.TaskHelpers.ToApm.AnonymousMethod__2_1(System.Threading.Tasks.Task antecedent, object obj) Line 113	C#

```

Adding `ConfigureAwait(false)` to the async methods along this path resolves the issue entirely - we no longer observe deadlocks. Based on our analysis, this does not appear to be a thread race condition. It’s also worth noting that switching the transport to `NetTcp` eliminates the problem without any code changes.

The reason I opted to add ConfigureAwait(false) broadly is that it’s a safe and defensive approach. It avoids the need to reason about whether a method is a public API / WCF entry point, or an internal method that is only ever called after the execution context has already been dropped. That distinction requires fairly deep knowledge of the WCF execution model, and it’s easy for a specific call path to be overlooked - which may be what happened here.

You may be right that our usage is incorrect; however, a few observations still stand:
 * The current WCF client implementation can resume execution on the UI thread, whereas the same code used in any other application with no synchronization context continues on a thread pool thread.
* The identical code path works without deadlocks on .NET Framework 4.8.

mconnew:
The call stack is going to help, I should be and to work out a fix.

mconnew:
I think the issue is I only implemented this for request/reply channel shapes, and it didn't get added to the IOutputChannel shaped channels, which we got away with due to SocketAwaitableEventArgs.SendAsync completing synchronously. Can you try adding `await TaskHelpers.EnsureDefaultTaskScheduler();` to `TransportDuplexSessionChannel.OnSendAsync` before the call to `await SendLock.WaitAsync`. This won't be sufficient to fix everything as this would still result in a deadlock on the sync code path, but as you're calling a Task based async method, it should fix it for your call path.

mconnew:
By the way, as a temporary workaround, you can do this:
```c#
var result = await Task.Run(() => channel.DoSomethingAsync());
```
This would effectively do the same thing we're doing deeper into our code.

mconnew:
Close it and I'll apply a different fix which covers sync and async scenarios. I just needed the validation that there wasn't anything else needed. Sometimes you fix one bug and it uncovers a similar one a few steps later. I can't see how that would be the case with this, but always worth doing the due diligence when possible.  
Did you try the workaround I suggested to unblock you until there's a release?

PR #5952, opened 2026-05-23
Un-skip CoreWCF-compatible tests; strengthen Mandatory transaction-fl…

afifi-ins:
After investigating every test gated by [Condition(nameof(Skip_CoreWCFService_FailedTest))], the following tests are now confirmed passing against the SelfHostedCoreWcfService and have their CoreWCF skip removed:

  * Client/ClientBase/ClientBaseTests.DefaultSettings_SetCookieOnServerSide
  * Extensibility/WebSockets/WebSocketTests.WebSocket_Https_Duplex_Buffered
  * Security/TransportSecurity/Tcp/Tcp_ClientCredentialTypeCertificateCanonicalNameTests
      - Certificate_With_CanonicalName_DomainName_Address_EchoString
      - Certificate_With_CanonicalName_Fqdn_Address_EchoString

WSHttpBinding_TransactionFlow_Mandatory_WithoutScope_Throws was originally observed to pass under CoreWCF as well, but that turned out to be a false positive: the test asserts only the client-side ProtocolException thrown by TransactionChannel before any wire activity (verified by running the test with no service at all, in 168ms). To make the test actually exercise CoreWCF's TransactionFlow support, the test has been:

  * Renamed to WSHttpBinding_TransactionFlow_Mandatory_RoundTrips_And_WithoutScope_Throws
  * Extended with a positive-case round-trip inside a TransactionScope that asserts the server sees the flowed transaction (IsTransactionFlowed returns true)
  * Kept gated by Skip_CoreWCFService_FailedTest because CoreWCF does not register the WSHttpTransactionFlowMandatory.svc endpoint (the host class is wrapped in #if !NET).

Verification matrix:
  * .NET Framework SelfHostedWcfService (net471): strengthened test PASSES in 3s
  * SelfHostedCoreWcfService (RunWithCoreWCF=true): strengthened test FAILS with EndpointNotFoundException at the new positive branch when the skip condition is temporarily removed, and is correctly SKIPPED with the condition in place.

### DIFF
--- a/src/System.Private.ServiceModel/tests/Scenarios/Binding/WS/TransactionFlow/WSHttpTransactionFlowTests.cs
+++ b/src/System.Private.ServiceModel/tests/Scenarios/Binding/WS/TransactionFlow/WSHttpTransactionFlowTests.cs
@@ -452,7 +452,7 @@ public class WSHttpTransactionFlowTests : ConditionalWcfTest
     [WcfFact]
     [Condition(nameof(Windows_Authentication_Available), nameof(Skip_CoreWCFService_FailedTest))]
     [OuterLoop]
-    public static void WSHttpBinding_TransactionFlow_Mandatory_WithoutScope_Throws()
+    public static void WSHttpBinding_TransactionFlow_Mandatory_RoundTrips_And_WithoutScope_Throws()
     {
         ChannelFactory<IWcfTransactionMandatoryService> factory = null;
         IWcfTransactionMandatoryService serviceProxy = null;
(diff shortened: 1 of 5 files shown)


mconnew:
@afifi-ins, this broke on the old windows test services. It resulted in the following error:
```
System.ServiceModel.Security.MessageSecurityException: Identity check failed for outgoing message. The expected DNS identity of the remote endpoint was 'wcfcoresrv23.westus3.cloudapp.azure.com' but the remote endpoint provided DNS claim 'wcfcoresrv23'. If this is a legitimate remote endpoint, you can fix the problem by explicitly specifying DNS identity 'wcfcoresrv23' as the Identity property of EndpointAddress when creating channel proxy.
```
I don't understand why this broke for dotnet-wcf-ci tests though as it looks like you only changed the certificate generation. But those tests use a certificate which is generated monthly, and uses the merged version from the main branch. Can you debug the failing test and work out what's going on.

mconnew (dismissed):

leculver (dismissed):

mconnew (approved):

PR #5959, opened 2026-06-13
Add client-side WebHttpBinding support (closes #1413, #46)

### LINKED ISSUE #46: Add support for HTTP requests which use the GET verb
`HttpTransportBindingElement` uses `HttpClient` to make the requests to the server. `HttpClient` will throw an exception if `HttpRequestMessage` has it's `Content` property not null when making a GET request. Currently we always set the Content property regardless of HTTP verb.

### LINKED ISSUE #1413: Add APIs to support WebHttpBinding
- [ ] T:System.ServiceModel.Channels.WebBodyFormatMessageProperty
- [ ] T:System.ServiceModel.Channels.WebContentTypeMapper
- [ ] T:System.ServiceModel.Channels.WebMessageEncodingBindingElement
- [ ] T:System.ServiceModel.Description.WebHttpBehavior
- [ ] T:System.ServiceModel.WebHttpBinding
- [ ] T:System.ServiceModel.WebHttpSecurity

afifi-ins:
Closes #1413
Closes #46

## Summary

Adds a new **client-side `System.ServiceModel.Web` package** for REST-style WCF clients on .NET, including:

- `WebHttpBinding`, `WebHttpSecurity`, and HTTP/HTTPS transport settings
- `WebHttpBehavior`, `WebGetAttribute`, and `WebInvokeAttribute`
- `WebChannelFactory<TChannel>` and client-side `WebOperationContext` request/response contexts
- `UriTemplate`, `UriTemplateTable`, matching/binding helpers, and `QueryStringConverter`
- XML, JSON, and raw `Stream` message encoding, including buffered and streamed paths
- XML/JSON reply formatter selection based on the response `Content-Type`

The implementation is sourced from the MIT-licensed .NET Framework reference source and CoreWCF, then adapted to current dotnet/wcf patterns and its client-only architecture.

## Scope

This PR implements the WCF **client** surface. Service hosting and dispatch remain CoreWCF responsibilities, so server-side dispatch selectors/formatters, help pages, service error handlers, JSONP, ASP.NET hosting integration, and `WebServiceHost` are intentionally excluded.

Raw octet-stream support is included. The reusable `ByteStreamMessage` and `ByteStreamMessageEncodingBindingElement` implementation lives in `System.ServiceModel.Primitives`; its public surface is exposed only for the current .NET target, matching the implementation TFMs.

## Compatibility and packaging

- Adds the shipping `System.ServiceModel.Web` project, reference assembly, unit-test project, package README, and solution wiring.
- Preserves compatibility with the runtime `System.ServiceModel.Web` facade by forwarding all 38 JSON and Syndication types exposed by that facade.
- Keeps the Web package free of `InternalsVisibleTo` dependencies on `System.ServiceModel.Http` and `System.ServiceModel.Primitives`.
- Documents deliberately retained .NET Framework `UriTemplate` and `QueryStringConverter` compatibility quirks.
- Uses centrally managed package versions and includes synchronized localized resources.

## Tests

- **312 `System.ServiceModel.Web` unit tests** covering bindings, security, URI templates, query conversion, attributes, operation context, behavior/formatters, XML/JSON/raw encoders, buffered/streamed paths, and facade compatibility.
- **12 WebHttp integration tests**: 8 local construction/loopback tests pass; 4 service-host scenarios are conditionally skipped when the external WCF test endpoint is unavailable.
- Focused `System.ServiceModel.Primitives` ByteStream tests cover buffering, offsets, ownership, repeated disposal, sync/async encoding, quotas, and validation.
- Full Release solution build completes with 0 warnings and 0 errors.

The self-hosted CoreWCF outerloop scenarios require elevated certificate installation locally and are expected to run in the CI/elevated test environment.

## Source attribution

Ported files retain the .NET Foundation MIT header. Commit history records whether each implementation came from the .NET Framework reference source or CoreWCF and documents client-specific adaptations.


src/System.ServiceModel.Web/src/System/ServiceModel/HttpTransportHelpers.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            DisableAuthentication(http);
+        }
+
+        private static void ConfigureAuthentication(HttpTransportBindingElement http, HttpTransportSecurity transportSecurity)
+        {
+            http.AuthenticationScheme = MapToAuthenticationScheme(transportSecurity.ClientCredentialType);
+            // Realm property is not exposed on dotnet/wcf's HttpTransportBindingElement /
+            // HttpTransportSecurity (server-side only). Skip in the client port.
+            http.ExtendedProtectionPolicy = transportSecurity.ExtendedProtectionPolicy;
```
mconnew, line 47: `ConfigureAuthentication` maps `ClientCredentialType` but never sets `http.ProxyAuthenticationScheme`. dotnet/wcf's own `HttpTransportSecurity.ConfigureAuthentication` does set it (`http.ProxyAuthenticationScheme = HttpProxyCredentialTypeHelper.MapToAuthenticationScheme(_proxyCredentialType)`), and `HttpChannelFactory` consumes it when an explicit proxy address is configured. As written, `Security.Transport.ProxyCredentialType` is silently dropped here.

This is compounded by `WebHttpBinding` not exposing `ProxyAddress`/`UseDefaultWebProxy`/`BypassProxyOnLocal` (all settable on the transport element), so the authenticated-proxy support that .NET Framework's `WebHttpBinding` had is effectively lost. Suggest mapping the proxy scheme here (mirroring the `MapToAuthenticationScheme` switch below, since `HttpProxyCredentialTypeHelper` is internal to System.ServiceModel.Http) and surfacing the proxy properties on the binding.

src/System.ServiceModel.Web/src/System/ServiceModel/WebHttpBinding.cs
```diff
@@ -0,0 +1,202 @@
...
+            _security.Mode = securityMode;
+        }
+
+        // This needs an update in the HttpTransportBindingElement to work.
+        //[DefaultValue(false)]
+        //public bool AllowCookies
+        //{
+        //    get { throw new PlatformNotSupportedException(); }
+        //    set { throw new PlatformNotSupportedException(); }
+        //}
```
mconnew: The comment here isn't accurate — `AllowCookies` is fully supported in dotnet/wcf. `HttpTransportBindingElement.AllowCookies` exists and `HttpChannelFactory` honors it (`clientHandler.UseCookies = AllowCookies`); `WSHttpBinding`/`HttpBindingBase` already wire it to both transports. It can be implemented the same way rather than left as a commented/throwing stub:

```csharp
[DefaultValue(HttpTransportDefaults.AllowCookies)]
public bool AllowCookies
{
    get { return _httpTransportBindingElement.AllowCookies; }
    set
    {
        _httpTransportBindingElement.AllowCookies = value;
        _httpsTransportBindingElement.AllowCookies = value;
    }
}
```

The same goes for `ProxyAddress`, `UseDefaultWebProxy`, and `BypassProxyOnLocal` — .NET FX exposed them and they're settable on the transport element, but they're omitted here. (`HostNameComparisonMode` genuinely isn't on the element yet, so leaving that one out is fair.)

src/System.ServiceModel.Web/src/System/ServiceModel/WebHttpBinding.cs
```diff
@@ -0,0 +1,202 @@
...
+        public bool CrossDomainScriptAccessEnabled
+        {
+            get { return _webMessageEncodingBindingElement.CrossDomainScriptAccessEnabled; }
+            set
+            {
+                _webMessageEncodingBindingElement.CrossDomainScriptAccessEnabled = value;
+            }
+        }
+
+        public override BindingElementCollection CreateBindingElements()
```
mconnew, line 182: The reference `WebHttpBinding` overrides `BuildChannelFactory` to throw `InvalidOperationException` up front when `Security.Mode` is `Transport`/`TransportCredentialOnly` and `Transport.ClientCredentialType == InheritedFromHost` (a value that's invalid on a client). That override isn't ported.

The case still fails — `InheritedFromHost` maps to `AuthenticationSchemes.None`, so `HttpTransportBindingElement.BuildChannelFactory` throws — but as an `ArgumentException` (`HttpAuthSchemeCannotBeNone`), a different exception type/message than .NET Framework produced. Minor, but porting the guard restores parity and gives a clearer error at factory-creation time.

src/System.ServiceModel.Web/src/System/ServiceModel/WebHttpBinding.cs
afifi-ins: Fixed: `AllowCookies`, `ProxyAddress`, `UseDefaultWebProxy`, and `BypassProxyOnLocal` now propagate to both HTTP and HTTPS transports; cookie round-trip is covered.

src/System.ServiceModel.Web/src/System/ServiceModel/HttpTransportHelpers.cs
afifi-ins: Fixed: proxy credentials now set and reset `ProxyAuthenticationScheme`, and proxy settings flow through `WebHttpBinding`. Covered by proxy mapping tests.

src/System.ServiceModel.Web/src/System/ServiceModel/WebHttpBinding.cs
afifi-ins: Fixed: `BuildChannelFactory` rejects `InheritedFromHost` for transport modes with the framework-compatible exception; positive and negative tests cover the guard.

release-notes/SupportedFeatures-v2.1.0.md
```diff
@@ -14,7 +14,7 @@ Supported features
 |              |CustomBinding          |:white_check_mark:                  |:white_check_mark:                          | :white_check_mark:                  |
 |              |NetHttpBinding         |:white_check_mark:                  |:white_check_mark:                          | :white_check_mark:                  |
 |              |NetTcpBinding          |:white_check_mark:                  |:white_check_mark:                          | :white_check_mark:    |
-|              |WebHttpBinding         |:no_entry_sign:                     |:no_entry_sign:                            | :no_entry_sign:                      |
```
mconnew, line 17: >|              |WebHttpBinding         |:no_entry_sign:            [](http://example.com/codeflow?start=0&length=66)

We shouldn't be modifying this file as it's the features released with a really old version.

src/System.ServiceModel.Http/src/System.ServiceModel.Http.csproj
```diff
@@ -15,6 +15,10 @@
     <ProjectReference Include="..\..\System.ServiceModel.Primitives\src\System.ServiceModel.Primitives.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == '$(DotNetVersion)'">
+    <InternalsVisibleTo Include="System.ServiceModel.Web" />
```
mconnew: Can you explain what internals are needed by System.ServiceModel.Web from System.ServiceModel.Http? Would any of them be reasonable to expose publicly? Basically should we keep this or can we work around all the issues? InternalsVisibleTo is generally not the best way to solve a problem unless absolutely needed.

mconnew:
>     {

Consider replacing with ObjectPool\<T\> from the runtime.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Pool.cs:9 in db4e189. [](commit_id = db4e1890aa117550df5480f85d8dadcbf4e22f7c, deletion_comment = False)

mconnew:
>     public class UriTemplate

No need for the type forward as this won't be consumed by assemblies compiled for .NET 3.5

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:19 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
> 

Update to use naming conventions for instance fields

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:22 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>             }

Use nameof for arguments/parameters that are referenced by exceptions.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:57 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>             : this(template, ignoreTrailingSlash, null)

Formatting, there should be a blank line between each method/constructor

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:45 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>             this.ignoreTrailingSlash = ignoreTrailingSlash;

`this.` shouldn't generally be needed and the current coding standards says to omit it if not needed.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:59 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>                     SR.UTBadBaseAddress));

SR.Format is only needed when there string resource is parameterized. When it isn't parameterized, you can use the string resource directly without calling SR.Format.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:337 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>         //  templates as such based on the structure of them and not based on the set of uri

Nit: Typo in comment. Intuative -> intuitive.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:410 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>                 // and warning suppression isn't working

This can probably be removed as it was included to stop a false positive on a static analysis rule in the .NET Framework source code. Replace with an Fx.Assert that they are not null to ensure the stated assertion holds true.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:429 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>             {

Formatting, we put blank lines after closing braces of code block, unless the next line is another closing brace.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:451 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>             Fx.Assert(segmentsCount >= this.firstOptionalSegment - 1, "How can that be? The Trie is constructed that way!");

Nit: extra space on final line of comment

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:582 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>         {

All methods should have their accessibility explicitly stated. We do rely on default accessibility.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:633 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>                 }

Update to use expression bodies where appropriate.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:981 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>             public int Count

User `#region` to group methods that implement a specific interface.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:1034 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>                 //  the nature of the last M-N path segments should be StringNature.Variable; otherwise,

There are multiple issues with extra spaces in comments. Fix them across the board.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:1414 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>         {

The one place this method was originally called has now been commented out, so remove this method.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplateHelpers.cs:21 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>                 }

Remove dead code. Have Copilot look for other commented out dead code and remove if appropriate

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplate.cs:562 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     using System.Runtime;

Move using's outside of namespace in all files.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplateLiteralPathSegment.cs:7 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
> 

Fix naming to match naming conventions for all private static fields. Also a reminder about explicit accessibility being specified.

---
Refers to: src/System.ServiceModel.Web/src/System/UriTemplateLiteralPathSegment.cs:19 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     {

I think there might be some async method overrides that need to be added to this class.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Channels/HttpStreamMessage.cs:11 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>         {

I couldn't find anywhere this was used

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Channels/MessageExtensions.cs:11 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>                     // Raw (Content-Type: application/octet-stream pass-through) support is deferred

ByteStreamMessageEncodingBindingElement should be ported into the Primitives package as it has utility outside of WebHttpBinding. Add it to this PR.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Channels/WebMessageEncoderFactory.cs:111 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>         {

It looks like this was ported over from the CoreWCF code base as it's referencing IServiceProvider. Have AI re-port this from the NetFx codebase as we don't need IServiceProvider for the client.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Description/WebHttpBehavior.cs:38 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     {

IDispatchMessageFormatter isn't needed for WebHttpBinding as it's only doing Request/Reply. IDispatchMessageFormatter is only used for duplex transports on the client.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Dispatcher/HttpStreamFormatter.cs:12 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>                 throw DiagnosticUtility.ExceptionUtility.ThrowHelperError(

Make sure this along with any other references is fixed once ByteStreamMessageEncodingBindingElement is ported.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Dispatcher/HttpStreamFormatter.cs:77 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
> 

I think an ImmutableHashSet\<Type\> would be a better type to use for storing the default supported query string types.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Dispatcher/QueryStringConverter.cs:19 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     {

Drop the IDispatchMessageFormatter as not needed on the client side

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Dispatcher/SingleBodyParameterMessageFormatter.cs:16 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     {

A performance issue was found with this code in CoreWCF and was replace with something else. I can't remember whether the replacement is usable here as I think it might have used an asp.net core class which the WCF Client can't reference. Take a look and see if it is something in the main runtime that I switched to. If not, then ask Copilot for an alternative. There might be something in the newer HttpClient headers api's which provides this parsing capability. Either way, we need to strip this code out and replace its usage with something else.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/HttpDateParse.cs:19 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     {

This is a service side class to represent an incoming request. On the client side we should only have IncomingWebResponseContext and OutgoingWebResponseContext

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/IncomingWebRequestContext.cs:16 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     {

Same for this class, this is a service side object representing sending a response to the client. It shouldn't be in this codebase

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/OutgoingWebResponseContext.cs:16 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     {

This interface is only used for server side code.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/IWebFaultException.cs:10 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     {

Once the unnecessary service side code has been removed, check this class to make sure everything is still needed.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/Utility.cs:13 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>     {

As far as I can tell, only used in service side code

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/WebFaultException.cs:15 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
> 

There's no incoming request on the client side

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/WebOperationContext.cs:57 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
> 

There's no outgoing response on the client side

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/WebOperationContext.cs:59 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>         public void Attach(OperationContext owner)

This is missing:
```c#
         public IncomingWebResponseContext IncomingResponse { get; }

        public OutgoingWebRequestContext OutgoingRequest { get; }
```

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/WebOperationContext.cs:60 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>         {

I believe the `CreateXXXXXXXResponse` methods are all unnecessary as the client doesn't create responses.

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/WebOperationContext.cs:69 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

mconnew:
>         {

Likely unneeded was the CreateXmlResponse methods go away

---
Refers to: src/System.ServiceModel.Web/src/System/ServiceModel/Web/WebOperationContext.cs:223 in a0df75f. [](commit_id = a0df75f31c20961c7e0e3498899ab2cd16b74cbd, deletion_comment = False)

src/System.ServiceModel.Http/src/System.ServiceModel.Http.csproj
afifi-ins: Fixed: removed the HTTP-to-Web `InternalsVisibleTo`; required defaults and proxy mapping are now local to Web.

release-notes/SupportedFeatures-v2.1.0.md
afifi-ins: Fixed: reverted the v2.1 release-note change; the file now matches `main`.

afifi-ins:
I investigated the `System.ServiceModel.Web` simple-name collision with the runtime facade and pushed a compatibility fix in `05abde92f`.

Findings:
- A normal `PackageReference` consumer builds cleanly on both `net10.0` and `net11.0`; NuGet/MSBuild selects this package assembly, so I could not reproduce `MSB3243` in the shipping-consumer path.
- The runtime still contains `System.ServiceModel.Web, Version=4.0.0.0, PublicKeyToken=31bf3856ad364e35`, while this package produces version 10.0.0.0 with token `b03f5f7f11d50a3a`. The default `AssemblyLoadContext` permits only one assembly per simple name, so the app-local package shadows the framework facade.
- I reproduced the resulting binary break with a legacy IL assembly whose signatures reference the v4 facade's `DataContractJsonSerializer` and `SyndicationFeed` types. Before the fix, reflecting the first signature threw `TypeLoadException` because our assembly had none of the facade's forwards.

The fix makes the new assembly a strict superset of the framework facade by preserving its exact 38 JSON/Syndication type forwards, adds the existing centrally-versioned `System.ServiceModel.Syndication` dependency, and locks the forward set with a unit test. The same legacy-binary probe now resolves JSON to `System.Private.DataContractSerialization` and feeds to `System.ServiceModel.Syndication` successfully.

afifi-ins:
All review feedback is resolved on squashed head `928e6d6ca`.

- [5072893121](https://github.com/dotnet/wcf/pull/5959#issuecomment-5072893121): replaced the custom pool with runtime `ObjectPool<T>`; removed `Pool.cs`.
- [5110479822](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110479822): removed unnecessary `TypeForwardedFrom` attributes.
- [5110483372](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110483372): normalized instance fields to `_camelCase`.
- [5110487977](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110487977): replaced argument-name literals with `nameof`.
- [5110490637](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110490637): added required member spacing.
- [5110494794](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110494794): removed unnecessary `this.` qualifiers.
- [5110507395](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110507395): use resources directly when no formatting is required.
- [5110513268](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110513268): corrected the comment typo.
- [5110526036](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110526036): removed the obsolete suppression and retained the invariant with `Fx.Assert`.
- [5110533365](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110533365): normalized blank lines after code blocks.
- [5110538647](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110538647): removed trailing comment whitespace.
- [5110544808](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110544808): made member accessibility explicit.
- [5110550914](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110550914): simplified the cited wrappers to get-only auto-properties.
- [5110557901](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110557901): grouped interface implementations with regions.
- [5110566784](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110566784): normalized comment spacing across Web source.
- [5110584133](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110584133): removed unused `AssertCanonical`.
- [5110587858](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110587858): removed commented-out dead code.
- [5110599002](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110599002): moved all `using` directives outside namespaces.
- [5110606854](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110606854): fixed static-field naming and accessibility.
- [5110642285](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110642285): removed `HttpStreamMessage`; raw streams now use the complete shared ByteStream implementation.
- [5110724902](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110724902): removed unused `MessageExtensions`.
- [5110915507](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110915507): ported ByteStream encoding to Primitives with reference API and tests.
- [5110929128](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110929128): re-ported `WebHttpBehavior` as client-only with no `IServiceProvider` dependency.
- [5110938838](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110938838): `HttpStreamFormatter` now implements only `IClientMessageFormatter`.
- [5110943378](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110943378): raw request/reply handling now uses the ByteStream encoder.
- [5110966047](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110966047): default query types now use a shared `FrozenSet<Type>`.
- [5110969586](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110969586): removed `IDispatchMessageFormatter` from the single-body formatter.
- [5110990620](https://github.com/dotnet/wcf/pull/5959#issuecomment-5110990620): removed unused server-only `HttpDateParse` after its consumers were removed.
- [5111010034](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111010034): removed server-only `IncomingWebRequestContext`.
- [5111016609](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111016609): removed server-only `OutgoingWebResponseContext`.
- [5111035074](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111035074): removed server-only `IWebFaultException`.
- [5111041710](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111041710): trimmed `Utility` to its remaining used helper.
- [5111054881](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111054881): removed server-only `WebFaultException`.
- [5111063192](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111063192): removed client-inapplicable `IncomingRequest`.
- [5111064806](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111064806): removed client-inapplicable `OutgoingResponse`.
- [5111077148](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111077148): added and wired `IncomingResponse` and `OutgoingRequest`.
- [5111088028](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111088028): removed client-inapplicable response factories.
- [5111093957](https://github.com/dotnet/wcf/pull/5959#issuecomment-5111093957): removed `CreateXmlResponse` and related response helpers.

The five inline threads also have concise final replies and are marked resolved. CI is green.

PR #5968, opened 2026-07-29
Fix Sync_PRService_Windows never invoking sync-pr.cmd

mconnew:
The Windows PR-service sync step never ran sync-pr.cmd, so the WCF test service was never synced to the PR on the Windows and macOS legs (macOS jobs orchestrate on Windows agents, so they take this same step).

The invocation was:

  & "$env:Build_SourcesDirectory\...\sync-pr.cmd "$env:WCFPRSERVICEID ...""

PowerShell parses this by juxtaposition, so the command name becomes "...\sync-pr.cmd " with a trailing space and the remaining text is folded in as arguments (the last one is not even expanded - it tokenizes as the literal CommandArgument "$env:branchNameorPrId"). The trailing space defeats PATHEXT matching, so PowerShell treats the target as a document rather than an application and ShellExecutes it. Under Windows PowerShell 5.1 - the shell the pipeline task uses - that is a silent no-op: no stdout, no error record, and $LASTEXITCODE is left untouched, so the step reports success while doing nothing.

This is observable in any build: sync-pr.cmd echoes a banner unconditionally before it validates its arguments, and that banner never appears in the Windows step log, while the Unix step log contains it along with "Making call to <sync url>".

The practical effect is that a PR which adds a new endpoint to IISHostedWcfService only gets that endpoint deployed on the slots synced by the Linux legs, and the scenario tests fail on the other legs against a stale service.

Invoke the script properly and surface a non-zero exit code as a pipeline warning so a future failure is not silent again.

### DIFF
--- a/eng/UpdatePRService.yml
+++ b/eng/UpdatePRService.yml
@@ -96,8 +96,11 @@ steps:
       Write-Host "Variable Operation is set to: $env:operation"
       Write-Host "Variable wcfPRServiceUri is set to: $env:WCFPRSERVICEURI"
       Write-Host "Variable branchNameorPrId is set to: $env:branchNameorPrId"
-      invoke-command -Scriptblock { & "$env:Build_SourcesDirectory\src\System.Private.ServiceModel\tools\scripts\sync-pr.cmd "$env:WCFPRSERVICEID $env:operation $env:WCFPRSERVICEURI $env:branchNameorPrId"" }
-      $LASTEXITCODE
+      & "$env:Build_SourcesDirectory\src\System.Private.ServiceModel\tools\scripts\sync-pr.cmd" $env:WCFPRSERVICEID $env:operation $env:WCFPRSERVICEURI $env:branchNameorPrId
+      Write-Host "sync-pr.cmd exit code: $LASTEXITCODE"
+      if ($LASTEXITCODE -ne 0) {
+        Write-Host "##vso[task.logissue type=warning]sync-pr.cmd failed with exit code $LASTEXITCODE - the test service was NOT synced to this PR."
+      }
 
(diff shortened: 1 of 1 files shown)


brianrob (approved):

eng/UpdatePRService.yml
```diff
@@ -96,8 +96,11 @@ steps:
       Write-Host "Variable Operation is set to: $env:operation"
       Write-Host "Variable wcfPRServiceUri is set to: $env:WCFPRSERVICEURI"
       Write-Host "Variable branchNameorPrId is set to: $env:branchNameorPrId"
-      invoke-command -Scriptblock { & "$env:Build_SourcesDirectory\src\System.Private.ServiceModel\tools\scripts\sync-pr.cmd "$env:WCFPRSERVICEID $env:operation $env:WCFPRSERVICEURI $env:branchNameorPrId"" }
-      $LASTEXITCODE
```
StephenMolloy, line 100: Does removing this cause this task to *not* return the exit code anymore? Is that intentional? Your PR comment said "and surface a non-zero exit code as a pipeline warning so a future failure is not silent again"... it's not clear to me whether you meant to swallow the non-zero exit code, or just add a warning message along with it.

mconnew: The changes cause it to emit a special string to stdout which the pipeline runner uses to interpret the success status of the script instead of using the exit code. Because most of the time running the script is basically a no-op and failure only matters if there are new service test endpoints created in the PR, a failure will often not be a problem. The script emits that as a warning which will show up in the pipeline steps view that there was a problem without preventing the rest of the pipeline from continuing and the tests actually being ran.

PR #5981, opened 2026-09-16
Remove stale SDL validation parameters

missymessa:
## Summary

- remove the obsolete `SDLValidationParameters` argument from the official post-build template invocation
- unblock YAML compilation for WCF production pipeline 318

## Context

Pipeline 318 currently fails before any jobs start because the flowed Arcade `post-build.yml` template no longer accepts `SDLValidationParameters`. This prevents production validation of the OneLoc GitHub App migration tracked by [AB#12306](https://dev.azure.com/dnceng/internal/_workitems/edit/12306).

The 1ES pipeline already configures SDL through its top-level `sdl` parameters, so this removes only the stale post-build argument.

## Validation

- parsed `azure-pipelines-arcade.yml` successfully with a YAML parser
- `git diff --check`

## Follow-up

After merge, run pipeline 318 from `main` and verify the `Get GitHub App installation token` and OneLoc tasks succeed.


missymessa:
@mconnew looks like there was a little cruft leftover that needed to be cleaned up after my last PR. This should fix the builds.

mconnew (approved):

